PR #102, opened 2020-02-24
Ak/tlcprinter

andrey-kuprianov:
Counterexample output is now possible in either a TLA format (default), or TLC format (if --tlc option is supplied to the check command).

As of now, source position reporting is not possible, blocked by https://github.com/konnov/apalache/issues/101. After the resolution of this issue, the source position tracking will be added.

### DIFF
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
@@ -45,4 +45,8 @@ class CheckCmd extends Command(name = "check",
     opt[String](name = "filter", default = "",
       description = "A sequence of regexes over transition numbers\n" +
         "to filter transitions at every step, e.g., (0|1),(1|2),4")
+  var tlc: Boolean = opt[Boolean](
+    name = "tlc", default = false,
+    description = "produce counterexample in TLC format,\n" +
+      "default: false")
 }
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
@@ -45,8 +45,4 @@ class CheckCmd extends Command(name = "check",
     opt[String](name = "filter", default = "",
       description = "A sequence of regexes over transition numbers\n" +
         "to filter transitions at every step, e.g., (0|1),(1|2),4")
-  var tlc: Boolean = opt[Boolean](
-    name = "tlc", default = false,
-    description = "produce counterexample in TLC format,\n" +
-      "default: false")
 }
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
@@ -37,8 +37,7 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
                    exprGradeStore: ExprGradeStore, sourceStore: SourceStore, checkerInput: CheckerInput,
                    searchStrategy: SearchStrategy,
                    tuningOptions: Map[String, String],
-                   debug: Boolean = false, profile: Boolean = false, checkRuntime: Boolean = false,
-                   tlc: Boolean = false)
+                   debug: Boolean = false, profile: Boolean = false, checkRuntime: Boolean = false)
   extends Checker with LazyLogging {
 
   import Checker._
@@ -157,8 +156,8 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
       case FinishOnDeadlock() =>
         logger.error(s"Deadlock detected.")
         if (solverContext.sat()) {
-          logger.error("Check an execution leading to a deadlock state in counterexample.txt")
-          dumpCounterexample(ValEx(TlaBool(true)) )
+          val filenames = dumpCounterexample(ValEx(TlaBool(true)) )
+          logger.error(s"Check an execution leading to a deadlock state in any of ${filenames.mkString(", ")}")
         } else {
           logger.error("No model found that would describe the deadlock")
         }
@@ -488,8 +487,8 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
           // introduce a dummy oracle to hold the transition index, we need it for the counterexample
           val oracle = new MockOracle(transitionNo)
           stack = (notInvState, oracle) +: stack
-          val filename = dumpCounterexample(notInv)
-          logger.error(s"Invariant is violated at depth $stepNo. Check the counterexample in $filename")
+          val filenames = dumpCounterexample(notInv)
+          logger.error(s"Invariant is violated at depth $stepNo. Check the counterexample in any of ${filenames.mkString(", ")}")
           if (debug) {
             logger.warn(s"Dumping the arena into smt.log. This may take some time...")
             // dump everything in the log
@@ -510,40 +509,16 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
     }
   }
 
-  private def dumpCounterexample(notInv: TlaEx): String = {
-    // Assumption: stack cannot be empty
-    val filename = if(tlc) "MC.out" else "counterexample.tla"
-    val writer = new PrintWriter(new FileWriter(filename, false))
-
-    val nextStates = new ListBuffer[NextState]()
+  // returns a list of files with counterexample written
+  private def dumpCounterexample(notInv: TlaEx): List[String] = {
+    val states = new ListBuffer[NextState]()
     for (((state, oracle), i) <- stack.reverse.zipWithIndex) {
       val decoder = new SymbStateDecoder(solverContext, rewriter)
       val transition = oracle.evalPosition(solverContext, state)
       val binding = decoder.decodeStateVariables(state)
-      nextStates += ((transition.toString, binding))
+      states += ((transition.toString, binding))
     }
-    if(tlc) {
-      writer.println(
-        s"""@!@!@STARTMSG 2262:0 @!@!@
-          |Created by Apalache on ${Calendar.getInstance().getTime}
-          |@!@!@ENDMSG 2262 @!@!@""".stripMargin
-      )
-      val tlcWriter = new TlcCounterexampleWriter(writer)
-      tlcWriter.write(notInv, nextStates.head._2, nextStates.tail.toList)
-    }
-    else {
-      writer.println("%s MODULE counterexample %s\n".format("-" * 25, "-" * 25))
-      writer.println("EXTENDS %s\n\n".format(checkerInput.rootModule.name))
-      val tlaWriter = new TlaCounterexampleWriter(writer)
-      tlaWriter.write(notInv, nextStates.head._2, nextStates.tail.toList)
-      writer.println("\n%s".format("=" * 80))
-      writer.println("\\* Created %s by Apalache".format(Calendar.getInstance().getTime))
-      writer.println("\\* https://github.com/konnov/apalache")
-
-    }
-
-    writer.close()
-    filename
+    CounterexampleWriter.writeAllFormats(checkerInput.rootModule, notInv, states.toList)
   }
 
   private def checkTypes(expr: TlaEx): Unit = {
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
@@ -1,7 +1,9 @@
 package at.forsyte.apalache.tla.bmcmt
 
 import java.io.{FileWriter, PrintWriter, StringWriter}
+import java.util.Calendar
 
+import at.forsyte.apalache.tla.assignments.FalseEx
 import at.forsyte.apalache.tla.bmcmt.analyses.{ExprGradeStore, FormulaHintsStore}
 import at.forsyte.apalache.tla.bmcmt.rewriter.{ConstSimplifierForSmt, RewriterConfig}
 import at.forsyte.apalache.tla.bmcmt.rules.aux.{CherryPick, MockOracle, Oracle}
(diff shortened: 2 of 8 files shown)


lemmy:
@andrey-kuprianov Can you describe the TLA format?  Does SANY accept it?  Background: There is something similar in TLC (MC_TE.tla) and we should probably make sure both are compatible.

andrey-kuprianov:
@lemmy the format is very simple: just the list of states leading the the invariant violation, like that:

(* Initial state *)
State0 == BIRTH_YEAR = 89 /\ LICENSE_AGE = 10

(* Transition 0 to State1 *)
State1 == BIRTH_YEAR = 89
  /\ LICENSE_AGE = 10
  /\ hasLicense = FALSE
  /\ year = 89
 ...

State13 == BIRTH_YEAR = 89
  /\ LICENSE_AGE = 10
  /\ hasLicense = TRUE
  /\ year = 0

(* The following formula holds true in State13 and violates the invariant *)
InvariantViolation == hasLicense /\ year - BIRTH_YEAR < LICENSE_AGE

I've checked -- it's not parsed as of now, but I will add `EXTENDS` decl to the original source, and then everything will work.

Thanks for pointing out about your module: where can I find the description for what TLC produces? And yes, it would be nice to have compatibility.

konnov:
@lemmy, it is a good idea to make it parseable by SANY. I was already happy to see that the syntax highlighting worked in vim.
@andrey-kuprianov, we could add `VARIABLES` and `CONSTANTS` in the counterexample. Figuring out what kind of modules should be extended is a bit tricky. We are not propagating this at the transformation phase. As all user operators are being expanded anyways, we could just extend all the standard modules: Integers, Sequences, FiniteSets, TLC, and, maybe, bags.

andrey-kuprianov:
@lemmy, @konnov : I've pushed the change to make TLA counterexample parseable: simply extended the root module that the checker works on...

@konnov : this should work in all cases, right? please correct me if I am wrong...

lemmy:
FWIW: TLC's ```TE.tla``` extends the spec that was model checked.  With https://github.com/tlaplus/tlaplus/issues/393, we introduced the concept of a monolith spec that is the union of the model-checked spec, ```TE.tla```, and the transitive closure of all extended modules.  The monolith spec serves the purpose of being a snapshot, i.e. the user has the version of the model-checked spec and its extended modules even if the actual spec evolves further.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
```diff
@@ -45,4 +45,8 @@ class CheckCmd extends Command(name = "check",
     opt[String](name = "filter", default = "",
       description = "A sequence of regexes over transition numbers\n" +
         "to filter transitions at every step, e.g., (0|1),(1|2),4")
+  var tlc: Boolean = opt[Boolean](
```
konnov: why don't we produce counterexamples in all supported formats: TLC's, TLA+, and JSON? Imagine that I have been running the tool for 1 day, just to realize that I forgot to specify `--tlc`. Dumping the counterexamples is cheap in comparison to model checking.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
```diff
@@ -506,22 +510,38 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
     }
   }
 
-  private def dumpCounterexample(): String = {
-    val filename = "counterexample.txt"
+  private def dumpCounterexample(notInv: TlaEx): String = {
+    // Assumption: stack cannot be empty
+    val filename = if(tlc) "MC.out" else "counterexample.tla"
```
konnov: Let's delegate the name choice to a specific implementation. Note that in the multicore version, we have to give a stem, as multiple core may produce multiple counterexamples for multiple invariants.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
```diff
@@ -506,22 +510,38 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
...
-      writer.println("--------")
       val binding = decoder.decodeStateVariables(state)
-      val args = binding.toList.sortBy(_._1).flatMap(p => List(ValEx(TlaStr(p._1)), p._2))
-      if (args.nonEmpty) {
-        val record = OperEx(TlaFunOper.enum, args :_*)
-        new PrettyWriter(writer).write(record)
-      }
-      writer.println("\n========\n")
+      nextStates += ((transition.toString, binding))
+    }
```
konnov: Nice! Shall we actually produce a `TlaModule` here instead of lists?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
```diff
@@ -506,22 +510,38 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
...
+    if(tlc) {
+      writer.println(
+        s"""@!@!@STARTMSG 2262:0 @!@!@
+          |Created by Apalache on ${Calendar.getInstance().getTime}
+          |@!@!@ENDMSG 2262 @!@!@""".stripMargin
+      )
```
konnov: I believe that this header should be implemented in TlcCounterexampleWriter. Perhaps, as a separate method, if you like

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
```diff
@@ -506,22 +510,38 @@ class ModelChecker(typeFinder: TypeFinder[CellT],
...
+      tlcWriter.write(notInv, nextStates.head._2, nextStates.tail.toList)
     }
+    else {
+      writer.println("%s MODULE counterexample %s\n".format("-" * 25, "-" * 25))
+      writer.println("EXTENDS %s\n\n".format(checkerInput.rootModule.name))
+      val tlaWriter = new TlaCounterexampleWriter(writer)
+      tlaWriter.write(notInv, nextStates.head._2, nextStates.tail.toList)
+      writer.println("\n%s".format("=" * 80))
+      writer.println("\\* Created %s by Apalache".format(Calendar.getInstance().getTime))
+      writer.println("\\* https://github.com/konnov/apalache")
```
konnov: Same here. If we just passed `TlaModule`, the whole output logic could be hidden in the implementation.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
```diff
@@ -73,10 +73,11 @@ class BoundedCheckerPassImpl @Inject() (val options: PassOptions,
         val random = tuning.getOrElse("search.randomDfs", "")
         new DfsStrategy(input, stepsBound, random.toLowerCase.equals("true"))
       }
+    val tlc = options.getOrElse("checker", "tlc", false)
 
     val checker: Checker =
         new ModelChecker(typeFinder, hintsStore, changeListener, exprGradeStore, sourceStore,
-          input, strategy, tuning, debug, profile, checkRuntime)
+          input, strategy, tuning, debug, profile, checkRuntime, tlc)
```
konnov: if we dumped the counterexamples in all formats, we would not need the extra parameter

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -0,0 +1,102 @@
...
+
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.oper._
+import at.forsyte.apalache.tla.lir.values._
+/**
+ * A printer for counterexamples, in various formats: TLA+ , as TLC output, ...
+ *
+ * @author Andrey Kuprianov
+ */
+abstract class CounterexampleWriter {
```
konnov: I would make this class a trait

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -0,0 +1,102 @@
...
+  // Print out invariant violation
+  def write(notInvariant: NotInvariant, init: State, nextStates: List[NextState] ) : Unit
+
+  def printState(pretty: PrettyWriter, state: State) = {
```
konnov: This method seems to be specialized for the outputs in TLC and Apalache format. It would not be useful for JSONWriter, which we will definitely have in the future.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -0,0 +1,102 @@
...
+      "(* Initial state *)",
+      s"State$i",
+      init)
+    for (nextState <- nextStates) {
+      i += 1
+      printNextState(
+        s"(* Transition ${nextState._1} to State$i *)",
+        s"State$i",
+        nextState._2)
+    }
```
konnov: Alternatively, you could do define `printNextState` to accept a state and an index and then just do`nextStates.zipWithIndex.foreach printNextState`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
konnov: We can fix the names of the declarations for the writers. E.g., `Init` for initialization, `ConstInit` for parameter initialization, `State<no>` for the states in the counterexample, `InvariantViolation` for the invariant negation.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/io/TestCounterexampleWriter.scala
```diff
@@ -0,0 +1,190 @@
...
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.FunSuite
+
+@RunWith(classOf[JUnitRunner])
+class TestCounterexampleWriter extends FunSuite {
+
+  def compareTla(notInvariant: NotInvariant, init: State, nextStates: List[NextState], expected: String): Unit = {
+    val stringWriter = new StringWriter()
+    val printWriter = new PrintWriter(stringWriter)
+    val writer = new TlaCounterexampleWriter(printWriter)
```
konnov: You could extend the test suite with `BeforeAndAfterEach`. Then you could set up the repetitive constructors before each test like this:
```scala

  override def beforeEach() {
    super.beforeEach()
    stringWriter = new StringWriter()
    ...
  }
```

When the number of tests is growing, it is getting harder to maintain (and refactor!) the test initialization. So it is better to do initialization once in one place.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/io/TestCounterexampleWriter.scala
```diff
@@ -0,0 +1,190 @@
...
+    writer.write(notInvariant, init, nextStates)
+    printWriter.flush()
+    assert(stringWriter.toString == expected)
+  }
+
+  def compareTlc(notInvariant: NotInvariant, init: State, nextStates: List[NextState], expected: String): Unit = {
```
konnov: or maybe you don't need it, as you have already extracted initialization here 👍

konnov:
> FWIW: TLC's `TE.tla` extends the spec that was model checked. With [tlaplus/tlaplus#393]

Let's try to be compatible with that solution. We definitely do not need a zoo of counterexamples.

andrey-kuprianov:
@lemmy thanks for pointing to https://github.com/tlaplus/tlaplus/issues/393! I will study what you've done, and try to make our solution compatible.

andrey-kuprianov:
@konnov Thanks for your nice comments and suggestions! I will implement them now, and as for the compatibility with https://github.com/tlaplus/tlaplus/issues/393, I suggest I do it on a separate branch/PR.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
andrey-kuprianov: yes, that's a nice suggestion -- and also removes some code, which is good;)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/ModelChecker.scala
andrey-kuprianov: In that respect, FWIW, I think it's better to keep specialized methods with specific signatures: packing everything into a TlaModule will introduce unnecessary code for packing/unpacking, and won't bring any real value. Let's maybe keep it as is for now, and we'll refactor later if needed

konnov: Well, the value is that the counterexample printers become just code printers. The TLC format is somewhat an exception here, as it contains source information.

andrey-kuprianov:
@konnov I'm done with this branch; could you maybe review the last commit? it is supposed to address all your previous comments... Thanks!

tlair/src/test/scala/at/forsyte/apalache/tla/lir/io/TestCounterexampleWriter.scala
```diff
@@ -0,0 +1,219 @@
...
+        |@!@!@ENDMSG 2262 @!@!@
+        |@!@!@STARTMSG 2110:1 @!@!@
+        |Invariant is violated.
+        |@!@!@ENDMSG 2110 @!@!@
+        |@!@!@STARTMSG 2121:1 @!@!@
+        |The behavior up to this point is:
+        |@!@!@ENDMSG 2121 @!@!@
+        |@!@!@STARTMSG 2217:4 @!@!@
+        |1: <Initial predicate>
+        |x = 0 /\ y = 8
```
konnov: Does not TLC print one variable assignment per line in the counterexample, e.g.,
```tla
/\ x = 0
/\ y = 8
```

andrey-kuprianov: yes, it does, but it doesn't matter -- both forms are valid and accepted by TLA Toolbox, I've checked.

konnov: It might be harder for third-party scripts to parse it though. Think of our work on testing.

andrey-kuprianov:
merged into unstable manually in https://github.com/konnov/apalache/commit/3e67ca4403861c187ff0502639d8ef1cb65264fb

PR #109, opened 2020-03-06
Ak/tlcconfig

andrey-kuprianov:
Integrated TLCConfigParser into the tool architecture


andrey-kuprianov:
@konnov I have added topological sorting of definitions; could you please review the PR? thanks!

tla-import/src/test/scala/at/forsyte/apalache/io/tlc/TestTlcConfigParser.scala
```diff
@@ -122,6 +137,23 @@ class TestTlcConfigParser extends FunSuite {
...
+//        |INIT Init
+//        |INVARIANTS Inv2
+//        |NEXT Next
+//      """.stripMargin
+//
+//    val config = TlcConfigParser(text)
+//    assert(config.behaviorSpec == InitNextSpec("Init", "Next"))
+//    assert(config.invariants == List("Inv1", "Inv2"))
+//  }
+
```
konnov, line 156: you could replace `test(...)` with `ignore(...)`

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -38,7 +43,47 @@ class ConfigurationPassImpl @Inject()(val options: PassOptions,
...
+    var module = tlaModule.get
+
+    // read TLC config if present
+    val filename =
+      options.getOrError("parser", "filename").asInstanceOf[String]
+      .replaceFirst("\\.tla$", "\\.cfg")
+    val basename = FilenameUtils.getName(filename)
+    try {
+      val config = TlcConfigParser.apply(new FileReader(filename))
+      module = new TlcConfigImporter(config, new IdleTracker())(module)
```
konnov, line 55: I don't like that the config file is read automatically if it is present in the file system. Let's add an option `--config <filename>`, similar to TLC. We should always give the user a way to override default options.

konnov:
Once you added the `--config` option, feel free to merge in unstable.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
lemmy: FWIW: TLC reads a config file `A.cfg` automatically if `A.cfg` is present in the same folder that `A.tla`; the spec that the user check.  In other words, if the config file has a matching name, one does not have to pass `--config ...`.

konnov: Wow! I did not know that :-)

Issue #116, opened 2020-03-20
[FEATURE] Counterexamples in the JSON format


konnov:
It should be relatively easy to output counterexamples in the JSON format. By doing so we will allow the other tools to use our counterexamples.

andrey-kuprianov:
Implemented in #117

PR #117, opened 2020-03-26
implements #116: Counterexamples in the JSON format

andrey-kuprianov:


tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -21,14 +21,16 @@ object CounterexampleWriter {
   def writeAllFormats(rootModule: TlaModule, notInvariant: NotInvariant, states: List[NextState] ): List[String] = {
     new TlaCounterexampleWriter(new PrintWriter(new FileWriter("counterexample.tla"))).write(rootModule, notInvariant, states)
     new TlcCounterexampleWriter(new PrintWriter(new FileWriter("MC.out"))).write(rootModule, notInvariant, states)
-    List("counterexample.tla", "MC.out")
+    new JsonCounterexampleWriter(new PrintWriter(new FileWriter("counterexample.json"))).write(rootModule, notInvariant, states)
+    List("counterexample.tla", "MC.out", "counterexample.json")
```
konnov: shall we close the writers in `try...finally`

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -114,3 +116,22 @@ class TlcCounterexampleWriter(writer: PrintWriter) extends TlaCounterexampleWrit
...
+        }
+        TlaOperDecl(s"State${i+1}", List(), OperEx(TlaBoolOper.and, state.toList: _*))
+
+    }
+
+    val mod = new TlaModule(
+      "counterexample",
+      tlaStates ++ List(TlaOperDecl(s"InvariantViolation", List(), notInvariant)))
+    json.write(mod)
+  }
```
konnov: I guess, the next logical step would be to extract the code that constructs `TlaModule`. Why shall we do that in three writers independently?

andrey-kuprianov: right now, I suppose, this is not necessary: it will not bring any additional value, as I would have to add decomposition code from TLA into raw lists in the other two classes (Tla/TlcCounterexampleWriter). JsonCounterexampleWriter only needs to construct TLA, in order to use the already existing JsonWriter. This is not the case for the other two formats.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
andrey-kuprianov: right; implemented in 7b06d47

PR #118, opened 2020-03-27
Cardinality optimizations

konnov:
This PR presents an optimization for `Cardinality(S) >= k`, where `k` is a constant expression.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/CardinalityConstRule.scala
```diff
@@ -0,0 +1,87 @@
...
+    val witnesses = 1.to(threshold).map(pick).toList
+
+    // create an inequality constraint
+    def mkEq(pair: (ArenaCell, ArenaCell)): TlaEx = {
+      nextState = rewriter.lazyEq.cacheOneEqConstraint(nextState, pair._1, pair._2)
+      tla.not(tla.eql(pair._1, pair._2))
+    }
+
+    // create the inequality predicate
+    val cons = (witnesses cross witnesses).filter(p => p._1.id < p._2.id).map(mkEq)
```
andrey-kuprianov: I am not sure about the efficiency, but at least it would make both the code, as well as the SMT encoding smaller, if we used the `distinct` predicate, available in smt-lib and in Z3 solver context (`mkDistinct`)

docs/smt/Cardinality.md
```diff
@@ -0,0 +1,209 @@
...
+
+Essentially, this encoding can be translated into the following formula
+`[ExistsForm]`:
+
+```tla
+\E x_1, ..., x_k \in S:
+  \A y, z \in {x_1, ..., x_k}:
+    y /= z
+```
+
```
andrey-kuprianov: I would add here an explanation about the complexity of this encoding. E.g., as far as I understand, it is linear in the number of set elements, but quadratic in the constant, is it true? Which also raises an interesting question on how this encoding will behave when we compare a small set to a large constant, and whether actually we should choose either one encoding or another depending on the relationship between the number of set cells and the constant.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestExprOptimizer.scala
```diff
@@ -118,11 +119,55 @@ class TestExprOptimizer extends FunSuite with BeforeAndAfterEach {
...
+              tla.letIn(
+                tla.exists(tla.name("t_6"), tla.appOp(tla.name("t_5")),
+                  tla.exists(tla.name("t_7"), tla.appOp(tla.name("t_5")),
+                    tla.eql(tla.name("t_6"), tla.name("t_7")))),
+                TlaOperDecl("t_5", List(), newSet)
+              )))))
+
+    val expected = tla.letIn(letBody, TlaOperDecl("t_8", List(), tla.name("S")))
+    assert(expected == output)
+  }
```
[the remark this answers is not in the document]
konnov: Yes, this test was written for another optimization. But it was producing too many formulas, due to universal quantifiers

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
```diff
@@ -71,40 +71,109 @@ class ExprOptimizer(nameGen: UniqueNameGenerator, tracker: TransformationTracker
...
+          case Nil => underlying
+          case varName :: tl => tla.forall(varName, letApp, mkForAll(underlying, tl))
+        }
+      }
+
+      tla.letIn(
+        mkForAll(letInExistsEq, allNames),
+        TlaOperDecl(setAlias, List(), set)
+      ) ///
+      */
```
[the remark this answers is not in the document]
konnov: Right. I have removed the code for `Cardinality(S) >= k`, but kept the commented code for `Cardinality(S) < k`, as we might want to use it one day.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/CardinalityConstRule.scala
konnov: Nice! I did not know about `distinct`. There might be other places in the code, where `distinct` could be used. Just to note, `mkEq` also creates equalities constraints, if needed.

docs/smt/Cardinality.md
konnov: To be precise, the encoding is linear in the length of the set approximation, which tends to grow faster than the set itself (e.g., `S' = S \ {x}` makes the set `S` smaller, but its approximation has the same size).

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/CardinalityConstRule.scala
konnov: I have implemented this test with `(distinct ...)`. However, our benchmarks have too small thresholds to see the effect.

PR #155, opened 2020-07-02
Suggested tweaks and elaborations to the manual

shonfeder:
This PR collections my suggestions and question as I work through the manual. I'm not quite finished yet, and my changes will benefit from a self-review prior to a final review, but I think there's already lots here that could benefit from feedback. I've tried to break it into logically distinct commits with messages that clearly explain the reasoning behind each change, so it might be worth reviewing by commit.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -284,9 +284,7 @@ become unwieldy.
 
 ## 5.2. Convention over configuration
 
-Alternatively, you can extend the base module and ...
-<!-- TODO(shonfeder): how does the following work exactly? Nothing obvious comes up -->
-<!-- when I search for "TLA+ override" -->
+Alternatively, you can extend the base module and use overrides:
 
 ```tla
 ---------------------------- MODULE y2k_override ----------------------------
@@ -325,9 +323,6 @@ $ apalache check --inv=Safety \
 ```
 
 ### Parameterized initialization
-
-<!-- TODO(shonfeder) How does this subsection relate to the example? how would it -->
-<!-- apply?  -->
 
 As a bonus of this approach, Apalache allows one to check a specification over a
 bounded set of parameters. For example:
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -171,6 +171,7 @@ $ docker image build -t apalache:0.7.0 .
 <a name="building-from-source"></a>
 ## 3.2. Building from source
 
+0. *optionally* install [direnv][] and run `direnv allow`
 1. Install `git`.
 2. Clone the git repository: `git clone https://github.com/informalsystems/apalache.git`
 3. Install OpenJDK8 or
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -154,7 +154,7 @@ Run it with the following command:
 $ docker run --rm -v <your-spec-directory>:/var/apalache apalache/mc:unstable <args>
 ```
 
-To create an alias pointing to `unstable` version:
+To create an alias pointing to the `unstable` version:
 
 ```bash
 $ alias apalache="docker run --rm -v $(pwd):/var/apalache apalache/mc:unstable"
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -623,6 +623,12 @@ Inv == FALSE
 ===============
 ```
 
+Running the checker with
+
+```bash
+apalache check Assignments20200309.tla
+```
+
 Apalache reports an error as follows:
 
 ```console
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -72,13 +72,10 @@ tool.
 # 3. Installation
 
 There are two ways to run Apalache: (1) Download and run a docker image, or (2)
-Build Apalache from sources and run the compiled package. If you just like to
-try the tool, we recommend to use the docker image. If you like to run the tool
-on daily basis or to contribute to the project, we recommend to build the project
-from the sources. In the following, we write `$APALACHE_HOME` to refer to the
-directory, where Apalache is cloned.
-
-<a name="useDocker"></a>   
+Build Apalache from sources and run the compiled package. If you just want to
+try the tool, we recommend using the docker image. If you would like to run the
+tool on a daily basis or [to contribute](../CONTRIBUTING.md) to the project, we
+recommend building the project from the source.
 
 <a name="useDocker"></a>
 ## 3.1. Using a docker image
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -479,7 +479,7 @@ apalache check --cinit=ConstInit --length=20 --inv=Safety y2k_cinit.tla
 ```
 
 This command checks, whether `Safety` can be violated in 20
-specification steps. The consants are initialized with the predicate
+specification steps. The constants are initialized with the predicate
 `ConstInit`, defined in `y2k_cinit.tla` as:
 
 ```tla
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -37,9 +37,10 @@ Apalache is working under the following assumptions:
  1. [An example of a TLA+ specification](#example)
  1. [Setting up specification parameters](#parameters)
  1. [Running the tool](#running)
- 1. [Understanding assignments](#assignments)
- 1. [Type annotations](#types)
- 1. [Recursive operators and functions](#recursion)
+ 1. [Principles of symbolic model checking with Apalache](#principles)
+    1. [Assignments and symbolic transitions](#assignments)
+    1. [Type annotations](#types)
+    1. [Recursive operators and functions](#recursion)
  1. [Five minutes of theory](#theory5)
  1. [Supported language features](#features)
 
@@ -557,8 +558,15 @@ In this case, Apalache performs the following steps:
 1. It pretty-prints the IR into `out-parser.tla`, see [Section 6.3](#detailed).
 
 
+<a name="principles"></a>
+# 7. Principles of symbolic model checking with Apalache
+
+In order to take advantage of Apalache's symbolic model checking, there are a
+few principles one must bear in mind when writing TLA.
+
 <a name="assignments"></a>
-# 7. Understanding assignments
+<a name="symbolicTransitions"></a>
+## 7.1 Assignments and symbolic transitions
 
 Let us go back to the example [`test/tla/y2k.tla`](../test/tla/y2k.tla) and
 run `apalache` against [`test/tla/y2k_override.tla`](../test/tla/y2k_override.tla):
@@ -666,7 +674,7 @@ unfortunately behind the Elsevier paywall, which will be lifted after the
 two-year embargo period.
 
 <a name="types"></a>
-# 8. Type annotations
+## 7.2 Type annotations
 
 **NOTE**: [Jure Kukovec](https://forsyte.at/people/kukovec/) is developing
 a completely automatic type inference engine. As soon as it is ready, type
@@ -677,7 +685,7 @@ Apalache requires two kinds of type annotations:
 - type annotations for empty sets and sequences, and
 - type annotations for records and sets of records.
 
-## 8.1. Empty sets and sequences
+### 7.2.1 Empty sets and sequences
 
 Consider the following example
 [`test/tla/NeedForTypes.tla`](../test/tla/NeedForTypes.tla):
@@ -737,7 +745,7 @@ To help the type checker, we have to introduce a few type annotations. But
 before doing that, we introduce the notation for type annotations in the
 specification.
 
-### Syntax for type annotations
+#### Syntax for type annotations
 
 Apalache reads any expression formed with the `<:` operator as an annotation of
 the value of the left hand side with the type on the right. E.g.,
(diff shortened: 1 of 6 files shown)


shonfeder:
Also:

- This PR adds the start of a CONTRIBUTING.md
- It presupposes #154, and should be rebased on master once that is merged.

andrey-kuprianov:
yes, @shonfeder, thanks for perfect corrections to our imperfect English :)

docs/manual.md
```diff
@@ -258,10 +278,15 @@ INSTANCE y2k WITH BIRTH_YEAR <- 80, LICENSE_AGE <- 18
...
 
 The downside of this approach is that you have to declare the variables of the
-extended specification.
+extended specification. This is easy with only two variables, but can quickly
+become unwieldy.
 
 ## 5.2. Convention over configuration
 
+Alternatively, you can extend the base module and ...
+<!-- TODO(shonfeder): how does the following work exactly? Nothing obvious comes up -->
```
shonfeder, line 287: @konnov or @andrey-kuprianov for my own edification, I'm hoping to get a bit of clarification on this point.

docs/manual.md
```diff
@@ -299,8 +323,13 @@ $ apalache check --inv=Safety \
   --length=20 --cinit=ConstInit y2k_cinit.tla
 ```
 
-**Parameterized initialization**. As a bonus of this approach, Apalache allows
-one to check a specification over a bounded set of parameters.  For example:
+### Parameterized initialization
+
+<!-- TODO(shonfeder) How does this subsection relate to the example? how would it -->
```
shonfeder, line 328: @konnov or @andrey-kuprianov for my own edification, I'm hoping to get a bit of clarification on this point.

docs/manual.md
```diff
@@ -538,48 +586,56 @@ Next$1 == year - 80 >= 18 /\ hasLicense' <- TRUE /\ (year' <- year)
...
    We call these operators **symbolic transitions**.
 
-The pure TLA+ does not have the notions of assignments and symbolic
+Pure TLA+ does not have the notions of assignments and symbolic
 transitions.  However, TLC sometimes treats expressions `x' = e` and `x' \in S`
 as if they were assigning a value to the variable `x'`. TLC does so
 dynamically, during the breadth-first search. Apalache looks statically for assignments
 among the expressions `x' = e` and `x' \in S`.
 
+<!-- TODO(shonfeder): The first stentence of the next passage isn't clear to me.
```
shonfeder, line 601: @konnov or @andrey-kuprianov for my own edification, I'm hoping to get a bit of clarification on this point.

docs/manual.md
```diff
@@ -851,14 +950,17 @@ SendAck ==
...
+As you can see, we have to annotate only those records that are subtypes of
+`MT`: since `ackno` has exactly the type `MT`, Apalache's type inference doesn't
+need any help. As soon as we have added the annotations, the type checker stops
+complaining and lets the model checker run. The annotated code can be found in
 [`test/tla/HandshakeWithTypes.tla`](../test/tla/HandshakeWithTypes.tla).
 
-Type annotations can be also applied to sets of records. For example:
+Type annotations can also be applied to sets of records. For example:
 
+<!-- TODO(shonfeder): I read this as saying that the record type on the lhs is a
```
shonfeder, line 961: @konnov or @andrey-kuprianov for my own edification, I'm hoping to get a bit of clarification on this point.

docs/manual.md
```diff
@@ -813,14 +885,41 @@ In our example, the set `msgs` may contain records of three kinds:
...
 
 From the perspective of the type checker, the three records shown above have
-three different types. Although we would love to reject this example as an
-ill-typed one, mixing records of different types is a widely-accepted idiom in
-TLA+, for instance, see [Lamport's specification of
+three different types. Mixing records of different types is a widely-accepted
+idiom in TLA+, for instance, see [Lamport's specification of
 Paxos](https://github.com/tlaplus/Examples/blob/master/specifications/Paxos/Paxos.tla).
-Think of records as of C unions, rather than C structs!
+Fortunately, there is a well established precedent for a subtyping relationship
```
shonfeder, line 891: @konnov or @andrey-kuprianov This is one of the two places where I suggested more substantive elaboration in the manual, I just wanted to call it out to be sure its scrutinized appropriately, since I'm just starting to figure out stuff here and so it's easy for me to misunderstand or make mistakes :)

docs/manual.md
```diff
@@ -663,27 +718,44 @@ The outcome is: Error                                             I@15:17:14.388
...
 In a somewhat obfuscated way, Apalache tells us the following. It has inferred
- that `Left` is a set of integers, that is, `FinSet[Int]`. First, it found that
+that `Left` is a set of integers, that is, `FinSet[Int]`. First, it found that
 `InSet` is a set of integers, by applying `ConstInit`. Second, as `Left = InSet`
 in `Init`, it inferred that `Left` is also a set of integers. Third, when
 applying `Next`, it processed `{}`, which is an empty set of any kind of
-objects. Hence, `{}` was assigned the type `FinSet[Unknown]`, that is, a set
-of some type. Finally, it found the expression `Left = {}`,
-and here the type checker has failed.
+objects. Hence, `{}` was assigned the type `FinSet[Unknown]`, that is, a set of
```
shonfeder, line 725: @konnov or @andrey-kuprianov This is the other place where I've suggested substantive changes. Does it seem accurate? Is the suggested change to using `v <: T == v` actually useful, or is it more confusing, since so much of the example code uses `a <: b == a`? (Note that I could change the examples to use the suggested convention, if we thought it enough of an improvement to be worth while).

docs/manual.md
```diff
@@ -538,48 +586,56 @@ Next$1 == year - 80 >= 18 /\ hasLicense' <- TRUE /\ (year' <- year)
...
+     It doesn't seem to be exemplified in the example code at all, and no further
+     example of how this is provided. -->
 Moreover, Apalache splits action operators `Init` and `Next` into disjunctions
 (e.g., `A_1 \/ ... \/ A_k`). The main contract between the assignments and
 symbolic transitions is as follows:
 
-> For every variable `x` declared with `VARIABLE`, there is exactly one assignment
-of the form `x' <- e` in every symbolic transition `A_j`.
+> For every variable `x` declared with `VARIABLE`, there is exactly one
+> assignment of the form `x' <- e` in every symbolic transition `T$n`.
```
shonfeder, line 609: Now that I'm down to reading the theory section, I think this change is incorrect. IIUC, we *are* using `A_i` to represent the `i`th step in a symbolic transition, which happens to have the concrete syntax `A$i`. I'll change this back, and instead add a note to the preceding paragraph which informs the reader that we'll use `A_i` in this way.

docs/manual.md
shonfeder: Now that I'm reading the theory, I think I understand this part now. It is a restatement of the second point from above,  where we currently have "it has factored the operator...", right? What "factoring the operator into n operators" consists in is replacing the single operator `Op` into a disjunction of `n` operators `Op$0 \/ Op$1 \/ ... Op$n`. But what that looks like in terms of concrete syntax is just

```
Op$0 == ...

Op$1 == ...

...

Op$n == ...
```

Is this the correct understanding?

shonfeder: I reworked my suggestion here, following the reasoning in this thread.

docs/manual.md
```diff
@@ -104,27 +115,28 @@ The following docker parameters are used:
...
 substantial number of improvements in the unstable branch before the next
-stable release biweekly.  Please refer to [Unstable
-Changes](https://github.com/informalsystems/apalache/blob/unstable/CHANGES.md) and
-[Unstable
-Manual](https://github.com/informalsystems/apalache/blob/unstable/docs/manual.md) for
-the description of new features.  **We recommend using the unstable version if
-you want to try all the exciting new features of Apalache. But be warned: It is
-called "unstable" for a reason**. To use `unstable`, just type
-`apalache/mc:unstable` instead of `apalache/mc` everywhere.
+biweekly stable release. Please refer to the [change
```
andrey-kuprianov, line 132: I would remove "biweekly" completely. I don't think we do have any schedule: sometimes the changes are very fast, sometimes we are distracted by other projects, and there may be no changes for the whole month

docs/manual.md
andrey-kuprianov: There is a special convention, that if smth with OVERRIDE_<x> prefix is found, it overrides the <x> declaration. Just a hack, actually, but quite easy to use in some circumstances

docs/manual.md
andrey-kuprianov: It's that you can initialize not with a single value, but with a constraint on initial values. Compared to TLC, this gives you more flexibility

docs/manual.md
```diff
@@ -442,41 +476,57 @@ a single specification step satisfies the invariant, when starting
...
+### Using a constant initializer:
 
 ```bash
 $ cd test/tla
 apalache check --cinit=ConstInit --length=20 --inv=Safety y2k_cinit.tla
 ```
 
 This command checks, whether `Safety` can be violated in 20
-specification steps. The consants are initialized with the predicate `ConstInit`:
+specification steps. The consants are initialized with the predicate
```
andrey-kuprianov, line 487: should be "The constants"

docs/manual.md
andrey-kuprianov: yes, you changes seem very reasonable to me!

docs/manual.md
andrey-kuprianov: This part I would actually postpone, and leave it as is for now. We can discuss this later, after the 0.7.0 release. My concern is that this brings a bit too much of theory, and might scare off some practice-oriented folks.

docs/manual.md
andrey-kuprianov: Let's discuss this later with @konnov ;)

docs/manual.md
shonfeder: Ah, I figured! So this is a hole in my TLA+ knowledge :) I looked for this in the TLA manual but didn't turn anything up. Thanks for explaining!

docs/manual.md
shonfeder: I understand now, enough to remove my question but not quite enough to add any additional clarification. I'll revisit this bit once I have more experience. Thanks :)

docs/manual.md
shonfeder: Cool! This makes sense. If it is useful, perhaps it's better suited  for the document on types. I still think the comparison to C unions is a bit misleading, but perhaps this will make more sense to me after more experience as well :)

docs/manual.md
konnov: It is not a standard TLA+ construct, but a convenient syntactic convention in Apalache. Nobody but me seems to like it though :-)

docs/manual.md
konnov: There are two things about the assignment finder. First, it finds expressions that can be used as assignments. Having a complete set of assignments, the assignment finder decomposes transition predicates, e.g., `Init` and `Next` into disjunctions, as you pointed out.

docs/manual.md
shonfeder: Ah ha! We may want to document this a bit more clearly then, unless it's covered somewhere else. I'll make an issue to do that in a followup. Thanks!

docs/manual.md
shonfeder: Sounds good. I'll make a personal note here and ask at some future time.

konnov: Hmmm. How can I see this comment in a diff?

docs/manual.md
konnov: It looks fine. Hopefully, this kind of type annotations goes away very soon :-)

shonfeder:
Changes addressed!

docs/manual.md
shonfeder: Oh, sorry, I dropped it from the PR already, you can see the comment by looking at https://github.com/informalsystems/apalache/pull/155/commits/e956b32a736604b456ac460d2bb2f6807a5fb3c3#diff-4b7e43801d0bf529f907fe63667db023R922

konnov:
@andrey-kuprianov, are you happy with the PR? If so, @shonfeder feel free to merge.

docs/manual.md
shonfeder: To elaborate a bit: 

This doesn't make sense to me

```tla
[syn |-> BOOLEAN, ack |-> BOOLEAN, seqno |-> Int] <: {MT}
```

For the same reason this doesn't make sense to me:

```tla
3 <: {Nat}
```

Since we earlier define the type alias

```tla
 MT == [syn |-> BOOLEAN, ack |-> BOOLEAN, seqno |-> Int, ackno |-> Int]
```

I'd think it should be

```tla
{[syn |-> BOOLEAN, ack |-> BOOLEAN, seqno |-> Int]} <: {MT}
```

I.e., the set containing a record that is a (sub)type of `MT` is has the type of a set of `MT`. But I think I may be missing something about the notation here.

PR #167, opened 2020-07-13
Remove CI branch restrictions

### LINKED ISSUE #165: [BUG] CI is not running on PRs or the unstable branch
I see that we purposefully stopped running on PRs in 7cae97cae5cfe9cecfb97e48504375a5bf7f7a4b, but in 118e55e2c891781eadc9c3c36dbd9f38c45a2de5 we added `unstable`, and I'm not sure why that's not building.

I also suggest we begin running CI on all PRs, and requiring they pass before merging. If there are good reasons we're not doing that now, let's document them, so we can fix those blockers and get the CI in a good state.

shonfeder:
Closes #165

I'll likely need to do some troubleshooting here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,11 +36,6 @@ before_install:
 #- sudo pip install --user release-me
 #- ./script/release --overwrite latest-unstable ./script/latest-message.txt
 
-branches:
-  only:
-  - master
-  - unstable
-
 script:
     - make integration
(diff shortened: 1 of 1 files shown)


shonfeder:
This looks to be working now.

shonfeder:
Was altered via commit directly to unstable.

PR #184, opened 2020-07-28
Add documentation on configuring Emacs for hacking

shonfeder:
This documents my current setup for writing Scala in Emacs. I think it's
worth having this kind of information documented for potential
contirbutors, and it can help lower the bar to entry significantly.


konnov:
Feel free to merge that whenever it is complete

PR #185, opened 2020-07-28
Close #153: make JSON TLA+ encoding more suitable for JsonPath queries

### LINKED ISSUE #153: [FEATURE] Change the JSON encoding for a seamless integration with JsonPath expressions
The current TLA+ JSON encoding works fine on itlself, but while trying to apply JsonPath expressions used in [Jsonatr](https://github.com/informalsystems/jsonatr), e.g. in [this spec](https://github.com/informalsystems/jsonatr/blob/master/tests/support/counterexample_to_test.json), it turned out that JsonPath expression do not work when special symbols ('+', '=', etc.) are used as object keys -- and exactly this we do in our current JSON encoding, to shorten it a bit.

As the JSON encoding is anyway used for automated processing, we need to change it slightly, to have a more straightforward structure without using special symbols as keys, albeit a bit inflated in size. The proposed encoding should look like [this file](https://github.com/informalsystems/jsonatr/blob/master/tests/support/counterexample.json).

andrey-kuprianov:
* get rid of special symbols in object keys
* use objects instead of arrays for binary operators
* use key-value objects instead of arrays for pairs of expressions

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 ## Unreleased #unstable
+ * Simplify JSON format to make it suitable for JsonPath queries, see #153
 
 ## 0.7.0 [RELEASE]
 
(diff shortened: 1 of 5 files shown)


andrey-kuprianov:
Not yet; but now we have the issue for that: #186

Issue #183, opened 2020-07-22
[FEATURE] Standard module for Apalache


konnov:
Following #181 and #157, we should finally introduce a standard module for Apalache that contains all handy operators, which until today have been introduced only by the Apalache transformations. We should expose these operators to the users, so they should be able to fix issues by hand. We have to expose all operators in [BmcOper](https://github.com/informalsystems/apalache/blob/unstable/tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/BmcOper.scala) in `Apalache.tla`, except for probably `<:`, which should be introduced in its own module `Types.tla`.

@lemmy, is there any standard practice for distributing non-SANY modules, or shall we just expect the user to download `Apalache.tla` from a known location and keep it up to date?

lemmy:
"Standard Modules" are those that appear in Specifying Systems.  For other modules that have broad applicability, I've created https://github.com/tlaplus/CommunityModules/

konnov:
Cool, I have opened an [issue](https://github.com/tlaplus/CommunityModules/issues/19) there. I think it is more of an issue for the SANY parser, but it is also related to the community libraries.

Issue #181, opened 2020-07-20
[FEATURE] Assignment annotations


konnov:
Sometimes the assignment finder fails to find assignments because it is buggy: #157

Apart from fixing the bug in #157, we should introduce assignment annotations and let the assignment solver consume them. This will give the users a way to work around bugs. Moreover, some users may find it more convenient to label some expressions as assignments.

konnov:
See the Apalche operators in https://github.com/informalsystems/apalache/blob/unstable/src/tla/Apalache.tla

konnov:
@Kukovec when shall we expect the PR?

PR #188, opened 2020-07-30
Apalache.tla: the standard module for Apalache

konnov:
Following #183, this PR introduces the standard module `Apalache.tla` that gives the user access to (almost all of) internal operators introduced by Apalache. Most importantly, it introduces the assignment operator `x := e`. This opens the way to deal with #181.

### DIFF
--- a/bin/apalache-mc
+++ b/bin/apalache-mc
@@ -32,18 +32,23 @@ case `cat $DISTS | wc -l | sed 's/[[:space:]]*//g'` in
 esac
 rm -f "$DISTS"
 
+# an additional lookup list for the TLA+ modules
+TLA_LIB="$DIR/src/tla"
+
+# This is a temporary solution for handling TLA_PATH via a Java variable.
+# See apalache/#187.
+# Once tlaplus/#490,#493 are released, remove this workaround.
+test -z "$TLA_PATH" || TLA_LIB="$TLA_LIB:$TLA_PATH"
+
 # set LD_LIBRARY_PATH to find z3 libraries
 export LD_LIBRARY_PATH="$DIR/3rdparty/lib:${LD_LIBRARY_PATH}"
 # 1. The maximum heap size, Z3 will use much more as a native library.
 # 2. Duplicating LD_LIBRARY_PATH, tell JVM to find the third-party libraries.
-JVM_ARGS="-Xmx4096m -Djava.library.path=$DIR/3rdparty/lib"
+JVM_ARGS="-Xmx4096m -Djava.library.path=$DIR/3rdparty/lib "
 # uncomment to track memory usage with: jcmd <pid> VM.native_memory summary
 #JVM_ARGS="${JVM_ARGS} -XX:NativeMemoryTracking=summary"
 
-# This is a temporary solution for handling TLA_PATH via a Java variable.
-# See apalache/#187.
-# Once tlaplus/#490,#493 are release, remove this workaround.
-test -z "$TLA_PATH" || JVM_ARGS="$JVM_ARGS -DTLA-Library=$TLA_PATH"
+JVM_ARGS="$JVM_ARGS -DTLA-Library=$TLA_LIB"
 
 echo "# JVM args: $JVM_ARGS"
 echo "#"
--- a/bin/apalache-mc
+++ b/bin/apalache-mc
@@ -38,7 +38,7 @@ TLA_LIB="$DIR/src/tla"
 # This is a temporary solution for handling TLA_PATH via a Java variable.
 # See apalache/#187.
 # Once tlaplus/#490,#493 are released, remove this workaround.
-test -z "$TLA_PATH" || TLA_LIB="$TLA_LIB:$TLA_PATH"
+(test -z "$TLA_PATH" && TLA_PATH="$TLA_LIB") || TLA_PATH="TLA_LIB:$TLA_PATH" 
 
 # set LD_LIBRARY_PATH to find z3 libraries
 export LD_LIBRARY_PATH="$DIR/3rdparty/lib:${LD_LIBRARY_PATH}"
@@ -67,4 +67,3 @@ trap sigterm SIGTERM SIGINT
 java $JVM_ARGS -jar "$JAR" "$@" &
 child=$!
 wait "$child"
-
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -1158,7 +1158,7 @@ operator.
 Similar to the `TLC` module, we provide the module called `Apalache`, which can
 be found in
 [src/tla](https://github.com/informalsystems/apalache/tree/unstable/src/tla).
-Most of the operators in that modules are introduce internally by Apalache,
+Most of the operators in that modules are introduced internally by Apalache,
 when it is rewriting a TLA+ specification.  It is useful to read the comments
 to the operators defined in `Apalache.tla`, as they will help you in
 understanding the [detailed output](#detailed) produced by the tool, see.
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -41,6 +41,7 @@ Apalache is working under the following assumptions:
     1. [Assignments and symbolic transitions](#assignments)
     1. [Type annotations](#types)
     1. [Recursive operators and functions](#recursion)
+ 1. [The Apalache module](#apalacheMod)
  1. [Five minutes of theory](#theory5)
  1. [Supported language features](#features)
 
@@ -611,16 +612,16 @@ VARIABLE hasLicense
 ASSUME(80 \in 0 .. 99)
 ASSUME(18 \in 1 .. 99)
 
-Init$0 == year' <- 80 /\ hasLicense' <- FALSE
-Next$0 == year' <- ((year + 1) % 100) /\ (hasLicense' <- hasLicense)
-Next$1 == year - 80 >= 18 /\ hasLicense' <- TRUE /\ (year' <- year)
+Init$0 == year' := 80 /\ hasLicense' := FALSE
+Next$0 == year' := ((year + 1) % 100) /\ (hasLicense' := hasLicense)
+Next$1 == year - 80 >= 18 /\ hasLicense' := TRUE /\ (year' := year)
 ===============
 ```
 
 As you can see, the model checker did two things:
 
-1. It has translated several expressions that look like `x' = e` into `x' <- e`.
-   For instance, you can see `year' <- 80` and `hasLicense' <- FALSE` in
+1. It has translated several expressions that look like `x' = e` into `x' := e`.
+   For instance, you can see `year' := 80` and `hasLicense' := FALSE` in
    `Init$0`. We call these expressions **assignments**.
 1. It has factored the operator `Next` into two operators `Next$0` and `Next$1`.
    We call these operators **symbolic transitions**.
@@ -646,7 +647,7 @@ The main contract between the assignments and symbolic transitions is as
 follows:
 
 > For every variable `x` declared with `VARIABLE`, there is exactly one
-> assignment of the form `x' <- e` in every symbolic transition `A_n`.
+> assignment of the form `x' := e` in every symbolic transition `A_n`.
 
 If Apalache cannot find expressions with the above properties, it fails.
 Consider the example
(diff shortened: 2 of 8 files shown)


docs/manual.md
```diff
@@ -505,8 +506,38 @@ State14 ==
...
+`tla2sany.StandardModules`, which is usually provided by `tla2tools.jar` that is
+included in the Java classpath.
+
+Additionally, Apalache includes the directory `$APALACHE_HOME/src/tla` into the
+SANY lookup list via the Java system variable called `TLA-Library`. (The script
+`bin/apalache-mc` automatically sets `APALACHE_HOME` to its parent directory.)
+
+On top of that, Apalache adds all the directories specified in the environment
+variable `TLA_PATH` into the SANY lookup list. See [issue
+187](https://github.com/informalsystems/apalache/issues/187).
```
shonfeder, line 525: This reads to me a bit more like implementor-focused documentation. We might get something a little more friendly for end users if we say something like:

```suggestion
In addition to the modules in the current working directory, Appalache provides

- a small standard library (located in `$APALACHE_HOME/src/tla`), and
- support for additional source directories specified in the environment variable `TLA_PATH`. `TLA_PATH` should be a list of paths to directories separated by `:`. 

(Directories in the `TLA_PATH` are provided to SANY via the `TLA-Library` Java system variable.)    
```

bin/apalache-mc
```diff
@@ -32,18 +32,23 @@ case `cat $DISTS | wc -l | sed 's/[[:space:]]*//g'` in
 esac
 rm -f "$DISTS"
 
+# an additional lookup list for the TLA+ modules
+TLA_LIB="$DIR/src/tla"
+
+# This is a temporary solution for handling TLA_PATH via a Java variable.
+# See apalache/#187.
+# Once tlaplus/#490,#493 are released, remove this workaround.
+test -z "$TLA_PATH" || TLA_LIB="$TLA_LIB:$TLA_PATH"
```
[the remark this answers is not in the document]
shonfeder: However, I see that since we call the system variable `TLA-LIbrary`, there is an argument for the present naming. But, given the documentation, it seems rather that `TLA-Library` should maybe be renamed to `TLA-Path` :).

Now that I see that, I feel even less strongly about the naming here, so please ignore if you're not convinced.

docs/manual.md
```diff
@@ -1122,8 +1153,27 @@ cardinality does not require `2^|NUMS|` constraints, when using a recursive
 operator.
 
 
+<a name="apalacheMod"></a>
+# 9. The Apalache module
+
+Similar to the `TLC` module, we provide the module called `Apalache`, which can
+be found in
+[src/tla](https://github.com/informalsystems/apalache/tree/unstable/src/tla).
+Most of the operators in that modules are introduce internally by Apalache,
```
shonfeder, line 1162:
```suggestion
Most of the operators in that modules are introduced internally by Apalache,
```

docs/manual.md
```diff
@@ -1122,8 +1153,27 @@ cardinality does not require `2^|NUMS|` constraints, when using a recursive
...
+when it is rewriting a TLA+ specification.  It is useful to read the comments
+to the operators defined in `Apalache.tla`, as they will help you in
+understanding the [detailed output](#detailed) produced by the tool, see.
```
shonfeder, line 1165: Was "see" here meant to be followed to a reference to something?

src/tla/Apalache.tla
```diff
@@ -0,0 +1,63 @@
...
+(* one would simply write x = e, or x \in {e}.                               *)
+(*                                                                           *)
+(* Apalache automatically converts some expressions of the form              *)
+(* x = e or x \in {e} into assignments. However, if you like to annotate     *)
+(* assignments by hand, you can use this operator.                           *)
+(*                                                                           *)
+(* For a further discussion on that matter, see:                             *)
+(* https://github.com/informalsystems/apalache/blob/ik/idiomatic-tla/docs/idiomatic/assignments.md *)
+(*****************************************************************************)
+x := e == x = e
```
shonfeder, line 29: Should we include the `<:` operator?

src/tla/Apalache.tla
```diff
@@ -0,0 +1,63 @@
...
+(* Although TLA+ does not have a concept of a variable assignment, we find   *)
+(* this concept extremely useful for symbolic model checking. In pure TLA+,  *)
+(* one would simply write x = e, or x \in {e}.                               *)
+(*                                                                           *)
+(* Apalache automatically converts some expressions of the form              *)
+(* x = e or x \in {e} into assignments. However, if you like to annotate     *)
+(* assignments by hand, you can use this operator.                           *)
+(*                                                                           *)
+(* For a further discussion on that matter, see:                             *)
+(* https://github.com/informalsystems/apalache/blob/ik/idiomatic-tla/docs/idiomatic/assignments.md *)
```
shonfeder, line 27: This URL looks unstable due to the `ik` branch.

src/tla/Apalache.tla
```diff
@@ -0,0 +1,63 @@
...
+(*****************************************************************************)
+x := e == x = e
+
+(*****************************************************************************)
+(* Annotating an expression \E x \in S: P as Skolemizable. That is, it can   *)
```
shonfeder, line 32: This just sounds better to my hear. Not even sure why :) Feel free to disregard.

```suggestion
(* Annotates an expression \E x \in S: P as Skolemizable. That is, it can   *)
```

src/tla/Apalache.tla
```diff
@@ -0,0 +1,63 @@
...
+(* places this hint by static analysis.                                      *)
+(*****************************************************************************)
+ConstCardinality(cardExpr) == cardExpr
+
+(****************************************************************************)
+(* Apalache internally defines the operator Distinct that requires all its  *)
+(* arguments to be distinct from each other. We do not know how to write    *)
+(* such an operator in a TLA+ module.                                       *)
+(****************************************************************************)
+(*
```
shonfeder, line 59: Hmm, so this is just a piece of documentation for the user, right? And not actually an operator they could use, thus it's commented out. I wonder if would make sense then to just move the operator head into the comment box as part of the documentation?

I don't feel strongly about this, but am generally wary of commented out source code.

tla-import/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporter.scala
```diff
@@ -1957,45 +1962,52 @@ class TestSanyImporter extends FunSuite {
...
+        |CC == ConstCardinality(Cardinality(S) >= 2)
         |================================
-        |""".stripMargin
+      """.stripMargin
 
-    val (rootName, modules) = new SanyImporter().loadFromSource("bmc", Source.fromString(text))
-    assert(2 == modules.size) // our module
-    // the root module and naturals
-    val root = modules(rootName)
+    // We have to set TLA-Library, in order to look up for Apalache.tla. This is done automatically in pom.xml.
```
shonfeder, line 1978:
```suggestion
    // We have to set TLA-Library, in order to look up Apalache.tla. This is done automatically in pom.xml.
```

tla-import/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporter.scala
```diff
@@ -1957,45 +1962,52 @@ class TestSanyImporter extends FunSuite {
...
-        _.name == expectedName
-      } match {
-        case Some(d: TlaOperDecl) =>
-          assert(expectedName == d.name)
-          assert(0 == d.formalParams.length)
-          assert(body == d.body)
+    val locationStore = new SourceStore
+    val (rootName, modules) = new SanyImporter(locationStore)
+      .loadFromSource("root", Source.fromString(text))
+    assert(6 == modules.size) // root, Apalache, Integers, FiniteSets, and whatever they import
```
shonfeder, line 1988: This test seems a bit brittle to me. iiuc, it requires there to be 6 modules to be imported (including transitive dependencies?). Do we actually want this to fail any time one of these modules adds a new dependency? If not, perhaps we should check for the presence of the modules that we know should be included?

shonfeder: I do see this is how we do it through the tests here, so probably better to leave just for consistency.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/BmcOper.scala
```diff
@@ -1,9 +1,10 @@
 package at.forsyte.apalache.tla.lir.oper
 
 /**
-  * The operators defined in the BMC module. This module give the users a facility to provide hints.
-  * To use this module, the user would have to copy the module somewhere. As our tool is not widely used,
-  * we just hijack some operators during the import stage, e.g., <:
+  * The operators defined in the module Apalache.tla. This module gives the users a facility to provide hints.
+  * The module Apalache is automatically looked up when Apalache is running.
+  *
+  * TODO: rename this class to ApalacheOper, once ik/multicore is merged into unstable.
```
shonfeder, line 7: Do we tend to open issues to track todos? I'm also happy with a "no" answer here. `grep TODO` is a fine solution :)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/BmcOper.scala
```diff
@@ -1,9 +1,10 @@
 package at.forsyte.apalache.tla.lir.oper
 
 /**
-  * The operators defined in the BMC module. This module give the users a facility to provide hints.
-  * To use this module, the user would have to copy the module somewhere. As our tool is not widely used,
-  * we just hijack some operators during the import stage, e.g., <:
+  * The operators defined in the module Apalache.tla. This module gives the users a facility to provide hints.
+  * The module Apalache is automatically looked up when Apalache is running.
```
shonfeder, line 5: Just to try to disambiguate the two tokens a bit:

```suggestion
  * The "Apalache" module is automatically looked up when Apalache is running.
```

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/BmcOper.scala
```diff
@@ -64,7 +65,9 @@ object BmcOper {
 
   /**
     * The distinct operator that is equivalent to (distinct ...) in SMT-LIB.
-    * Formally, BMC!Distinct(x_1, ..., x_n) is equivalent to \A i, j \in 1..n: i /= j => x_i /= x_j
+    * Formally, BMC!Distinct(x_1, ..., x_n) is equivalent to \A i, j \in 1..n: i /= j => x_i /= x_j.
+    *
+    * XXX: there seems to be no way of defining a user-defined variadic operator in Apalache.tla.
```
shonfeder, line 70: is this a feature we might upstream? Since you specify `user-defined` it makes it sound like this might be feasible in the model checker and AST, but just isn't supported at the level of parsing?

docs/manual.md
```diff
@@ -505,8 +506,38 @@ State14 ==
 InvariantViolation == hasLicense /\ year - BIRTH_YEAR < LICENSE_AGE
 ```
 
+<a name="lookup"></a>
+## 6.3. Module lookup
+
+Apalache uses [the SANY
+parser](https://lamport.azurewebsites.net/tla/tools.html), which is the
+standard parser of TLC and TLA+ Toolbox. By default, SANY is looking for the
```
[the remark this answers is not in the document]
konnov: Not really. TLC and TLA+ Toolbox are two different programs, although they are compiled in one megaproject

bin/apalache-mc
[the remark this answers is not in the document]
konnov: Yeah, `TLA-Library` is a feature specific to SANY: https://github.com/tlaplus/tlaplus/pull/493
`TLA_PATH` is a very new feature that hopefully will propagate in tla2tools soon.

tla-import/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporter.scala
konnov: Good point. However, the standard TLA+ modules rarely change. So it may be good if the tests fail with a new version of `tla2tools`. This will at least indicate a structural change.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/BmcOper.scala
konnov: In this case, it is a convenience thing. Nothing really depends on that name, it is just an inconsistency :)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/BmcOper.scala
konnov: It does not seem to be supported by the TLA+ parser. Apparently, variadic operators are like a special form in TLA+. They are like other operators, but not exactly.

konnov:
Thanks for the comments, @shonfeder!

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/BmcOper.scala
shonfeder: Thanks!

docs/manual.md
shonfeder: @konnov I think this question may have gotten hidden by Github eliding some of the comments.

src/tla/Apalache.tla
shonfeder: This one was also hidden. Is it OK that we may end up with a broken link here? Or is that branch permanent?

shonfeder:
Thanks for the replies!

konnov:
I will merge this pull request

PR #190, opened 2020-08-01
Update dependencies

shonfeder:
This PR updates most of our dependencies, as well as the maven-compiler and scala-maven plugins.

See c101e3a for details on the dependency upgrades.

This also adds two new phony make targets, optimized for quicker build/compile times.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ QUICK_MAVEN_OPTS := "-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -Xverify:non
 # - run up to 4 threads per core (4C): https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3
 QUICK_MAVEN_ARGS := -DskipTests -Dscoverage.skip=true -T 4C
 
-.PHONY: all, apalache, build, test, integration, clean
+.PHONY: all, apalache, compile, build, test, integration, clean
 
 all: apalache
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -4,14 +4,33 @@ DEPDIR=3rdparty
...
 	# tell maven to load the binary libraries and build the package
 	$(ENV) mvn package
 
+# Just compile with quick settings
+compile: $(DEPS)
+	MAVEN_OPTS=$(QUICK_MAVEN_OPTS) mvn $(QUICK_MAVEN_ARGS) compile
+
+# Build with quick settings, but and skip the tests
+build: $(DEPS)
+	MAVEN_OPTS=$(QUICK_MAVEN_OPTS) mvn $(QUICK_MAVEN_ARGS) package
```
konnov, line 32: shall we call it `quick_build`?

pom.xml
```diff
@@ -85,19 +81,42 @@
...
+                <groupId>org.slf4j</groupId>
+                <artifactId>slf4j-api</artifactId>
+                <version>1.7.30</version>
             </dependency>
 
+            <!-- <dependency> -->
+            <!--     <groupId>org.slf4j</groupId> -->
+            <!--     <artifactId>slf4j-simple</artifactId> -->
+            <!--     <version>1.7.30</version> -->
+            <!-- </dependency> -->
```
konnov, line 113: why do we replace `slf4j-simple` with `slf4j-api`? I guess, it deserves a comment.

shonfeder: Oops, my mistake here! I shouldn't have left this commented out. Thanks for the catch. So, iiuc, we are using logback as our logging backend, not slf4j-simple. Including both `simple` and logback backends in dependencies trigger a runtime error fro self4j. I referred to the error and the SO answer explaining the needed configuration on https://github.com/informalsystems/apalache/pull/190/files#diff-600376dffeb79835ede4a0b285078036R88-R90

With this commented out dependency removed, would you still like additional documentation in the comments?

shonfeder:
The CI failure here is just because I merged the PR and deleted it after the CI got started but before it could pull the branch. CI wasn't necessary since the final commit just renamed a makefile target.

PR #195, opened 2020-08-04
Make Dockerfile staged

shonfeder:
This builds the vendored dependencies and apalache itself in two
separate stages (as per https://docs.docker.com/develop/develop-images/multistage-build/). 
This lets us reuse previously built images for
z3 and box, so long as the content of the ./3rdparty direcotry
hasn't been changed. In those cases, this reduces the time for building an 
image for our code considerably.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
 FROM maven:3.6.3-jdk-8-slim AS thirdparty
+# This is image for building vendored dependencies
+#
 # We build the vendored dependencies in their own image so we only need to
 # rebuild this image when those dependencies change, and so we don't need to
 # keep their build dependencies.
@@ -19,7 +21,9 @@ ENV _JAVA_OPTIONS="-Djdk.net.URLClassPath.disableClassPathURLCheck=true"
 
 RUN 3rdparty/install-local.sh --nocache
 
-FROM maven:3.6.3-jdk-8-slim
+FROM maven:3.6.3-jdk-8-slim AS builder
+
+# This is the image for building apalache itself
 
 ADD . /opt/apalache/
 WORKDIR /opt/apalache/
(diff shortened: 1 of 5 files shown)


shonfeder:
Thanks!

This also has the benefit of reducing the size of the image:

```
apalache/mc                                       latest       333ec0458ef9        8 hours ago         1.43GB
apalache/mc                                       staged              d00960c6582a        10 hours ago        1.14GB
```

However, last night I realized we might get a better reduction if we add a final stage to just pull the jars out into a minimal image. I'll give that a go here before merging this in.

Dockerfile
```diff
@@ -7,25 +12,87 @@ RUN apt-get update && apt-get install -y wget \
     git \
     python
 
-ADD . /opt/apalache/
+ADD ./3rdparty /opt/apalache/3rdparty
 WORKDIR /opt/apalache/
 
 # Workaround for Surefire not finding ForkedBooter
 # (see https://stackoverflow.com/questions/53010200/maven-surefire-could-not-find-forkedbooter-class)
 ENV _JAVA_OPTIONS="-Djdk.net.URLClassPath.disableClassPathURLCheck=true"
```
shonfeder, line 20: @konnov or @andrey-kuprianov do you know if this is needed in theory for building our 3rd party deps?

Dockerfile
```diff
@@ -7,25 +12,87 @@ RUN apt-get update && apt-get install -y wget \
...
+# Vendored binaries
+COPY --from=thirdparty /opt/apalache/3rdparty/bin /opt/apalache/3rdparty/bin
+# Vendored libraies
+COPY --from=thirdparty /opt/apalache/3rdparty/lib /opt/apalache/3rdparty/lib
+# The maven repository
+COPY --from=thirdparty /root/.m2 /root/.m2
+
+# Workaround for Surefire not finding ForkedBooter
+# (see https://stackoverflow.com/questions/53010200/maven-surefire-could-not-find-forkedbooter-class)
+ENV _JAVA_OPTIONS="-Djdk.net.URLClassPath.disableClassPathURLCheck=true"
```
shonfeder, line 40: Similar question, @konnov or @andrey-kuprianov: Are these runtime dependencies or something we need here in the build stage?

Dockerfile
```diff
@@ -7,25 +12,87 @@ RUN apt-get update && apt-get install -y wget \
...
+#
+# To prepare it, we
+#
+# - copy over all the artifacts needed to run
+# - prepare the enviroment
+# - declare the entrypoint
+
+WORKDIR /opt/apalache/
+
+# Vendored binaries
```
shonfeder, line 59: Having all these copies isn't pretty, but it's a symptom of us not collecting all our build artifacts under one directory and necessary in the current state, afaict. I'm Not sure if that is idiomatic Java/Scala practice, or just a quirk of our project. But it's something we can address fairly easily in followups if we find it worth while.

Dockerfile
konnov: The installation script `3rdparty/install-local.sh` publishes `z3.jar` in the local maven repo. Apalache needs that jar to compile the code.

Dockerfile
konnov:
```docker
# The maven repository
COPY --from=thirdparty /root/.m2 /root/.m2
```

This one is definitely needed for the build stage. The other two are probably needed for the runtime (e.g., dynamic libraries).

Dockerfile
konnov: Ideally, this command should copy the source modules, `bin`, `*.md`, `LICENSE`, and `test`. It should not copy the build-specific and IDE-specific directories such as `target` or `workspace`. This should not be an issue if you publish a docker image from a clean repo.

konnov: We can have fine-grained copying commands. Java/Scala does not need the compiled classes. Once we have a distribution jar, we need that jar, `src/tla`, `bin`, `3rdparty/{bin,lib}`, and docs, license, etc.

Dockerfile
shonfeder: Sorry, I think I'm missing the connection. Is the surefire workaround here related to the `z3.jar`?

Dockerfile
shonfeder: Sorry, I'm only asking about the environment variable here `_JAVA_OPTIONS=...`, not about copying the maven repository.

Dockerfile
shonfeder: >Ideally, this command should copy the source modules, bin, *.md, LICENSE, and test

Could you explain the motive for adding the markdown and test files? iiuc, those shouldn't be needed for the executable in a minimal app image. Do we intend the docker container to be something that users would work inside of, rather than as a way of wrapping the executable?

Did I capture everything you'd advised here?

Dockerfile
konnov: Oh. The Surefire option is probably needed for building the project inside docker. It has been for several years there, so probably it is not needed anymore.

Dockerfile
konnov: Yeah, it was an option needed by maven to make the build. But I am not actually sure, whether it is needed anymore.

Dockerfile
konnov: You are right. The docs can be checked at github any time. I don't think the docs and tests are actually required.

konnov (approved):
Looks good. I think we can merge it. If you like to investigate on whether `_JAVA_OPTIONS` is still needed, please go ahead.

Dockerfile
shonfeder: Turns out we don't need it. I ran the integration tests against the app image built with this flag removed and everything worked as expected.

shonfeder:
This should be ready for another look (once the tests pass).

It also adds a wrapper script for docker, and a tweak to the integration tests that enables running the integration tests against a docker container build. I used to to verify that the new, svelt app image (only 367mb) behaves as expected.

shonfeder:
Do we consider the docker container a proper part of the product, such that this PR should include an entry in the changelog?

konnov (approved):
Looks great! Let's add the entry in CHANGELOG.

PR #198, opened 2020-08-05
Clarification on Naturals in the manual

konnov:
@istoilkovska pointed that there is no designated type for naturals. I have added a clarification in the manual.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -994,7 +994,7 @@ of `42` be an integer?
 
 Apalache still allows you to extend `Naturals`. However, it will treat all
 number-like literals as integers. This is consistent with the view that the naturals are
-a subset of integers, and integers are a subset of reals.  Classically, one
+a subset of the integers, and the integers are a subset of the reals.  Classically, one
 would not define subtraction for naturals. However, the module `Naturals`
 defines binary minus, which can easily drive a variable outside of `Nat`. For
 instance, see the following example:
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -993,7 +993,7 @@ somebody else extends your module and also `Integers`, should be the type
 of `42` be an integer?
 
 Apalache still allows you to extend `Naturals`. However, it will treat all
-number-like literals as integers. This is consistent with that the naturals are
+number-like literals as integers. This is consistent with the view that the naturals are
 a subset of integers, and integers are a subset of reals.  Classically, one
 would not define subtraction for naturals. However, the module `Naturals`
 defines binary minus, which can easily drive a variable outside of `Nat`. For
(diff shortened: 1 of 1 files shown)


istoilkovska:
Looks good to me, thanks!

docs/manual.md
```diff
@@ -978,6 +981,46 @@ Type annotations can be also applied to sets of records. For example:
...
+will find that there is no designated type for naturals. Indeed, one can just
+use the type `Int`, whenever a natural number is required. If we introduced a
+special type for naturals, that would cause a lot of confusion for the type
+checker. What would be the type of the literal `42`? That depends on, whether
+you extend `Naturals` or `Integers`. And if you extend `Naturals` and later
+somebody else extends your module and also `Integers`, should be the type
+of `42` be an integer?
+
+Apalache still allows you to extend `Naturals`. However, it will treat all
+number-like literals as integers. This is consistent with that the naturals are
```
shonfeder, line 996:
```suggestion
number-like literals as integers. This is consistent with the view that the naturals are
```

docs/manual.md
```diff
@@ -978,6 +981,46 @@ Type annotations can be also applied to sets of records. For example:
...
+a subset of integers, and integers are a subset of reals.  Classically, one
```
shonfeder, line 997:
```suggestion
a subset of the integers, and the integers are a subset of the reals.  Classically, one
```

docs/manual.md
```diff
@@ -978,6 +981,46 @@ Type annotations can be also applied to sets of records. For example:
...
+would not define subtraction for naturals. However, the module `Naturals`
```
shonfeder, line 998: Is this right? Or would subtraction just be defined as a partial function? (Asking for my own edification here.)

konnov: I also thought of Naturals!- as a partial function. But it is a total function. You can imagine that 4-10 returns an undefined value when applied to naturals, and this value happens to be -6 :-)

shonfeder: Ah. Iiuc, you're saying that while `(-) : Nat x Nat -> Nat` is partial, we could also define it as `(-) : Nat x Nat -> Int` to make it total?

konnov: Since there are no types in TLA+, `(-): Nat x Nat -> Nat` does not make a lot of sense in TLA+, at least for now. There is one `-` defined for Naturals, Integers, and Reals. But since we know that integers are closed under `-`, we pretend that there are no reals, when dealing with integers. The same trick applies to `+` with naturals.

shonfeder: Ah, I see :) Thanks!

PR #205, opened 2020-08-07
Release 0.7.2

shonfeder:
## 0.7.2 [RELEASE]

 * fixed an omitted version number update

--- 

**Note** that this cherry picks in just the bug fixes and changelog updates from unstable, and does not merge all of unstable into master.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## Unreleased #unstable
 
+ * fixed an omitted version number update
+
 ## 0.7.1 [RELEASE]
 
  * safe checks for the user options in ConfigurationPassImpl, see #193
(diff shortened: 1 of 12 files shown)


shonfeder:
Looks like `mvn versions:set -DremoveSnapshot` should that just be in this branch, or in unstable and then cherry picked in? How have you been doing it to this point?

konnov:
Our releases were so rare that I don't remember :-)

shonfeder:
Ah! OK! We'll just approach it as designing out a new process then. Let me do a little bit of research on standard release cycle for java/scala projects, and I'll loop back here.

shonfeder:
Ok, using the approach with our current set up, I think the `-RELEASE` suffix had to be added in `unstable` or else we'll need to merge the release branch back into `unstable`, which it would be better to avoid. So I've done that.

PR #206, opened 2020-08-17
Update release process documentation

shonfeder:
This is to take account of missing steps discovered from the last release, and to make some of the implicit steps more explicit.

This also updates the TOC and refactors the structure of the


konnov (approved):
Awesome! Thanks for documenting the process!

Issue #208, opened 2020-08-21
[BUG] Could not parse TLC configuration in BlockingQueue.cfg: string matching regex '\z' expected but '{' found


lemmy:
```bash
kuppe@mkuppe-z440:~/src/TLA/blockingqueue$ ../apalache/bin/apalache-mc check BlockingQueue.tla
# Tool home: /home/mkuppe/src/TLA/apalache
# Package:   /home/mkuppe/src/TLA/apalache/mod-distribution/target/apalache-pkg-0.7.3-SNAPSHOT-full.jar
# JVM args: -Xmx4096m -Djava.library.path=/home/mkuppe/src/TLA/apalache/3rdparty/lib  -DTLA-Library=/home/mkuppe/src/TLA/apalache/src/tla:/home/mkuppe/src/TLA/blockingqueue/
#
# APALACHE version 0.7.3-SNAPSHOT build v0.7.0-28-gdd171c8
#
# WARNING: This tool is in the experimental stage.
#          Please report bugs at: [https://github.com/informalsystems/apalache/issues]

Checker options: filename=BlockingQueue.tla, init=, next=, inv=   I@19:04:13.850
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/home/mkuppe/src/TLA/apalache/mod-distribution/target/apalache-pkg-0.7.3-SNAPSHOT-full.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
PASS #0: SanyParser                                               I@19:04:14.139
Parsing file /home/mkuppe/src/TLA/blockingqueue/BlockingQueue.tla
Parsing file /tmp/Naturals.tla
Parsing file /tmp/Sequences.tla
Parsing file /tmp/FiniteSets.tla
Parsing file /home/mkuppe/src/TLA/blockingqueue/TLAPS.tla
PASS #1: ConfigurationPass                                        I@19:04:14.482
  > Loading TLC configuration from BlockingQueue.cfg              I@19:04:14.483
[...]

PR #209, opened 2020-08-24
Add parser position to TLCConfigurationError message

shonfeder:
Useful for users and for us when debugging (as in the case of  #208).

With this change, lexer errors now look like:

```
Configuration error (see the manual):   > Could not parse TLC configuration in BlockingQueue.cfg at position 4.17: string matching regex '\z' expected but '{' found E@00:14:23.070
```


konnov (approved):
That definitely improves usability!

PR #211, opened 2020-08-24
Update to scala 2.12.12

shonfeder:
Now we're only one minor version and three patches behind the latest Scala :)

This also removes a profile configuration that seems to be an unneeded complication in the build configuration, afaict. See 27311da for details.


konnov (approved):
Perfect! Thanks for bringing order to the scala dependencies!

PR #214, opened 2020-09-16
Add distclean target to remove 3rd party artifacts

shonfeder:
Eventually this target should return the project to the
state of a fresh install (removing test artifacts etc),
but this at least handles the case of needing to clean build
our vendored deps.


konnov (approved):
That makes sense

shonfeder:
#230 will make this irrelevant :) :tada:

PR #216, opened 2020-09-18
Embarrassingly simple type checker (ETC): phase 2 

konnov:
This PR contains plenty of code, as I had to come up with a consistent way to do type checking. The main action is happening in:

- [x] `EtcExpr` and its case classes present a simple lambda calculus that is built around types. This calculus cannot be used to write programs, but the type checker is running against this language. The language is rather minimalistic, its purpose is to express the type-checking problems for TLA+ expressions.

 - [x] `ToEtcExpr` translates TLA+ expressions into `EtcExpr`. For most of the expressions, the translation is trivial, e.g., `x + y` becomes `((Int, Int) => Int) x y`. However, it contains non-trivial, though totally natural, translation rules for set comprehensions, quantifiers, and recursive functions.

 - [x] `EtcTypeChecker` does the actual type checking for `EtcExpr`.

 - [x] The unit tests `TestToEtcExpr` and `TestEtcTypeChecker` are the best source for understanding the expected behavior.

This is the core of the future type checker. The next steps (phase 3) include:

 - [x] translating the operators in `Typing` as type annotations
 - [ ] making a standalone tool
 - [ ] adding support for in-comment annotations

These will be added later in the PR. Use the chance for review before it gets really lots of code.


shonfeder:
Is it at all possible to do the additional work in a separate PR? This PR is already an order of magnitude larger than is ideal for high quality code review.

konnov:
Yep, I will try to stop working on this branch after fixing a few issues. TLA+ is always full of surprises. The current code is useless for `{}` and `<<>>`.

konnov:
This should be the final commit in this PR. Actually, `{ }` and `<< >>` work in most cases, when we treat them as polymorphic constants, not operators.

konnov:
The rest will be in ik/typecheck3

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+  * Lambda abstraction that introduces an operator type
+  * of multiple arguments that range over their respective types Set(*).
+  *
+  * @param body          the function body
+  * @param paramsAndDoms parameter names and type expressions that encode sets of values
+  * @param sourceRef     the identifier of the TLA+ expression that resulted in this STCExpr (ignored in equals).
+  */
+case class EtcAbs(body: EtcExpr, paramsAndDoms: (String, EtcExpr)*)(val sourceRef: EtcRef) extends EtcExpr {
```
shonfeder, line 52: Why does this take body before the arguments? It seems counter intuitive to me given the usual syntax for lambdas, which always take the argument first.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
```diff
@@ -0,0 +1,117 @@
...
+  override def toString: String = {
+    val bindings = paramsAndDoms.map(p => "%s ∈ %s".format(p._1, p._2))
+    "%s@λ %s. %s".format(sourceRef, String.join(", ", bindings: _*), body)
+  }
+}
+
+/**
+  * Application of an operator whose signature is known. An operator may have several overloaded polytypes, e.g., f[e].
```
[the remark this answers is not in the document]
shonfeder: Reading the `toEtcExpr` code I can see that this is because abstractions are used to encode operators. However, I'd suggest keeping the abstraction barrier in place here, and just documenting `EtcExpr` as a lambda calculus, being careful *not* to conflate the representation here with it's intended use. Then we can describe the intended use as an additional comment. Like

```suggestion
  * Application of an EtcExpr term (which must be an abstraction) whose signature is known. An EtcExpr abstraction may have several overloaded polytypes, e.g., f[e].
  
  We abstractions to represent TLA operators, so this application models the application of an operator to its arguments.
```

Also, I don't know how to parse `f[e]` as an example of an operator with several overloaded polytypes. Expansion there would be helpful.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+    */
+  def apply(decl: TlaOperDecl, inScopeEx: EtcExpr): EtcExpr = {
+    val paramsAndDoms = formalParamsToTypeVars(0, decl.formalParams).map {
+      case (paramName, paramType) =>
+        (paramName, mkConst(BlameRef(decl.body.ID), SetT1(paramType)) )
+    }
+
+    if (decl.name.startsWith("TypeAssumptions")) {
+      // the special form for type annotations over state variables and constants
+      parseTypeAssumptions(decl, inScopeEx)
```
shonfeder, line 38: Breaking this out into a helper function makes it easy to separate the control flow from the actions dispatched on the different branches of the match statement, and the function name helps self document the intent. I like it! 

I think you could remove the `case 1: ....` and `case 2: ...` comments below if you also refactored those out into helper functions. That would make this section of code more consistent (why is `parseTypeAssumptions` a helper function but there's no helper for `parseTypedOperBody`?) and imo help with readability in general.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+                mkTypeDecl(ExactRef(decl.body.ID), name, tt, terminal)
+              } catch {
+                case e: Type1ParseError =>
+                  throw new TypingInputException(s"Parser error in type annotation of $name: ${e.msg}")
+              }
+          }
+        }
+
+      case _ =>
+        throw new TypingInputException(s"""Error in ${decl.name}: Expected /\ AssumeType(varName, "typeString") /\ ...""")
```
shonfeder, line 87: I could see users getting caught up here pretty easily.

Ideally it would link to the documentation on this convention.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+      }
+    }
+  }
+
+  // parse the type annotations inside TypeAssumptions
+  private def parseTypeAssumptions(decl: TlaOperDecl, inScopeEx: EtcExpr): EtcExpr = {
+    decl.body match {
+      case OperEx(TlaBoolOper.and, args @ _*) =>
+        val annotations =
+          args.collect { case OperEx(TypingOper.assumeType, NameEx(name), ValEx(TlaStr(typeText))) =>  (name, typeText) }
```
shonfeder, line 70: Won't this raise a run-time match error if any conjunct isn't an instance of  `AssumeType`? iiuc, that would make the subsequent, indirect check for length match unreachable, and it would return an helpful error to the user (and raise noisy warnings during compilation! :)).

How about a wild card match here that raises the `TypingInputException` if the conjunct isn't an `AssumeType`, along with the information that the user needs to identify the specific conjunct?

shonfeder (changes requested):
@konnov I've made some progress on reviewing, but I'm afraid I still have lots and lots to go.

Myst of my comments are just questions to get situated with the intent and approach used here. I'd be happy to sync up to talk over in real time if that would help.

I'll resume reviewing ASAP, but wanted to share what I have so far, since there are at least a few concrete, uncontroversial changes needed (e.g., the references to `STCExpr`).

src/tla/Typing.tla
```diff
@@ -46,4 +46,16 @@ AssumeType(name, tp) == TRUE
 (****************************************************************************)
 type_str :> ex == ex
 
+(****************************************************************************)
```
shonfeder, line 49: fwiw, it seems to me it would be much nicer to just have an operator that let people make inline type annotations. Having to add adhoc value constructors as special cases seems a bit smelly to me.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+              }
+          }
+        }
+
+      case _ =>
+        throw new TypingInputException(s"""Error in ${decl.name}: Expected /\ AssumeType(varName, "typeString") /\ ...""")
+    }
+  }
+
+  // TODO: a long string of translation rules. Can we decompose it?
```
shonfeder, line 91: I'm not sure I understand this TODO. Is the "long string" meaning the many cases to match? 

If so, one way to decompose would be to replace the comment blocks (`LITERALS`, `GENERAL OPERATORS`, `SETS`) with self-documenting helper functions that translate the appropriate kinds of terms. So we'd end up with

```
case ValEx(val) => typeValEx(ex, val)
case OperEx(op) => typeOperEx(ex, op)
 // ... 
```

Aside from the self documentation, I'm not sure if the indirection adds much, but maybe there are some additional benefits that I'm not considering?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
```diff
@@ -0,0 +1,117 @@
+package at.forsyte.apalache.tla.typecheck.etc
+
+import at.forsyte.apalache.tla.typecheck.TlaType1
+
+/**
+  * An expression in a simple typed lambda calculus. Here we do not care about the concrete values,
```
shonfeder, line 6: Since TLA expressions are referred to as `TlaEx`, wouldn't it be more consistent to call these `EtcEx`? Otherwise we have two different abbreviations for the same word.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+    case OperEx(op, args @ _*) if op == TlaOper.eq || op == TlaOper.ne =>
+      // x = y, x /= y
+      val opsig = OperT1(Seq(VarT1("a"), VarT1("a")), BoolT1())
+      mkApp(ex.ID, opsig, args)
+
+    case OperEx(TlaOper.apply, NameEx(name), args @ _*) =>
+      // F(e_1, ..., e_n)
+      mkAppByName(ExactRef(ex.ID), name, args.map(this(_)) :_*)
+
+    case OperEx(TlaOper.apply, opName, args @ _*) =>
```
shonfeder, line 163: How would this arise?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+        mkConst(ExactRef(ex.ID), SeqT1(elemType))
+      } catch {
+        case e: Type1ParseError =>
+          throw new TypingInputException(s"Parser error in Typing!EmptySeq($elemTypeText): ${e.msg}")
+      }
+
+    //******************************************** GENERAL OPERATORS **************************************************
+    case OperEx(op, args @ _*) if op == TlaOper.eq || op == TlaOper.ne =>
+      // x = y, x /= y
+      val opsig = OperT1(Seq(VarT1("a"), VarT1("a")), BoolT1())
```
shonfeder, line 156: Why does this hard code in type variable name `"a"`? Couldn't this conflict with other type variables in scope? Also, why don't we get the actual types of the arguments here in stead of just positing variables? Wouldn't this fail to produce the desired type checking error if someone passed in, e.g., `1 = "1"`, since it doesn't actually check whether the operands are the same type?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcBuilder.scala
```diff
@@ -0,0 +1,98 @@
+package at.forsyte.apalache.tla.typecheck.etc
+
+import at.forsyte.apalache.tla.lir.UID
+import at.forsyte.apalache.tla.typecheck._
+
+/**
+  * A builder trait to conveniently construct instances STCExpr.
```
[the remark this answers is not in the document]
konnov: Yes, they are instances of `Etc`. It is a refactring leftover.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+  * Application of an operator whose signature is known. An operator may have several overloaded polytypes, e.g., f[e].
+  *
+  * @param operTypes  an STC expression that represents an operator type
+  * @param args  operator arguments
+  * @param sourceRef the identifier of the TLA+ expression that resulted in this STCExpr (ignored in equals).
+  */
+case class EtcApp(operTypes: Seq[TlaType1], args: EtcExpr*)(val sourceRef: EtcRef) extends EtcExpr {
```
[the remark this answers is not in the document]
konnov: Yeah. My initial attempt was to use a term instead of `operTypes` too. But then you would have strange constant terms that represent overloaded signatures: `Int` or `Bool`. The type checking started to be a nightmare very quickly, so I reverted to this solution, which is what is sufficient for TLA+ type checking.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+  * Application of an operator by its name. The operator type should be resolved with a type signature that is
+  * encoded in a type context.
+  *
+  * @param name operator name
+  * @param args operator arguments
+  * @param sourceRef identifier of the TLA+ expression that resulted in this STCExpr (ignored in equals).
+  */
+case class EtcAppByName(name: String, args: EtcExpr*)(val sourceRef: EtcRef) extends EtcExpr {
```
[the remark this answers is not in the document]
konnov: That is an artifact of the solution above. If we just had a term, there would be no need for this case. Now, there are essentially two cases: an application of an operator with known polymorphic signatures, or an application of an operator by its name, e.g., originating from let. This separation comes from the fact that the built-in operators are inherently overloaded in our types system, but the user-defined operators are not overloaded. I deliberately excluded overloading from the type grammar.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
```diff
@@ -0,0 +1,117 @@
...
+/**
+  * Bind an expression to a name. To bind an operator of multiple arguments, use STCAbs.
+  * The operator type should be resolved with a type signature that is encoded in a type context.
+  *
+  * @param name  an expression name in the let-in binding
+  * @param bound the expression to bind
+  * @param body  the expression the binding applies to
+  * @param sourceRef the identifier of the TLA+ expression that resulted in this STCExpr (ignored in equals).
+  */
+case class EtcLet(name: String, bound: EtcExpr, body: EtcExpr)(val sourceRef: EtcRef) extends EtcExpr {
```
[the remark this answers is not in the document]
konnov: We do not expect anybody to write in this language. But I believe it is a nice and (rather) minimal language to explain how our type checking works. Actually, I'd like to write a 2-3 page paper on explaining type checking for this language. So the users will be able to see the mechanics without diving too much into SMT solving or whatever.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+  * This translator is an extension of the ideas that appear in SignatureGenerator by Jure Kukovec.
+  *
+  * @see at.forsyte.apalache.tla.types.SignatureGenerator
+  *
+  * @author Igor Konnov
+  */
+class ToEtcExpr extends EtcBuilder {
+  private val type1Parser = DefaultType1Parser
+
+  // TODO: add support for recursive functions
```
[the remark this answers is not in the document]
konnov: yep, https://github.com/informalsystems/apalache/issues/228

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+
+  /**
+    * Translate an operator declaration.
+    *
+    * @param decl an operator declaration
+    * @param inScopeEx an expression to chain in the bottom of let-definition, it should have been translated
+    * @return the translated let-in expression with inScopeEx attached
```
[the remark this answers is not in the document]
konnov: There are two reasons: (1) recursive operators and (2) exponential complexity of operator substitution. For the recursive operators, it is more reasonable to bind the type to name and later use this name. Re: (2), it is somewhat a stretch here, but there are small made up programs in ML that demonstrate an exponential blow-up, if one replaces declarations with their bodies.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcBuilder.scala
```diff
@@ -0,0 +1,98 @@
...
+  * Mix this trait to your class to construct STC expressions without pain.
+  * This class shields the user from the weird syntax of case classes that have two kinds of fields:
+  * the fields counted in equals, and the fields that are ignored in equals.
```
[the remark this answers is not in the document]
konnov: That is the Java/Scala method that checks two objects for equality. Objets of case classes are tested for structural equality by default, in contrast to normal classes. However, a declaration like `case Foo(x: Int)(y: Int)` excludes `y` from the default equality test.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+      // There are two cases: (1) the definition body contains a type annotation, and (2) no type annotation
+      decl.body match {
+        case OperEx(TypingOper.withType, ValEx(TlaStr(typeText)), body) =>
+          // case 1: the definition body contains a type annotation
+          val parsedType =
+            try {
+              type1Parser(typeText)
+            } catch {
+              case e: Type1ParseError =>
+                throw new TypingInputException(s"Parser error in type annotation of ${decl.name}: ${e.msg}")
```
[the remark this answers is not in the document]
konnov: This exception will not give you the position, so `e.toString` would not help.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -0,0 +1,594 @@
...
+    * Translate an operator declaration.
+    *
+    * @param decl an operator declaration
+    * @param inScopeEx an expression to chain in the bottom of let-definition, it should have been translated
+    * @return the translated let-in expression with inScopeEx attached
+    */
+  def apply(decl: TlaOperDecl, inScopeEx: EtcExpr): EtcExpr = {
+    val paramsAndDoms = formalParamsToTypeVars(0, decl.formalParams).map {
+      case (paramName, paramType) =>
+        (paramName, mkConst(BlameRef(decl.body.ID), SetT1(paramType)) )
```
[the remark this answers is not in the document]
konnov: That is another quirk that is coming from the TLA+ expressions. We are using lambdas to translate TLA+ expressions like `\A x \in S: P`. So it is natural to use sets here. Alternatively, we would have to introduce a set destructor like `Set(a) => a` and then write `x: ((Set(a) => a) S)`. Once I have written that, I actually like this alternative approach better :-)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
konnov: Right. That is our standard hack to use variadic operators in Scala. You can construct `EtcAbs` with at least one argument.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: That is the feature of `.collect`. It calls a partial function. If the function is not defined on a sequence element, it would silently ignore the element. It is quite handy for filtering noise by pattern matching. But then you have to check, whether some elements were actually noise.

src/tla/Typing.tla
konnov: Yes and no. I think it is a good idea to limit type annotations only to the declarations: variable declarations, constant declarations, and operator declarations. You can always make a workaround in the form of `LET foo: "Int => Int" :> myexpr IN foo`. However, the user expectations of what can be and should be annotated are much lower with the current approach. It is also often counterintuitive how one should treat an annotation of an arbitrary expression in the presence of polymorphism. I guess, that is why languages like Scala also have very limited scoping for type information.

konnov: Arbitrary annotations also raise the issue of contravariance, which I totally fail to understand in the TLA+ context.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: Yes, it is quite a long method. Alternatively, we could have several partial functions. I am not sure about what is better here.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
konnov: That is a good point :-)

konnov: I created a separate issue, so it does not get lost in this huge PR.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: That is usually a call to a user-defined operator. We call them by names. In the past, we were passing the declaration directly, but that was quite inconvenient for multiple preprocessing and analyses phases.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: All signatures have simple variable names, for convenience and simplicity. They get renamed in the type checking code.

src/tla/Typing.tla
shonfeder: Ok. Thanks for explaining your motivation.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TlaType1.scala
```diff
@@ -36,15 +86,68 @@ case class StrT1() extends TlaType1 {
...
 }
 
 /**
-  * A type variable.
+  * A type variable. Instead of using strings for names, we are just using integers, which makes it easier
+  * to process them. To make vars user-friendly, we assign the names a..z to the numbers 0..25.
+  * The rest are called a27, a28, etc.
   *
-  * @param name a variable name that should be assigned a type
+  * @param no the variable number
```
shonfeder, line 100: `no` reads as "no" as in the opposite of "yes" to me. I suggest using `n`, `num`, or `number`.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcBuilder.scala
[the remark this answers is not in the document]
shonfeder: Ok, thanks! So this would be obvious to an experienced Scala dev?

For my part, something that makes it clear "equals" here is referring to a Scala method and not equality checks on lambda terms would be helpful, but maybe I'll feel differently once I have some more Scala under my belt :)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
shonfeder: Correcting the naming here won't impact the readability or review-ability of the huge PR, because you're just introducing the code for the first time.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
shonfeder: Oh. I see. Why not just have `EtcAbs` take a sequence/list for the args?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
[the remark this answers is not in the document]
shonfeder: I'm afraid I don't understand. But if you're sure this is the most sound and elegant way, I won't push the point. :)

konnov:
@shonfeder if you have not started to look at `EtcTypeChecker`, don't do that. I will push the next PR soon. `EtcTypeChecker` has changed there, as the real TLA+ specs are a bit different from what I expected.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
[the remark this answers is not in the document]
shonfeder: I'd add that to the doc string for future readers. Ditto about the rational for why you use TlaType1 directly.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
shonfeder: Ok. This part seems more significant:

> Wouldn't this fail to produce the desired type checking error if someone passed in, e.g., 1 = "1", since it doesn't actually check whether the operands are the same type?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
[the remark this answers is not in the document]
shonfeder: That makes sense. Thanks for explaining.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
shonfeder: Ah. It seems unhelpful here, IMO, since we lose specificity and then have to do an indirect check in the following line.

shonfeder:
Summary of a discussion with @konnov: #244 contains all of the changes here, including additional changes, so we'll close this PR.

PR #218, opened 2020-09-18
Manual assignments

### LINKED ISSUE #181: [FEATURE] Assignment annotations

Kukovec:
Closes #181 .

### DIFF
--- a/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/ManualAssignments.scala
+++ b/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/ManualAssignments.scala
@@ -1,0 +1,38 @@
+package at.forsyte.apalache.tla.assignments
+
+import at.forsyte.apalache.tla.lir.oper.{BmcOper, TlaActionOper, TlaOper}
+import at.forsyte.apalache.tla.lir._
+
+/**
+  * Instead of relying on the automatic inference of assignment sites, users may instead choose to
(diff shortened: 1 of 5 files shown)


tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/ManualAssignments.scala
```diff
@@ -0,0 +1,38 @@
...
+  * manually annotate where they expect assignments in their specifications (with BmcOper.assign)
+  *
+  * This changes automatic assignment detection in the following way:
+  * If, for a given (primed) variable `v`, there exists at least one manual assignment
+  * [ v' := ... ], then all expressions of the form [ v' = ... ] are ignored as assignment candidates
+  * Manual assignments are still required to satisfy both the acyclicity and the covering properties,
+  * or assignment finding will fail.
+  *
+  * Any variable, for which no manual assignemt exists, is handled in the standard way.
+  */
```
konnov, line 17:
```suggestion
*
* @author Jure Kukovec
*/ 
```

konnov (approved):
Looks great! In general, we have to think about reducing the code complexity (or maybe adding more comments) in the assignment finder. The code still looks too complex to me.

PR #222, opened 2020-09-20
Enable dependabot for automatic dep updates

shonfeder:
This should help us keep our dependencies up to date.

See https://docs.github.com/en/github/administering-a-repository/about-github-dependabot-version-updates

(When we switch to SBT we can instead use https://github.com/scala-steward-org/scala-steward)


konnov (approved):
I like dependabot! It found security issues in our dependencies a few times. Let's enable it.

PR #239, opened 2020-09-23
Use github actions for CI

shonfeder:
Closes #220, #221

Contributes to #192 and #237

Aside from removing dependency on an external service provider and being able to make use of the github actions ecosystem, this PR also adds the following features to our CI:

- Add macOS to our build matrix
- Run and (integration) test the Dockerfile

PR #241, opened 2020-09-23
Remove travis config

### LINKED ISSUE #171: Notifications in .travis.yml
The notifications section in `.travis.yml` points to an old email alias at `forsyte.at`, which is still working, but we cannot add new people. What is the best approach there?

shonfeder:
We're now using GitHub Actions (presupposes #239 and will need rebase once that's in).

Closes #171
(by virtue of obviation)


konnov (approved):
Yes! Let's remove travis.

Issue #172, opened 2020-07-13
Reduce CI times to under 10 minutes


shonfeder:
CI currently takes about 30 minutes to complete. I'm optimistic that with some more deliberate caching, use of pre-built containers, and perhaps a smarter incremental/build test strategy we might be able to achieve significant improvements here.

shonfeder:
(This is a more or less arbitrary target time :)

shonfeder:
Getting closer!

![2020-09-23-170301_2105x657_scrot](https://user-images.githubusercontent.com/19820422/94069477-ac076200-fdbe-11ea-888e-4a2f069d1081.png)

shonfeder:
If we can split the integrations to run them in parallel, I think we can nearer to 5 minutes.

konnov:
While the CI times are still high, this is not as pressing issue as it used to be. The development velocity has decreased and there is no congestion on trying to merge the PRs.

PR #243, opened 2020-09-23
Run integration tests in parallel

shonfeder:
Towards #172 

This is an attempt to improve iteration time by reducing the time of our CI pipeline.

Instead of running the unit tests and integration sequentially on each platform, this breaks the unit tests and integration tests into two separate jobs, each running in parallel. 

- When running the unit tests, we use `mvn test`. This job is clocking in at around 5 minutes. If this does fail, we'll likely end up with quicker feedback than we've been receiving.
- When running integration tests, we run `mvn package` with `skipTests`, to get the artifacts as quickly as possible, and then run the integration tests using `NO_MVN`. This job is coming in at 12 mins on ubuntu and 15 mins (fastest times so far) on macOs

My unscientific observations so far indicate that this *might* save us about 3 minutes, and the faster feedback on unit tests could be helpful. But I'm not sure it's worth the added complexity in the build configuration.

Opening this just to get feedback from others on whether it seems worth merging in.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,7 @@ on: [push]
 name: Apalache Continuous Integration
 
 jobs:
-  build-and-test:
+  unit-tests:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


shonfeder:
Oh yeah, this also adds flags to the mvn commands to reduce the noise in their output. We should probably add those changes in regardless of whether we want to parallelize the integration and unit tests.

konnov (approved):
Sure. That gives us faster feedback. There is no additional burden on the user. Let's add it.

PR #250, opened 2020-09-30
Use cram-style integration tests

### LINKED ISSUE #192: [FEATURE] Improve integration test harness
We currently just use a bash script for running our integration tests. This is brittle, requires reading and writing bash, and has a lot of limits (e.g., cannot be integrated into proper incremental build/test cycle, has no support for running specific tests). It would improve our CI and testing discipline to use a more mature test harness for our integration tests.

shonfeder:
Closes #192 

This PR uses `mdx` for writing and running cram-style CLI integration tests. 

The tests are embedded in code blocks in a markdown file. Here's an example of what some tests look like:

2020-09-30-150212_2335x1555_scrot

See https://github.com/informalsystems/apalache/blob/4a01124f3bdf692fbc6ba356e1077676c52531dc/test/tla/cli-integration-tests.md for the full suite of cram-style integration tests, and for the documentation describing how to run and write the tests.

---

*Note on impact on CI build times.*

**tl;dr**; Note that building the `mdx` dependency will add about 7 minutes to the integrations tests in CI *the first time* they run on `unstable`. But it should only add a few dozen seconds to fetch the catch after that, and this saving should be inherited by all branches based on `unstable`. 

This is all according to [GitHub's documentation on caching in actions](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache). Once the dependency is cached on the default branch -- `unstable` -- the cached artifact should be used by all branches based on the default. This holds for any branch *based off of* one that has already saved the cache. I tested this in #252,  which confirmed that the dependency was fetched from the cache stored from **this** branch.

If it becomes an issue, we can always dockerize the dependency, prebuild a standalone artifact for our test environments, or use a different tool that achieves the same purpose (tho I am unaware of an actively maintained replacement).


shonfeder:
I triggered a failure, to confirm that the new integration tests catch the failure in CI in parity with the existing integration tests.

See https://github.com/informalsystems/apalache/runs/1188178566?check_suite_focus=true#step:9:11 for the logs reporting the failure, showing both how the failures appear in CI (as a diff) and serving as evidence of the failure.

shonfeder:
A substantive improvement to the integrations tests would be and explanation in each section describing what is being tested, but I don't have the context to provide that yet, and this PR is already longer than would be ideal.

test/mdx-test.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
+#
+# Run an integration test suite using mdx
+# See the tla/cli-integration-tests.md for details.
+#
+# Shon Feder, 2020
```
konnov, line 6: Could you comment on why you need such a long script to run the tests? Does it mean that we probably should change the format of our tests?

.github/workflows/main.yml
```diff
@@ -31,6 +31,18 @@ jobs:
         operating-system: [ubuntu-latest, macos-latest]
     steps:
       - uses: actions/checkout@v2
+      - name: Cache local opam repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.opam
+          key: ${{ runner.os }}-opam-4.11.0
+      - name: Set up opam
```
konnov, line 39: I am a bit worried that we add `opam` as a dependency. I had problems with the stability of `opam` packages a few years ago, though this may have been improved since then. Are there alternatives, in case we experience a problem with `mdx`?

konnov (approved):
The cram tests are much cleaner than the handwritten ones! My only worry is the opam dependency (see the comment).

test/mdx-test.py
shonfeder: The script maybe overly engineered or overly defensive.

This script does a few things:

1. Sequences two command executions
2. Ensures pretty feedback on test failure
3. Provides some defensive automation to protect against misuse or common errors
4. Manages paths in a general way, to prevent polluting our working directory with test artifacts

If don't care about those 2-3, then we could just hardcode our target into the Makefile with two command invocations.

Are there any parts of the wrapper script find unnecessary, hard to understand, or otherwise objectionable?

.github/workflows/main.yml
shonfeder: My main hesitation with `cram` is that it appears to be abandoned: https://github.com/brodie/cram/issues/32 and it has not had a new release since 2016.

test/mdx-test.py
konnov: No objections. Just wondered about the code size :)

.github/workflows/main.yml
konnov: I see. Then let's stay with `mdx` and merge!

konnov:
Feel free to merge

test/mdx-test.py
shonfeder: It's a good question. I should probably put this comment in the code, for when I look back in a month and wonder "why is this so complicated"? :)

PR #254, opened 2020-10-07
talks related to Apalache

konnov:
The youtube videos on Apalache are hard to find. I have added them to the front page.


josef-widder:
should we also add the FMBC talk?

konnov:
Yes! I knew that I was missing something

andrey-kuprianov (approved):
Great point! We should definitely have those links :+1:

PR #244, opened 2020-09-23
WIP: The type checker: phases 2 & 3

konnov:
This PR implements:
 
 - [x] a type checking command: `apalache typecheck`, see #160 
 - [x] annotations from `Typing`, see #217
 - [x] bugfixes in the type checker and the translator into `EtcExpr`.

The type checker worked on a few annotated specs: `CarTalkPuzzleTyped.tla`, `GameOfLifeTyped.tla`, `MissionariesAndCannibalsTyped.tla`. A more thorough is needed.

PR #255, opened 2020-10-08
Fixing the TLA_PATH

shonfeder:
Extracted from #244

@konnov I'd be interested to know what this fixes, actually, since it looks like the code should be strictly equivalent to what it's replacing :D

### DIFF
--- a/bin/apalache-mc
+++ b/bin/apalache-mc
@@ -38,7 +38,11 @@ TLA_LIB="$DIR/src/tla"
 # This is a temporary solution for handling TLA_PATH via a Java variable.
 # See apalache/#187.
 # Once tlaplus/#490,#493 are released, remove this workaround.
-(test -z "$TLA_PATH" && TLA_PATH="$TLA_LIB") || TLA_PATH="$TLA_LIB:$TLA_PATH"
+if [ -z "$TLA_PATH" ]; then
+    TLA_PATH="$TLA_LIB"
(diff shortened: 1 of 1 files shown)


shonfeder:
Once this is merged in, we'll rebase #255 on to `unstable` to pick up the change. This is a first, very trivial, inroad to staging #255 into a sequence of logically distinct changesets.

bin/apalache-mc
```diff
@@ -38,7 +38,11 @@ TLA_LIB="$DIR/src/tla"
 # This is a temporary solution for handling TLA_PATH via a Java variable.
 # See apalache/#187.
 # Once tlaplus/#490,#493 are released, remove this workaround.
-(test -z "$TLA_PATH" && TLA_PATH="$TLA_LIB") || TLA_PATH="$TLA_LIB:$TLA_PATH"
```
konnov, line 41: I found that these two pieces of code are actually not equivalent. The expression in parentheses spawns a new shell, sets a variable and then terminates the shell. Effectively, it never updates `TLA_PATH`. That is why I introduced the simple `if-else-fi`.

konnov (approved):
See the comment.

bin/apalache-mc
shonfeder: ah! The subshell thing! Good catch. Thank you for explaining :)

PR #256, opened 2020-10-08
Update ADR2 to address type ascriptions for empty collections

shonfeder:
dbcf4d2 also suggests some minor tweaks to the wording.

Note that this will require resolving a tiny merge conflict when rebasing #244 on top of this change.


konnov (approved):
Theses changes look good to me

PR #258, opened 2020-10-08
Types fix

Kukovec:
Fixes a bug where nullary operator types are printed incorrectly to the output file.

### DIFF
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/types/DependencyGraph.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/types/DependencyGraph.scala
@@ -63,7 +63,7 @@ object DependencyGraph {
                 val operParams = params.withFilter( _.isInstanceOf[OperFormalParam] ).map {
                   case OperFormalParam( paramName, _ ) => paramName
                 }
-                // ... and scan it for any operator calls (that arent params)
+                // ... and scan it for any operator calls (that aren't params)
                 val children = subCalls( body ) -- operParams
                 // Any operator that is called in such a way, cannot be a basic root,
                 // (though it may possibly be chosen as a non-basic root, in the mutual recursion setting)
@@ -77,6 +77,8 @@ object DependencyGraph {
                 constructed += name -> node
                 explorationStack = explorationStack.tail
               case None =>
+                // The body of the operator is unknown, so it's either a CONSTANT or a higher-order
+                // operator parameter. Should be treated as a leaf.
                 throw new IllegalArgumentException( s"The body of operator $name is unknown" )
             }
           case n =>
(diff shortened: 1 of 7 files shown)


shonfeder:
@Kukovec could you rebase this on top of the latest unstable so it picks up the new CI configuration?

shonfeder (changes requested):
@Kukovec could you rebase this on top of the latest unstable so it picks up the new CI configuration?

Kukovec:
> @Kukovec could you rebase this on top of the latest unstable so it picks up the new CI configuration?

Not sure that's a good idea; `types` itself diverges too much from `unstable` for that. We're still a few steps away from integrating type inference into the standard Apalache workflow.

shonfeder:
Oh my bad! I missed the base branch, and assumed this was a PR into unstable. Please ignore :)

tla-types/src/test/scala/at/forsyte/apalache/tla/types/TestTypeAnnotationPass.scala
```diff
@@ -81,12 +81,21 @@ class TestTypeAnnotationPass extends FunSuite with TestingPredefs with BeforeAnd
...
 
-//  test( "Test: Blockchain" ) {
+//  ignore( "Test: Blockchain" ) {
 //    testFromFile( "realSpecs/BlockchainP1.tla" )
 //  }
 //
-//  test( "Test: test7.tla (Lightclient)" ) {
+//  ignore( "Test: test7.tla (Lightclient)" ) {
 //    testFromFile( "test7.tla" )
 //  }
```
konnov, line 90: Shall we exclude these tests from the base? I think they should be the integration tests.

konnov (approved):
Good catch. That is why we have to feed the inferred types to the type checker.

konnov:
Actually, what prevents us from integrating the inference in unstable? It does not imply that the model checker will use the types immediately. What I have in mind: The type inferer would produce type annotations of all declarations for the type checker. The type checker will fill the typing gaps, while the model checker will pick up the types from the run of the type checker (not sure exactly about how it will work).

konnov:
I think, your branch is quite independent. Why don't you merge the changed **from unstable**? The build system has changed to better quite a bit on unstable.

tla-types/src/main/scala/at/forsyte/apalache/tla/types/DependencyGraph.scala
```diff
@@ -150,13 +152,20 @@ object DependencyGraph {
         _ ++ _
       }
     case LetInEx( letInBody, defs@_* ) =>
-      ( defs map { case TlaOperDecl( _, _, b ) => subCalls( b ) } ).foldLeft(
+      ( defs map {
+        case TlaOperDecl( _, params, b ) =>
```
shonfeder, line 156: I think a small helper function to self-document what's being done here could be helpful here. From the perspective of an unfamiliar reader (this one, any how :) ), the level of tested operations here is pretty hard to parse out.

shonfeder (approved):

shonfeder:
I just noticed that this PR hasn't seen any action in a while. I wanted to check in to see whether it's waiting on something, is still a WIP, or has just been overlooked.

konnov:
Good point. Jure @Kukovec, I guess it requires just a bit of action to get merged.

Kukovec:
I'm sitting on it because we still haven't agreed on how exactly we want to handle the types branch. As it stands, it's incompatible with the current unstable and it would take several PRs to get to a point where the model checker is actually using the output of the type inferer.
I'm also not sure what to do about the CI tests that inevitably fail for this branch.

konnov:
Let's decompose the problem into pieces. We are also merging the type checker that is not integrated in the model checker. The first goal is to merge unstable into your branch, so it is compatible with all the recent features.

konnov:
If you like to merge this one in `types`, just go ahead

PR #260, opened 2020-10-08
Add operators for type annotated empty collections

shonfeder:
Broken out of #244

See
https://github.com/informalsystems/apalache/blob/unstable/docs/adr/002adr-types.md#23-dealing-with-bound-variables
for the motivation

Original Author: Igor Konnov <igor@informal.systems>

### DIFF
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
@@ -3,8 +3,8 @@ package at.forsyte.apalache.tla.lir.oper
 import at.forsyte.apalache.tla.lir.{OperEx, TlaEx}
 
 /**
- * Function operators.
- */
+  * Function operators.
+  */
 abstract class TlaFunOper extends TlaOper {
   override def interpretation: Interpretation.Value = Interpretation.Predefined
 }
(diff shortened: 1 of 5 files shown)


tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
```diff
@@ -103,27 +105,29 @@ object TlaFunOper {
...
+    * NameEx("n"),
+    * OperEx(TlaFunOper.app,
+    * OperEx(recFunRef),
+    * OperEx(TlaArithOper.minus, NameEx(n), ValEx(TlaInt(1)))
+    * )
+    * )
+    * ),
+    * NameEx("n"),
+    * TlaIntSet
+    * )
```
konnov, line 122: Shall we exclude this reformatting glitch?

konnov (approved):
Looks great to me

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
shonfeder: Yeah, I'd asked about this reformatting on #244 as well, tho this is probably the proper place!

Is the rest of the reformatting in this changeset as intended?

shonfeder: I fixed just this block.

PR #269, opened 2020-10-12
Fix logging for integration tests

shonfeder:
This is a followup to #250 -- I had misconfigured the logger,
and so nothing was being logged to the files our stdout.'

Thanks to @konnov  for reporting the broken behavior.

Also adds a flag to enable debugging output to stdout

With this change, `test/mdx-test.py` can be run with the `--debug` flag to get streaming output like the following:

```
[sf@comp apalache-core]$ ./test/mdx-test.py "check Bug20190118 succeeds" --debug
[2020-10-12 16:13:04,922] DEBUG:Output level: DEBUG
[2020-10-12 16:13:04,923] DEBUG:running command: ['ocaml-mdx', 'test', '--verbosity', 'info', '--root', '/home/sf/Sync/informal-systems/apalache/apalache-core/test/tla', '--output', '/home/sf/Sync/informal-systems/apalache/apalache-core/target/test/tla/cli-integration-tests.md.corrected', '/home/sf/Sync/informal-systems/apalache/apalache-core/test/tla/cli-integration-tests.md', '--section', 'check Bug20190118 succeeds']
[2020-10-12 16:13:04,996] INFO:b'ocaml-mdx-test: [INFO] exec: "apalache-mc check --length=1 --init=Init --next=Next --inv=Inv Bug20190118.tla | sed \'s/I@.*//\'"'
[2020-10-12 16:13:07,556] INFO:b'ocaml-mdx-test: [INFO] exec: "apalache-mc check --length=1 Bug20190118.tla | sed \'s/I@.*//\'"'
[2020-10-12 16:13:10,128] INFO:b''
```

It's not the prettiest, and it doesn't include the output of the apalache in real time. This latter is something we can work towards, if we decide its important, but I'd like to get some clarity on the use case that makes the latter necessary before we spend any time on it.

### DIFF
--- a/test/mdx-test.py
+++ b/test/mdx-test.py
@@ -101,8 +101,6 @@ if __name__ == "__main__":
 
     args = cli_parser(working_dir).parse_args()
 
-    configure_logger(args.debug)
-
     # The file holding the integration tests
     test_file = args.test_file.resolve()
 
@@ -117,6 +115,9 @@ if __name__ == "__main__":
     working_dir_str = str(working_dir)
     test_file_str = str(corrected_file)
     corrected_file_str = str(test_file)
+
+    # Must come after the `test_target_dir` is created
+    configure_logger(args.debug)
 
     # Run the mdx tests to generate the corrected file
     cmd = [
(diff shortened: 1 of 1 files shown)


konnov (approved):
Having verbose output is nice!

shonfeder:
Sorry it's not yet as verbose as you may like! :)

Issue #266, opened 2020-10-10
[BUG] Property in .cfg silently ignored


lemmy:
Apalache should, perhaps, raise a big fat warning instead of silently ignoring `Properties` in .cfg?

```tla
----------------------------- MODULE Github484 -----------------------------
VARIABLE x

Init == 
  /\ x \in {1}
  
Next == 
  /\ x' \in {1}

Spec ==
  /\ Init
  /\ [][Next]_x

Prop ==
  [](x # 1)
=============================================================================
```

```
SPECIFICATION
Spec
PROPERTIES   
Prop
```

```bash
markus@avocado:~/src/TLA/_specs/models/tlcbugs/Github484/Github484.toolbox/Model_1$ /opt/apalache/bin/apalache-mc --config=MC.cfg check MC
# Tool home: /opt/apalache
# Package:   /opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar
# JVM args: -Xmx4096m -Djava.library.path=/usr/lib/x86_64-linux-gnu/jni/  -DTLA-Library=
#
# APALACHE version 0.7.2-RELEASE build v0.7.1-5-g5b85083
#
# WARNING: This tool is in the experimental stage.
#          Please report bugs at: [https://github.com/informalsystems/apalache/issues]

Checker options: filename=MC, init=, next=, inv=                  I@20:40:09.896
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
PASS #0: SanyParser                                               I@20:40:10.231
Parsing file /home/markus/src/TLA/_specs/models/tlcbugs/Github484/Github484.toolbox/Model_1/MC.tla
Parsing file /home/markus/src/TLA/_specs/models/tlcbugs/Github484/Github484.toolbox/Model_1/Github484.tla
Parsing file /tmp/TLC.tla
Parsing file /tmp/Naturals.tla
[...]

PR #270, opened 2020-10-12
Better diagnostics in ConfigurationPassImpl

konnov:
- [x] Following #266, I have added more detailed diagnostic messages.
 - [x] Initializing init and next from a canonical formula `Init /\ [Next]_vars /\ ...` that is specified with `SPECIFICATION`.
 - [x] Multiple integration tests for command-line arguments and TLC config.


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -45,36 +46,39 @@ class ConfigurationPassImpl @Inject()(val options: WriteablePassOptions,
     * @return true, if the pass was successful
     */
   override def execute(): Boolean = {
+    val inModule = tlaModule.get
```
shonfeder, line 49: I initially read `inModule` as a predicate that would be true if we are in a module. But, iiuc, this is returning the current module (that we are in), so maybe

```suggestion
    val currentModule = tlaModule.get
```

would be clearer?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -86,35 +90,66 @@ class ConfigurationPassImpl @Inject()(val options: WriteablePassOptions,
...
+
+    def setInit(init: String): Unit = {
+      options.get[String]("checker","init") match {
+        case None =>
+          // In general, passes should not override options. This is a reasonable exception to this rule.
+          logger.info(s"  > Using the init predicate $init from the TLC config")
+          options.set("checker.init", init)
+
+        case Some(cmdInit) =>
+          logger.warn(s"  > $basename: Init operator is set both in TLC config and via --init option; using $cmdInit")
```
shonfeder, line 102: To make it immediately clear which is taking precedence, and to be consistent with the messaging for `--next`:

```suggestion
          logger.warn(s"  > $basename: Init operator is set in TLC config but overridden via --init command line option; using $cmdInit")
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -86,35 +90,66 @@ class ConfigurationPassImpl @Inject()(val options: WriteablePassOptions,
...
+
+    def setNext(next: String): Unit = {
+      options.get[String]("checker", "next") match {
+        case None =>
+          // In general, passes should not override options. This is a reasonable exception to this rule.
+          logger.info(s"  > Using the next predicate $next from the TLC config")
+          options.set("checker.next", next)
+
+        case Some(cmdNext) =>
+          val msg = s"  > $basename: Next operator is set both in TLC config and via --next command line option; using $cmdNext"
```
shonfeder, line 114:
```suggestion
          val msg = s"  > $basename: Next operator is set in TLC config but overridden via --next command line option; using $cmdNext"
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -86,35 +90,66 @@ class ConfigurationPassImpl @Inject()(val options: WriteablePassOptions,
...
-          // In general, passes should not override options. This is a reasonable exception to this rule.
-          options.set("checker.inv", config.invariants)
-        }
-        else {
-          logger.warn("  > Invariants are set both in " + basename + " and via --inv option; using the latter")
+      if (config.invariants.nonEmpty) {
+        options.get[List[String]]("checker", "inv") match {
+          case None =>
+            // In general, passes should not override options. This is a reasonable exception to this rule.
+            logger.info(s"  > $basename: found INVARIANTS: " + String.join(", ", config.invariants :_*))
```
shonfeder, line 137: This line appears in both branches of the `match` (here and on line 141), so maybe better to just construct the message once up top then log it at the appropriate level in each branch. This Will help make clear the few parts that are really essentially different in the two branches.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -123,19 +158,17 @@ class ConfigurationPassImpl @Inject()(val options: WriteablePassOptions,
           logger.info("  > No TLC configuration found. Skipping.")
         }
         else {
-          throw new TLCConfigurationError("  > Could not find TLC config file " + basename + " given via --config option")
+          throw new TLCConfigurationError(s"  > $basename: TLC config is provided with --config, but not found")
```
shonfeder, line 161: I think supplying the name of the config file that failed to load would be helpful.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -169,6 +202,47 @@ class ConfigurationPassImpl @Inject()(val options: WriteablePassOptions,
...
+        OperEx(TlaBoolOper.and,
+          // Init
+          OperEx(TlaOper.apply, NameEx(init)),
+          // [Next]_vars
+          OperEx(TlaActionOper.stutter,
+            // Next
+            OperEx(TlaOper.apply, NameEx(next)),
+            // vars
+            _*
+          ), ///
```
shonfeder, line 237: Out of curiosity, what's the `///` meant to signal here?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -169,6 +202,47 @@ class ConfigurationPassImpl @Inject()(val options: WriteablePassOptions,
...
+          _*))) =>
+        (init, next)
+
+      case Some(d) =>
+        val msg = s"$contextName: Expected $specName to be in the canonical form Init /\\ [Next]_vars /\\ ..."
+        logger.error(s"Operator $specName of ${d.formalParams.length} arguments is defined as: " + d.body)
```
shonfeder, line 243: Since `msg` isn't used in between, seems to make sense to bind it right before the use site? 
```suggestion
        logger.error(s"Operator $specName of ${d.formalParams.length} arguments is defined as: " + d.body)
        val msg = s"$contextName: Expected $specName to be in the canonical form Init /\\ [Next]_vars /\\ ..."
```

shonfeder (approved):
I have some feedback, but none of it's blocking or critical :)

test/tla/cli-integration-tests.md
```diff
@@ -381,3 +381,106 @@ $ TLA_PATH=./tla-path-tests apalache-mc check ./tla-path-tests/ImportingModule.t
...
+### configure missing property in TLC config 
+
+```sh
+$ apalache-mc check --config=Config3.cfg Config.tla | sed 's/[IE]@.*//'
+...
+  > Loading TLC configuration from Config3.cfg
+...
+Configuration error (see the manual): Operator NoLiveness not found (used as a temporal property)
+...
+EXITCODE: ERROR (99)
```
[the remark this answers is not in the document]
konnov: The error code does not propagate when the output is preprocessed with `sed`. I am not sure how one could solve that without having an additional script.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
konnov: I sometimes use it to make the closing parenthesis visible

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
konnov: Yep, fixed that

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
konnov: That is exactly what is kept in `basename`

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
shonfeder: Look visible to me :P

test/tla/cli-integration-tests.md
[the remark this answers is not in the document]
shonfeder: Ah yeah. My bad.

I think the cleanest fix -- and one with other benefits -- would be to remove the nondeterministic noise on all the output lines so we don't need to clean it up in all the tests. I'd be in favor of not including the timestamps by default, as I suspect that they are *usually* not useful, and only of interest in some special occasions. But it would also make sense to leave the timestamps in by default and add a flag to disable them (perhaps also configurable by environment variable?).

If you think this change is sensible, I can open issue for it, and would be happy to have it in my backlog.

A hack for the moment, if we **need** to capture the exit codes while filtering is still necessary, might be something like

```sh
out=$(cmd 2>&1); res=$?; sed 's/foo/bar/' <<< out
```

but that's quite gnarly, and not probably worth the effort since apalache reports the exit code in any case :D

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
shonfeder: Ah I see. My mistake.

test/tla/cli-integration-tests.md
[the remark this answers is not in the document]
konnov: Yeah. People like to see how long it takes to check an individual step. It may take some time to run the model checker. If you like to change the logging output, have a look at `at.forsyte.apalache.infra.log.LogbackConfigurator`. At some point, I gave up on the XML configuration, as it was too unpredictable. So logging is set up in the code.

konnov:
I will merge this PR. Feel free to open an issue for less verbose logging.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -6,8 +6,9 @@ import java.nio.file.Path
 import at.forsyte.apalache.infra.passes.{Pass, TlaModuleMixin, WriteablePassOptions}
 import at.forsyte.apalache.io.tlc.TlcConfigParser
 import at.forsyte.apalache.io.tlc.config._
-import at.forsyte.apalache.tla.lir.TlaModule
+import at.forsyte.apalache.tla.lir._
```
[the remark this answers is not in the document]
konnov: Most likely, it is some formatting rule in Idea :-)

PR #271, opened 2020-10-12
a slice from ik/multicore: operators made Serializable

konnov:
This is a first changeset from `ik/multicore`. The operators and other basic classes in `lir` made `Serializable`.

### DIFF
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
@@ -3,8 +3,8 @@ package at.forsyte.apalache.tla.lir.oper
 import at.forsyte.apalache.tla.lir.{OperEx, TlaEx}
 
 /**
- * Function operators.
- */
+  * Function operators.
+  */
 abstract class TlaFunOper extends TlaOper {
   override def interpretation: Interpretation.Value = Interpretation.Predefined
 }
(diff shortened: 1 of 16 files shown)


tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
```diff
@@ -82,15 +81,15 @@ object TlaFunOper {
...
+    *   <li>function body of type TlaEx that may refer to the function via recFunRef,</li>
+    *   <li>NameEx(variableName),</li>
+    *   <li>variable domain of type TlaEx.</li>
     * </ul>
     *
     * <p>Hence, a declaration of a recursive operator looks like a nullary operator declaration,
-    * whose body contains the constructor of a recursive function. The body of a recursive function may
-    * refer to the function itself by using the operator recFunRef (see below).
-    * Note that the output methods should convert this intermediate representation to the standard TLA+ form.</p>
+    *    whose body contains the constructor of a recursive function. The body of a recursive function may
```
shonfeder, line 90: I wonder if it would be worth just running the formatter you're using on the entire code base and submitting that as a large (trivial) change? It would help reduce some noise from subsequent work.

Let's also put the formatting check CI? If you share the tool and config you're using for the formatting, I can make it happen :)

shonfeder (approved):
Nice! :)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaFunOper.scala
konnov: Well, I am just using Intellij Idea. But it is slowly becoming a problem, as we have a mix of different code styles. Shall we switch to https://scalameta.org/scalafmt/ ?

Issue #274, opened 2020-10-13
[BUG] Unroller is missing the documentation


konnov:
The class `at.forsyte.apalache.tla.pp.Unroller` is missing comments and the class description

Issue #272, opened 2020-10-13
[BUG] Better error reporting for recursive operators


konnov:
The following specification produces exceptions:

```tla
--------------------------- MODULE Rec10 -------------------------------------
EXTENDS Integers

VARIABLES f

RECURSIVE Fact(_)

Fact(n) ==
  IF n <= 1
  THEN 1
  ELSE n * Fact(n - 1)

Init == f = Fact(4)

Next == f' = Fact(7)

Inv == f >= 1
========================================================================
```

While the messages in the exceptions are right, they should be reported as normal errors. Additionally, a pointer to the manual should be given.

PR #275, opened 2020-10-13
closing #272: better diagnostics for recursive operators

konnov:
This PR adds better messages and exceptions to `Unroller`, following #272. The tool does report normal errors instead of throwing exceptions when annotations are missing. The documentation for `Unroller` is scheduled for another issue #274 .

### DIFF
--- a/test/tla/Rec11.tla
+++ b/test/tla/Rec11.tla
@@ -1,4 +1,4 @@
---------------------------- MODULE Rec10 -------------------------------------
+--------------------------- MODULE Rec11 -------------------------------------
 (*
  * A test for unfolding recursive definitions.
  *
@@ -14,6 +14,12 @@ Fact(n) ==
   IF n <= 1
   THEN 1
   ELSE n * Fact(n - 1)
+
+\* define the default value
+UNROLL_DEFAULT_Fact == 0
+
+\* but keep UNROLL_TIMES_Fact undefined
+\*UNROLL_TIMES_Fact == 4
 
 Init ==
     f = Fact(4)
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
@@ -9,7 +9,6 @@ import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModu
 import at.forsyte.apalache.tla.lir.values.TlaInt
 
 class Unroller( tracker : TransformationTracker ) extends TlaModuleTransformation {
-
   import Unroller._
 
   def unrollLetIn(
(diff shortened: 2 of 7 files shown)


test/tla/Rec11.tla
```diff
@@ -0,0 +1,33 @@
+--------------------------- MODULE Rec11 -------------------------------------
+(*
+ * A test for unfolding recursive definitions.
```
shonfeder, line 3: notes here on what this is testing vs. the previous 10 `Rec`s might help in the future :)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
```diff
@@ -65,20 +64,22 @@ class Unroller( tracker : TransformationTracker ) extends TlaModuleTransformatio
     bodyMap.get( unrollLimitOperName ) match {
       case Some( unrollLimitDecl ) =>
         // The unrollLimit operator must not be recursive ...
+        val msg = s"Expected an integer bound in $unrollLimitOperName, found a recursive operator. See: " + MANUAL_LINK
         if ( unrollLimitDecl.isRecursive )
-          FailWith( new Exception( s"Unroll-limit operator $unrollLimitOperName is recursive." ) )
+          FailWith( new TlaInputError( msg ) )
         else
```
shonfeder, line 70: Since `msg` is only used in one branch below, maybe best to just put it directly above the use site (as you do on line 81)?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
```diff
@@ -87,13 +88,14 @@ class Unroller( tracker : TransformationTracker ) extends TlaModuleTransformatio
     bodyMap.get( defaultOperName ) match {
       case Some( defaultDecl ) =>
         // ... which must not be recursive ...
+        val msg = s"Expected a default value in $defaultOperName, found a recursive operator. See: " + MANUAL_LINK
         if ( defaultDecl.isRecursive )
-          FailWith( new Exception( s"Default body operator $defaultOperName is recursive." ) )
+          FailWith( new TlaInputError( msg ) )
```
shonfeder, line 93: Same suggestion here re: not computing the message unless we actually need it.

shonfeder (approved):
Two tiny suggestions, but LGTM :)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
konnov: This is interesting. Idea did not want to introduce curly braces when one writes: `if (...) a else b`. That is another reason for always using curly braces.

shonfeder: lol! 

I can't wait for the day we're all doing structured editing :D

konnov: I have a childhood trauma with structured editing. I think it was QBasic that enforced structured editing instead of having a normal parser :D

PR #282, opened 2020-10-14
ETC-5: Update Typing's Operator typing operator from :> to ##

shonfeder:
I'd missed this update in my previous port of these changes in #260


konnov (approved):
This is an important change, as `:>` is unfortunately used by TLC

PR #259, opened 2020-10-08
WIP: Source for refactoring patches out of #244

shonfeder:
This PR is just to bring visibility to the branch being used to track the refactoring of #244 into smaller change sets.

The branch should shrink over time, as pieces are broken out and merged into unstable.

PR #284, opened 2020-10-15
ETC-6: Add the EtcExpr module

shonfeder:
This encodes the AST of the labda-calculus like language used as the basis of type checking in the Etc.

Authored by @konnov and broken out of #259

### DIFF
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
@@ -32,7 +32,8 @@ case class EtcConst(polytype: TlaType1)(val sourceRef: EtcRef) extends EtcExpr {
 
 /**
   * A reference to a name, which can be introduced in the initial type context, or with EtcAbs.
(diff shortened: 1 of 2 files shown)


tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EtcExpr.scala
```diff
@@ -0,0 +1,117 @@
...
+  */
+case class EtcConst(polytype: TlaType1)(val sourceRef: EtcRef) extends EtcExpr {
+  override def toString: String = {
+    s"$sourceRef@$polytype"
+  }
+}
+
+/**
+  * A reference to a name, which can be introduced in the initial type context, or with EtcAbs.
+  * Note that name is not a type variable, but rather a TLA+ name that carries a type.
```
shonfeder, line 35: How does the TLA+ name carry a type?

konnov: I guess, it is not the right wording. We annotate names (variables, constants, operators) with types. The name itself does not carry any type, but it is carried in the type context.

konnov (approved):
I still like this mini-language :-)

Issue #285, opened 2020-10-16
[BUG] at.forsyte.apalache.tla.pp.OverridingError


lemmy:
```bash
markus@avocado:~/src/TLA/_specs/models/mine/PageQueue/PageQueue.toolbox/w19(master)$ /opt/apalache/bin/apalache-mc --cinit=ConstInit --init=Init --next=Next --length=50 check /home/markus/src/TLA/_specs/models/mine/PageQueue/PageQueue.toolbox/w19/MC.tla
# Tool home: /opt/apalache
# Package:   /opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar
# JVM args: -Xmx4096m -Djava.library.path=/usr/lib/x86_64-linux-gnu/jni/  -DTLA-Library=
#
# APALACHE version 0.7.2-RELEASE build v0.7.1-5-g5b85083
#
# WARNING: This tool is in the experimental stage.
#          Please report bugs at: [https://github.com/informalsystems/apalache/issues]

Checker options: filename=/home/markus/src/TLA/_specs/models/mine/PageQueue/PageQueue.toolbox/w19/MC.tla, init=Init, next=Next, inv= I@16:58:02.799
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
PASS #0: SanyParser                                               I@16:58:03.180
Parsing file /home/markus/src/TLA/_specs/models/mine/PageQueue/PageQueue.toolbox/w19/MC.tla
Parsing file /home/markus/src/TLA/_specs/models/mine/PageQueue/PageQueue.toolbox/w19/PageQueue.tla
Parsing file /tmp/TLC.tla
Parsing file /tmp/Integers.tla
Parsing file /tmp/Sequences.tla
Parsing file /home/markus/src/TLA/_specs/models/mine/PageQueue/PageQueue.toolbox/w19/SequencesExt.tla
[...]

PR #289, opened 2020-10-19
Fixing 285: step 1

konnov:
This PR does not completely fix the problem discovered by Markus in #285, as the spec uncovers multiple layers of issues with parsing advanced specs. This commit fixes the long-known issue with LAMBDAs. 

@Kukovec can you have a look at the changes and tell us if this solution potentially breaks your code. I can imagine that having a LET-definition in the operator parameters may come a bit unexpected.

### DIFF
--- a/tla-import/src/main/scala/at/forsyte/apalache/tla/imp/ExprOrOpArgNodeTranslator.scala
+++ b/tla-import/src/main/scala/at/forsyte/apalache/tla/imp/ExprOrOpArgNodeTranslator.scala
@@ -126,7 +126,7 @@ class ExprOrOpArgNodeTranslator(sourceStore: SourceStore,
       case defNode: OpDefNode if name == "LAMBDA" =>
         val decl = OpDefTranslator(sourceStore, context).translate(defNode)
         // e.g., LET Foo(x) == e1 in Foo
-        LetInEx(name, decl)
+        LetInEx(NameEx(name), decl)
 
       // an operator that is passed as an argument to another operator
       case _: OpDefNode =>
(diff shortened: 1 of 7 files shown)


tla-import/src/main/scala/at/forsyte/apalache/tla/imp/ExprOrOpArgNodeTranslator.scala
```diff
@@ -103,6 +115,29 @@ class ExprOrOpArgNodeTranslator(sourceStore: SourceStore,
...
+    // Instead of extending the IR with a new expression type, we simply introduce a local LET-IN definition.
+    // Although this is a well-defined expression in the IR, it does not correspond to a well-defined TLA+ expression.
+    // Hence, one has to take care of this, when printing the output to the user.
+    val name = opArgNode.getName.toString
+    opArgNode.getOp match {
+      // a lambda-definition is passed as an argument
+      case defNode: OpDefNode if name == "LAMBDA" =>
+        val decl = OpDefTranslator(sourceStore, context).translate(defNode)
+        // e.g., LET Foo(x) == e1 in Foo
+        LetInEx(NameEx(opArgNode.getName.toString), decl)
```
shonfeder, line 129: Since you already have `name` in scope from line 123:

```suggestion
        LetInEx(name, decl)
```

shonfeder (approved):

konnov:
That was too fast. One can also write something like `A(LAMBDA x: A(LAMBDA y: y, x), z)`. This means that we have to take care of nested lambdas.

shonfeder:
oop, sorry for the mistaken suggestion :)

konnov:
I will merge this one, as it contains useful bugfixes and handling of LAMBDAs. Then I will come back to #285

Kukovec:
> 
> 
> @Kukovec can you have a look at the changes and tell us if this solution potentially breaks your code. I can imagine that having a LET-definition in the operator parameters may come a bit unexpected.

I'm not sure which code you're referring to, but I can't think of any that would assume operator arguments to be leaf expressions off the top of my head, so this shouldn't break anything.

PR #291, opened 2020-10-19
ETC7: Add Type1 parser and lexer changes

shonfeder:
Broken out of #259

Originally authored by Igor Konnov <konnov@forsyte.at>

Also includes a white space change to `TlaType1.scala` that I'd not picked up before.


konnov (approved):
looks good!

PR #294, opened 2020-10-20
ETC-8: Add Substition and Unification modules

shonfeder:
Broken out of #259

 @konnov please see esp commits after https://github.com/informalsystems/apalache/commit/b7dcbe6f81df10ffff3c5e9f1202aaed356ad753, which are suggestions I'm proposing.


konnov (approved):
Looks good!

shonfeder:
Thanks! I also forgot to hit submit in a few questions, which I’ve just posted now. Most of them are just for my own edification :)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeUnifier.scala
```diff
@@ -0,0 +1,311 @@
+package at.forsyte.apalache.tla.typecheck.etc
+
+import at.forsyte.apalache.tla.typecheck._
+import at.forsyte.apalache.tla.typecheck.etc.TypeUnifier.CycleDetected
```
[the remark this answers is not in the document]
konnov: I don't think so. That might be something done automatically by Idea.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeUnifier.scala
```diff
@@ -0,0 +1,311 @@
...
+  *
+  * <p>This class is not designed for concurrency. Use different instances in different threads.</p>
+  *
+  * @author Igor Konnov
+  */
+class TypeUnifier {
+  // a partial solution to the unification problem is stored here during unification
+  private var solution: Map[Int, TlaType1] = Map.empty
+
+  def unify(substitution: Substitution, lhs: TlaType1, rhs: TlaType1): Option[(Substitution, TlaType1)] = {
```
konnov, line 20:
```suggestion
  /**
    * Try to unify lhs and rhs by starting with the given substitution. If successful, it returns Some(mgu, t),
    * where mgu is the solution set showing how to unify lhs and rhs and t is the type resulting from
    * successfully unifying lhs and rhs using mgu. Note that apart from variable substitution, our unification
    * also involves merging record types. When there is no unifier, it returns None.
    */
  def unify(substitution: Substitution, lhs: TlaType1, rhs: TlaType1): Option[(Substitution, TlaType1)] = {
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeUnifier.scala
```diff
@@ -0,0 +1,311 @@
...
+          val result =
+            if (isCyclic) {
+              None
+            } else {
+              computeClosureWhenAcyclic()
+              val substitution = new Substitution(solution)
+              Some((substitution, substitution(unifiedType)))
+            }
+
+          solution = Map.empty // let GC collect the solution map later
```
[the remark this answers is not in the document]
konnov: Good point. Can you add emptying on the other branches?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeUnifier.scala
```diff
@@ -0,0 +1,311 @@
...
+          result
+      }
+    } catch {
+      case _: CycleDetected =>
+        solution = Map.empty // let GC collect the solution map later
+        None
+    }
+  }
+
+  private def computeOptions(lhs: Option[TlaType1], rhs: Option[TlaType1]): Option[TlaType1] = {
```
konnov, line 49: This function is deprecated in the following commits. So let's keep it like that and then remove in the end.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeUnifier.scala
```diff
@@ -0,0 +1,266 @@
...
+
+    // if the fix-point is empty, there is no cycle
+    next.nonEmpty
+  }
+
+  // Compute the transitive closure of the solution, provided that the solution is acyclic.
+  // We separate the closure computation from the acyclicity checking, as acyclicity checking is a bit more efficient.
+  // (Though both methods are not optimized at all.)
+  //
+  // Shall we propagate the result in insert, as soon as they are available?
```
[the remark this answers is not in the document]
konnov: It was a side thought. I believe `isCyclic` and `computeClosureWhenAcyclic` could be computed more efficiently by using a textbook unification algorithm. But I did not have one at hand. The current solution is just a straightforward fixpoint computation.

konnov (approved):
Looks good. I quickly baked `TypeUnifier`, so if you have ideas about its improvements, feel free to PR.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeUnifier.scala
shonfeder: This is great. Thanks :)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeUnifier.scala
shonfeder: I don't see any sign of the function being deprecated on your initial prototyping branch: https://github.com/informalsystems/apalache/pull/244/files#diff-12379bb1bd9390b219a628ad755afb666f5fcb18f250fc21dc66d094353f84b2R68-R74

konnov: Oh, sorry. I thought of another version of `unify`. Yeah, `computeOptions` is a bit weird.

shonfeder: No worries! There was a deprecated `unify` method in here, but I already removed it. :)

> computeOptions is a bit weird

What do you think of my suggestion to just rework it to 

```scala
def computeRecordField(field: string, lhsRecord: RecT1, rhsRecord: Rect1): Option[TlaType1] ...
```

konnov: I think you are right. It was probably used in another piece of code that is gone by now. Please go ahead and rework it.

shonfeder:
Ok, I've made the discussed changes. @konnov if you could just take a peek at 564a34f and fc0e04d when you get a chance, and confirm that it makes sense to you, then we should be good to go here.

konnov:
The updates look great! Let's merge

Issue #286, opened 2020-10-16
[BUG] Error by TLA+ parser: Unexpected structure of EXCEPT


lemmy:
```bash
markus@avocado:~/src/TLA/_community/formal-methods-playground/tla/swim(master)$ /opt/apalache/bin/apalache-mc --cinit=ConstInit --init=Init --next=Next --length=50 --config=swim_stats_sim_v3_false_dead_pg1.cfg check  swim_stats.tla
# Tool home: /opt/apalache
# Package:   /opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar
# JVM args: -Xmx4096m -Djava.library.path=/usr/lib/x86_64-linux-gnu/jni/  -DTLA-Library=
#
# APALACHE version 0.7.2-RELEASE build v0.7.1-5-g5b85083
#
# WARNING: This tool is in the experimental stage.
#          Please report bugs at: [https://github.com/informalsystems/apalache/issues]

Checker options: filename=swim_stats.tla, init=Init, next=Next, inv= I@17:19:50.167
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
PASS #0: SanyParser                                               I@17:19:50.552
Parsing file /home/markus/src/TLA/_community/formal-methods-playground/tla/swim/swim_stats.tla
Parsing file /tmp/Naturals.tla
Parsing file /tmp/FiniteSets.tla
Parsing file /home/markus/src/TLA/_community/formal-methods-playground/tla/swim/FiniteSetsExt.tla
Parsing file /tmp/Sequences.tla
Parsing file /tmp/TLC.tla
Parsing file /home/markus/src/TLA/_community/formal-methods-playground/tla/swim/TLCExt.tla
Parsing file /tmp/Integers.tla
Error by TLA+ parser: Unexpected structure of EXCEPT              E@17:19:51.061
[...]

PR #302, opened 2020-10-27
Accept expressions as the first arguments of EXCEPT

konnov:
This PR fixes issue #286. The importer expected an operator application as the first argument of `EXCEPT`, but actually other kinds of expression may also occur there, e.g., the reference `@`. The solution is simply a one-liner. The job of checking expression compatibility for `EXCEPT` has been done by SANY already.


shonfeder (approved):

shonfeder:
LGTM

PR #304, opened 2020-10-28
fix #303: friendly messages on unexpected expressions (broken)

### LINKED ISSUE #303: [FEATURE] FlatLanguagePred and KeraLanguagePred should produce friendly messages
Currently, `FlatLanguagePred` and `KeraLanguagePred` spit out large blocks of problematic expressions when the input does not fit into the expected fragment. It is impossible to figure the root of the problem when running apalache on a real spec. They should point to the cause of the problem.

konnov:
Better diagnostic messages when preprocessing passes reject a piece of TLA+ spec. There is code duplication between `FlatLanguagePred` and `KeraLanguagePred`. But I would like to keep it, to make the predicates isolated, without introducing additional abstractions.


konnov:
I don't know what happened, but idea has committed all of my uncommitted files in the project. Need to roll back.

PR #306, opened 2020-10-28
fix #143: correct translation of chained substitutions in INSTANCE

### LINKED ISSUE #143: [BUG] Incorrect parsing on global module and INSTANCE name collision
**Describe the bug**
The spec `Paxos.tla` defines both a nullary operator `votes` and an instance `V` of `Voting` (defined in `Voting.tla`). 
The spec `Voting.tla` uses a VARIABLE `votes` and an operator 
```
Spec == Init /\ [][Next]_<<votes, maxBal>>
```

In the pass immediately after parsing, the internal representation of `V!Spec` is read as
```
V!Init() ∧ []([V!Next()]_(<<votes(), maxBal>>))
```
where `votes()` is an operator application (to the locally defined nullary `votes`) instead of a variable.

**To Reproduce**
 - The command-line parameters that you use to run the tool : 
`check --length=10 --debug --inv=Inv <dir>/Paxos.tla`
 - The problematic specification (if this cannot be submitted, try to come up with a minimal working example):
https://github.com/tlaplus/Examples/tree/master/specifications/Paxos

**Expected behavior**
What did you expect to see?
`V!Spec` as 
```
V!Init() ∧ []([V!Next()]_(<<V!votes, V!maxBal>>))
```
or similar.

konnov:
I have finally figured out substitutions in instantiated modules in SANY. They actually make sense. I just did not expect that the substitutions are stacked. Here is the fix for the long-standing issue.


Kukovec:
LGTM.

Kukovec (approved):

Issue #130, opened 2020-04-20
[BUG] Unroller produces stack overflow due to a bug in SANY


konnov:
Following #67, we tried to run the unstable version against the [light client spec](https://github.com/informalsystems/verification/tree/develop/spec/light-client) as follows:

```tla
apalache-mc check --config=MC3-PrecisionInv.cfg MC2nodes3blocks.tla
```

`Blockchain.tla` contains a recursive operator `PowerOfSet`, which we did not have produce annotations for. The tool should have complained about missing annotations. Instead, it throws a stack overflow:

```tla
PASS #1: ConfigurationPass                                        I@11:30:37.199
  > Loading TLC configuration from MC2nodes3blocks.tla            I@11:30:37.201
  > Could not parse TLC configuration in MC2nodes3blocks.tla: unexpected character E@11:30:37.238
PASS #2: UnrollPass                                               I@11:30:37.297
  > Unroller                                                      I@11:30:37.299
PASS #3: InlinePass                                               I@11:30:37.325
  > InlinerOfUserOper                                             I@11:30:37.327
(Please report an issue at: [https://github.com/konnov/apalache/issues],java.lang.StackOverflowError)
Unhandled exception                                               E@11:30:37.576
java.lang.StackOverflowError: null
	at scala.collection.immutable.HashMap.get(HashMap.scala:53)
	at at.forsyte.apalache.tla.lir.transformations.standard.InlinerOfUserOper.$anonfun$transform$1(InlinerOfUserOper.scala:34)
```

Kukovec:
The issue is not with `Unroller` but with `INSTANCE`-imported recursive operators not being marked recursive (i.e. having `.isRecursive = false`). See attached minimal example.

Kukovec:
[test.txt](https://github.com/konnov/apalache/files/4503258/test.txt)

konnov:
@Kukovec do you know whether this has been fixed?

konnov:
This is still an issue. For the moment, we will introduce a workaround. This should have been fixed in SANY, but we have to check that.

Kukovec:
Redirecting this to @konnov, since it's not an Unroller issue.

konnov:
We should reproduce it again and submit to tlaplus, if the bug is still there.

konnov:
We are going to stop supporting recursive operators and functions. The users should use folds instead. Closing this issue.

thpani:
Reopening, there's still clean-up notes for this:

[...]

PR #315, opened 2020-10-30
Workaround for issue #130

konnov:
This PR introduces a temporary workaround for issue #130, which needs a fix in SANY

### DIFF
--- a/tla-import/src/main/scala/at/forsyte/apalache/tla/imp/ModuleTranslator.scala
+++ b/tla-import/src/main/scala/at/forsyte/apalache/tla/imp/ModuleTranslator.scala
@@ -2,6 +2,7 @@ package at.forsyte.apalache.tla.imp
 
 import at.forsyte.apalache.tla.imp.src.SourceStore
 import at.forsyte.apalache.tla.lir._
+import com.typesafe.scalalogging.LazyLogging
 import tla2sany.semantic.{InstanceNode, ModuleNode, OpDefNode}
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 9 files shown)


tla-import/src/main/scala/at/forsyte/apalache/tla/imp/ModuleTranslator.scala
```diff
@@ -57,6 +58,10 @@ class ModuleTranslator(sourceStore: SourceStore) {
           decl.isRecursive = updatedDecl.isRecursive
           decl.body = updatedDecl.body
 
+          // temporary bugfix for the issue #130. Remove as soon as it is fixed in tla2tools.
```
shonfeder, line 61:
```suggestion
          // TODO(igor) temporary bugfix for the issue #130. Remove as soon as it is fixed in tla2tools.
```

tla-import/src/main/scala/at/forsyte/apalache/tla/imp/ModuleTranslator.scala
```diff
@@ -114,6 +119,42 @@ class ModuleTranslator(sourceStore: SourceStore) {
 
     node.getInstances.toList.foldLeft(newCtx)(eachInstance)
   }
+
+  /**
+    * This is a temporary bugfix for the issue 130: https://github.com/informalsystems/apalache/issues/130
+    * Once it is fixed in SANY, remove this method. This method does not detect mutually recursive operators,
```
shonfeder, line 125:
```suggestion
    * TODO(igor) Once it is fixed in SANY, remove this method. This method does not detect mutually recursive operators,
```

shonfeder (approved):
LGTM

Suggested adding `TODO` string to comments to make them easier to spot and find in the future, but that's no important really.

Issue #318, opened 2020-11-04
[FEATURE] Seed the randomness for deterministic model based test generation 


ebuchman:
**Is your feature request related to a problem? Please describe.**

Currently every time we regenerate tests we get new ones.

**Describe the solution you'd like**

A way to regenerate the same set of tests, typically in a case where we've changed some data structure on the Rust side but otherwise want to be testing the same scenario.

andrey-kuprianov:
To do the testing in the proper way, we need to have it both ways:
 - allow Z3 to choose it's random seed, and then provide it to the user in the output (for exploratory testing)
 - accept the seed in the user input, and pass it to Z3 (for test reproducibility)

konnov:
I am reading the z3 docs, and the developers say that z3 should be deterministic. The default seed is 0. Are we sure that we get non-determinism from z3? @andrey-kuprianov,  you can produce the smt log with `--debug`, run it with z3 several times and see, whether the results are different.

I will add an option to set the seed. But it may happen that non-determinism is also on our side.

andrey-kuprianov:
well, the outcome is that smt logging seems to be broken. first, it's `log0.smt` that's being produced. Second, after some initial valid SMT commands, it contains some printout, which is not an SMT-LIB:

```
(error "line 44536 column 1: invalid command, '(' expected")
(error "line 44538 column 7: unexpected character")
(error "line 44538 column 19: unexpected character")
(error "line 44550 column 8: unexpected character")
```

```
(assert (= $C$1994 (>= $C$1776 $C$1993)))
; } IntCmpRule returns $C$1994 [1995 arena cells])
;; assert $C$1994
(assert $C$1994)
(check-sat)
;; sat = true
$C$0 = FALSE
$C$1 = TRUE
$C$2 = {FALSE, TRUE}
$C$3 = Nat
```
```

konnov:
That's weird. It used to work pretty well. Are you feeding this back to z3?

konnov:
`log0.smt` is fine. There are multiple solver instances in the parallel version.

konnov:
Oh wait. The last four lines are the output of the sat model. Just delete all these things.

konnov:
I have added #320 on that issue. But if you remove the model output at the end of the log, it should work.

konnov:
That command seems to produce stable results:

konnov:
[...]

PR #321, opened 2020-11-06
Implement 318: random seeds in z3

konnov:
This PR implements #318:

 - [x] setting z3's random seeds to 0 by default

 - [x] letting the user to choose the random seed by defining `smt.randomSeed` in `tuning.properties`, which can be passed via `--tuning=tuning.properties`.

The current conclusion is that non-determinism is caused by JRE8. The solution is to use JRE9.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -17,18 +17,33 @@ import com.microsoft.z3.enumerations.Z3_lbool
...
   private val logWriter: PrintWriter = initLog()
 
+  // dump the configuration parameters in the log
+  // set the global configuration parameters for z3 modules
+  Z3SolverContext.RANDOM_SEED_PARAMS.foreach {
+    p =>
+      Global.setParameter(p, config.randomSeed.toString)
+      logWriter.println(";; %s = %s".format(p, config.randomSeed))
+//    the following fails with an exception: java.lang.NoSuchFieldError: value
+//      logWriter.println(";; %s = %s".format(p, Global.getParameter(p)))
```
shonfeder, line 44: Is there value to this comment? Looks me like it might just be a left over...

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -175,14 +190,18 @@ class Z3SolverContext(debug: Boolean = false, profile: Boolean = false) extends
     log(s"(assert ${z3expr.toString})")
     z3solver.add(z3expr.asInstanceOf[BoolExpr])
 
-    if (profile) {
+    /**
```
shonfeder, line 193: Seems like we should just remove this?

shonfeder (approved):

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
konnov: Right. I wanted to print out the value as reported by z3 itself. But the current version produces an exception. So I left it for the future, or just as documentation for the future if someone (like me) wants to write the same code again :-)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
konnov: Yes

PR #326, opened 2020-11-09
fix #320: print the SMT model in comments rather than in plain output

### LINKED ISSUE #320: [BUG] write the SMT model as comments in the SMT log
The current version dumps the SMT log (when the option `--debug` is passed). If there is a counterexample, it also decodes the SMT model and dumps it in the log. We should write the model in the SMT comments, not like plain constants in SMT, which produce invalid SMT.

konnov:
This is a two-line fix


shonfeder (approved):
LGTM

PR #308, opened 2020-10-29
WIP Improved cover analysis output

Kukovec:
For example, given
```
--------------------------- MODULE test -------------------------------------

EXTENDS Integers

VARIABLE x,y

Init == x = 1 /\ y = 1

A1 == x' = 1 \/ x' = 2
A2 == x' = 3 \/ x' = 4
B1 == y' = 1 \/ y = 2
B2 == x' = 1 /\ (y' = 3 \/ y' = 4)

Next == 
    \/ A1 /\ B1
    \/ A1 /\ B2
    \/ A2 /\ B1
    \/ A2 /\ B2

=============================================================================
```

the new output is
```
test.tla:7:9-7:13   : Init may update x
test.tla:7:18-7:22  : Init may update y
test.tla:9:7-9:12   :   A1 may update x
test.tla:9:17-9:22  :   A1 may update x
test.tla:10:7-10:12 :   A2 may update x
test.tla:10:17-10:22:   A2 may update x
test.tla:11:7-11:12 :   B1 may update y
test.tla:12:18-12:23:   B2 may update y
test.tla:12:28-12:33:   B2 may update y

test.tla:7:9-7:22   : Init always updates: x, y
test.tla:9:7-9:22   :   A1 always updates: x
test.tla:10:7-10:22 :   A2 always updates: x
test.tla:11:7-11:21 :   B1 sometimes does not update: y 	 [WARNING!] 
test.tla:12:7-12:34 :   B2 always updates: y
```

Closes #298.

PR #327, opened 2020-11-10
Passes reordered + keep (C)initPrime after InlinePass

Kukovec:
Changes the order of passes, so `CoverAnalysisPass` comes before inlining.


Kukovec:
Once this passes I can merge it into #308 and close that.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/InlinePassImpl.scala
```diff
@@ -65,10 +65,17 @@ class InlinePassImpl @Inject()(val options: PassOptions,
     // Remove the operators that are not needed,
     // as some of them may contain higher-order operators that cannot be substituted
     val relevantOperators = NormalizedNames.userOperatorNamesFromOptions(options).toSet
```
konnov, line 67: why don't you extend `userOperatorNamesFromOptions` with the operators for `cinit`, `initPrime`, etc.? These names should keep the relevant operators, so later passes would not remove them.

konnov (approved):
One comment. The rest looks good to me.

shonfeder:
Hi @Kukovec! Just checking in here, as it looks like this PR might be nearly ready for merge, which would help unblock other WIP. LMT if there's anything I can do to help it along :)

PR #328, opened 2020-11-10
Fix 208: support advanced TLC configuration files

### LINKED ISSUE #208: [BUG] Could not parse TLC configuration in BlockingQueue.cfg: string matching regex '\z' expected but '{' found

konnov:
Fix #208. This PR is larger than I expected. It took me a day of work instead of a few hours.

 - [x] detailed manual on TLC configuration options in `docs/tlc-config.md`
 - [x] extended parser for `CONSTANTS`
 - [x] ignoring unsupported features without failing on them
 - [x] fix the overrides in `ConfigurationPassImpl`, which was accidentally disabled by previous refactorings
 - [x] added integration tests for assignments and replacements


tla-import/src/main/scala/at/forsyte/apalache/io/tlc/config/TlcConfig.scala
```diff
@@ -20,13 +25,60 @@ case class InitNextSpec(init: String, next: String) extends BehaviorSpec
...
+  def toTlaEx: TlaEx
+}
+
+/**
+  * A TLC model value, that is, a unique identifier that is treated as an uninterpreted constant.
+  * @param name the name of a model value
+  */
+case class ConfigModelValue(name: String) extends ConfigConstExpr {
+  override def toTlaEx: TlaEx = ValEx(TlaStr(name))
+}
```
andrey-kuprianov, line 50: I see a problem here: from the user perspective, model values and string constants are different entities; However, as implemented, a model value A, and a string value "A" result in the same LIR expression. What about encoding model values with some special symbol, e.g. $?

```suggestion
case class ConfigModelValue(name: String) extends ConfigConstExpr {
  override def toTlaEx: TlaEx = ValEx(TlaStr(name))
}
```
```suggestion
case class ConfigModelValue(name: String) extends ConfigConstExpr {
  override def toTlaEx: TlaEx = ValEx(TlaStr("$" + name))
}
```

docs/tlc-config.md
```diff
@@ -0,0 +1,157 @@
...
+
+// Recent feature: https://lamport.azurewebsites.net/tla/current-tools.pdf
+// APALACHE IGNORES THIS CONFIGURATION OPTION.
+Postcondition ::=
+    "POSTCONDITION" ident
+
+// Very recent feature: https://github.com/tlaplus/tlaplus/issues/485
+// APALACHE IGNORES THIS CONFIGURATION OPTION.
+Alias ::=
+    "ALIAS" ident
```
andrey-kuprianov, line 148: I doubt it's a good idea to just silently ignore so many configuration parameters. The user will expect certain behavior, and the tool will probably just fail in another place, e.g. due to missing invariant, not explaining to the user that actually his PROPERTIES definition was ignored.

What about adding to the parser smth. like two string variables, warnings, and errors? And for whatever not supported, add lines to one of those? Later, when the parsed configuration is imported, check on these, and if errors are non-empty -> abort, if warnings are non-empty -> print them out.

andrey-kuprianov (changes requested):
Great work, @konnov!

Only a couple of minor comments from my side; see above.
What would be also nice to have is some more tests, in particular to have tests that include the ignored config features.

docs/tlc-config.md
konnov: Well, `ConfigurationPassImpl` warns the user about the ignored properties. However, I do not want to make the configuration model too complex, just to complain about unsupported options later. We can add warnings right in the parser. Apart from `PROPERTY`, many other ignored options are TLC optimizations, which we will most probably never support.

tla-import/src/main/scala/at/forsyte/apalache/io/tlc/config/TlcConfig.scala
konnov: Good point, I will add a prefix like `ModelValue_`. I see model values more like a hack, as they are semantically just uninterpreted constants, not so different from strings in TLA+.

andrey-kuprianov: agree; it's a hack:) but still good to differentiate between the two types, such that the user doesn't run into some unexpected errors.

docs/tlc-config.md
andrey-kuprianov: Right, warnings right in the parser is better, I only was not sure whether this is possible from the architectural point of view.

Btw, there is a warning issued about PROPERTIES in the importer, but this is already outdated, the leftover from the previous ConfigParser implementation, is it right? then it should be probably deleted?

shonfeder:
Hmm. The merge conflicts in the integration tests should be easy to resolve, but that's a warning sign IMO that we should probably get this merged in lest we hit more troublesome conflicts.

Let me know if I can help here.

tla-import/src/main/scala/at/forsyte/apalache/io/tlc/config/TlcConfig.scala
andrey-kuprianov: Returning back to this: I think adding the `ModelValue_` prefix would be an overkill. Imagine Apalache returning counterexamples like `your_set = {"ModelValue_A", "ModelValue_B", "ModelValue_C"}` 

I would really prefer to have strings and model values differentiated, but in a non-overwhelming way, like with a small prefix: `$`, `#`, `_` ... But even leaving the strings clashing with the model values would be still better than the `ModelValue_` prefix.

docs/tlc-config.md
lemmy: Don't forget about [typed model values](https://tla.msr-inria.inria.fr/tlatoolbox/doc/model/model-values.html).

konnov: I have added warnings for the options that are completely ignored by the parser such as "VIEW", "SYMMETRY", etc. We might want to extend the parser later, to propagate these options, but I do not see any value in doing that for now. The "PROPERTIES" are propagated and then ignored by the model checker with a warning, as we like to implement liveness checking later.

konnov: > Don't forget about typed model values.

From the perspective of the config parser, there is no difference between model values and typed model values, right?

tla-import/src/main/scala/at/forsyte/apalache/io/tlc/config/TlcConfig.scala
konnov: We are already using `$` when renaming variables. I believe lots of unusual ASCII symbols confuse users a lot. Here it is clear that the string is coming from the config as a model value :)

PR #329, opened 2020-11-11
ETC-11: Add TypeChecker command

shonfeder:
Also adds type checking passes into parser

This adds the final components of #244


tla-import/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporter.scala
```diff
@@ -1499,6 +1499,27 @@ class TestSanyImporter extends FunSuite {
     assert(aOfM.isInstanceOf[TlaOperDecl] && aOfM.asInstanceOf[TlaOperDecl].formalParams.size == 1)
   }
 
+  test("substitution in INSTANCE") {
```
shonfeder, line 1502: @konnov is this related to the Etc work? Is it because of a regression that was introduced in the etc stuff, or is it a misplaced change?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
```diff
@@ -0,0 +1,28 @@
...
+import at.forsyte.apalache.tla.typecheck.etc._
+
+/**
+  * The API to the type checker. It first translates a TLA+ module into EtcExpr and then does the type checking.
+  *
+  * @author Igor Konnov
+  */
+class TypeCheckerTool {
+  def check(listener: TypeCheckerListener, module: TlaModule): Boolean = {
+    // TODO: remember to type check the assumptions!
```
shonfeder, line 13: @konnov Is this something we should implement as part of this PR, or an additional feature we want to add later?

konnov: This is an additional feature. We have not been paying attention to `ASSUME(...)` so far, but we have to do it at some point.

tla-import/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporter.scala
konnov: I think it was just a quick fix/test for a type-checking benchmark that caused a failure in the parser.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
```diff
@@ -0,0 +1,69 @@
+package at.forsyte.apalache.tla.typecheck.etc
```
shonfeder, line 1: @konnov Seems like we should also add tests for the many Tla modules that are part of this PR. 

Do you think it makes sense to do those as integration tests via our mdx setup? Or should I put them here?

konnov: The PR has several `*.tla` files that can be used for integration testing. Yes, it would be good to have them integrated in the mdx setup.

shonfeder: So currently type checking is failing on these files:

konnov: Yes, it was definitely failing on `SlidingPuzzlesTyped` and `QueensTyped`. `QueensTyped` definitely require another operator that turns a function into a sequence. I don't remember about `TwoPhaseTyped.tla`; that is a bit surprising, but something was probably tricky there. Let's not fix these ones.

konnov (approved):
Looks good to me!

test/tla/SlidingPuzzlesTyped.tla
```diff
@@ -0,0 +1,69 @@
...
+
+(***************************************************************************)
+(* Given a position and a unit translation vector return a pair of         *)
+(* pieces, before and after translation in opposite this vector direction  *)
+(***************************************************************************)      
+move(p, d) == "(<<Int, Int>>, <<Int, Int>>) => <<Int, Int>>" ##
+              LET s == "<<Int, Int>>" ##
+                    <<p[1] + d[1], p[2] + d[2]>>
+                  pc == ChooseOne(board, LAMBDA pc : s \in pc)
+              IN <<pc, {<<q[1] - d[1], q[2] - d[2]>> : q \in pc}>>
```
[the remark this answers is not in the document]
konnov: Are you talking about the type of `move(p, d)` or the type of `s`?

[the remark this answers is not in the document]
shonfeder: The type of `move`. 

It's annotated as `"(<<Int, Int>>, <<Int, Int>>) => <<Int, Int>>"` but the `LET` binding evaluates to 

```
<<pc, {<<q[1] - d[1], q[2] - d[2]>> : q \in pc}>>
```

`pc` is the result of apply `ChooseOne`, whose return type is `Set(<<Int, Int>>)`, and the second element of the tuple is also evidently a `Set(<<Int, Int>>)`. So I'm not sure where the `<<Int, Int>>` in the return type of `move`s annotation is supposed to come in. But also, from the way `move` is actually used, it looks like it should yield an  `<<Int, Int>>`. So that makes me wonder if I'm just misreading the spec in some basic way?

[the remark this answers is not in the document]
shonfeder: lol. I was misreading the usage of `move`. The type annotation was just mistaken :) I corrected it and things are working as expected now. :+1:

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
shonfeder: Opened an issue to track this #330

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
shonfeder: I fixed the errors on `SlidingPuzzles` and `TwoPhased` and added a failing test for the `Queens` spec, which we can fix once support for the needed sub-typing or coercion is supplied.

PR #334, opened 2020-11-18
Fix 333: The optimizer in Z3SolverContext killed the basic assumptions

konnov:
This is somewhat embarrassing. The optimization in `Z3SolverContext` replaced the basic assertions about cells 0 and 1 with false and true. All of a sudden the solver lost the knowledge that `$C$0 = FALSE` and `$C$1 = TRUE`. Interestingly, this bug popped up in very rare cases, as the optimizer was rewriting `$C$0` and `$C$1` most of the time.


shonfeder (approved):
LGTM

PR #335, opened 2020-11-19
grammar fix

josef-widder:
I believe the old was wrong. Not sure. Should be checked.


shonfeder:
I don't think the grammar was incorrect, but I can see how this could help clarify the phrasing. 

LGTM

shonfeder (approved):

konnov:
Yes! That reads much better!

PR #339, opened 2020-11-26
Manual on the standard TLA+ operators

konnov:
This is the first part of the manual on TLA+ operators with operational semantics in mind. This is a large PR, but it is text, not code.

Feedback on the content, language, and readability is welcome!


docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+
+```tla
+VARIABLES x, y
+Init == x == 0 /\ y == 0
+Next ==
+    \/ x >= 0 /\ y' = x /\ x' = x + 1
+    \/ x <= 0 /\ y' = -x /\ x' = -(x + 1)
+```
+
+As you can see, the operator `Next` is non-deterministic: The both clauses may
```
shonfeder, line 387:
```suggestion
As you can see, the operator `Next` is non-deterministic since both clauses may
```

shonfeder (commented):
I'm about halfway done reviewing, but submitting feedback now.

I'm finding it very illuminating! Thanks for this awesome work :)  I'll be done reviewing the rest by tomorrow.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+By following [semantics of conjunction](./booleans.md), we see that `/\` is
+evaluated from left-to-right. The left-hand side equality `x = 0` is treated
+as an assignment to `x`, since `x` has value `Null` in the partial state 
+`[x |-> Null, y |-> Null]`. The expression `x = 0` effects in the partial state 
+`[x |-> 0, y |-> Null]`. Likewise, the right-hand side equality `y = 0` is also
+treated as an assignment to `y`. Hence, the expression `y = 0` effects in the
+partial state `[x |-> 0, y |-> 0]`.
+
+Let's see how to evaluate the body of the operator `Next` starting with the
+partial state `[x |-> 3, y |-> 3, x' |-> Null, y' |-> Null]`:
```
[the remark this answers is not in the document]
Kukovec: Agreed, this left-to-right evaluation, if you really want to be pedantic, is the following relation, parameterized by the initial state:
Take `s_1` as the initial state, `s_Null` as the completely unassigned partial state and the syntax `s_i --R--> s_j` to mean `R(s_i,s_j)`:
```
s_Null --R(s_1)--> [x |-> 4, y |-> Null] 
[x |-> 4, y |-> Null]  --R(s_1)--> [x |-> 4, y |-> 4] 
```
With the property that `\A s_1,s_2 \in States: Next(s_1,s_2) <=> R(s_1)*(s_Null, s_2)` (where * is the transitive closure).

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+
+ 1. `GUESS S` can be simply the user input. In this case, the user resolves
+ non-determinism.
+
+ 1. `GUESS S` can be controlled by an adversary, who is trying to break the
+ system.
+
+ 1. `GUESS S` can pick an element by calling a random number generator.
+ However, note that RNG is a very special way of resolving non-determinism: It
+ assumes probabilistic distribution of elements (usually, it is close to the
```
Kukovec, line 254: Technically, RNG is also deterministic.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+
+As you see, there are multiple sources of non-determinism. With `GUESS S` we can
+model all of them. As TLA+ does not introduce special primitives for different
+kinds of non-determinism, neither do we fix any implementation of `GUESS S`.
+
+**Halting.** Note that `GUESS {}` halts the evaluation. What does it mean? The
+evaluation cannot continue in the current partial state. It does not imply that
+we have found a deadlock in our TLA+ specification. It simply means that we
+made wrong choices on the way. If we like to enumerate all possible state
+successors, like TLC does, we have to backtrack. In general, the course of
```
Kukovec, line 269:
```suggestion
made wrong choices on the way. If we would like to enumerate all possible state
successors, like TLC does, we have to backtrack. In general, the course of
```

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+
+<a name="nondetExists"></a>
+### Non-determinism in `\E x \in S: P`
+
+We only have to consider the following case: `\E x \in S: P` is evaluated in a
+partial state `s`, and there is a primed state variable `y'` that satisfies two
+conditions:
+
+ 1. The predicate `P` refers to `y'`, that is, `P` has to assign a value to `y'`.
+ 2. The value of `y'` is not defined yet, that is, `s.y' = Null`.
```
Kukovec, line 281: Continuation of the above, don't treat `s'` as if it's a variable and part of the same state as `s.y` , it's a different state's value `s_2.y`.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+It is the job of `GUESS S` to tell us what value of `x` should be
+evaluated. There are three possible outcomes:
+
+ 1. Predicate `P` evaluates to `TRUE` when using the provided value of `x`.
+ In this case, `P` assigns the value of an expression `e` to `y'` as soon as
+ the evaluator meets the expression `y' = e`.
+ The evaluation may continue.
+ 1. Predicate `P` evaluates to `FALSE` when using the provided value of `x`.
+ Well, that was a wrong guess. According to our semantics, the evaluation
+ halts. See the above discussion on "halting".
```
Kukovec, line 303: The above discussion does not go into depth on this. What happens is that the next guess is made as follows:
`GUESS S \{x} IN P ` until `S` is exhausted. Only then is the entire `\E` expression evaluated as `FALSE`.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+```tla
+  \/ P_1
+  \/ P_2
+  ...
+  \/ P_n
+```
+
+Assume that we evaluate the disjunction in a partial state `s`. Further,
+let us say that `Unassigned(s)` is the set of state variables that evaluate
+to `Null` in `s`. For every `P_i` we construct the set of state variables
```
Kukovec, line 344: Again, find a better way to differentiate between `s.x` and `s_Next.x` (aka `s.x'`). This absorption of primed variable names into partial states is very confusing at times.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+ 2. Two clauses disagree on the set of variables to be assigned.
+    Formally, there is a pair `i, j \in 1..n` that satisfy the inequality:
+    `Use_i \intersect Unassigned(s) /= Use_j \intersect Unassigned(s)`.
+    In this case, the specification is ill-structured. TLC would
+    raise an error when it found a partial state like this.
+    Apalache would detect this problem when preprocessing the specification.
+ 3. The clauses do not assign values to the primed variables.
+    Formally, `Use_i \intersect Unassigned(s) = {}` for `i \in 1..n`.
+    This is the deterministic case. It can be evaluated by using the
+    deterministic semantics of [Boolean operators](./boolean.md).
```
Kukovec, line 360: Some pedanticism, but 3 isn't mutually exclusive with 1.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+conditionals](./conditionals.md).
+
+Consider an `IF-THEN-ELSE` expression to be evaluated in a partial state `s`:
+
+```tla
+IF A THEN B ELSE C
+```
+
+Here we assume that both `B` and `C` produce Boolean results and `B` and `C`
+refer to a primed variable `y'` that is undefined in `s`. Otherwise, the
```
Kukovec, line 429:
```suggestion
refer to at least one primed variable `y'` that is undefined in `s`. Otherwise, the
```

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+This operator is equivalent to the following disjunction:
+
+```tla
+\/ P_1 /\ e_1
+\/ P_2 /\ e_2
+...
+\/ P_n /\ e_n
+```
+
+_Similar to IF-THEN-ELSE, we do not recommend you using CASE for expressing
```
Kukovec, line 477:
```suggestion
_Similar to IF-THEN-ELSE, we do not recommend using CASE for expressing
```

Kukovec (commented):
Too much to review in one session, this is review 1/n.

docs/lang/control-and-nondeterminism.md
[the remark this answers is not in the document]
shonfeder: Yes, that's helps clarify! It is more dense, but I think it's worth being pedantic here. iiuc, we actually have two different levels of state transitions at play: (1) the incremental updating of the "partial state" (which is actually building the context of variable bindings, seemingly very similar to building the context for unification), and (1) the update of the actual state variables of the computation being modeled with reference to that constructed context.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+ - If `v_1` and `v_2` are functions, tuples, records, or sequences,
+    then `e_1 = e_2` evaluates to `TRUE`
+    if and only if the following expression evaluates to `TRUE`:
+
+    ```tla
+      DOMAIN v_1 = DOMAIN v_2 /\ \A x \in DOMAIN v_1: v_1[x] = v_2[x]
+    ```
+
+ - In other cases, `e_1 = e_2` evaluates to `FALSE`. TLC and Apalache report an
+   error, as the values have incomparable types.
```
shonfeder, line 206: So this be

```suggestion
 - In other cases, `e_1 = e_2` evaluates to `FALSE` if the values have comparable types.
 - TLC and Apalache report an error, if the values have incomparable types.
```

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+   
+**Determinism:** Deterministic, unless `e_1` has the form `x'`, which can be
+interpreted an assignment to the variable `x'`.  For the non-deterministic
```
shonfeder, line 209:
```suggestion
interpreted as an assignment to the variable `x'`.  For the non-deterministic
```

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+_This operator causes a lot of confusion. Read carefully!_
+
+**Notation:** `CHOOSE x \in S: P`
+
+**LaTeX notation:** ![choose-bounded](./img/choose-bounded.png)
+
+**Arguments:** Three arguments: a variable name, a set, and an
+expression.
+
+**Effect:** This operator is a magical device by the definition, no kidding. It
```
shonfeder, line 298: I don't understand this sentence.

shonfeder (commented):
I'm up to records now... :)

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+
+ - `\A x \in S: P` evaluates to `TRUE`, if for every element `e` of `S`, the
+   expression `P` evaluates to `TRUE` against the binding `[x |-> e]`.
+ - Conversely, `\A x \in S: P` evaluates to `FALSE`, if there exists an element
+   `e` of `S` that makes the expression `P` evaluate to `FALSE` against the
+   binding `[x |-> e]`.
+
+ _Importantly, when `S = {}`, the expression `\A x \ in S: P` evaluates to
+ `TRUE`, independently of what is written in `P`. Likewise, when `{x \in S: P}
+ = {}`, the expression `\A x \ in S: P` evaluates to `TRUE`._
```
Kukovec, line 40:
```
Likewise, when `{x \in S: P} = {}`, the expression `\A x \ in S: P` evaluates to `TRUE`.
```
This is not true, take `S == {1}`, `P == x = 2`. While it is true that `{x \in {1}: x = 2}` equals `{}`, the formula `\A x \in {1}: x = 2` is `FALSE`. 
If you want to state quantification as filtering, you have the equivalence between `{x \in S: P} = S` and ``\A x \ in S: P`.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+   
+**Determinism:** Deterministic.
+
+**Errors:** Pure TLA+ does not restrict the operator arguments.  TLC flags a
+model checking error, if `S` is infinite.  Apalache produces a static type
+error, if the type of elements of `S` is not compatible in the context of `P`
+when an element of `S` is bound to `x`.
```
Kukovec, line 47: This sentence is too confusing (imprecise).

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+syntax to unpack variables from a Cartesian product, see [Tuples](./tuples.md).
+For instance, one can write `\A <<x, y>> \in S: P`. In this case, for every
+element `e` of `S`, the variable `x` is bound to `e[1]` and `y` is bound to
+`e[2]`. The predicate `P` is evaluated against this binding.
+
+Moreover, instead of introducing one variable, one can quantify over several
+sets. For instance, you can write: `\A x \in S, y \in T: P`. This form is
+simply syntax sugar for the form with nested quantifiers: `\A x \in S: \A y
+\in T: P`.
+
```
Kukovec, line 59: \+ Similarly, `\A x,y \in S: P` is syntax sugar for `\A x \in S : \A y \in S: P`.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+ - `\E x \in S: P` evaluates to `TRUE`, if there is an element `e` of `S`
+    that makes the expression `P` evaluate to `TRUE` against the binding
+    `[x |-> e]`.
+ - Conversely, `\E x \in S: P` evaluates to `FALSE`, if for all elements
+   `e` of `S`, the expression `P` evaluate to `FALSE` against the
+   binding `[x |-> e]`.
+
+ _Importantly, when `S = {}`, the expression `\E x \ in S: P` evaluates to
+ `FALSE`, independently of what is written in `P`. Likewise, when `{x \in S: P}
+ = {}`, the expression `\A x \ in S: P` evaluates to `FALSE`._
```
Kukovec, line 111: This is true, but I don't know why you're giving an example for `\A` in the section on `\E`.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+
+ As you probably have noticed, `\E x \in S: P` is equivalent to `~(\A x \in S:
+ P)`, and `\A x \in S: P` is equivalent to `~(\E x \in S: P)`. This is called
+ _duality_ in logic. But take care!  If `P` contains primes inside, `\E x \in
```
Kukovec, line 115: Again, the categorization with "contains primes" is incomplete.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+ S: P` may work as a non-deterministic assignment. In this case, duality does
+ not work anymore!  See [Control Flow and Non-determinism].
+   
+**Determinism:** Deterministic, unless `P` contains primes.
```
Kukovec, line 119: See above.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+**Advanced syntax:** Instead of a single variable `x`, you can use the tuple
+syntax to unpack variables from a Cartesian product, see [Tuples](./tuples.md).
+For instance, one can write `\E <<x, y>> \in S: P`. In this case, for every
+element `e` of `S`, the variable `x` is bound to `e[1]` and `y` is bound to
+`e[2]`. The predicate `P` is evaluated against this binding.
+
+Moreover, instead of introducing one variable, one can quantify over several
+sets. For instance, you can write: `\E x \in S, y \in T: P`. This form is
+simply syntax sugar for the form with nested quantifiers: `\E x \in S: \E y
+\in T: P`.
```
Kukovec, line 136: See comment on `\E`

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+
+ - In other cases, `e_1 = e_2` evaluates to `FALSE`. TLC and Apalache report an
+   error, as the values have incomparable types.
+   
+**Determinism:** Deterministic, unless `e_1` has the form `x'`, which can be
+interpreted an assignment to the variable `x'`.  For the non-deterministic
+case, see [Control Flow and Non-determinism].
+
+**Errors:** Pure TLA+ does not restrict the operator arguments. TLC flags a
+model checking error, `e_1` and `e_2` evaluate to incomparable values.
```
Kukovec, line 213:
```suggestion
model checking error, if `e_1` and `e_2` evaluate to incomparable values.
```

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+Apalache produces a static type error, if the type of `e_1` and `e_2` do not
```
Kukovec, line 214:
```suggestion
Apalache produces a static type error, if the types of `e_1` and `e_2` do not
```

docs/lang/logic.md
Kukovec: Agreed, tone down the theatrics in a manual.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+element `e`.
+
+Did we say "an algorithm"? Yes! `CHOOSE x \in S: P` is deterministic.  If you
+give it two equal sets and two equivalent predicates, it will give you the
+same value. More formally, the only known property of `CHOOSE` is as follows:
+
+```tla
+  {x \in S: P} = {y \in T: Q} =>
+      (CHOOSE x \in S: P) = (CHOOSE y \in T: Q)
+```
```
Kukovec, line 310: Is this really true? Take `S == {1}` and `T == {1,2}`, then `P == TRUE` and `Q == y = 1`. Obviously `S /= T` and `~(P <=> Q) `, even though `{x\ in {1}: TRUE} = {y \in {1,2}: y = 1}`. 
My understanding is that for `P <=> Q` it holds that `(CHOOSE x \in S: P) = (CHOOSE x \in S: Q)` and for `S = T`, then `(CHOOSE x \in S: P) = (CHOOSE x \in T: P)`, but I'm not sure you can combine those two properties.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+
+Importantly, when `{x \in S: P} = {}`, the expression `CHOOSE x \ in S: P`
+    evaluates to an undefined value.
+
+How does it work? We don't know. Nobody knows. Maybe it returns the first
+element of the set? Well, sets are not ordered, so there is no first element.
```
Kukovec, line 316: Again, a bit theatrical.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+
+Why shall you use it? Actually, you should not. Unless you have no other
```
Kukovec, line 318:
```suggestion
Why should you use it? Actually, you should not. Unless you have no other
```

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+Apalache picks a set element that satisfies the predicate `P`, but it does not
+guarantee the repeatability property of CHOOSE. It does not guarantee
+non-determinism either. Interestingly, this behavior does not really make a
+difference for the use cases 1 and 2. If you believe that this causes a problem
+in your specification, [open an issue...]
+
+**Errors:** Pure TLA+ does not restrict the operator arguments.  TLC flags a
+model checking error, if `S` is infinite.  Apalache produces a static type
+error, if the type of elements of `S` is not compatible in the context of `P`
+when an element of `S` is bound to `x`.
```
Kukovec, line 348: See previous.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+  CHOOSE f \in [ 1..10 -> BOOLEAN ]:
+    \E x, y \in DOMAIN f:
+      f[x] /\ ~f[y]
+  \* some Boolean function from 1..10 that covers FALSE and TRUE
+```
+
+**Example in Python:** Python does not have anything similar to `CHOOSE`.
+The closest possible solution is to sort the filtered set by the string values
+and pick the first one (or the last one). So we have introduced a particular
+way of implementing CHOOSE, see [choose.py](./examples/choose.py):
```
Kukovec, line 367: You can do:
```python
def get(s):
  for e in s:
    return e

print(get({"1", 1, (1,)}))
```
under the hood, set is iterable, so you get the 1st element of that iteration.

docs/lang/logic.md
```diff
@@ -0,0 +1,458 @@
...
+expression. 
+
+**Effect:** This operator evaluates to some value `v` in the logical universe
+that evaluates `P` to `TRUE` against the binding `[x |-> v]`.
+
+Neither TLC, nor Apalache support this operator. It is impossible to give
+operational semantics for this operator, unless we explicitly introduce the
+universe and introduce a fixed rule for enumerating its elements.
+
+Congratulations! You have reached the bottom of this page. If you like to learn
```
Kukovec, line 448:
```suggestion
Congratulations! You have reached the bottom of this page. If you want to learn
```

Kukovec (commented):
Review 2/n.

docs/lang/standard-operators.md
```diff
@@ -0,0 +1,225 @@
...
+In this document, we present the semantics of TLA+, as if it was executed on a
+computer that is equipped with an additional device that we call an _oracle_.
+Most of the TLA+ operators are understood as deterministic operators, so they
+can be executed on your computer. A few operators are non-deterministic, so
+they require the oracle to resolve non-determinism, see [Control Flow and
+Non-determinism]. This is one of the most important features that makes TLA+
+distinct from programming languages.  Wherever possible, we complement the
+English semantics with code in [Python](https://www.python.org/). Although our
+semantics is more restrictive than the denotational semantics in Chapter 16 of
+[Specifying Systems], it is very close to the treatment of TLA+ by the model
```
Kukovec, line 32:
```suggestion
semantics are more restrictive than the denotational semantics in Chapter 16 of
[Specifying Systems], they are very close to the treatment of TLA+ by the model
```

docs/lang/tuples.md
```diff
@@ -0,0 +1,179 @@
+# Tuples
+
+[[Back to all operators]](./standard-operators.md)
+
+Tuples in TLA+ are a special kind of [functions](./functions.md) that satisfy one
```
Kukovec, line 5:
```suggestion
Tuples in TLA+ are special kinds of [functions](./functions.md) that satisfy one
```

docs/lang/tuples.md
```diff
@@ -0,0 +1,179 @@
...
+of the following properties:
+
+ - The domain is either empty, that is, `{}`, or
+ - The domain is `1..n` for some `n > 0`.
+
+That is right. You can construct the empty tuple `<<>>`in TLA+ as well as a
```
Kukovec, line 11:
```suggestion
That is right. You can construct the empty tuple `<<>>` in TLA+ as well as a
```

docs/lang/tuples.md
```diff
@@ -0,0 +1,179 @@
...
+single-element tuple, e.g., `<<1>>`. You can also construct pairs, triples, an
+so on, e.g., `<<1, TRUE>>`, `<<"Hello", "world", 2020>>`. If you think that
+empty tuples do not make sense: In TLA+, there is no difference between tuples
+and [sequences](./sequences.md).  Again, it is duck typing: Any function with
+the domain `1..n` can be also treated as a tuple (or a sequence!), and vice
+versa, tuples and sequences are also functions. So you can use all function
+operators on tuples.
+
+Importantly, the domain of a tuple is `1..n` for some `n > 0`. So tuples never
```
Kukovec, line 20:
```suggestion
Importantly, the domain of a nonempty tuple is `1..n` for some `n > 0`. So tuples never
```

docs/lang/tuples.md
```diff
@@ -0,0 +1,179 @@
...
+  { "Alice", "Bob" } \X (1900..2000)
+```
+
+The expression in the above example constructs a set of tuples `<<n, y>>`: the
+first field `n` is set to either "Alice" or "Bob", and the second field `y` is set
+to an integer from 1900 to 2000.
+
+**Application.** Simply use function application, e.g., `t[2]`.
+
+**Immutability**. As tuples are a special kind of [functions](./functions.md),
```
Kukovec, line 45:
```suggestion
**Immutability**. As tuples are special kinds of [functions](./functions.md),
```

Kukovec (commented):
Review 3/3. Apparently, n=3 all along.

docs/lang/standard-operators.md
```diff
@@ -0,0 +1,225 @@
...
+   - [`S \ T`](./sets.md#setminus)
+ - Set predicates:
+    - [`x \in S`](./sets.md#in) and [`x \notin S`](./sets.md#notin),
+    - [`S \subset T`](./sets.md#subset),
+        [`S \subseteq T`](./sets.md#subseteq),
+        [`S \supset T`](./sets.md#supset),
+        [`S \supseteq T`](./sets.md#supseteq)
+ - Set filter: [`{ x \in S: p }`](./sets.md#filter)
+ - Set map: [`{ e: x \in S }`](./sets.md#map)
+ - Powers: [`SUBSET S`](./sets.md#powerset) and [`UNION S`](./sets.md#fold)
```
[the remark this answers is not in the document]
konnov: That makes sense. I also realized that I have missed the basic set constructor `{e_1, ..., e_n}` in this list ;-)

docs/lang/standard-operators.md
```diff
@@ -0,0 +1,225 @@
...
+ - All operators of [functions](./functions.md)
+ - [Record constructor](./records.md#recCtor): `[ h_1 |-> e_1, ..., h_n |-> e_n ]`
+ - [Set of records](./records.md#recSetCtor): `[ h_1: S_1, ..., h_n: S_n ]`
+ - [Access by field name](./records.md#recApp): `e.h`
+
+### Tuples :triangular_ruler:
+
+_Well, tuples_, indexed with 1, 2, 3... [Learn more...](./tuples.md)
+
+  - All operators of [functions](./functions.md)
```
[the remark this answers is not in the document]
konnov: Right. It is a confusing concept. I have written literally: Records are functions, tuples are functions, sequences are functions.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+**LaTeX notation:** ![land](./img/land.png)
+
+**Arguments:** Two or more arbitrary expressions.
+
+**Effect:** The binary case `F /\ G` evaluates to:
+    
+ - `TRUE`, if both `F` and `G` evaluate to `TRUE`.
+
+ - `FALSE`, if `F` evaluates to `FALSE`,
+      or `F` evaluates to `TRUE` and `G` evaluates to `FALSE`.
```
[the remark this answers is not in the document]
konnov: "either ..., or ..."  is sometimes understood as "exclusive or". So I would avoid it.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+<a name="and"></a>
+### And (conjunction)
+
+**Notation:** `F /\ G` or `F \land G`
+
+**LaTeX notation:** ![land](./img/land.png)
+
+**Arguments:** Two or more arbitrary expressions.
+
+**Effect:** The binary case `F /\ G` evaluates to:
```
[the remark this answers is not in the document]
konnov: Yes, by effect, I understand how it is evaluated.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+----------------------------------------------------------------------------
+
+<a name="and"></a>
+### And (conjunction)
+
+**Notation:** `F /\ G` or `F \land G`
+
+**LaTeX notation:** ![land](./img/land.png)
+
+**Arguments:** Two or more arbitrary expressions.
```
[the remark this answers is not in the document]
konnov: Well, because TLA+ does not enforce types. You can pass arbitrary expressions. It's up to semantics and a tool how to evaluate it. To make things worse, there are three kinds of TLA+ semantics, when types are concerned.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+
+**Effect:**
+
+The binary case `F \/ G` evaluates to:
+    
+ - `FALSE`, if both `F` and `G` evaluate to `FALSE`.
+
+ - `TRUE`, if `F` evaluates to `TRUE`,
+      or `F` evaluates to `FALSE` and `G` evaluates to `TRUE`.
```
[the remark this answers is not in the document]
konnov: because "either, or" is ambiguous. Maybe not to the English speakers, but I found that German speakers understand it as "exclusive or" quite often.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+
+The general case `F_1 \/ ... \/ F_n` can be understood by evaluating
+the expression `F_1 \/ (F_2 \/ ... \/ (F_{n-1} \/ F_n)...)`.
+
+**Determinism:** deterministic, if the arguments do not contain primes.
+If the arguments contain primes, disjunctions produce non-determinism,
+see [Control Flow and Non-determinism].
```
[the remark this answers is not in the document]
konnov: True. To be more precise, Leslie defines constant expressions, state expressions, and action expressions.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+
+<a name="implies"></a>
+### Implication
+
+**Notation:** `F => G`
+
+**LaTeX notation:** ![implies](./img/implies.png)
+
+**Arguments:** Two arguments. Although they can be arbitrary expressions, the
+result is only defined when both arguments are evaluated to Boolean values.
```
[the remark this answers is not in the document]
konnov: Yes, this is pointed in the "Errors" section.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+You find these operators in every programming language and every textbook on
+logic. These operators form _propositional logic_.
+
+<a name="const"></a>
+## Constants
+
+TLA+ contains three special constants: `TRUE`, `FALSE`, and `BOOLEAN`.
+The constant `BOOLEAN` is defined as the set `{FALSE, TRUE}`.
+
+In theory, `TRUE` and `FALSE` are also sets, but in practice they are treated
```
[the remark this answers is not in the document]
konnov: Added a note for set-theory purists :)

docs/lang/booleans.md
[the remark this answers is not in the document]
konnov: It's hard to write it precisely, without going into definitions. For the moment, I wrote: "If the arguments may not update primed variables". In the future, we should refer to the new algorithms for identifying assignments.

docs/lang/booleans.md
```diff
@@ -0,0 +1,328 @@
...
+**Example in Python:**
+
+Assuming that both expressions are Boolean, `F <=> G` is equivalent to `F = G`.
+
+```python
+False == True
+True  == True
+False == False
+True  == False
+```
```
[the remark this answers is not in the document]
konnov: I am not sure about python `is`. Maybe it works as expected in the Boolean case. But it probably does not in the case of sets. There are also three interpretations of TLA+: the theoretical (as you know, there are also three theoretical interpretations), TLC, and Apalache.

docs/lang/conditionals.md
```diff
@@ -0,0 +1,241 @@
+# Deterministic conditionals
+
+[[Back to all operators]](./standard-operators.md)
+
+In this section, we consider the instances of `IF-THEN-ELSE` and `CASE` that do
+not contain unassigned primed variables inside. For the case, when the
```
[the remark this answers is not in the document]
konnov: a bit better wording, still not perfect: `that may not update primed variables`

docs/lang/conditionals.md
```diff
@@ -0,0 +1,241 @@
...
+<a name="ite"></a>
+## Deterministic IF-THEN-ELSE
+
+_Use it when choosing between two values, not to structure your code._
+
+**Notation:** `IF A THEN B ELSE C`
+
+**LaTeX notation:** the same
+
+**Arguments:** a Boolean expression `A` and two expressions `B` and `C`
```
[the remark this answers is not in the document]
konnov: Right. However, this is only the requirement of Apalache. I will add a note in errors.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+the mathematical depths of precisely defining TLA+ values. Due to the
+background theory of ZFC, this set is well-defined and is not subject to
+logical paradoxes. Basically, the values are Booleans, integers, strings, sets,
+functions, etc.
+
+In the above example, the operator `Init` evaluates to `TRUE` on exactly one
+state, which we can conveniently write using the record constructor as follows:
+`[x |-> 0, y |-> 0]`.
+
+The operator `Next` contains primes (`'`) and thus accepts pairs of states,
```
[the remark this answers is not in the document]
konnov: Yes, it does define a binary relation, but it also a predicate.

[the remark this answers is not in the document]
konnov: I changed it to "represents"

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+which we call _transitions_. An operator over unprimed and primed variables
+is called an _action_ in TLA+. Intuitively, the operator `Next` in our example
+evaluates to `TRUE` on infinitely many pairs of states that look like follows:
+
+```tla
+<<[x |-> 0, y |-> 0], [x |-> 1, y |-> 1]>>
+<<[x |-> 1, y |-> 1], [x |-> 2, y |-> 2]>>
+<<[x |-> 2, y |-> 2], [x |-> 3, y |-> 3]>>
+...
+```
```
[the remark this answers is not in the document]
konnov: True. I have added a note in the end.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+
+When we look at the operators like `Init` and `Next` in our examples, we can
+guess the states and transitions. If we could ask our logician friend to guess
+the states and transitions for us every time we read a TLA+ specification, that
+would be great. But this approach does not scale well.
+
+*Can we explain non-determinism to a computer?* It turns out that we can.
+TLC and Apalache work because their authors managed to write programs
+that are reasoning about
+non-determinism. Of course, this comes with constraints on the structure of the
```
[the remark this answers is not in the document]
konnov: Sure, we are not the first ones. I will try to explain it.

docs/lang/control-and-nondeterminism.md
```diff
@@ -0,0 +1,509 @@
...
+Null]`, we can see how to automatically evaluate the body of the operator
+`Init`:
+
+```tla
+x = 0 /\ y = 0 
+```
+
+By following [semantics of conjunction](./booleans.md), we see that `/\` is
+evaluated from left-to-right. The left-hand side equality `x = 0` is treated
+as an assignment to `x`, since `x` has value `Null` in the partial state
```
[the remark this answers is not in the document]
konnov: It is not about primed variables, but about whether a variable (primed or not) is bound to a value. In Apalache, we use `x' = e` even in `Init` just because it is more convenient to process such expressions.

docs/lang/control-and-nondeterminism.md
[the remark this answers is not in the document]
konnov: Thanks! I see how it can be confusing.

docs/lang/control-and-nondeterminism.md
konnov: OK, I am talking about pseudo-random generators here :-)

docs/lang/control-and-nondeterminism.md
konnov: I have a simpler semantics in mind. It is not about enumerating all states but like a non-deterministic Turing machine.

docs/lang/control-and-nondeterminism.md
konnov: That should be better now with the bindings

docs/lang/functions.md
```diff
@@ -0,0 +1,529 @@
...
+For instance, one can write `[ <<x, y>> \in S |-> x + y ]`. In this case, for
+every element `e` of `S`, the variable `x` is bound to `e[1]` and `y` is bound
+to `e[2]`. The function constructor maps the tuples from `S` to the values
+that are computed under such a binding.
+
+**Example in TLA+:**
+
+```tla
+  [ x \in 1..3 |-> 2 * x ]  \* a function that maps 1, 2, 3 to 2, 4, 6
+  [ x, y \in 1..3 |-> x * y ]
```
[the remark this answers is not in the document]
konnov: I find this pattern matching quite confusing in TLA+, though mathematicians do it all the time too :-)

docs/lang/functions.md
```diff
@@ -0,0 +1,529 @@
...
+_Importantly, `g` is a new function, the function `f` is not modified!_
+
+**Determinism:** Deterministic.
+
+**Errors:** When `a_i \notin DOMAIN f` for some `i \in 1..n`,
+TLC flags a model checking error.
+
+When `a_1, ..., a_n` are not type-compatible with the type of `DOMAIN f`,
+Apalache flags a type error. When `a_i \notin DOMAIN f`, Apalache ignores this
+update. This is consistent with the semantics of TLA+ in [Specifying Systems].
```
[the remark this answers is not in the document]
konnov: True. Replaced it with "argument".

docs/lang/logic.md
konnov: Fixed

docs/lang/logic.md
konnov: It was also wrong. Fixed!

docs/lang/logic.md
konnov: The book defines `CHOOSE x \in S` as `CHOOSE (x \in S) /\ P`, so it should be true. I am not sure about TLC though. Of course, what is written in English is slightly more restrictive.

docs/lang/logic.md
konnov: Fixed

docs/lang/logic.md
konnov: Yes, but that would not guarantee the determinism of CHOOSE. I am enforcing determinism by sorting the elements with their string representation as a key. It is quite inefficient, but sound.

docs/lang/sets.md
```diff
@@ -0,0 +1,802 @@
...
+                            \* static type error in Apalache
+```
+
+**Example in Python:** Python conveniently offers us `not in`:
+
+```python
+  1 not in frozenset({1, 2, 3})     # False
+  10 not in frozenset({1, 2, 3})    # True
+  1 not in frozenset({"a", "b"})    # True
+```
```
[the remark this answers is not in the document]
konnov: Good point. Added a sentence on that.

docs/lang/sets.md
```diff
@@ -0,0 +1,802 @@
...
+```tla
+  { 0, 1, 2 } \ { 1, 2, 3 }     \* { }
+          { } \ { 1, 2, 3 }     \* { }
+  { 1, 2, 3 } \ { }             \* { }
+      { {1} } \ { 1, 2 }        \* { {1} }, model checking error in TLC
+                                \* static type error in Apalache
+```
+
+**Example in Python:** Python conveniently offers us `intersection`, which
+    can be also written as `&`:
```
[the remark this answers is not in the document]
konnov: Indeed. Fixed

PR #342, opened 2020-11-27
Improve the Dockerfile documentation

shonfeder:
Make the two phase build clearer, and explain the reason for the
different phases.

Follows from conversation with @andrey-kuprianov and @konnov on slack.


shonfeder:
Thanks, Andrey!

Merging, as the dockerhub build error is spurious.

PR #343, opened 2020-11-29
Reordered assignments

### LINKED ISSUE #157: [BUG] Out of order assignments
Something has changed in the behavior of the assignment finder and type checker between 0.6.0 and unstable.

Now the following test fails:

```bash
cd $HOME/devl/apalache-tests/performance/bcastByz
$HOME/devl/apalache/bin/apalache-mc check --length=1 \
  --init=IndInv_Unforg_NoBcast --inv=IndInv_Unforg_NoBcast --cinit=ConstInit4 APAbcastByz.tla
```

Apalache fails with the message:

```
Step 0, level 0: checking if 1 transition(s) are enabled and violate the invariant I@15:55:37.782
APAbcastByz.tla:182:6-182:36, '(...), type error: Failed to find type of variable sent' E@15:55:37.798
APAbcastByz.tla:182:6-182:36, =(...), type error: Expected a Boolean, found: None E@15:55:37.803
APAbcastByz.tla:183:34-183:50, '(...), type error: Failed to find type of variable sent' E@15:55:37.808
APAbcastByz.tla:183:15-183:54, [S -> T](...), type error: Expected a finite set, found: None E@15:55:37.811
The outcome is: Error                                             I@15:55:37.843
Checker has found an error                                        I@15:55:37.843
It took me 0 days  0 hours  0 min  2 sec                          I@15:55:37.846
Total time: 2.121 sec                                             I@15:55:37.846
EXITCODE: OK
```

The reason for that is that `sent'` is assigned a value after it is used:

```tla
Init$0 ==
  (BMC!Skolem((\E t_18$1 \in [Corr -> { "V0", "V1", "SE", "AC" }]: pc' <- t_18$1))
      /\ Corr \in SUBSET (Corr \union Faulty)
      /\ Faulty \in SUBSET (Corr \union Faulty)
      /\ BMC!Skolem((\E t_1b$1 \in SUBSET ({
        <<t_19$1, t_1a$1>>:
          t_19$1 \in Corr \union Faulty,
          t_1a$1 \in {"ECHO"}
      }):
        sent' = t_1b$1))
      /\ BMC!Skolem((\E t_1e$1 \in [Corr
        -> SUBSET (sent'
          \union { <<t_1c$1, t_1d$1>>: t_1c$1 \in Faulty, t_1d$1 \in {"ECHO"} })]:
        rcvd' <- t_1e$1)))
    /\ sent' <- ({} <: {<<Int, STRING>>})
    /\ pc' = [ i$1 \in Corr |-> "V0" ]
```

We have been discussing for some time that the assignment finder should change the order of subformulas so that the assignments always come first. For some reason, the assignment finder always picks the assignment under the existential,
even if we swap the subformulas.

### LINKED ISSUE #338: [BUG] Operators in Init are not inlined before priming
This is a recent bug introduced in the refactoring 57d95aefcbcbf992079a662cd54fd970ea8e7ae7.

If `Init` is using another operator `Foo`, then the state variables inside `Foo` are not replaced with their primed versions.

Jure @Kukovec will add an integration test and fix the bug :-)

Kukovec:
closes #157 
closes #338

### DIFF
--- a/test/tla/APAbcastByz.tla
+++ b/test/tla/APAbcastByz.tla
@@ -1,0 +1,1072 @@
+------------------------------ MODULE APAbcastByz ------------------------------
+
+(* TLA+ encoding of a parameterized model of the broadcast distributed  
+   algorithm with Byzantine faults.
+  
+   This is a one-round version of asynchronous reliable broadcast (Fig. 7) from:
+  
+   [1] T. K. Srikanth, Sam Toueg. Simulating authenticated broadcasts to derive
+   simple fault-tolerant algorithms. Distributed Computing 1987,
+   Volume 2, Issue 2, pp 80-94
+                                                             
+   A short description of the parameterized model is described in: Gmeiner,   
+   Annu, et al. "Tutorial on parameterized model checking of fault-tolerant   
+   distributed algorithms." International School on Formal Methods for the  
+   Design of Computer, Communication and Software Systems. Springer  
+   International Publishing, 2014.                   
+  
+   This specification has a TLAPS proof for property Unforgeability: if process p 
+   is correct and does not broadcast a message m, then no correct process ever 
+   accepts m. The formula InitNoBcast represents that the transmitter does not 
+   broadcast any message. So, our goal is to prove the  formula
+        (InitNoBcast /\ [][Next]_vars) => []Unforg                    
+  
+   We can use TLC to check two properties (for fixed parameters N, T, and F):
+    - Correctness: if a correct process broadcasts, then every correct process accepts,
+    - Replay: if a correct process accepts, then every correct process accepts.  
+  
+   Igor Konnov, Thanh Hai Tran, Josef Widder, 2016
+  
+   This file is a subject to the license that is bundled together with this package 
+   and can be found in the file LICENSE.
+ *)
+
+
+EXTENDS Integers, \*Naturals, 
+        FiniteSets
+        (*Functions,
+        FunctionTheorems, 
+        FiniteSetTheorems,
+        NaturalsInduction,
+        SequenceTheorems*)
+
+\* APALACHE-BEGIN
+a <: b == a
+MT == <<Int, STRING>>
+\* APALACHE-END
+        
+CONSTANTS N, T, F,
+    Corr,  \* the correct processes, their identities are not used
+    Faulty \* the Byzantine processes, their identities are not used
+    
+ConstInit4 ==
+    /\ N \in {4} /\ T \in {1} /\ F \in {1} /\ Corr \in {1..3} /\ Faulty \in {{4}}     
+    
+ConstInit6 ==
+    /\ N \in {6} /\ T \in {1} /\ F \in {1} /\ Corr \in {1..5} /\ Faulty \in {{6}}     
+
+VARIABLE pc             (* the control state of each process *)
+VARIABLE rcvd           (* the messages received by each process *)
+VARIABLE sent           (* the messages sent by all correct processes *)
+ASSUME NTF == N \in Nat /\ T \in Nat /\ F \in Nat /\ (N > 3 * T) /\ (T >= F) /\ (F >= 0)
+
+Proc == Corr \cup Faulty         (* all processess, including the faulty ones    *)
+M == { "ECHO" }
+(* ByzMsgs == { <<p, "ECHO">> : p \in Faulty }: quite complicated to write a TLAPS proof 
+   for the cardinality of the expression { e : x \in S}
+ *)
+ByzMsgs == Faulty \X M
+                            
+vars == << pc, rcvd, sent >>
+
+(* Instead of modeling a broadcaster explicitly, two initial values V0 and V1 at correct
+   processes are used to model whether a process has received the INIT message from the
+   broadcaster or not, respectively. Then the precondition of correctness can be modeled 
+   that all correct processes initially have value V1, while the precondition of unforgeability  
+   that all correct processes initially have value V0.
+ *)
+Init == 
+  /\ sent = {} <: {MT}                          (* No messages sent initially *)
+  /\ pc \in [ Corr -> {"V0", "V1"} ]    (* Some processes received INIT messages, some didn't *)
+  /\ rcvd = [ i \in Corr |-> {} <: {MT} ]       (* No messages received initially *)
+        
+(* This formula specifies restricted initial states: all correct processes initially have value V0.
+   (This corresponds to the case when no correct process received an INIT message from a broadcaster.)
+   Notice that in our modeling Byzantine processes also start in the local state V0.
+ *)
+InitNoBcast == pc \in [ Corr -> {"V0"} ] /\ Init
+
+(* A correct process can receive all ECHO messages sent by the other correct processes,
+   i.e., a subset of sent, and all possible ECHO messages from the Byzantine processes,
+   i.e., a subset of ByzMsgs.
+ *)
+Receive(self) ==
+  \E newMessages \in SUBSET ( sent \cup ByzMsgs ) :
+    rcvd' = [ i \in Corr |-> IF i # self THEN rcvd[i] ELSE rcvd[self] \cup newMessages ]
+                             
+(* The first if-then expression in Figure 7 [1]: If process p received an INIT message and
+   did not send <ECHO> before, then process p sends ECHO to all.
+ *)
+UponV1(self) ==
+  /\ pc[self] = "V1"
+  /\ pc' = [pc EXCEPT ![self] = "SE"]
+  /\ sent' = sent \cup { <<self, "ECHO">> }
+
+(* The 3rd if-then expression in Fig. 7 [1]: If correct process p received ECHO messages 
+   from at least N - 2*T distinct processes and did not send ECHO before, then process p sends
+   ECHO messages to all. 
+  
+   Since processes send only ECHO messages, the number of messages in rcvd[self] equals the   
+   number of distinct processes from which process self received ECHO messages. 
+  
+   The 3rd conjuction "Cardinality(rcvd'[self]) < N - T" ensures that process p cannot accept 
+   or not execute the 2nd if-then expression in Fig. 7 [1]. If process p received ECHO messages
+   from at least N - T distinct processes, the formula UponAcceptNotSentBefore is called.
+ *)         
+UponNonFaulty(self) ==
+  /\ pc[self] \in { "V0", "V1" }
+  /\ Cardinality(rcvd'[self]) >= N - 2*T  
+  /\ Cardinality(rcvd'[self]) < N - T
+  /\ pc' = [ pc EXCEPT ![self] = "SE" ]
+  /\ sent' = sent \cup { <<self, "ECHO">> }
+        
+(* The 2nd and 3rd if-then expressions in Figure 7 [1]: If process p received <ECHO> from at 
+   least N - T distinct processes and did not send ECHO message before, then process p accepts       
+   and sends <ECHO> to all.                  
+ *)        
+UponAcceptNotSentBefore(self) ==
+  /\ pc[self] \in { "V0", "V1" }
+  /\ Cardinality(rcvd'[self]) >= N - T
+  /\ pc' = [ pc EXCEPT ![self] = "AC" ]
+  /\ sent' = sent \cup { <<self, "ECHO">> }
+        
+(* Only the 2nd if-then expression in Fig. 7 [1]:  if process p sent ECHO messages and received 
+   ECHO messages from at least N - T distinct processes, it accepts.
+  
+   As pc[self] = "SE", the 3rd if-then expression cannot be executed.
+ *)        
+UponAcceptSentBefore(self) ==
+  /\ pc[self] = "SE"
+  /\ Cardinality(rcvd'[self]) >= N - T
+  /\ pc' = [pc EXCEPT ![self] = "AC"]
+  /\ sent' = sent
+
+(* All possible process steps.*)                
+Step(self) == 
+  /\ Receive(self)
+  /\ \/ UponV1(self)
+     \/ UponNonFaulty(self)
+     \/ UponAcceptNotSentBefore(self)
+     \/ UponAcceptSentBefore(self)
+
+(* Some correct process does a transition step.*)                         
+Next ==
+     \/ \E self \in Corr: Step(self)                     
+     \/ UNCHANGED vars (* add a self-loop for terminating computations *)
+
+(* Add weak fairness condition since we want to check liveness properties. We require that 
+   if UponV1 (or UponNonFaulty, UponAcceptNotSentBefore, UponAcceptSentBefore) ever becomes 
+   forever enabled, then this step must eventually occur.      
+ *)
+Spec == Init /\ [][Next]_vars 
+             /\ WF_vars(\E self \in Corr: /\ Receive(self)
+                                          /\ \/ UponV1(self)
+                                             \/ UponNonFaulty(self)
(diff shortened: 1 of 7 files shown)


tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
```diff
@@ -0,0 +1,50 @@
...
+import at.forsyte.apalache.tla.lir.{LetInEx, OperEx, TlaEx, TlaOperDecl}
+import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TransformationTracker}
+
+class Reordering[T]( ord : Ordering[T], rankingFn : TlaEx => T, tracker : TransformationTracker ) {
+
+  def reorder : TlaExTransformation = tracker.track {
+    // For now, only /\ and \/ are reordered, but the same process works for all commutative operators
+    case ex@OperEx( op@( TlaBoolOper.and | TlaBoolOper.or ), args@_* ) =>
+      val reorderedArgs = args.sortBy( rankingFn )( ord ) map reorder
+      if ( args == reorderedArgs ) ex
```
shonfeder, line 13: What's the purpose of this check? iiuc, whenever `args == reorderedArgs` it will also be true that `ex == OperEx( op, reorderedArgs : _* )`. Right? So can skip the conditional here and on 18 and just return the constructed `OperEx` in the `else`?

The motive here (in addition to possibly removing unnecessary code) is the assumption that it's cheaper to just construct a new `OperEx` than to check all the arguments for equality.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
```diff
@@ -0,0 +1,50 @@
...
+    case ex@LetInEx( letInBody, defs@_* ) =>
+      val newDefs = defs map {
+        // Assignments can only appear in nullary operators
+        case TlaOperDecl( name, Nil, declBody ) =>
+          TlaOperDecl( name, Nil, reorder( declBody ) )
+        case d => d
+      }
+      val newBody = reorder( letInBody )
+      if ( defs == newDefs && letInBody == newBody ) ex
+      else LetInEx( newBody, newDefs : _* )
```
shonfeder, line 30: Same question here about why we bother with the equality check.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
```diff
@@ -0,0 +1,50 @@
...
+
+
+    case ex => ex
+
+  }
+}
+
+object Reordering {
+
+  // The order we want, for x < y, is Some(x) < Some(y) < None
```
shonfeder, line 40: Perhaps add a note here indicating why we want this ordering?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -247,12 +255,32 @@ class SymbTransGenerator( tracker : TransformationTracker ) {
...
-      selections( ex.ID ).map( s =>
-        (
-          mkOrdered( s, strategy ),
-          sliceWith( s, selections )( ex )
-        )
-      ).toSeq
+      selections( ex.ID ).map { s =>
+        val ordered = mkOrdered( s, strategy )
+        def rankingFn( tlaEx: TlaEx) : Option[Int] = {
+          // the union of all sets in selections( tlaEx.ID ) lists all the assignments below
```
shonfeder, line 261: What is it that we are `below` here?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -247,12 +255,32 @@ class SymbTransGenerator( tracker : TransformationTracker ) {
                         strategy : StrategyType,
                         selections : SelMapType
                       ) : Seq[SymbTrans] =
-      selections( ex.ID ).map( s =>
-        (
-          mkOrdered( s, strategy ),
-          sliceWith( s, selections )( ex )
-        )
-      ).toSeq
+      selections( ex.ID ).map { s =>
```
shonfeder, line 258: This block is pretty hairy. Do there's an relatively way way of refactoring this a bit so that the reader doesn't have to grok so much activity within this lambda? Perhaps by breaking it into a few named functions (for the sake of self-documentation) or at least one named function defined prior to actually calling the function on the selections?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/PrimingPassImpl.scala
```diff
@@ -19,14 +19,18 @@ import com.typesafe.scalalogging.LazyLogging
...
-      extends PrimingPass with LazyLogging {
+  extends PrimingPass with LazyLogging {
   /**
     * The name of the pass
     *
     * @return the name associated with the pass
     */
   override def name: String = "PrimingPass"
 
+  private def trSeq( seq: Seq[TlaExTransformation] ) : TlaExTransformation = {
```
shonfeder, line 30: Assuming `tr` is short for `translate`, using the full word would help with readability imo.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
```diff
@@ -0,0 +1,50 @@
+package at.forsyte.apalache.tla.assignments
+
+import at.forsyte.apalache.tla.lir.oper.TlaBoolOper
+import at.forsyte.apalache.tla.lir.{LetInEx, OperEx, TlaEx, TlaOperDecl}
+import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TransformationTracker}
+
+class Reordering[T]( ord : Ordering[T], rankingFn : TlaEx => T, tracker : TransformationTracker ) {
+
+  def reorder : TlaExTransformation = tracker.track {
```
shonfeder, line 9: Seems like we might benefit from a unit test or two for this method. Does that seem reasonable?

shonfeder (commented):
I left some feedback! I'm mainly reading for my own comprehension. My few suggestions are mostly just things I think might help me with readability, and are arguably just stylistic, so feel free to ignore (naturally ;)).

I do have two substantive question: the one about why the spec used for the integration test is so long, and the one about the seemingly redundant equality checks.

test/tla/APAbcastByz.tla
```diff
@@ -0,0 +1,1072 @@
+------------------------------ MODULE APAbcastByz ------------------------------
```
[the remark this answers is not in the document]
Kukovec: Possibly, but Igor explicitly requested I include this one.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
```diff
@@ -0,0 +1,50 @@
...
+    // For now, only /\ and \/ are reordered, but the same process works for all commutative operators
```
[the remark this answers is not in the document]
Kukovec: No plans atm, the only other control-flow relevant operator I can think of is `CASE`, which is commutative in base TLA+, but not actually commutative in Apalache. 
The opposite actually, it might happen that we end up not reordering `\/`. But for now, no TODO, this is complete as is.

test/tla/APAbcastByz.tla
```diff
@@ -0,0 +1,1072 @@
...
+   example, additional constraints are  relationships between the number of messages, pc, and 
+   the number of faulty processes.    
+ *)                
+IndInv_Unforg_NoBcast ==  
+  /\ TypeOK
+  \*/\ FCConstraints
+  /\ sent = {} <: {MT}
+  /\ pc = [ i \in Corr |-> "V0" ]
+
+IndInv_Unforg_NoBcast4 ==
```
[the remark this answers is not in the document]
Kukovec: That's Igor's spec from the apalache-tests repository, I didn't write or edit it in any way, so I'll defer to @konnov.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
Kukovec: This is a canonical check, in practice, it is better to not introduce a new expression not for performance's sake, but because preserving the expression also preserves the UID. This matters, because a lot of auxiliary information is passed as a `UID -> X` map, so by keeping the expression whenever possible, you don't invalidate those maps.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
```diff
@@ -0,0 +1,50 @@
+package at.forsyte.apalache.tla.assignments
+
+import at.forsyte.apalache.tla.lir.oper.TlaBoolOper
+import at.forsyte.apalache.tla.lir.{LetInEx, OperEx, TlaEx, TlaOperDecl}
+import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TransformationTracker}
+
+class Reordering[T]( ord : Ordering[T], rankingFn : TlaEx => T, tracker : TransformationTracker ) {
```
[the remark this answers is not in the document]
Kukovec: I figured this class was tiny enough, but sure, can do.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
Kukovec: Below in the syntax-tree sense, i.e. subexpressions.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
Kukovec: If you want, sure, should be easy enough to split.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/PrimingPassImpl.scala
Kukovec: `tr` is standard shorthand for `transform`, as in apply a `TlaExTransformation`. But sure, easy to rename.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
Kukovec: I can add some, but this entire class is basically just boilerplate around line 12.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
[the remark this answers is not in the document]
shonfeder: Yeah, the class is tiny enough that it's clear what its doing. But for a reader without the full context (which most will be) there's no clear indication of *why* this would need to be done, imo. Just a line or two indicating why we need a reordering class should be enough of a pointer for future readers I think.

test/tla/APAbcastByz.tla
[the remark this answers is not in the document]
shonfeder: I see. In that case the content of this spec shouldn't really be part of the review I guess. Thanks :)

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
shonfeder: You're definitely a better judge than I am of where unit tests would be applicable, but it does seem to me intuitively that we should have *some* kind of unit tests to ensure the reordering is happening as we expect. Does that seem sensible?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/PrimingPassImpl.scala
shonfeder: Ah! Where is this standardized? I'm not familiar but would interesting to learn the context :)

Given the amount of meanings for `tr` recorded by [wiktionary](https://en.wiktionary.org/wiki/tr.), I think the disambiguation is probably worth a few extra characters :)

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
shonfeder: I know it would help with readability for me. I suspect it'd help future readers too.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
shonfeder: Ah! How about:

```suggestion
          // the union of all sets in selections( tlaEx.ID ) lists all the assignments in the subexpressions of TlaEx
```

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
shonfeder: Ah! I see thanks for explaining :)

In that case, would it be valid to preserve the UUID even when we reorder the args? Or does that break the logic we need for the expression identification?

test/tla/cli-integration-tests.md
```diff
@@ -421,6 +421,16 @@ The outcome is: NoError
 ...
 ```
 
+### check APAbcastByz.tla succeeds
+
```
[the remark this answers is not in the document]
shonfeder: @konnov perhaps this would also be a good place to document why we want to test this complicated spec specifically? Perhaps you could provide the context needed for that?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/PrimingPassImpl.scala
shonfeder: Duh. nm here. I finally our point about the relation between `tr` and the type. Sorry for being dense here. I think that should suffice for someone reading more carefully than I have here :)

test/tla/APAbcastByz.tla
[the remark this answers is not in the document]
konnov: What we can do. Shorten the spec by removing the long header and THEOREMs and refer to the full spec: https://github.com/informalsystems/apalache-tests/blob/master/performance/bcastByz/APAbcastByz.tla

test/tla/cli-integration-tests.md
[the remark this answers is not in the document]
konnov: Let's give a link to https://github.com/informalsystems/apalache/issues/157

konnov (commented):
@shonfeder has already done the hard work of reviewing this PR. The rest looks good to me, just a few comments.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/Reordering.scala
Kukovec: I would advise against that, UIDs are, by definition, unique, and copying them over opens us up to a host of new, potentially very well hidden, bugs.

PR #344, opened 2020-11-29
Unroller fix

### LINKED ISSUE #129: Unrolling recursive LET-IN definitions
Following #127, we should add the unrolling feature to the recursive operators that are defined with LET-IN, not globally. This will complete support for recursive operators.

### LINKED ISSUE #131: [BUG] Unroller produces a recursive definition
See the example Rec2.tla. Unroller produces the leaf definition `LET T$3 == T$3() \in {x$3()} IN 1 + 0`.

Unhandled exception                                               E@23:22:56.186
at.forsyte.apalache.tla.lir.UnexpectedLanguageError: Predicate FlatLanguagePred does not accept the expression:

```tla
IF (S = (BMC!<:({}, {Int}))) THEN 0 ELSE LET x$1 ≜ CHOOSE y$1 ∈ S : TRUE IN LET T$1 ≜ S ∖ {x$1()} IN 1 + (IF (T$1() = (BMC!<:({}, {Int}))) THEN 0 ELSE LET x$2 ≜ CHOOSE y$2 ∈ T$1() : TRUE IN LET T$2 ≜ T$2() ∖ {x$2()} IN 1 + (IF (T$2() = (BMC!<:({}, {Int}))) THEN 0 ELSE LET x$3 ≜ CHOOSE y$3 ∈ T$2() : TRUE IN LET T$3 ≜ T$3() ∖ {x$3()} IN 1 + 0))
```

[Rec2.txt](https://github.com/konnov/apalache/files/4506279/Rec2.txt)

### LINKED ISSUE #274: [BUG] Unroller is missing the documentation

### LINKED ISSUE #309: [BUG] Unroller needs unit tests
I just realized that we don't have a single unit test for `Unroller`. Since many specifications are using recursive operators, we need unit tests, at least for the standard behaviors. Otherwise, we might get swamped by similar-looking error reports. For instance, #130 and #131 are very simple examples of failing specs.

Kukovec:
closes #129
closes #131 
closes #274
closes #309

### DIFF
--- a/test/tla/Rec2.tla
+++ b/test/tla/Rec2.tla
@@ -18,7 +18,11 @@ Card(S) ==
   IF S = IntSet({})
   THEN 0
   ELSE
-    LET T == S \ {CHOOSE y \in S: TRUE} IN
+    \* CHOOSE is introduced with a LET definition, to fix its value, whatever it may be
+    \* Note that CHOOSE in APALACHE is non-deterministic and therefore
+    \* S \ { CHOOSE x \in S : TRUE } leads to a false-positive C.E.
+    LET x == CHOOSE y \in S: TRUE IN
+    LET T == S \ {x} IN
     1 + Card(T)
 
 \* unfold the operator Card up to 10 times
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/AppWrap.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/AppWrap.scala
@@ -1,0 +1,85 @@
+package at.forsyte.apalache.tla.pp
+
+import at.forsyte.apalache.tla.lir.oper.{BmcOper, TlaOper}
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModuleTransformation, TransformationTracker}
+
+/** Replaces instances of user-defined operator applications with a LET-IN wrapper.
+  *
+  * Example:
+  * A(x,y) ~~> LET App_1 == A(x,y) IN App_1
+  *
+  * Operator constants and formal parameters are ignored.
+  */
+class AppWrap(
+               nameGenerator : UniqueNameGenerator,
+               tracker : TransformationTracker
+             ) {
+
+  import AppWrap.NAME_PREFIX
+
+  private def setTracking( oldEx : => TlaEx, newEx : => TlaEx ) : TlaEx = {
+    val tr = tracker.track {
+      _ => newEx
+    }
+    tr( oldEx )
+  }
+
+  def wrap( wrappableNames : Set[String] ) : TlaExTransformation = tracker.track {
+    case ex@OperEx( TlaOper.apply, NameEx( opName ), args@_* ) if wrappableNames.contains( opName ) && args.nonEmpty =>
+      val newArgs = args map wrap( wrappableNames )
+      val newEx =
+        if ( args == newArgs ) ex
+        else setTracking( ex, Builder.appOp( NameEx( opName ), newArgs : _* ) )
+
+      val newName = s"${NAME_PREFIX}_${opName}_${nameGenerator.newName()}"
+      val newDecl = TlaOperDecl( newName, List.empty, newEx )
+      LetInEx( Builder.appDecl( newDecl ), newDecl )
+      // On type annot. ignore the RHS
+    case ex@OperEx( BmcOper.withType, argEx, typeEx ) =>
+      val newArg = wrap( wrappableNames )(argEx)
+      if ( argEx == newArg )
+        ex
+      else
+        OperEx( BmcOper.withType, newArg, typeEx )
+    case ex@OperEx( oper, args@_* ) =>
+      val newArgs = args map wrap( wrappableNames )
+      if ( args == newArgs )
+        ex
+      else
+        OperEx( oper, newArgs : _* )
+    case ex@LetInEx( body, defs@_* ) =>
+      // Assumption: let-in defined operators are defined in dependency order
+      val (newWrappable, newDefs) = defs.foldLeft( (wrappableNames, Seq.empty[TlaOperDecl]) ) {
+        case ((partialSet, partialDecls), decl) =>
+          val newDecl = decl.copy( body = wrap( partialSet )( decl.body ) )
+          newDecl.isRecursive = decl.isRecursive
+          (partialSet + decl.name, partialDecls :+ newDecl)
+      }
+      val newBody = wrap( newWrappable )( body )
+      if ( body == newBody && defs == newDefs )
+        ex
+      else
+        LetInEx( newBody, newDefs : _* )
+    case ex => ex
+  }
+
+  def moduleTransform( wrappableNames : Set[String] ): TlaModuleTransformation = { m =>
+
+    val newDecls = m.declarations map {
+      case TlaOperDecl( name, params, body ) =>
+        TlaOperDecl( name, params, wrap( wrappableNames )(body) )
+      case d => d
+    }
+    new TlaModule( m.name, newDecls )
+  }
+}
+
+object AppWrap {
+  val NAME_PREFIX = "APP"
+
+  def apply(
+             nameGenerator : UniqueNameGenerator,
+             tracker : TransformationTracker
+           ) = new AppWrap( nameGenerator, tracker )
+}
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ParameterNormalizer.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ParameterNormalizer.scala
@@ -5,31 +5,38 @@ import at.forsyte.apalache.tla.lir.transformations.standard.ReplaceFixed
 import at.forsyte.apalache.tla.lir._
 import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModuleTransformation, TransformationTracker}
 
+/**
+  * Transforms a declaration A(x,y(_,_)) == e
+  * into parameter-normal form, i.e.
+  * A(x,y(_,_)) == LET x_new == x
+  *               y_new(p1, p2) == y(p1,p2)
+  *           IN e[ x_new/x, y_new/y ]
+  * This allows us to limit the number of substitutions when inlining A.
+  * @param decisionFn Normal-form transformation will only be applied to operator declarations (both top-level and LET-IN),
+  *                   for which `decisionFn` evaluates to true. Default: returns true for all recursive operators.
+  */
 class ParameterNormalizer(
                            nameGenerator: UniqueNameGenerator,
                            tracker: TransformationTracker,
                            decisionFn: TlaOperDecl => Boolean
                          ) {
-
-  // Transforms a declaration A(x,y(_,_)) == e
-  // into A(x,y) == LET x_new == x
-  //                    y_new( p1, p2) == y(p_1,p_2)
-  //                IN e[ x_new/x, y_new/y ]
-  // This allows us to limit the number of substitutions when inlining A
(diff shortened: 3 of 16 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ParameterNormalizer.scala
```diff
@@ -0,0 +1,116 @@
...
+            )
+            val letInDef = TlaOperDecl( paramOperName, inventedParams map SimpleFormalParam, newBody )
+            LetInEx( replaced, letInDef )
+        }
+    }
+  }
+
+  def normalize : TlaModuleTransformation = { m =>
+    // Iterates over all declarations and replaces those for which decisionFn returns true
+    // (by default, for all recursive)
```
shonfeder, line 92: Should this maybe be a document comment, since its quite informative re: this public method?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
```diff
@@ -146,25 +197,32 @@ class Unroller( tracker : TransformationTracker ) extends TlaModuleTransformatio
   }
 
   override def apply( module : TlaModule ) : TlaModule = {
-    val operDecls = module.operDeclarations
+    // Before unrolling we transform all recursive operators into parameter-normal form
+    // This reduces the number of inlining calls when operator arguments aren't primitive
+    val parameterNormalizer = ParameterNormalizer( nameGenerator, tracker, ParameterNormalizer.DecisionFn.recursive )
```
shonfeder, line 202: This makes me wonder about the API for the `ParameterNormalizer`. Perhaps instead of requiring client code to access these `DecisionFn` functions, it should provide methods like `ParameterNormalizer.recursive`, `ParameterNormalizer.all`?

shonfeder (commented):
Looks pretty good to me! I have just a couple questions and a suggestion or two.

test/tla/Rec2.tla
```diff
@@ -18,7 +18,11 @@ Card(S) ==
   IF S = IntSet({})
   THEN 0
   ELSE
-    LET T == S \ {CHOOSE y \in S: TRUE} IN
+    \* CHOOSE is introduced with a LET definition, to fix its value, whatever it may be
+    \* Note that CHOOSE in APALACHE is non-deterministic and therefore
```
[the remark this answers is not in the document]
konnov: This is one of the (two or three?) common patterns of using CHOOSE. Basically, we iterate over a set by using a fixed but unknown iteration order. Since we remove one element per iteration, it is not essential that CHOOSE is deterministic here.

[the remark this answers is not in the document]
konnov: But since Apalache implements a non-deterministic version of CHOOSE (for efficiency), we have to call `CHOOSE` once in this code. Otherwise, Apalache produces a false-positive by picking two different elements from the same set `S`.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/AppWrap.scala
```diff
@@ -0,0 +1,85 @@
...
+import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModuleTransformation, TransformationTracker}
+
+/** Replaces instances of user-defined operator applications with a LET-IN wrapper.
+  *
+  * Example:
+  * A(x,y) ~~> LET App_1 == A(x,y) IN App_1
+  *
+  * Operator constants and formal parameters are ignored.
+  */
+class AppWrap(
```
konnov, line 14: Can we give this class a more intuitive name? Like `OperAppToLetInDef`? Essentially, it introduces a symbolic constant for an operator application.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ParameterNormalizer.scala
```diff
@@ -0,0 +1,116 @@
...
+  //                IN e[ x_new/x, y_new/y ]
+  // This allows us to limit the number of substitutions when inlining A
+  def mkParamNormalForm( decl : TlaOperDecl ) : TlaOperDecl = {
+    val normalizedLetIn = normalizeInternalLetIn( decl.body )
+    val newBody =
+      if ( decisionFn( decl ) ) paramNormal( decl.formalParams )( normalizedLetIn )
+      else normalizedLetIn
+    val newDecl = decl.copy( body = newBody )
+    // Copy doesn't preserve .isRecursive!
+    newDecl.isRecursive = decl.isRecursive
```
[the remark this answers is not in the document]
konnov: True. Is there any reason for copy not to preserve `isRecursive`? I think it is just a bug :)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ParameterNormalizer.scala
```diff
@@ -0,0 +1,116 @@
+package at.forsyte.apalache.tla.pp
+
+import at.forsyte.apalache.tla.lir.oper.TlaOper
+import at.forsyte.apalache.tla.lir.transformations.standard.ReplaceFixed
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModuleTransformation, TransformationTracker}
+
+class ParameterNormalizer(
```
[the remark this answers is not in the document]
konnov: After reading the code, I agree that this transformation deserves a written explanation and motivation. Most likely, we will forget what it is doing in a few months.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
```diff
@@ -4,38 +4,84 @@ import at.forsyte.apalache.tla.lir.aux.{ExceptionOrValue, FailWith, SucceedWith}
...
 import at.forsyte.apalache.tla.lir.oper.TlaOper
 import at.forsyte.apalache.tla.lir.storage.{BodyMap, BodyMapFactory}
-import at.forsyte.apalache.tla.lir.transformations.standard.InlinerOfUserOper
+import at.forsyte.apalache.tla.lir.transformations.standard.{IncrementalRenaming, InlinerOfUserOper, ReplaceFixed}
 import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModuleTransformation, TransformationTracker}
 import at.forsyte.apalache.tla.lir.values.TlaInt
 
-class Unroller( tracker : TransformationTracker ) extends TlaModuleTransformation {
+/**
+  * Replaces definitions of user-defined recursive operators with a finite-step unrolling.
```
konnov, line 12:
```suggestion
  * Replaces definitions of user-defined recursive operators with a bounded chain of the operator bodies.
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
```diff
@@ -4,38 +4,84 @@ import at.forsyte.apalache.tla.lir.aux.{ExceptionOrValue, FailWith, SucceedWith}
...
+  *
+  * For each recursive operator A the user must define two additional TLA+ operators,
+  *  - [UNROLL_TIMES_PREFIX]_A
+  *  - [UNROLL_DEFAULT_PREFIX]_A
+  *
```
konnov, line 17: Does this markdown-like commenting work in javadoc? I think you have to use a bit of HTML here, e.g., `<p>...</p>` and `<ul><li>...</li></ul>`.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
```diff
@@ -57,85 +100,93 @@ class Unroller( tracker : TransformationTracker ) extends TlaModuleTransformatio
     case ex => ex
   }
 
-
-  def getUnrollLimit( name : String, bodyMap : BodyMap ) : ExceptionOrValue[BigInt] = {
+  /** Given an operator name, A, extracts the value of [UNROLL_TIMES_PREFIX]_A or propmts a `TlaInputError` */
```
konnov, line 103:
```suggestion
  /** Given an operator name, A, extracts the value of [UNROLL_TIMES_PREFIX]_A or throws a `TlaInputError` */
```

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestAppWrap.scala
```diff
@@ -0,0 +1,79 @@
...
+    val assertCond1 = newEx1 == ex
+    val assertCond2 = newEx2 match {
+      case LetInEx(
+      OperEx( TlaOper.apply, NameEx( someName ) ),
+      TlaOperDecl( declName, Nil, defEx ) ) if declName == someName =>
+        defEx == ex
+      case _ => false
+    }
+
+    assert( assertCond1 && assertCond2 )
```
konnov, line 49: why do you postpone the evaluation of `assertCond1`? I prefer unit tests to fail asap.

konnov (commented):
Cool. That brings us one step closer to using the new type checker!

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/AppWrap.scala
```diff
@@ -0,0 +1,85 @@
...
+      case TlaOperDecl( name, params, body ) =>
+        TlaOperDecl( name, params, wrap( wrappableNames )(body) )
+      case d => d
+    }
+    new TlaModule( m.name, newDecls )
+  }
+}
+
+object AppWrap {
+  val NAME_PREFIX = "APP"
```
konnov, line 79: Could you rename "APP" to "CALL"?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ParameterNormalizer.scala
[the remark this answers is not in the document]
Kukovec: `_.copy(...)` is the default copy-constructor defined by the `TlaOperDecl` case class, so it only copies constructor arguments. The current implementation of the `_.isRecursive` field, for what I believe are legacy reasons, is an internal var, set to false by default, so it does not get copied. This is likely not how we want the class to behave in the future, but for the purposes of this PR, the above code is correct and necessary.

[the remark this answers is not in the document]
shonfeder: I saw the issue opened to track this #345. That seems like the right approach to me: the fix should be separate from the PR :+1:

test/tla/Rec2.tla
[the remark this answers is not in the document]
shonfeder: Thanks for explaining!

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
Kukovec: If I understand this correctly, the behavior you describe is that `ParameterNormalizer.recursive(x, y)` is equivalent to `ParameterNormalizer(x,y, DecisionFn.recursive)`? 
If so, I disagree, because this gives us more leeway in the future, w.r.t. introducing custom decision functions. For example, the current implementation allows us (or any other user) to seamlessly call normalization only on nullary operators, or only on operators from one particular module or even only on exactly one named operator. Having modes that we have to pre-program into the normalizer seems much more restrictive.

shonfeder: I wasn't suggesting remove the definition of `apply`, just that having to call with this predefined functions internally is clunky, and since you are providing them, it seems like it would be a more elegant API to write method to support that application. No biggie here tho.

Issue #345, opened 2020-12-02
[FEATURE] Change `TlaOperDecl` to use `isRecursive` as a case-class parameter


Kukovec:
Currently, `TlaOperDecl` delcares an `isRecursive` var. The problem is, it does not get copied with the default case-class copy constructor. This makes it easy to accidentally remove the recursive flag, when modifying operator declarations:
```
val d = TlaOperDecl("Decl",...)
d.isRecursive = true
val d2 = d.copy(name = "DeclCopy", ...) // d2.isRecursive == false !
```

Kukovec:
On a similar topic, we should remove the `var` tag from `TlaOperDecl.body`

konnov:
This seems to require a more major refactoring that we have thought

konnov:
Shall we prioritize this refactoring? I just found that `at.forsyte.apalache.tla.lir.transformations.impl.Lift` hits the same bug.

PR #347, opened 2020-12-04
Added renaming before the first Unroller call.

### LINKED ISSUE #313: [BUG] Inlining over the boundaries of recursive functions
This issue demonstrates a minimal working example that pops up in #297 after the fixes.

```tla
----------------- MODULE Rec13 ----------------------                           
EXTENDS Integers

VARIABLE y

LOCAL Send[x \in { 1, 2}] ==
    x \* at this point, we expect x = 2, not x = 1

SendAll(x) == Send[x + 1]

Init == y = SendAll(1)

Next == UNCHANGED y

Inv == y = 2
=====================================================
```

Here is the result of inlining, which is obviously broken:

```tla
----- MODULE Rec13 -----                                                        
VARIABLE y

Init ==
  y = LET Send == [ 1 \in { 1, 2 } |-> 1 ] IN [ x \in { 1, 2 } |-> x ][(1 + 1)]

Next == UNCHANGED y

Inv == y = 2
===============
```

I suspect that `Inliner` does not take into account the parameter definitions inside `TlaFunOper.recFunDef`. Instead, it erroneously propagates `1` from the upper call.

Kukovec:
closes #313 

Adds a temporary workaround, to address the problems described in #345 (unexpected behavior of `TlaOperDecl.copy` when the declaration is recursive)

### DIFF
--- a/test/tla/Rec13.tla
+++ b/test/tla/Rec13.tla
@@ -1,0 +1,16 @@
+----------------- MODULE Rec13 ----------------------                           
+EXTENDS Integers
+
+VARIABLE y
+
+LOCAL Send[x \in { 1, 2}] ==
+    x \* at this point, we expect x = 2, not x = 1
+
+SendAll(x) == Send[x + 1]
+
+Init == y = SendAll(1)
+
+Next == UNCHANGED y
+
+Inv == y = 2
+=====================================================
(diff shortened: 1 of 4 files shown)


shonfeder (approved):
LGTM

konnov (approved):
LGTM

PR #363, opened 2020-12-06
thoughts for the planning meeting

konnov:
Please read, comment, extend, etc.


shonfeder:
We agreed to move the notes to the wiki: https://github.com/informalsystems/apalache/wiki/Meeting-Notes

Issue #348, opened 2020-12-04
Change the definition of an assignment candidate 


Kukovec:
Currently, the notion of assignment candidates used is derived from the ABZ paper. In particular, an expression `x' = e` qualifies as an assignment candidate, regardless of its position as a subexpression of `Next`. Concretely, in
```
Next ==
  /\ x' = 1
  /\ y' = x' + 1
  /\ x' = 2
```
both `x' = 1` and `x' = 2` qualify as assignment candidates, and the extracted transition (in which expressions are reordered) would be either 
```
Next ==
  /\ x' := 1
  /\ y' := x' + 1
  /\ x' = 2
```
or
```
Next ==
  /\ x' := 2
  /\ y' := x' + 1
  /\ x' = 1
```
Unfortunately, we cannot predict which of the expressions `x' = 1` or `x' = 2` the SMT solver will (arbitrarily) select as the assignment to `x`, which makes some specifications harder to test and, consequently, some bugs harder to find.
We wish to move towards a more syntax-driven definition of assignment candidates, which uses the `assignment-before-use` criterion, i.e. an expression `x' = e` is only considered an assignment candidate if no preceding expression (in the left-to-right syntax order) uses the value of `x'`. Concretely, this would remove `x' = 2` as an assignment candidate in the above example, which more closely matches user expectations.

This change would be accompanied by the introduction of an analysis pass, similar to `CoverAnalysisPass`, in which instances of assignments after use trigger a warning to the user.

PR #364, opened 2020-12-07
Added minimal breaking retorderTest, disabled APABcastByz test

Kukovec:
Replaces the monolithic `APABcastByz.tla` test with a minimal problem-isolating spec.

This PR is expected to fail the integration tests, as it does not fix the problems with SMT-based assignment finding and reordering, but merely improves the tests themselves. 

Tests are expected to pass after #348.

### DIFF
--- a/test/tla/reorderTest.tla
+++ b/test/tla/reorderTest.tla
@@ -28,7 +28,7 @@ MayFail == /\ v1' = 1     \* (1)
            
 
 \* MustFail cannot be reordered, because (2) has a use-case dependency
-\* on (4) (and (3) on (1)), but the \E blocks are reordered as a single unit
+\* on (3) (and (4) on (1)), but the \E blocks are reordered as a single unit
 
 \* In left-to-right processing, assignments do not exist, due to the 
 \* assignment-before-use paradigm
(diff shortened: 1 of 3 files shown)


test/tla/reorderTest.tla
```diff
@@ -0,0 +1,44 @@
...
+\* is already correct
+MayFail == /\ v1' = 1     \* (1)
+           /\ \E x \in S: 
+             /\ v2' = x   \* (2)
+             /\ v1' > 0   \* (3)
+           /\ v1' = v2'   \* (4)
+           
+
+\* MustFail cannot be reordered, because (2) has a use-case dependency
+\* on (4) (and (3) on (1)), but the \E blocks are reordered as a single unit
```
shonfeder, line 31: It's not clear to me how (2) can be dependent on (4). I can see how (2) is dependent on (3), but don't know why `v2' > 0` could depend on `v1' > 0`. Could you clarify for the sake of my understanding?

shonfeder (approved):
I have one question (for my own edification), but LGTM

test/tla/reorderTest.tla
Kukovec: Ah, that's a typo, it should be (2) and (3) / (4) and (1) as you've observed.

Issue #346, opened 2020-12-04
[BUG] Failed to find assignments and symbolic transitions in InitPrimed


romac:
**Describe the bug**

Running Apalache v0.7.3-SNAPSHOT (03fdeff1af778172e8f9a8d344ac7d482a69c59b) against the [latest Tendermint light client specification](https://github.com/tendermint/spec/blob/master/rust-spec/lightclient/verification/Lightclient_003_draft.tla) results in the following error:

```
Assignment error: Failed to find assignments and symbolic transitions in InitPrimed E@15:05:27.381
```

**To reproduce**

1. Clone and build latest Apalache (currently 03fdeff1af778172e8f9a8d344ac7d482a69c59b)

```
$ git clone https://github.com/informalsystems/apalache/
$ cd apalache/
$ git checkout 03fdeff1af778172e8f9a8d344ac7d482a69c59b
$ mvn package
```

2. Clone the Tendermint specifications repository

```
$ git clone https://github.com/tendermint/spec/
$ cd spec/
$ git checkout 6abcb13dab6ef82f9ebf65b02ae7b861289d5742
$ cd rust-spec/lightclient/verification/
```

3. Run Apalache against [`MC4_3_correct.tla`](https://github.com/tendermint/spec/blob/master/rust-spec/lightclient/verification/MC4_3_correct.tla)

```
$ /path/to/apalache-mc check --inv=CorrectnessInv --length=5 MC4_3_correct.tla
```

(the choice of the invariant and length does not matter)

4. Apalache exits with:

```
Assignment error: Failed to find assignments and symbolic transitions in InitPrimed E@15:05:27.381
```

**Expected behavior**
Apalache exits with `EXITCODE: OK`

**Log files**

- [Tool output](https://gist.github.com/romac/7b4c6f057c35a343ece1833e0dce6f5f#file-output-log)
- [`detailed.log` ](https://gist.github.com/romac/7b4c6f057c35a343ece1833e0dce6f5f#file-detailed-log)

**Desktop**
 - OS: macOS 10.15.7
 - JDK version: OpenJDK 1.8.0_275 (`OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_275-b01)`)
 - Version: 0.7.3-SNAPSHOT (03fdeff1af778172e8f9a8d344ac7d482a69c59b)

**Additional context**

@konnov pointed out that this is related to #338.

**Temporary solution**

[...]

Issue #298, opened 2020-10-23
[FEATURE] Friendly assignment analysis


konnov:
The new `CoverAnalysisPass` produces extremely useful output about the state variables assigned in every operator. However, it is mostly unreadable on large examples:

```
TRUSTING_PERIOD covers: 
...
SwitchToPrimary covers: Faulty, blockchain, commonHeight, evidences, fetchedLightBlocks1, fetchedLightBlocks1b, fetchedLightBlocks2, lightBlockStatus1, lightBlockStatus1b, lightBlockStatus2, nextHeightToTry, now, state
Next covers: Faulty, blockchain, commonHeight, evidences, fetchedLightBlocks1, fetchedLightBlocks1b, fetchedLightBlocks2, lightBlockStatus1, lightBlockStatus1b, lightBlockStatus2, nextHeightToTry, now, state
TARGET_HEIGHT covers: 
...
```

(Note that `TRUSTING_PERIOD` and `TARGET_HEIGHT` are constants in this example, so the users are not interested in that output.)

Moreover, the users would not know what `cover` means and expect branch coverage, statement coverage, or similar.

A more informative output for the possible assignments may look like follows:

```
Foo.tla:200:8-200:75: SwitchToPrimary may update nextHeightToTry' 
Foo.tla:201:8-201:47: SwitchToPrimary may update state' 
Foo.tla:202:8-202:50: SwitchToPrimary may update blockchain' 
```

If you pad this info into three columns, that will be extremely useful to see, even if there are no bugs in the assignments.

Kukovec:
Closed by #366.

Issue #279, opened 2020-10-14
[FEATURE] Preprocessing-independent cover analysis 


Kukovec:
Currently, cover analysis relies on preprocessing transformations (e.g. inlining). We should be able to do it on unprocessed specifications.

Kukovec:
#366 removed the need for cover analysis.

PR #366, opened 2020-12-07
Reworked assignment search

### LINKED ISSUE #348: Change the definition of an assignment candidate

### LINKED ISSUE #353: Feature FAF: Stabilizing the assignment finder
This is a meta-issue that summarizes the work to be done to stabilize the assignment finder. All issues for this feature are labeled with [FAF](https://github.com/informalsystems/apalache/issues?q=is%3Aopen+is%3Aissue+label%3AFAF).

As the assignment finder is one of the key preprocessing steps, we have to finally stabilize it and test rigorously.

 - [x] A much simpler approach to finding assignments without SMT: #366
 - [x] A manual chapter: #368 
 - [x] A tutorial #369

### LINKED ISSUE #354: [FEATURE] Improve unit tests and integration tests for the assignment finder and transition slicer
We have been experiencing instability issues with the assignment finder and transition slicer for some time (due to refactoring and new features). We have to improve its coverage with unit tests and integration tests to detect these issues as early as possible.

Kukovec:
closes #348
closes #354 (together with #364)

As this approach to assignments makes cover analysis irrelevant, this PR implicitly closes/invalidates the following:
issues: #279, #298, #353, #111
open pr: #308

Potentially resolves issue #346: Under this build, the solver returns to expected behavior (EXITCODE: OK), but finds a deadlock in the listed specification. Author clarification required to close.

See also:  #62

### DIFF
--- a/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
+++ b/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
@@ -1,0 +1,174 @@
+package at.forsyte.apalache.tla.assignments
+
+import at.forsyte.apalache.tla.lir.oper._
+import at.forsyte.apalache.tla.lir.storage.{BodyMap, BodyMapFactory, SourceLocator}
+import at.forsyte.apalache.tla.lir.{LetInEx, NameEx, OperEx, TlaEx}
+import at.forsyte.apalache.tla.lir.transformations.TransformationTracker
+
+
+/**
+  * Performs the complete procedure of finding symbolic transitions from the TLA+ implementation.
+  *
+  * Assignment candidates follow the assignment-before-use paradigm.
+  *
+  * @see [[SymbTransGenerator]]
+  *
+  */
+class SmtFreeSymbolicTransitionExtractor(
+                                          tracker: TransformationTracker,
+                                          sourceLoc: SourceLocator
+                                        ) {
+
+  private def getLocString( ex: TlaEx ) = sourceLoc.sourceOf( ex ).getOrElse( "<[UNKNOWN]>" )
+
+  /** Checks whether expressions, which cannot contain assignments, use unassigned variables */
+  private def throwOnUseBeforeAssignment( unassignedVars: Set[String] ) : TlaEx => Unit = {
+    case ex@OperEx( TlaActionOper.prime, NameEx( name ) ) if unassignedVars.contains( name ) =>
+      val locString = getLocString( ex )
+      throw new AssignmentException( s"$locString: $name' is used before it is assigned." )
+    case OperEx( _, args@_* ) =>
+      args foreach throwOnUseBeforeAssignment( unassignedVars )
+    case LetInEx( body, defs@_* ) =>
+      defs foreach { d => throwOnUseBeforeAssignment( unassignedVars )( d.body ) }
+      throwOnUseBeforeAssignment( unassignedVars )( body )
+    case _ =>
+  }
+
+  // Partial state in the computation, with some subset of all variables already assigned and a partial strategy found
+  private case class PartialState( unassignedVars: Set[String], partialStrat: StrategyType )
+
+  // Since disjunction and ITE behave exactly the same w.r.t. assignments, we write a unified routine
+  private def handleDisjunctionOrITE( s: PartialState, letInOperMap: BodyMap , args: Seq[TlaEx] ): PartialState = {
+    // We independently process each disjunct
+    val childStates = args map { getStrategyOptInternal(s, letInOperMap)(_) }
+    val unassignedVarsSeq = childStates map { _.unassignedVars }
+    // We need to see whether all branches assign exactly the same variables
+    val unassignedEverywhere = unassignedVarsSeq.foldLeft( s.unassignedVars ){ _.intersect( _ ) }
+    val unassignedSomewhere = unassignedVarsSeq.foldLeft( Set.empty[String] ) { _.union( _ ) }
+    // Variables that are unassigned on at least one branch and assigned on all others
+    val imbalancedAssignments = unassignedSomewhere -- unassignedEverywhere
+    if (imbalancedAssignments.nonEmpty) {
+      // Report problms for each disjunct, if any
+      val problems : Seq[String] = args.zip( childStates ) flatMap {
+        case (childEx, PartialState( childUnassignedVars, _ )) =>
+          val s = childUnassignedVars.intersect( imbalancedAssignments )
+          if ( s.isEmpty ) None
+          else {
+            val locString = getLocString( childEx )
+            Some( s"$locString - Missing assignments to: ${s.mkString( "," )}" )
+          }
+      }
+
+      throw new AssignmentException( problems.mkString("\n") )
+    }
+    // Assuming all branches assign the same variables, works correctly if args.isEmpty (for whatever reason)
+    val unifiedStrat = (childStates map { _.partialStrat }).foldLeft( s.partialStrat ) { _ ++ _ }
+    PartialState( unassignedEverywhere, unifiedStrat )
+  }
+
+  // Transition method between partial states
+  private def getStrategyOptInternal( s: PartialState, letInOperMap: BodyMap ): TlaEx => PartialState = {
+    /** Base case, assignment candidates  */
+    case ex@OperEx( TlaOper.eq, OperEx( TlaActionOper.prime, NameEx( name ) ), star ) =>
+      // First, check if star contains unassigned varaible access
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      // if `name` is not yet assigned, this spot becomes an assignment
+      if ( s.unassignedVars.contains( name )) PartialState( s.unassignedVars - name, s.partialStrat :+ ex.ID )
+      else s
+
+    /** Base case, manual assignments */
+    case ex@OperEx( BmcOper.assign, OperEx( TlaActionOper.prime, NameEx( name ) ), star ) =>
+      // Similar to above, except manual assignments throw if spurious
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      if ( s.unassignedVars.contains( name )) PartialState( s.unassignedVars - name, s.partialStrat :+ ex.ID )
+      else {
+        val locString = getLocString( ex )
+        throw new AssignmentException( s"Manual assignment at $locString is spurious: $name is already assigned!" )
+      }
+
+    /** Conjunction */
+    case OperEx( TlaBoolOper.and, args@_* ) =>
+      // We sequentially update the partial state
+      args.foldLeft( s ) { getStrategyOptInternal(_, letInOperMap)(_) }
+    /** Disjunction */
+    case ex@OperEx( TlaBoolOper.or, args@_* ) =>
+      handleDisjunctionOrITE( s, letInOperMap,  args)
+
+    /** \E quantifier */
+    case OperEx( TlaBoolOper.exists, NameEx( _ ), star, subEx ) =>
+      // We need to check that star does not contain unassigned primes.
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      // if no exception is thrown, continue in setEx
+      getStrategyOptInternal( s, letInOperMap )( subEx )
+
+
+    /** ITE */
+    case OperEx( TlaControlOper.ifThenElse, star, thenAndElse@_* ) =>
+      // We need to check that star does not contain unassigned primes.
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      // The rest is equivalent to the disjunction case
+      handleDisjunctionOrITE( s, letInOperMap, thenAndElse)
+
+    /** Nullary LetIn */
+    case LetInEx( body, defs@_* ) =>
+      // Sanity check, all operators must be nullary and non-recursive
+      assert( defs.forall { d => d.formalParams.isEmpty && !d.isRecursive } )
+      // We memorize the let-in operators, to recall when we see an apply
+      val newMap = BodyMapFactory.makeFromDecls( defs, letInOperMap)
+      // finally, we check the let-in body as well
+      getStrategyOptInternal( s, newMap )( body )
+
+    /** Nullary apply */
+    case OperEx( TlaOper.apply, NameEx(operName) ) =>
+      // Apply may appear in higher order operators, so it might not be possible to pre-analyze
+      val knownBodyOpt = letInOperMap.get( operName ) map { _.body }
+      val newStateOpt = knownBodyOpt map { getStrategyOptInternal(s, letInOperMap)(_) }
+
+      newStateOpt.getOrElse( s )
+
+    /** Misc. operator */
+    case OperEx( _, args@_* ) =>
+      // For other operators, make sure they don't use unassigned variables ...
+      args foreach throwOnUseBeforeAssignment( s.unassignedVars )
+      // ... but don't update assignments
+      s
+    /** In the other cases, return the default args */
+    case _ => s
+  }
+
+  def getStrategy(vars: Set[String], actionEx: TlaEx): StrategyType = {
+    // The strategy can be extracted from the state obtained by starting from the initial state, where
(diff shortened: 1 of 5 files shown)


tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+      args.foldLeft( s ) { getStrategyOptInternal(_, letInOperMap)(_) }
+    /** Disjunction */
+    case ex@OperEx( TlaBoolOper.or, args@_* ) =>
+      handleDisjunctionOrITE( s, letInOperMap,  args)
+
+    /** \E quantifier */
+    case OperEx( TlaBoolOper.exists, NameEx( _ ), star, subEx ) =>
+      // We need to check that star does not contain unassigned primes.
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      // if no exception is thrown, continue in setEx
```
shonfeder, line 101:
```suggestion
      // if no exception is thrown, continue in subEx
```

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+    case OperEx( TlaControlOper.ifThenElse, star, thenAndElse@_* ) =>
+      // We need to check that star does not contain unassigned primes.
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      // The rest is equivalent to the disjunction case
+      handleDisjunctionOrITE( s, letInOperMap, thenAndElse)
+
+    /** Nullary LetIn */
+    case LetInEx( body, defs@_* ) =>
+      // Sanity check, all operators must be nullary and non-recursive
+      assert( defs.forall { d => d.formalParams.isEmpty && !d.isRecursive } )
```
shonfeder, line 115: I assume we just assert here because there are earlier checks that should guarantee this property by the time we get to this part of the process?

shonfeder (approved):
LGTM!

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+    /** Base case, assignment candidates  */
+    case ex@OperEx( TlaOper.eq, OperEx( TlaActionOper.prime, NameEx( name ) ), star ) =>
+      // First, check if star contains unassigned varaible access
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      // if `name` is not yet assigned, this spot becomes an assignment
+      if ( s.unassignedVars.contains( name )) PartialState( s.unassignedVars - name, s.partialStrat :+ ex.ID )
+      else s
+
+    /** Base case, manual assignments */
+    case ex@OperEx( BmcOper.assign, OperEx( TlaActionOper.prime, NameEx( name ) ), star ) =>
```
konnov, line 80: what does `star` mean here?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+    }
+    // Assuming all branches assign the same variables, works correctly if args.isEmpty (for whatever reason)
+    val unifiedStrat = (childStates map { _.partialStrat }).foldLeft( s.partialStrat ) { _ ++ _ }
+    PartialState( unassignedEverywhere, unifiedStrat )
+  }
+
+  // Transition method between partial states
+  private def getStrategyOptInternal( s: PartialState, letInOperMap: BodyMap ): TlaEx => PartialState = {
+    /** Base case, assignment candidates  */
+    case ex@OperEx( TlaOper.eq, OperEx( TlaActionOper.prime, NameEx( name ) ), star ) =>
```
konnov, line 72: what does `star` mean here?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+      else s
+
+    /** Base case, manual assignments */
+    case ex@OperEx( BmcOper.assign, OperEx( TlaActionOper.prime, NameEx( name ) ), star ) =>
+      // Similar to above, except manual assignments throw if spurious
+      throwOnUseBeforeAssignment( s.unassignedVars )( star )
+      if ( s.unassignedVars.contains( name )) PartialState( s.unassignedVars - name, s.partialStrat :+ ex.ID )
+      else {
+        val locString = getLocString( ex )
+        throw new AssignmentException( s"Manual assignment at $locString is spurious: $name is already assigned!" )
```
konnov, line 86:
```suggestion
        throw new AssignmentException( s"Manual assignment at $locString is spurious: $name is already assigned!" )
```
```suggestion
        throw new AssignmentException( s"$locString: manual assignment is spurious: $name is already assigned!" )
```

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+class SmtFreeSymbolicTransitionExtractor(
+                                          tracker: TransformationTracker,
+                                          sourceLoc: SourceLocator
+                                        ) {
+
+  private def getLocString( ex: TlaEx ) = sourceLoc.sourceOf( ex ).getOrElse( "<[UNKNOWN]>" )
+
+  /** Checks whether expressions, which cannot contain assignments, use unassigned variables */
+  private def throwOnUseBeforeAssignment( unassignedVars: Set[String] ) : TlaEx => Unit = {
+    case ex@OperEx( TlaActionOper.prime, NameEx( name ) ) if unassignedVars.contains( name ) =>
```
konnov, line 26: shall we add a case, where an assignment-like expression (or a manual assignment) appears in an expression that is not supposed to have assignments? I understand that line 26 covers that, but it does so in a subtle way. A direct warning to the user would be more helpful.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+      // Sanity check, all operators must be nullary and non-recursive
+      assert( defs.forall { d => d.formalParams.isEmpty && !d.isRecursive } )
+      // We memorize the let-in operators, to recall when we see an apply
+      val newMap = BodyMapFactory.makeFromDecls( defs, letInOperMap)
+      // finally, we check the let-in body as well
+      getStrategyOptInternal( s, newMap )( body )
+
+    /** Nullary apply */
+    case OperEx( TlaOper.apply, NameEx(operName) ) =>
+      // Apply may appear in higher order operators, so it might not be possible to pre-analyze
```
konnov, line 123: Can you elaborate on this comment? I don't understand it.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+      val newStateOpt = knownBodyOpt map { getStrategyOptInternal(s, letInOperMap)(_) }
+
+      newStateOpt.getOrElse( s )
+
+    /** Misc. operator */
+    case OperEx( _, args@_* ) =>
+      // For other operators, make sure they don't use unassigned variables ...
+      args foreach throwOnUseBeforeAssignment( s.unassignedVars )
+      // ... but don't update assignments
+      s
```
konnov, line 134: It is easy to miss `s` here. Actually, it is quite a short parameter name for such a long function. Shall we call it `state` or `partialState`?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -242,42 +234,6 @@ class SymbTransGenerator( tracker : TransformationTracker ) {
...
-      // Of all the assignment candidates, the one with the lowest position in `orderedAssignments` determines the rank
-      filtered.foldLeft( None : Option[Int] ) {
-        case (rankCand, uid) =>
-          val uidRank = orderedAssignments.find( _ == uid ) map {
-            orderedAssignments.indexOf
-          }
-          Reordering.IntOptOrdering.min( rankCand, uidRank )
-      }
-    }
-
```
konnov, line 280: Does it mean that we have completely get rid of the SMT version? I would keep it on a side and mark it as deprecated. It is still useful to have that code. We might want to use it in the future, perhaps, when we have a different SMT encoding in the model checker.

konnov (approved):
That looks just great! The code is much simpler to follow than the SMT solution. Let's merge and play with this new implementation of assignments!

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -0,0 +1,174 @@
...
+  *
+  * @see [[SymbTransGenerator]]
+  *
+  */
+class SmtFreeSymbolicTransitionExtractor(
+                                          tracker: TransformationTracker,
+                                          sourceLoc: SourceLocator
+                                        ) {
+
+  private def getLocString( ex: TlaEx ) = sourceLoc.sourceOf( ex ).getOrElse( "<[UNKNOWN]>" )
```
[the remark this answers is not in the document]
shonfeder: Cool. Thanks for the info :)

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
shonfeder: If there is a specific format we want for these kinds of messages, then I suggest we create a function to generate the desired formatting rather than having to enforce it manually. (This would be for a separate PR obviously).

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
shonfeder: If we decide to keep the code as deprecated, then how about wrapping this deprecated solver in a feature flag (that can be passed via CLI)? Otherwise, if the code is not going to be called at all by the application I really don't think we should we keep it. We can always recover it from the git history (i.e., we already have files with this code in it), but leaving dead code in the code base is liable to lead to errors down the line.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
Kukovec: Correct. After inlining, depending on an option, we either end up with no let-in operators at all, or with only nullary ones.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
Kukovec: Recall: ★ was the symbol we used in the ABZ paper for "expressions that do not contain assignments".

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
Kukovec: See above

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
Kukovec: "Assignment-like" is too broad, since it's not unreasonable to use something like `x' = v` in, say, an ITE condition. I could see doing that for manual assignments though.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
konnov: If the new version works for us, let's drop the support for the old implementation. But in order to find easily, we may want to introduce a clear tag that shows where the divergence has begun.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
Kukovec: This was intended to comment on cases where `operName` is an operator parameter or a CONSTANT -defined operator, but thinking about it, is it even possible to define nullary operator parameters? I imagine the parser would treat nullary parameters as simple?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
Kukovec: Just to comment on this a bit, the code removed here relates only to reordering, which is now unnecessary. I've opened a separate issue for removing SMT-related code in the future (#367)

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
shonfeder: Oh! lol, I assumed it just meant "wild card". Please make this meaning clear through some conspicuous comments in the file, or chose a more self-evident name :) :pray:

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
Kukovec: I've removed this comment for now, not sure it adds value

PR #407, opened 2021-01-08
Table of contents for manual.md

konnov:
Added TOC directly in SUMMARY.md, so the readers see the important sections in the sidebar.


konnov:
Forgot that it should be done on master. Rolling back.

PR #408, opened 2021-01-08
added manual TOC directly in the summary

konnov:
Added the contents of manual.md directly in the sidebar.


shonfeder:
Unfortunately this approach doesn't work. mdbook requires each of the chapters listed in the SUMMARY.md to be a file. When I compile the manual on this branch, I get

```
mdbook serve
2021-01-08 09:07:50 [ERROR] (mdbook::utils): Error: Chapter file not found, ./manual.md#apalacheOrTlc
2021-01-08 09:07:50 [ERROR] (mdbook::utils): 	Caused By: No such file or directory (os error 2)
```

The way to do with with mdbook (as is, we could open an issue for this if we think its a bug) is to break up each subchapter into it's own file. If you're OK with it, I can restructure as needed to achieve this outcome.

docs/src/SUMMARY.md
```diff
@@ -4,6 +4,19 @@
 
 - [Introduction](./README.md)
 - [Getting Started](./manual.md)
+    1. [Shall I use Apalache or TLC?](./manual.md#apalacheOrTlc)
```
shonfeder, line 7: Unfortunately this won't work. We'll need each subchapter to be in its own file.

shonfeder (changes requested):

docs/src/SUMMARY.md
konnov: I see. Let's split the manual into several chapters then.

shonfeder: I'll do this in a separate PR, since it's an orthogonal change, and get the change on top of #409

shonfeder:
The relevant issue on mdbook is here: https://github.com/rust-lang/mdBook/issues/167

PR #416, opened 2021-01-09
fixed the instructions for windows

konnov:
- [x] OpenJDK is hard to install, so now we recommend AdoptOpenJDK
 - [x] The command-line parameters did not work, now using `-jar` instead of `-cp`.


shonfeder (approved):

shonfeder:
LGTM, if you've tested that the `-jar` command works as expected.

PR #428, opened 2021-01-12
Added new manual section on assignments

### LINKED ISSUE #368: [FEATURE] Update the section on assignments in the user manual
Hopefully, we will use #366 soon. This means that the relevant text in the user manual has to be updated.

Kukovec:
closes #368


docs/src/idiomatic/README.md
```diff
@@ -51,7 +51,7 @@ reduce your discomfort when learning more advanced TLA+.
 __Idiom 0:__
     [Keep state variables to the minimum](000keep-minimum-state-variables.md) :battery:
 
-__Idiom 1:__ [Update state variables with assignments](001assignments.md) :date:
+__Idiom 1:__ [Update state variables with assignments](assignments.md) :date:
```
shonfeder, line 54: Should this be in the apalache user manual rather than in the idioms guidelines? It seems more suitable to the former.

Kukovec: Yeah, Igor and I already discussed this, it's probably getting moved to somewhere in the main body. I'm waiting on his review to figure out where exactly.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
+# Idiom 1: Update state variables with assignments
+For each TLA+ variable `x` defined in a specification, Apalache requires that every possible  transition defines an assignment to `x'`.
```
shonfeder, line 2: I think it would be helpful to also capture the intuitive concept here, not just the operational requirement. So maybe with an additional sentence like:

```suggestion
For each TLA+ variable `x` defined in a specification, Apalache requires that every possible  transition defines an assignment to `x'`. I.e., every possible transition must specify the next value after the transition for each TLA+ variable.
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Assignments and Assignment Candidates
+An _assignment candidate_ for `x` is an expression that has one of the following forms:
+- `x' = e`, 
+- `x' \in e`,
+- `UNCHANGED x`, or
+- `x' := e`
```
shonfeder, line 9:
```suggestion
- `x' := e` (this operator is specific to Apalache)
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+```tla
+A == x' > 0 /\ x' = 1
+B == y' = x' + 2 /\ x' = 1
+```
+In `A`, the expression `x' > 0` precedes any assignment to `x` and in `B`, while `y' = x' + 2` is an assignment candidate for `y`, it precedes any assignment to `x`, so both expressions are inadmissible (and would trigger exceptions). 
+
+Note that this only holds true if `A` (resp. `B`) is chosen as the transition operator. If `A` is called inside another transition operator, for example in `Next == x' = 1 /\ A`, no exception is thrown.
+
+### Balance
+In cases of disjunction, both (resp. all) disjuncts must have assignments for the same set of variables. In particular, if one disjunct contains an assignment candidate and another does not, such as in this example:
```
shonfeder, line 49: Seems to get the same point:

```suggestion
In cases of disjunction, all disjuncts must have assignments for the same set of variables. In particular, if one disjunct contains an assignment candidate and another does not, such as in this example:
```

docs/src/idiomatic/assignments.md
konnov, line 2: Can you also add a quick comparison to TLC? I think we are pretty close to how TLC does it right now.

konnov: Also, the standard terminology for transitions in TLA+ is *action*. So shall we talk about actions instead of transitions in this document? In the end, we construct symbolic transitions. But we can also view them as a particular case of actions.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+/\ \/ y = 1
+   \/ y' = 2
+/\ y' = 3
+```
+the assignment finder will throw an exception of the following shape:
+```
+Assignment error: test.tla:10:15-10:19: Missing assignments to: y
+```
+notifying the user of any variables for which assignments exist in some, but not all, disjuncts. 
+Note that if we correct the above example to
```
shonfeder, line 60: This would be clearer IMO if it only showed the disjuncts, and not also the conjunction. Then the conjunction demonstrating the syntax order could be introduced in the next paragraph.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
+# Idiom 1: Update state variables with assignments
+For each TLA+ variable `x` defined in a specification, Apalache requires that every possible  transition defines an assignment to `x'`.
+
+## Assignments and Assignment Candidates
+An _assignment candidate_ for `x` is an expression that has one of the following forms:
```
konnov, line 5:
```suggestion
An _assignment candidate_ for `x` is a TLA+ expression that has one of the following forms:
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+     ELSE y' = 2
+
+E == /\ x' = 2 
+     /\ \A s \in { t \in 1..10 : x' > t }: y' = s
+```
+Operator `A` contains assignments to both `x` and `y`; while `x' > t` uses `x'`, it does not violate the assignment-before-use rule, since the assignment to `x` precedes the expression, w.r.t. syntax order.
+In operator `B`, the assignment to `x` is missing, therefore `x' > t` produces an exception, as it violates assignment-before-use.
+The case in `C` is similar to `A`; conditions of the if-then-else operator may not contain assignments to `x`, so `x' = 0` can never be one, but they may use `x'`, since a preceding expression (`x' = 1`) qualifies as an assignment.
+The operator `D` produces an exception, for the same reason as `B`; even though `x' = 0` is an assignment candidate, if-conditions are assignment-free, so `x' = 0` cannot be chosen as an assignment to `x`.
+Lastly, while `E` looks almost identical to `A`, the key difference is that expressions under universal quantifiers may not contain assignments. Therefore, `y' = s` is *not* an assignment to `y` and thus violates assignment-before-use.
```
shonfeder, line 112: I think this would be easier to parse if you interleave the examples and the analysis, and and clearly flag in advance which examples are meant to be valid and which invalid, likely with an inline comment. So like

> Examples:
>
> ```tla
> A == /\ x' = 2 
>     /\ \E s \in { t \in 1..10 : x' > t }: y' = s
> ```
>
> Operator `A` contains assignments to both `x` and `y`; while `x' > t` uses `x'`, it does not violate the assignment-before-use rule, since the assignment to `x` precedes the expression, w.r.t. syntax order.
> 
> ```tla
> (* INVALID *)
> B == \E s \in { t \in 1..10 : x' > t }: y' = s
> ```
> 
> In operator `B`, the assignment to `x` is missing, therefore `x' > t` produces an exception, as it violates assignment-before-use.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
+# Idiom 1: Update state variables with assignments
+For each TLA+ variable `x` defined in a specification, Apalache requires that every possible  transition defines an assignment to `x'`.
+
+## Assignments and Assignment Candidates
+An _assignment candidate_ for `x` is an expression that has one of the following forms:
+- `x' = e`, 
+- `x' \in e`,
```
konnov, line 7:
```suggestion
- `x' \in S`,
```

docs/src/idiomatic/assignments.md
konnov, line 9:
```suggestion
- `x' := e` (note that `:=` is the operator defined in `Apalache.tla`)
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+- `UNCHANGED x`, or
+- `x' := e`
+
+While a transition may contain multiple assignment candidates for the same variable, not all of them are chosen as _assignments_ by Apalache. The sections below describe how the assignments are selected.
+
+### Minimality
+Assignments aren't spurious; each variable must have at least one assignment per transition operator, but no more than necessary to satisfy all of the additional constraints below.
```
konnov, line 14: This is not clear. A normal user of TLA+ has no idea about transition operators.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+Unlike other shapes of assignment candidates, whenever a manual assignment is used in a position where the assignment candidate would not be chosen as an assignment (either within assignment-free expressions or in violation of, for example, the syntax order rule) an  exception of the following shape is thrown:
+```
+Assignment error: test.tla:10:12-10:18: Manual assignment is spurious, x is already assigned!
+```
+or 
+```
+Assignment error: test.tla:10:15-10:21: Illegal assignment inside an assignment-free expression.
+```
+
+The benefit of using manual assignments, we believe, lies in synchronizing the user's and the tool's understanding of where assignments happen and can help prevent unexpected results, where the user's expectations or intuition regarding assignment positions are incorrect.
```
shonfeder, line 127:
```suggestion
The benefit of using manual assignments, we believe, lies in synchronizing the user's and the tool's understanding of where assignments happen. This helps prevent unexpected results, where the user's expectations or intuition regarding assignment positions are incorrect.
```

shonfeder (changes requested):
I like it!

docs/src/idiomatic/README.md
shonfeder: Seems like an `assignments.md` chapter under `apalache` would make the most sense to me.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+- `x' \in e`,
+- `UNCHANGED x`, or
+- `x' := e`
+
+While a transition may contain multiple assignment candidates for the same variable, not all of them are chosen as _assignments_ by Apalache. The sections below describe how the assignments are selected.
+
+### Minimality
+Assignments aren't spurious; each variable must have at least one assignment per transition operator, but no more than necessary to satisfy all of the additional constraints below.
+
+If all possible transitions fail to assign one or more variables, an exception of the following shape is thrown:
```
konnov, line 16: We really have to be clear about what transitions are. I would say that we start with actions, which is a standard terminology of TLA+. Then we can say that sometimes actions have different options. So we split actions into smaller ones, to have one assignment per refined action. This is what we call symbolic transitions. Perhaps, we should get rid of the transition terminology at all. Let's invent a name, like 'minimal actions' or something.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+```
+Assignment error: No assignments found for: x, z 
+```
+
+Such exceptions are usually the result of adding a `VARIABLE` without any accompanying TLA+ code relating to it.
```
konnov, line 21:
```suggestion
Such errors are usually the result of adding a `VARIABLE` without any accompanying TLA+ code relating to it. 
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+If all possible transitions fail to assign one or more variables, an exception of the following shape is thrown:
+```
+Assignment error: No assignments found for: x, z 
+```
+
+Such exceptions are usually the result of adding a `VARIABLE` without any accompanying TLA+ code relating to it. 
+The case where at least one transition, but not all of them, fails to assign a variable is shown below.
+
+### Syntax Order
+For the purpose of evaluating assignments, Apalache considers the left-to-right syntax order of conjuncts.
```
konnov, line 25: This is not a friendly language for TLA+ users. Can we say "and (/\\)" instead of conjunct? The point of our manual is not to teach them first-order logic, but to explain what we are doing.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+For the purpose of evaluating assignments, Apalache considers the left-to-right syntax order of conjuncts. 
+Therefore, as many assignments as possible are selected from the first (w.r.t. syntax order) conjunct, then from the second, and so on.
+
+Example:
+```tla
+Next == x' = 1 /\ x' = 2
+```
+In the above example, `x' = 1` would be chosen as an assignment to `x`, over `x' = 2`.
+
+### Assignment-before-use Paradigm
```
konnov, line 34:
```suggestion
### Assignment-before-use convention
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+If, in the syntax order defined above, an expression containing a primed variable `x'` precedes an assignment to `x`, the assignment finder throws an exception of the following shape:
```
konnov, line 35:
```suggestion
If, in the syntax order defined above, an expression containing a primed variable `x'` syntactically precedes an assignment to `x`, the assignment finder throws an exception of the following shape:
```

docs/src/idiomatic/assignments.md
Kukovec: True, but then I can't use it as a running example in the next 2 cases.

docs/src/idiomatic/assignments.md
Kukovec: Makes sense

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+```
+Assignment error: test.tla:10:16-10:17: x' is used before it is assigned.
+```
+notifying the user of any variables used before assignment. In particular, right-hand-sides of assignment candidates are subject to this restriction as well. Consider:
```
konnov, line 39:
```suggestion
notifying the user of any variables used before assignment. In particular, right-hand-sides of assignment candidates (e.g., TBD) are subject to this restriction as well. Consider:
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+```tla
+A == x' > 0 /\ x' = 1
+B == y' = x' + 2 /\ x' = 1
+```
+In `A`, the expression `x' > 0` precedes any assignment to `x` and in `B`, while `y' = x' + 2` is an assignment candidate for `y`, it precedes any assignment to `x`, so both expressions are inadmissible (and would trigger exceptions). 
+
+Note that this only holds true if `A` (resp. `B`) is chosen as the transition operator. If `A` is called inside another transition operator, for example in `Next == x' = 1 /\ A`, no exception is thrown.
```
konnov, line 46: "Transition operator" is really a non-standard terminology. I know what you are talking about, but a normal reader would be lost.

docs/src/idiomatic/assignments.md
konnov, line 49: user-friendly language: or (\\/)

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+the only assignment has to be `y' = 3`. While one of the disjuncts is an assignment candidate and the other is not, the balance requirement is not violated here, since neither disjunct is chosen as an assignment.
+
+Similar rules apply to if-then-else: both the `THEN _` and `ELSE _` branch must assign the same variables, however, the `IF _` condition is ignored when determining assignments.
+
+### Assignment-free Expressions
+Not all expressions may contain assignments. Given a transition operator `A`, based on the shape of `A`, the following holds:
+- If `A` has the shape `A_1 /\ ... /\ A_n`, then assignments are selected from `A_1, ... , A_n` sequentially, subject to the syntax-order rule.
+- If `A` has the shape `A_1 \/ ... \/ A_n`, then assignments are selected in all `A_1, ... , A_n` independently, subject to the balance rule.
+- If `A` has the shape `IF p THEN A_1 ELSE A_2`, then:
+    - `p` may not contain assignments. Any assignment candidates in `p` are subject to the assignment-before-use rule.
```
konnov, line 82: This may be confusing. We still allow the expressions that look like assignments, but we do not recognize them as such. The current writing may read like you cannot have those expressions at all.

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+
+D == IF x' = 0 
+     THEN y' = 1 
+     ELSE y' = 2
+
+E == /\ x' = 2 
+     /\ \A s \in { t \in 1..10 : x' > t }: y' = s
+```
+Operator `A` contains assignments to both `x` and `y`; while `x' > t` uses `x'`, it does not violate the assignment-before-use rule, since the assignment to `x` precedes the expression, w.r.t. syntax order.
+In operator `B`, the assignment to `x` is missing, therefore `x' > t` produces an exception, as it violates assignment-before-use.
```
konnov, line 109:
```suggestion
In operator `B`, the assignment to `x` is missing, therefore `x' > t` produces an error, as it violates assignment-before-use.
```

konnov, line 109:
```suggestion
In operator `B`, the assignment to `x` is missing, therefore `x' > t` produces an error, as it violates assignment-before-use.
```

docs/src/idiomatic/assignments.md
```diff
@@ -0,0 +1,129 @@
...
+The operator `D` produces an exception, for the same reason as `B`; even though `x' = 0` is an assignment candidate, if-conditions are assignment-free, so `x' = 0` cannot be chosen as an assignment to `x`.
+Lastly, while `E` looks almost identical to `A`, the key difference is that expressions under universal quantifiers may not contain assignments. Therefore, `y' = s` is *not* an assignment to `y` and thus violates assignment-before-use.
+
+## Manual Assignments
+Users may choose, but aren't required, to use manual assignments `x' := e` in place of `x' = e`.
+While the use of this operator does not change Apalache's internal search for assignments (in particular, using manual assignment annotations is *not* a way of circumventing the syntax order requirement), we encourage the use of manual assignments for clarity.
+
+Unlike other shapes of assignment candidates, whenever a manual assignment is used in a position where the assignment candidate would not be chosen as an assignment (either within assignment-free expressions or in violation of, for example, the syntax order rule) an  exception of the following shape is thrown:
```
konnov, line 118:
```suggestion
Unlike other shapes of assignment candidates, whenever a manual assignment is used in a position where the assignment candidate would not be chosen as an assignment (either within assignment-free expressions or in violation of, for example, the syntax order rule) an error like the one below is reported:
```

konnov (commented):
Great text! I have added some comments. Also, as we discussed, let's keep the idiom as before and add this new document to the manual.

docs/src/idiomatic/assignments.md
Kukovec: The problem here is granularity. From Lamport's book: 
```
... is an ordinary mathematical formula, except that it contains primed as well as unprimed variables.  Such a formula is called an _action_.  An action is true or false of a step
```
In this sense, `x' = 1 \/ x' = 2` is an action, whereas, in our terminology, it describes two transitions. In Apalache, the assignment constraints must hold for transitions, not just for actions.

konnov: Yes, that is why we have a special form of actions.

docs/src/idiomatic/assignments.md
Kukovec: In either case, since this is not something that is defined in Lamport's book, it seems like there needs to be some sort of chapter dedicated to explaining transitions/minimal actions, but I think this chapter should be its own PR.

docs/src/idiomatic/assignments.md
Kukovec: I think that's hard, you're not talking about `/\`, but about the sub-formulas that make up a `/\` (for which a name already exists and is used here - conjunct).

docs/src/idiomatic/assignments.md
Kukovec: See discussion on `/\`

docs/src/idiomatic/assignments.md
Kukovec: I disagree, there is a defined and stressed difference between _assignments_ and _assignment candidates_. The text here in no way implies that assignment candidates are forbidden, in fact, it mentions them explicitly, e.g.
```
    - `p` may not contain assignments. Any assignment candidates in `p` are subject to the assignment-before-use rule.
```

docs/src/idiomatic/assignments.md
shonfeder: I think users should be trusted to look up words they don't know, and that we should use the right language for what we're talking about.

But, if we don't think users can be trusted to google words they don't know, then perhaps we can add something along the lines of one of the two following fixes to address Igor's point?

```suggestion
For the purpose of evaluating assignments, Apalache considers the left-to-right syntax order of [conjuncts](https://en.wikipedia.org/wiki/Logical_conjunction). 
```

or

```suggestion
For the purpose of evaluating assignments, Apalache considers the left-to-right syntax order of conjuncts (the terms conjoined in a conjunction, e.g., `a` and `b` in `a /\ b`). 
```

docs/src/idiomatic/assignments.md
shonfeder: I would break this into two expository steps. 

First, make it clear and simple what you mean by `Balance` by giving an example that only includes the parts relevant.

Second, present your complicating example, introducing it by saying this will help explain some nuances/complications.

konnov:
What is the status of this PR? Shall I review it again?

Kukovec:
> 
> 
> What is the status of this PR? Shall I review it again?

I've added a commit (and a whole chapter) that hasn't been commented on yet.

shonfeder:
@Kukovec I don't see the chapter on actions here. Is it possible it didn't get added?

shonfeder (approved):

Kukovec:
> @Kukovec I don't see the chapter on actions here. Is it possible it didn't get added?

I can see it, it's part of a61b13e

shonfeder:
Oh! My bad. When you said "chapter" I was looking for a separate file, but I see it's a new section with the existing file. :+1:

docs/src/SUMMARY.md
```diff
@@ -26,6 +26,7 @@
 - [TLA+ Preprocessing](./apalache/preprocessing.md)
 - [Fine Tuning](./apalache/tuning.md)
 - [KerA: kernel logic of actions](./apalache/kera.md)
+- [Assignments in Apalache](./apalache/assignments.md)
```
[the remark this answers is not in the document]
konnov: Agree. This section is quite important. How about moving it right after "Specification parameters"?

docs/src/apalache/assignments.md
```diff
@@ -0,0 +1,233 @@
...
+Using some basic math, we see that action `B` can be equivalently written as `(y' - 1)*(y' - 1) = 0`, so it describes the exact same successor state, in which the new value of `y` is `1`. 
+What makes it different from action `A` is the fact that this is far from immediately obvious from the syntax. 
+The fact that there happened to be a closed-form solution for which gave us an integer value for `y'`, is a lucky coincidence, as `B` could have been, for example, `y' * y' + 1 = 0`, with no real roots. 
+To avoid cases like this, we require that transition operators explicitly declare the values of state variables in successor states.
+
+We call syntactic forms, which explicitly represent successor state values, _assignment candidates_. An assignment candidate for `x` is a TLA+ expression that has one of the following forms:
+- `x' = e`, 
+- `x' \in S`,
+- `UNCHANGED x`, or
+- `x' := e` (note that `:=` is the operator defined in `Apalache.tla`)
```
konnov, line 46:
```suggestion
- `x' := e` (note that `:=` is the operator defined in [Apalache.tla](https://github.com/informalsystems/apalache/blob/unstable/src/tla/Apalache.tla))
```

docs/src/apalache/assignments.md
```diff
@@ -0,0 +1,233 @@
...
+When writing non-trivial specifications, authors often end up with something similar to the following:
+```tla
+EventA == ...
+EventB == ...
+...
+
+Next == \/ EventA
+        \/ EventB
+```
+Specifically, `EventA` and `EventB` often represent mutually exclusive possibilities of execution.
```
konnov, line 60: This kind of suggests that we do not have non-determinism, which is not true.

docs/src/apalache/assignments.md
```diff
@@ -0,0 +1,233 @@
...
+B == x' = 1 \/ x' = 2 (* arguments are actions *)
+```
+Simply put, if all arguments to an operator `\/` are actions, then that operator is an action-or, otherwise it is a value-or. Similarly, if both the `THEN _` and `ELSE _` subexpressions of if-then-else are actions, it is an action-ITE, otherwise it is a value-ITE (in particular, a value-ITE can be non-Boolean).
+
+Using these two operators we can define the following terms:
+A _minimal action_ is an action which contains no action-or and no action-ITE.
+Conversely, a _compound action_ is an action which contains at least one action-or or at least one action-ITE.
+
+### Slices
+Given a transition operator, which is most commonly a compound action, we can decompose it into as many minimal actions as possible. We call this process _slicing_ and the resulting minimal actions _slices_.
```
konnov, line 79: We don't decompose it into "as many actions as possible". We decompose it into an equivalent set of minimal actions. And in practice it is a rather small set. Otherwise, we could blast Next by computing DNF.

konnov (approved):
This version of the manual looks great! I only added a few comments.

docs/src/apalache/assignments.md
Kukovec: I disagree, the "often" is there to suggest the exact opposite, that there are cases where this isn't true.

docs/src/apalache/assignments.md
Kukovec: "As many as possible" and "equivalent set of" aren't mutually exclusive. And neither are "as many as possible" and "few" (re: "small set")

PR #433, opened 2021-01-13
Deploy manual from unstable

shonfeder:
While we are working rapidly on the manual, catching up on the current
state of things, we want to deploy manual updates to unstable
immediately without having to wait for a release to master.

As discussed in our meeting on monday.


konnov (approved):
Looks good!

Issue #388, opened 2021-01-04
[FEATURE] Configure automated weekly releases


shonfeder:
As discussed in our strategy meeting today, we plan to cut weekly releases of the ongoing development on `unstable`, which we can then promote to major/minor/patch releases whenever we think its justified.

I'll update here with details of the planned release process before I begin implementation.

shonfeder:
This sketches out the general principle we&rsquo;ll employ. Technical details will be
added subsequently.

We will implement [trunk based development](https://trunkbaseddevelopment.com/), with the following features:

<a id="org1d6d81a"></a>

## Branching models

-   A single trunk for all development (this is just our current `unstable`,
    should we rename?)
-   Since we work remotely and asynchronously, instead of committing directly to
    trunk, we will work via **small**, **short lived** (i.e., 1-3 days) feature branches (we are
    already aiming for this), enabling
    -   Code Review
    -   CI to protect against breaking master

<a id="orgd41fccc"></a>

## Releases

We will follow [late creation of release branches](https://trunkbaseddevelopment.com/branch-for-release/#late-creation-of-release-branches).

-   Releases are cut from a tag off the trunk.
-   When bugs are reported:
    -   Branch or no branch?
        -   If possible, we just fix the bug in the trunk and cut a new release (note,
            this assumes that breaks to master are extraordinary, which they should be!).
        -   If a release needs ongoing support, because users cannot update to the
            latest version, **then** we cut a branch off of the tag.
            -   Reproduce the bug in the trunk.
            -   Add a test.
            -   Fix the bug the bug with a feature branch.
            -   Once it is integrated into trunk, cherry pick the fix into the release branch.

shonfeder:
## Implementation plan

-   Keep unreleased changes in a file `UNRELEASED.md`.
-   Configure a github workflow triggered by a manual event:
    <https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#manual-events>
    -   This will take two inputs, `releaseVersion` and `nextVersion`. See a dummy example here:
        <https://github.com/shonfeder/apalache/actions?query=workflow%3Arelease>
-   The workflow will trigger the following sequence:
    -   build and run the tests (including integration)
    -   Prepare the release
        -   Run `mvn versions:set -DnewVersion={{releaseVersion}}-RELEASE`
        -   Move the `UNRELEASED.md` into a file `RELEASE-NOTES.md`
        -   Add an empty `UNRELEASED.md`
[...]

PR #449, opened 2021-01-18
[release] 0.8.1

shonfeder:
This is a pilot run of the first stage of the new release process detailed in #388.

Once this is merged into `unstable`, I'll cut a release from 90f69bd using the existing release script, as the second stage is mostly just putting that process into CI.


CHANGES.md
```diff
@@ -2,6 +2,28 @@
      This file is generated. Do not write release notes here.
      Notes for unreleased changes go in ./UNRELEASED.md -->
 
+## 0.8.1
```
shonfeder, line 5: @konnov Are you OK with us no longer having this `[release]` suffix in the changes file?

mod-distribution/pom.xml
```diff
@@ -8,11 +8,11 @@
     <parent>
         <groupId>at.forsyte.apalache</groupId>
         <artifactId>apalache</artifactId>
-        <version>0.8.1-SNAPSHOT</version>
+        <version>0.8.1</version>
```
shonfeder: @konnov Note that I've also configured the release process to simply version releases as the plain version number, without the `-RELEASE` suffix. This follows other mature projects I'm aware of, wherein you know a version is a principle release through the absence of any qualifiers or suffixes. Just want to be sure you're OK with this :)

konnov: Yes. That sounds great. But maybe we should check with maven central or similar repos. It may be a maven thing to add `RELEASE`.

CHANGES.md
konnov: True. We do not need the suffix any longer.

mod-distribution/pom.xml
shonfeder: Here are a couple examples of release versions:

- https://github.com/google/guice/blob/4.2.3/pom.xml#L30
- https://github.com/scoverage/scoverage-maven-plugin/blob/scoverage-maven-plugin-1.4.1/pom.xml#L31

Here's an explainer on Maven version conventions, seeming to indicate that plain `major.minor.patch` versions are standard (but also explaining when we'd want to use qualifiers).

Seems we are safe to drop it? Unless you prefer otherwise.

PR #472, opened 2021-01-22
Replace double with single quotes in alias command

Alexander-N:
The alias does not work with double quotes since it means `$(pwd)` gets
expanded at the time the alias is created and so the mounted volume
stays the same wherever the command is run.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 
 name: build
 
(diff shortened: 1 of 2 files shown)


shonfeder (approved):
LGTM

Thanks for the fix, @Alexander-N! 

Please let us know if/as you encounter any further complications :)

shonfeder:
CI is not kicking off here, so I will fiddle with the PR a bit to try to get it going.

.github/workflows/main.yml
```diff
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
```
shonfeder, line 1: I'm sneaking this change into your PR @Alexander-N, since without it it seems our CI won't run your contribution.

This is also a valuable contribution for you, because you're the first outside contributor to help us since we switched from travis to github actions :)

PR #473, opened 2021-01-22
Strategy element duplication fix

Kukovec:
Fixed a bug where assignment strategies contained duplicate elements, causing noticeable slowdowns in `TransitionFinderPass` on formulas with many disjunctions.

### DIFF
--- a/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
+++ b/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
@@ -70,7 +70,9 @@ class SmtFreeSymbolicTransitionExtractor(
       throw new AssignmentException( problems.mkString("\n") )
     }
     // Assuming all branches assign the same variables, works correctly if args.isEmpty (for whatever reason)
-    val unifiedStrat = (childStates map { _.partialStrat }).foldLeft( s.partialStrat ) { _ ++ _ }
+    val unifiedStrat = (childStates map { _.partialStrat }).foldLeft( s.partialStrat ) {
+      case (a,b) => a ++ b.filterNot( a.contains )
+    }
     PartialState( unassignedEverywhere, unifiedStrat )
   }
 


shonfeder (changes requested):
Code looks good, but should we also add a unit test to ensure the behavior and protect against regressions?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
```diff
@@ -70,7 +70,9 @@ class SmtFreeSymbolicTransitionExtractor(
       throw new AssignmentException( problems.mkString("\n") )
     }
     // Assuming all branches assign the same variables, works correctly if args.isEmpty (for whatever reason)
-    val unifiedStrat = (childStates map { _.partialStrat }).foldLeft( s.partialStrat ) { _ ++ _ }
+    val unifiedStrat = (childStates map { _.partialStrat }).foldLeft( s.partialStrat ) {
+      case (a,b) => a ++ b.filterNot( a.contains )
+    }
```
konnov, line 75: I think it works as a fix, but it looks like quadratic complexity. Is there a reason for not using sets or maybe sorted sets?

konnov (approved):
Let's merge it, but think about maybe improving the data structure.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
Kukovec: The reason for not using just normal sets is that you care about order. For sorted sets, you'd have to specify what you're sorting by (syntax position), which is tedious.
While you're right about the quadratic size, this is, at worst, quadratic in the number of assignments in a spec, which is generally well under 500. 
In contrast, the bug this fixes possibly doubled, or tripled, the list size on every disjunction.
If you're adamant, I can go and make this `n log(n)`, but I don't think it's need at this point.

Kukovec:
> 
> 
> Code looks good, but should we also add a unit test to ensure the behavior and protect against regressions?

Unit tests wouldn't help here, as this is a performance issue, not a correctness issue.

shonfeder:
Ah I see, this is just an internal value in a private method, so not really susceptible for tests. :+1:

shonfeder (approved):

PR #484, opened 2021-01-24
Renamed track to trackEx

### LINKED ISSUE #439: [FEATURE] Rename TransformationTracker.track to TransformationTracker.trackEx
As we are adding `TransformationTracker.trackDecl` for transformations of declarations.

The change is trivial, but it touches upon tens of classes. So it needs an individual PR.

Kukovec:
closes  #439


konnov (approved):
LGTM. Thanks for taking care of it.

PR #470, opened 2021-01-21
Finish automating relases through CI

shonfeder:
Closes #388 
Closes #442

This PR adds the finishing touches needed to push the release automation through.  It also expands the documentation, in the hopes that I've provided enough pointers that anyone can pickup maintenance and manage the releases without needing to retrace my steps.

Please especially review [the documentation on releasing](https://github.com/informalsystems/apalache/blob/3f0e0bd8779cc3d862ec4ecb88328c81c301364a/CONTRIBUTING.md#releases), to be sure it all seems clear.

I've tested the pipeline on my fork. See 

- The automatically created release PR (release notes suppressed to reduce noise during testing): https://github.com/shonfeder/apalache/pull/11
- And the release generated on merging the PR: https://github.com/shonfeder/apalache/releases/tag/v0.0.2-test

- [ ] ~Tests added for any new code~
- [X] Documentation added for any new functionality
- [ ] ~Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality~

PR #487, opened 2021-01-25
Fix release cron schedule

shonfeder:
The release preparation job has been trying to run continuously since it became Monday wherever the server is located.

I apparently forgot my cron syntax. 

With this fix, it should only run once per week, instead of trying to run constantly on Monday.

Followup to #470


shonfeder:
Pull this in now so we don't waste compute cycles for no reason.

PR #495, opened 2021-01-25
Update readme to reflect new release process

shonfeder:
We don't need the master CI badge any more, since we're not releasing from master.

Also updated the section on releases and removed a duplicate link to the website added in #489

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,16 @@ manual][user-manual-installation].
 
 ## Website
 
-[apalache.informal.systems][] is the main website of the project.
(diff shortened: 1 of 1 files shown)


konnov (approved):
Looks good. One question: Are the links case-insensitive?

shonfeder:
> Looks good. One question: Are the links case-insensitive?

Link labels are not case sensitive (see the note on normalizing in https://github.github.com/gfm/#link-label), nor are URLs :) I believe this is why my auto formatter lower-cases the link labels.

PR #496, opened 2021-01-25
Add formatting automation

shonfeder:
- Adds formatting checks to our build configuration.
- Reformats all source files.
- Adds the formatting checks into our CI

Best to review by individual commits. The one massive formatting commit in 9608b88  only needs to be reviewed to decide if we are OK with his formatting, or if we'd like to tweak the configs. scalafmt is very configurable, See https://scalameta.org/scalafmt/docs/configuration.html (I've just added one config which they say is "popular").


shonfeder:
Tested to ensure the CI fails on incorrect formatting.

https://github.com/informalsystems/apalache/pull/496/checks?check_run_id=1765999521

It also prints instructions on how to fix it.

konnov (commented):
Do we really need maxColumn=80? E.g., idea has maxColumn=120 by default. The reformatted code is significantly longer in many case.

shonfeder:
> Do we really need maxColumn=80? E.g., idea has maxColumn=120 by default. The reformatted code is significantly longer in many case.

I don't have any preferences re: our formatting. I'll change the maxColumn to 120. Anything else you'd like to see tweaked about the configuration?

konnov:
Could you re-run it with 120 chars? Maybe all of my problems come from that

konnov:
I am trying `preset=Scala.js`. This is probably the standard convention: https://docs.scala-lang.org/style/

konnov:
These settings look good to me:

```
maxColumn = 120
preset = Scala.js
```

shonfeder:
Configured as specified.

konnov (approved):
The settings look good to me and they actually make more sense than what I used before

Issue #226, opened 2020-09-23
[FEATURE] Parse annotations from the SANY comments


konnov:
As discussed in #162, it should be possible to write type annotations in comments:

```tla
VARIABLE
  \* type: Set(Int)
  S

\* type: (Int) => Bool
Mem(e) == \E x \in S: x = e
```

We should integrate annotations in `SanyImporter` and propagate them to the type checker.

konnov:
It is actually a good idea to hide the annotations in comments. Our current approach with extending the Apalache module introduces an additional dependency.

konnov:
This is implemented in #503.

PR #503, opened 2021-01-28
Java-like annotations in comments

konnov:
This PR addresses two related issues: #262 and #226. Instead of developing a special solution for type annotations, we produce a simple and general syntax of annotations, like in Java. To make use of the annotations, this PR also extends `SanyImporter` to parse annotations in the comments that are written in front of declarations. The annotation syntax and examples are given in ADR004 (this PR).

On top of that, `TestAnnotationParser` is the first property-based test in Apalache :-)


UNRELEASED.md
```diff
@@ -10,3 +10,11 @@
          * Another change description, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
+### SANY importer
```
shonfeder, line 13: I wonder if this shouldn't just be under `Features`?

docs/src/adr/004adr-annotations.md
```diff
@@ -0,0 +1,136 @@
...
+## 1. What can be annotated
+
+Annotations should be written in comments that are written in front of a
+declaration. The following declarations are supported:
+ 
+ 1. Constant declarations, e.g., `CONSTANT N`.
+ 1. Variable declarations, e.g., `VARIABLE x`.
+ 1. Top-level operator declarations, e.g., `Foo(x) == e`.
+ 1. LET-IN operator declarations, e.g., `Foo(x) == LET Inner(y) == e IN f`.
+ 1. Recursive functions, e.g., `foo[i \in Int] == e`.
```
shonfeder, line 21: Is this distinct from the LET-IN declarations? I've only seen functions defined in LET-IN decs.

docs/src/adr/004adr-annotations.md
```diff
@@ -0,0 +1,136 @@
...
+ 1. Recursive operators, that are declared either at the top-level or by using
```
shonfeder, line 22: This also seems like not it's own item, but something that is a sub-item under, respectively, top level operator declarations and LET-IN declarations.

docs/src/adr/004adr-annotations.md
```diff
@@ -0,0 +1,136 @@
...
+  - Any character except double quotes, control characters or backslash `\`
+  - A backslash followed by another backslash, a single or double quote,
+    or one of the letters `b`, `f`, `n`, `r` or `t`
+  - `\` followed by u followed by four hexadecimal digits
+  
+**Examples.** The following strings are examples of syntactically correct
+annotations:
+
+ - `@tailrec`
+ - `@type("(Int, Int) => Int")`
```
shonfeder, line 54: since we're making up our own syntax, can't we do away with these ugly types as strings thing?

docs/src/adr/004adr-annotations.md
```diff
@@ -0,0 +1,136 @@
...
+## 5. Discussion
+
+Most likely, this topic does not deserve much discussion, as we are using
+the pretty standard syntax of Java annotations. So we are following the
+principle of the least surprise.
+
+A more concise syntax for type annotations could be as follows:
+
+```tla
+\* @type: Int => Int
```
shonfeder, line 125: Would we have the same problem with `@type(Int => Int)`?

docs/src/adr/004adr-annotations.md
```diff
@@ -0,0 +1,136 @@
...
+Inc(n) == n + 1
+```
+
+However, this concise syntax is technically harder to parse (SANY is prunning
```
shonfeder, line 129: How much harder to parse?

test/tla/Annotations.tla
```diff
@@ -0,0 +1,37 @@
+-------- MODULE Annotations ----------
```
shonfeder, line 1: Should it include annotations of functions?

konnov:
@shonfeder the formatting check fails on a markdown file. What shall we do?

shonfeder (commented):
I left some comments, but I'm hoping the size of this PR will be very reduced once the formatting is fixed (as per the failing CI job) so I'll hold of on reviewing more until then.

shonfeder:
@konnov did you already try `make fmt-fix`?

konnov:
interesting. I forgot to do it. Can we add it to a PR template.

docs/src/adr/004adr-annotations.md
konnov: I agree that it would be good to have a nicer syntax. But that seems to be hard, as SANY simply merges all strings in a comment together by removing `\n`.

docs/src/adr/004adr-annotations.md
konnov: yes, the types may contain parentheses, that will confuse the annotation parser a lot. These are two separate parsers: one of java-like annotations, and one for types.

docs/src/adr/004adr-annotations.md
konnov: See the answer above. I was about to do it and then realized that it is nearly impossible with the current AST that we get from SANY

test/tla/Annotations.tla
konnov: Yes, `FactFun` is a recursive function. It is often hard to distinguish functions from operators in TLA+

docs/src/adr/004adr-annotations.md
konnov: The reason why I'd like to use java annotations is that we can use them for other hints, e.g., `@pure` to indicate pure functions.

UNRELEASED.md
konnov: It should. I just followed the template in `UNRELEASED.md` :)

konnov:
There are plenty of small replacements in files, basically adding one more parameter `annotationStore: AnnotationStore`.

UNRELEASED.md
shonfeder: If you think the example in `UNRELEASED.md` could be improved for clarity, a fix to the preamble in https://github.com/shonfeder/apalache/blob/shon/fix-release-of-zip/script/release-notes.sh#L19-L30 would be appreciated :)

docs/src/adr/004adr-annotations.md
shonfeder: Too bad :,(

shonfeder (approved):
LGTM

docs/src/adr/004adr-annotations.md
konnov: I know. If people start using our annotations, we can talk about extending the TLA+ parser :-)

tla-import/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporter.scala
```diff
@@ -1149,64 +1165,64 @@ class TestSanyImporter extends FunSuite {
...
+        "E3",
+        OperEx(
+            TlaFunOper.except,
+            NameEx("f"),
+            TlaFunOper.mkTuple(
+                TlaFunOper
+                  .mkTuple(ValEx(TlaInt(0)), ValEx(TlaInt(1)), ValEx(TlaInt(2)))
+            ),
+            ValEx(TlaInt(3))
+        )
```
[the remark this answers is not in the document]
konnov: That is strange indeed. I've changed `TestSanyImporter`, but not that part of it

docs/src/adr/004adr-annotations.md
konnov: The problem is that we don't have an end marker, which is usually `\n`. How about add the infamous semicolon in the end of a quote-free annotation:

```tla
\* @type: (Int, Int) => Int ;
```

konnov:
I have resolved the conflicts. The last commit is massive, as it merges the formatting reversal from unstable

docs/src/adr/004adr-annotations.md
konnov: There are operators and a special form for functions, usually, recursive ones. I've grouped them differently.

docs/src/adr/004adr-annotations.md
konnov: I have added this special form to the annotations. It definitely has less clutter when writing type annotations.

PR #511, opened 2021-01-28
Revert global scalafmt changes

shonfeder:
We're finding that the formatting changes applied in #496 are leading to massive merge conflicts. The hop is that by reverting those formatting changes, and instead configuring our checks to add incremental formatting checks per file, we can avoid the nightmare of massive merge conflict resolutions.


shonfeder:
It's expected that the formatting checks will fail here, because we're changing the files relative to unstable, to revert them back to their pre-formatted state.

konnov (approved):
Let's revert

PR #514, opened 2021-01-29
Fix CI formatting check

shonfeder:
We "ratchet" our formatting checks, which means we only check files that
have changed relative to `origin/unstable`. This makes sure we have
fetched the latest changes before running those checks, so we're getting
an accurate report.

Apparently the checkout action doesn't fetch our remote branches, but we need
unstable for our ratcheted format checking.

This is causing the formatting check to fail on on #503  and #467

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ test/tla/%.md: target/test/tla/%.md.corrected
 	cp -f $< $@
 
 fmt-check:
+	git fetch origin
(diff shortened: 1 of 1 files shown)


konnov (approved):
LGTM

shonfeder:
This only affects the formatting check, so I only waited for that to pass in CI.

PR #516, opened 2021-01-29
Fix scalafmt config

shonfeder:
The docs led me astry, and the nonsense config was just ignored.
Igor fixed it.


konnov (approved):
Yes! That's perfect!

PR #541, opened 2021-02-03
Some fixes for the manual

Alexander-N:
I think one "errors" section was misplaced. I removed it completely for now, since nothing seemed better than something confusing.


konnov (approved):
Thank you for spotting the errors! All changes are to the point.

PR #545, opened 2021-02-05
Correct flags on apalache-jar make target

shonfeder:
This is meant to be a speedy build of the jar without running tests,
but I had the flags wrong I think.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,7 +81,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
       - name: Build apalache
-        run: make apalache-jar
+        run: make build-quick
       - name: Run integration tests
         run: ./test/run-integration
 
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ apalache:
 	mvn package
 
 apalache-jar:
-	mvn --batch-mode --no-transfer-progress -DskipTests package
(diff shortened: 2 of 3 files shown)


konnov (approved):
Looks good. Shall we add a line on that target in the manual? Otherwise, the users will never find it by themselves.

shonfeder:
Turns out the flags like this don't make any difference in running time, and the tests are skipped either way. So I was wrong about the need for this PR. Still, this flag arrangement seems as good as the other.

Where in the manual do you think this should go? (It's currently used in CI).

konnov:
Here looks like a good place: https://apalache.informal.systems/docs/apalache/installation/source.html

shonfeder:
Do you want me to replace the currently advised `make` target that runs the tests with the `make apalache-jar` target?

konnov:
I would say that we should have `make` by default. But you can make a note that if the user wants to bypass the tests and make the build faster (e.g., they do it on regular basis), they can use the special target.

shonfeder:
I don't think we should complicate the installation instructions with superfluous details. I think there should just be one way to install in the user documentation.

Perhaps this additional should note should go in the contributing docs? But then, any contributor who's gonna build should just look at the makefile.

shonfeder:
Tho I guess we already have some optional items in the install instructions. OK I'll add this there.

PR #552, opened 2021-02-05
Emojify docs

### LINKED ISSUE #479: [BUG] Emojis are not displayed in the manual
The github emojis are not displayed in the mdbook manual, for example in https://apalache.informal.systems/docs/lang/index.html there are only the text representations like `:electric_plug:` and `:bulb:`. Maybe they could simply be replaced with unicode characters?

shonfeder:
Closes #479 

Last weekend I hacked together a little command line utility for replacing and looking up GitHub emoji names, because I couldn't find any CLIs utilities to this in a unixy way and it seemed useful. You can find it at https://github.com/shonfeder/emojitsu .

This PR adds that utility into our deployment of the mdbook docs, so that the emoji names will be replaced with their unicode equivalents during deployment (if available).

I cheated a deployment into the CI for this PR, and you can see the results here: https://apalache.informal.systems/docs/idiomatic/index.html#the-idioms

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,15 +20,6 @@ jobs:
           mdbook-version: "0.4.5"
           # mdbook-version: 'latest'
 
-      # - name: Use OCaml 4.11.0
-      #   uses: actions-ml/setup-ocaml@8047ff86bc363ab5a0ba4323ac99e1800daaa622
-      #   with:
-      #     ocaml-version: 4.11.0
-
-      # - name: Install emojitsu
-      #   run: |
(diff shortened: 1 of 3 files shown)


konnov (approved):
Looks great! You should probably send this to mdbook developers too, so other people would not have to repeat your work.

shonfeder:
@konnov Yep, will do.

shonfeder:
I posted a link on mdBook's open issue for emoji suport, and added instructions for use as an mdBook preprocessor (we can also integrate the utility into our own mdBook config if we want, instead of just running it in CLI.

PR #553, opened 2021-02-07
Docs + reformat

### LINKED ISSUE #438: [FEATURE] ReplaceFixed needs documentation
The standard transformation `at.forsyte.apalache.tla.lir.transformations.standard.ReplaceFixed` has no documentation. We should fix that.

### LINKED ISSUE #566: [FEATURE] Refactor ReplaceFixed
How about the following refactoring:

1. Instead of the object, introduce a class that received `TransformationTracker` as a constructor parameter. You can also introduce a companion object for convenience.

2. The method `apply` of class `at.forsyte.apalache.tla.lir.transformations.standard.ReplaceFixed` has the following signature:

```scala
  def apply(
      replacedEx: TlaEx,
      newEx: => TlaEx,
      tracker: TransformationTracker
  ): TlaExTransformation
```

By doing step 1, we can get rid of `tracker`. I would also prefer having a matching predicate (maybe even a partial function?) instead of a concrete expression, that is:

```scala
  def apply(
      needsReplacement: TlaEx => Boolean,
      replaceWithEx: => TlaEx
  ): TlaExTransformation
```

Kukovec:
closes #438 
closes #566


konnov (approved):
Looks good. Since you are on it, could you extract the tests for `ReplaceFixed` from `TestTransformations`? We should upgrade our test for transformations. There are just a few.

Kukovec:
> 
> 
> Looks good. Since you are on it, could you extract the tests for `ReplaceFixed` from `TestTransformations`? We should upgrade our test for transformations. There are just a few.

When you say "extract", do you mean moving them to a file containing _only_ `ReplaceFixed`-related tests?

konnov:
> When you say "extract", do you mean moving them to a file containing only ReplaceFixed-related tests?

Exactly. Now all transformation tests are mixed in one file. When you have a designated file per transformation, you will see how few tests there are.

Kukovec:
@konnov , could you look over this again please? I decided to include some changes to also close #566 with this, after you had already approved.

konnov (approved):
Looks good!

Issue #530, opened 2021-02-01
Improve JSON format for counterexamples


andrey-kuprianov:
The current JSON format for counterexamples follows the general TLA+ to JSON formatting of Apalache. As correctly pointed out by @vitorenesduarte, counterexamples have much simpler structure than the general TLA+ -- only ground formulas. So the facilities provided by JSON (maps and arrays) should be essentially enough to represent everything needed, without the special syntax.

It would be nice to implement this specialised syntax for counterexamples -- this would simplify the task of integrating with Apalache and counterexample parsing substantially.

konnov:
Yes. It would be a nice feature. Let's discuss at the planning meeting when we like to schedule some work on it.

konnov:
Shall we produce the same JSON structure as TLC: https://github.com/tlaplus/CommunityModules/blob/master/modules/Json.tla

konnov:
Following the discussion in #599, that does not look like a priority any longer. Do you still want it implemented in Apalache or shall we close the issue?

andrey-kuprianov:
I think for usability it would be still nice to generate JSON counterexamples directly from Apalache. But taking into account that there is a workaround present, it becomes less urgent, so let's move this issue to a backlog, may be.

konnov:
If there is a need for this, we could implement a simplifier in an external script, e.g., in python. However, since JSON is not human-readable, I do not see value in supporting two formats. It only can lead to more bugs.

konnov:
I have reopened the issue. In the hindsight, it is actually an important issue. It will let us to decouple testing tools from Apalache/TLC.

konnov:
This is done: https://apalache.informal.systems/docs/adr/015adr-trace.html

PR #555, opened 2021-02-07
Counterexample formatting

### LINKED ISSUE #265: [BUG] Checking FALSE invariant produces funny counterexample
```tla
------------------------- MODULE counterexample -------------------------

EXTENDS MC

(* Initial state *)

State1 ==
TRUE                             <--- What happened to the variable `x`?
(* Transition 0 to State2 *)     <--- Strange that states are index by Nat \ {0}, but transitions by Nat

State2 ==
/\ x = 1

(* The following formula holds true in the last state and violates the invariant *)  <--- Why doesn't State1 already violate FALSE?

InvariantViolation == TRUE
```

```tla
----------------------------- MODULE Github484 -----------------------------
VARIABLE x

T == TRUE
F == FALSE

Init == 
  /\ x \in {1}
  
Next == 
  /\ x' \in {1}
```

```bash
markus@avocado:~/src/TLA/_specs/models/tlcbugs/Github484/Github484.toolbox/Model_1$ /opt/apalache/bin/apalache-mc --inv=F --init=Init --next=Next check MC
# Tool home: /opt/apalache
# Package:   /opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar
# JVM args: -Xmx4096m -Djava.library.path=/usr/lib/x86_64-linux-gnu/jni/  -DTLA-Library=
#
# APALACHE version 0.7.2-RELEASE build v0.7.1-5-g5b85083
#
# WARNING: This tool is in the experimental stage.
#          Please report bugs at: [https://github.com/informalsystems/apalache/issues]

Checker options: filename=MC, init=Init, next=Next, inv=F         I@17:48:34.761
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/opt/apalache/mod-distribution/target/apalache-pkg-0.7.2-RELEASE-full.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
PASS #0: SanyParser                                               I@17:48:35.124
Parsing file /home/markus/src/TLA/_specs/models/tlcbugs/Github484/Github484.toolbox/Model_1/MC.tla
Parsing file /home/markus/src/TLA/_specs/models/tlcbugs/Github484/Github484.toolbox/Model_1/Github484.tla
Parsing file /tmp/TLC.tla
Parsing file /tmp/Naturals.tla
Parsing file /tmp/Sequences.tla
Parsing file /tmp/FiniteSets.tla
PASS #1: ConfigurationPass                                        I@17:48:35.395
  > Loading TLC configuration from MC                             I@17:48:35.397
  > No TLC configuration found. Skipping.                         I@17:48:35.401
PASS #2: UnrollPass                                               I@17:48:35.423
  > Unroller                                                      I@17:48:35.427
PASS #3: InlinePass                                               I@17:48:35.445
  > InlinerOfUserOper                                             I@17:48:35.447
  > LetInExpander                                                 I@17:48:35.450
PASS #4: PrimingPass                                              I@17:48:35.455
  > Introducing InitPrimed for Init'                              I@17:48:35.459
PASS #5: VCGen                                                    I@17:48:35.474
  > Producing verification conditions from the invariant F        I@17:48:35.475
  > VCGen produced 1 verification condition(s)                    I@17:48:35.476
PASS #6: PreprocessingPass                                        I@17:48:35.484
  > Before preprocessing: unique renaming                         I@17:48:35.485
 > Applying standard transformations:                             I@17:48:35.494
  > Desugarer                                                     I@17:48:35.496
  > UniqueRenamer                                                 I@17:48:35.503
  > Normalizer                                                    I@17:48:35.519
  > Keramelizer                                                   I@17:48:35.522
  > After preprocessing: UniqueRenamer                            I@17:48:35.529
PASS #7: TransitionFinderPass                                     I@17:48:35.548
  > Found 1 initializing transitions                              I@17:48:35.628
  > Found 1 transitions                                           I@17:48:35.643
  > No constant initializer                                       I@17:48:35.643
  > Applying unique renaming                                      I@17:48:35.644
PASS #8: OptimizationPass                                         I@17:48:35.655
 > Applying optimizations:                                        I@17:48:35.659
  > ConstSimplifier                                               I@17:48:35.660
  > ExprOptimizer                                                 I@17:48:35.664
  > ConstSimplifier                                               I@17:48:35.665
PASS #9: AnalysisPass                                             I@17:48:35.672
 > Marking skolemizable existentials and sets to be expanded...   I@17:48:35.674
  > SkolemizationMarker                                           I@17:48:35.675
  > ExpansionMarker                                               I@17:48:35.676
 > Running analyzers...                                           I@17:48:35.679
  > Introduced expression grades                                  I@17:48:35.687
  > Introduced 2 formula hints                                    I@17:48:35.688
PASS #10: BoundedChecker                                          I@17:48:35.688
No CONSTANT initializer given                                     I@17:48:35.764
Step 0, level 0: checking if 1 transition(s) are enabled and violate the invariant I@17:48:35.768
Invariant is violated at depth 0. Check the counterexample in any of counterexample.tla, MC.out, counterexample.json E@17:48:35.856
The outcome is: Error                                             I@17:48:35.864
Checker has found an error                                        I@17:48:35.864
It took me 0 days  0 hours  0 min  1 sec                          I@17:48:35.865
Total time: 1.165 sec                                             I@17:48:35.865
EXITCODE: OK
```

Kukovec:
closes #265


andrey-kuprianov:
@Kukovec, nice:)

If you are working on this anyway, I would like to point to something that bothered me about Apalache counterexamples a bit. If an invariant contains quantifiers, then `InvariantViolation` will contain skolem constants with `$`, which doesn't fit the TLA+ syntax and prevents the counterexample from loading in the TLA+ Toolbox for example.

Would be very nice if you could fix this:)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -57,38 +57,41 @@ class TlaCounterexampleWriter(writer: PrintWriter) extends CounterexampleWriter
...
-  override def write(rootModule: TlaModule, notInvariant: NotInvariant, states: List[NextState] ) : Unit = {
+  override def write(rootModule: TlaModule, notInvariant: NotInvariant, states: List[NextState]): Unit = {
     val pretty = new PrettyWriter(writer)
 
     writer.println("%s MODULE counterexample %s\n".format("-" * 25, "-" * 25))
     writer.println("EXTENDS %s\n".format(rootModule.name))
 
     states.zipWithIndex.foreach {
+      case (state, 0) =>
+        writer.println("(* Constant initialization state *)")
```
konnov, line 76: It is strange that we mix `PrettyWriter` and `PrintWriter` in this code. `PrettyWriter` may have its own print buffer and it does know when to flush the buffer. I would prefer to extend `PrettyWriter.toDoc(mod: TlaModule)` with an optional parameter that receives the modules to be extended.

Of course, that will require some fiddling with kyama `PrettyPrinter`, but the result is rewarding. The closest code is probably in `CASE`, though you will need "," instead of "[]" as a separator:

```scala
        val doc = group(text("CASE") <> nest(space <> folddoc(pairsWithOther.toList, _ <> line <> "[]" <> space <> _)))
```

konnov: I have checked `PrettyWriter` and you must be safe w.r.t. buffers as long as you call `write` methods. However, moving the `EXTENDS` code to `PrettyWriter.toDoc` is still valuable, as you can reuse it.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -57,38 +57,41 @@ class TlaCounterexampleWriter(writer: PrintWriter) extends CounterexampleWriter
...
+        writer.println(s"\nStateConst ==")
```
konnov, line 77:
```suggestion
        writer.println(s"\nConstInit ==")
```

konnov: We usually call a constant initializer `ConstInit` in the specs and the manual.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -57,38 +57,41 @@ class TlaCounterexampleWriter(writer: PrintWriter) extends CounterexampleWriter
...
+        printStateFormula(pretty, state._2)
+        writer.println()
       case (state, j) =>
-        val i = j+1 // start counting from 1, for compatibility with TLC counterexamples
-        if(i == 1) {
+        // start counting from 0, for compatibility with TLC counterexamples
```
konnov, line 81: This comment does not help much. Shall we say that index 0 is reserved for `ConstInit`, whereas indices `1..n` are reserved for states. However, users expect `State0` to be the initial state, so we count states from 0.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -98,55 +101,48 @@ class TlaCounterexampleWriter(writer: PrintWriter) extends CounterexampleWriter
...
+                    |@!@!@STARTMSG 2110:1 @!@!@
+                    |Invariant is violated.
+                    |@!@!@ENDMSG 2110 @!@!@
+                    |@!@!@STARTMSG 2121:1 @!@!@
+                    |The behavior up to this point is:
+                    |@!@!@ENDMSG 2121 @!@!@
+                    |""".stripMargin)
+
+    states.zipWithIndex.foreach { case (state, j) =>
+      val i = j + 1 // start counting from 1, for compatibility with TLC counterexamples
```
konnov, line 119: There is probably something wrong here, as index 0 would correspond to the constant initializer.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/io/TestCounterexampleWriter.scala
```diff
@@ -10,30 +10,32 @@ import org.scalatest.FunSuite
...
+        ),
+        """------------------------- MODULE counterexample -------------------------
         |
         |EXTENDS test
         |
-        |(* Initial state *)
+        |(* Constant initialization state *)
         |
-        |State1 ==
+        |StateConst ==
```
konnov, line 38: It's a strange test, as `x` is not a constant here but a variable.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/io/TestCounterexampleWriter.scala
```diff
@@ -83,37 +86,38 @@ class TestCounterexampleWriter extends FunSuite {
...
```
konnov, line 108: again, the constant initializer initializes a variable here, instead of a constant

konnov (commented):
Good start. Still some things to fix.

konnov:
> If you are working on this anyway, I would like to point to something that bothered me about Apalache counterexamples a bit. If an invariant contains quantifiers, then InvariantViolation will contain skolem constants with $, which doesn't fit the TLA+ syntax and prevents the counterexample from loading in the TLA+ Toolbox for example.

That's a good point. Shall we simply replace `$` with a special sequence, e.g., `_AP_`? Of course, this can lead to name collisions but this is very unlikely. We can do this substitution either in `PrettyWriter` or in the renamer. Perhaps, it is safer to do it in `PrettyWriter` to minimize the damage in the case of a collision.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
Kukovec: I noticed that it was using both, but I assumed you must have had some reason to do it this way at the time, so I just extended the extra case in the same way.
As for your `.toDoc` method suggestion, I'm not exactly sure what you're trying to do, could you elaborate more?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
Kukovec: I copied the name from your comment in #265, but that's an easy fix.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
Kukovec: Sure

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
Kukovec: Hmm, I didn't modify `TlcCounterexampleWriter` (beyond reformatting), since it wasn't mentioned in the issue, but you're probably right. There are actually quite a few `i +1`s peppered throughout the file, now that I looked at it.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/io/TestCounterexampleWriter.scala
Kukovec: That's because your test is written in a way that doesn't reflect real-case use. The 1st element of `states`, in the test, is the `x=1` state, wehreas in the real tool case it would be `TRUE` or whatever `CInit` does.

Kukovec:
> 
> 
> > If you are working on this anyway, I would like to point to something that bothered me about Apalache counterexamples a bit. If an invariant contains quantifiers, then InvariantViolation will contain skolem constants with $, which doesn't fit the TLA+ syntax and prevents the counterexample from loading in the TLA+ Toolbox for example.
> 
> That's a good point. Shall we simply replace `$` with a special sequence, e.g., `_AP_`? Of course, this can lead to name collisions but this is very unlikely. We can do this substitution either in `PrettyWriter` or in the renamer. Perhaps, it is safer to do it in `PrettyWriter` to minimize the damage in the case of a collision.

Yes, but I wonder whether this should be its own PR. If we are going to name variables something that might cause a name collision, it would be better to define "reserved" name patterns and implement a pass, close to parsing, which warns about such names written by the user. After that, we can go ahead and re-insert these names in CE outputs.

andrey-kuprianov:
@konnov , @Kukovec is this going to be merged? Wrt. to `InvariantViolation` encoding I am fine with addressing this in a separate PR. 

I would like to do #530, but there is a dependency on this one.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
konnov: I suggested to move the `EXTENDS...` code into `PrettyWriter`. That means that you have to extend the method `toDoc` for `TlaModule`.

konnov:
Let's introduce a separate issue for the problem with `$`. As soon as Jure fixes the few small issues, we can merge.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -52,59 +52,69 @@ object CounterexampleWriter {
...
+    pretty.writeln()
+    pretty.prettyWriteDoc(pretty.moduleExtendsDoc(rootModule.name))
+    pretty.writeln()
 
     states.zipWithIndex.foreach {
       case (state, 0) =>
-        writer.println("(* Constant initialization state *)")
-        writer.println(s"\nStateConst ==")
+        pretty.writeln("(* Constant initialization state *)")
+        pretty.writeln(s"\nConstInit ==")
```
konnov, line 79: Why don't use use `write(decl: TlaOperDecl)`? It would do all pretty printing for you.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -52,59 +52,69 @@ object CounterexampleWriter {
...
+    pretty.prettyWriteDoc(pretty.moduleNameDoc("counterexample", 25))
+    pretty.writeln()
+    pretty.prettyWriteDoc(pretty.moduleExtendsDoc(rootModule.name))
+    pretty.writeln()
 
     states.zipWithIndex.foreach {
       case (state, 0) =>
-        writer.println("(* Constant initialization state *)")
-        writer.println(s"\nStateConst ==")
+        pretty.writeln("(* Constant initialization state *)")
```
konnov, line 78: I would object against introducing `writeln` in the pretty printer. Can you introduce a specialized method for writing comments instead? It could do some special processing of comments, like wrapping and indentation.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
```diff
@@ -52,59 +52,69 @@ object CounterexampleWriter {
...
-
   override def write(rootModule: TlaModule, notInvariant: NotInvariant, states: List[NextState]): Unit = {
-    writer.print(s"""@!@!@STARTMSG 2262:0 @!@!@
+    // `states` must always contain at least 1 state: the constant initialization
+    // This makes `states.tail` safe, since we have a nonempty sequence
+    assert(states.nonEmpty)
+
+    val pretty = new PrettyWriter(writer)
+
+    pretty.write(s"""@!@!@STARTMSG 2262:0 @!@!@
```
konnov, line 117: In case of TLC, I would rather use a `PrintWriter`, not `PrettyWriter`. The TLC format does not have to be pretty printed.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
```diff
@@ -32,24 +33,38 @@ class PrettyWriter(writer: PrintWriter, textWidth: Int = 80, indent: Int = 2) ex
...
 
-  def write(mod: TlaModule): Unit = {
-    writer.write(pretty(toDoc(mod), textWidth).layout)
-  }
+  def prettyWriteDoc(doc: Doc): Unit = writer.write(pretty(doc, textWidth).layout)
 
-  def write(decl: TlaOperDecl): Unit = {
-    writer.write(pretty(toDoc(decl), textWidth).layout)
-  }
+  def write(string: String): Unit = prettyWriteDoc(text(string))
```
konnov, line 38: This is a way to mess with the pretty printer.

konnov (commented):
I think we are converging. I have added a few comments on the use of pretty writer.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
Kukovec: Not the comments, but mostly because this was a batch 'find-replace'. Was there a reason this originally wasn't printed as a decl?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
Kukovec: Can you be more specific with exactly what kind of "special processing" you want?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
Kukovec: In the original code, TLC writer was already using prettyprinting, in `printStateFormula(new PrettyWriter(writer), state._2)`

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
Kukovec: Perhaps, but also the only way to force comments in at the desired line, since comments are invisible in the internal representation.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
konnov: I just want to have a method `writeComment(text: String)` that wraps `text` into `(*` and `*)`. In the future, we may add smarter code. My main objection is that if you add `writeln`, it is a recepy for pushing unformatted text in the middle of the pretty printed text.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
konnov: I have no idea. Maybe because somebody wanted to write code quick %)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
konnov: I believe that comments deserve a method of their own. No big deal if they are printed on a new line. At least, they will not ruin pretty printing.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/CounterexampleWriter.scala
konnov: Then let's not use `PrettyWriter` for TLC counterexamples. TLC counterexamples obviously enforce their own formatting, which we have to beatify later.

Kukovec: Just to copy our Slack discussion here:
We cannot get rid of a `PrettyWriter` call here, because `PrintWriter` does not know how to display TLA+ code.
Possible TODO in the future, but not as part of this PR.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/io/TestCounterexampleWriter.scala
```diff
@@ -118,30 +104,18 @@ class TestCounterexampleWriter extends FunSuite {
...
-        |/\ x = 1
-        |/\ y = 9
+        |State1 == x = 1 /\ y = 9
         |
         |(* Transition Trans2 to State2 *)
-        |
-        |State2 ==
-        |/\ x = 2
-        |/\ y = 10
+        |State2 == x = 2 /\ y = 10
```
konnov: actually, I hope that the Toolbox parses that! The leading `/\` might be from the TLC counterexample format.

Kukovec: `State2 == x = 2 /\ y = 10` is a legal TLA+ formula, so it should be fine. Leading `/\` is just for multiline parsing. If, however, you're referring to the examples in the `TLCCounterexampleWriter`, I checked it before, only in cases with 2 or more variables is there a leading `/\`

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
```diff
@@ -32,24 +33,42 @@ class PrettyWriter(writer: PrintWriter, textWidth: Int = 80, indent: Int = 2) ex
...
-  def write(mod: TlaModule): Unit = {
-    writer.write(pretty(toDoc(mod), textWidth).layout)
-  }
+  def prettyWriteDoc(doc: Doc): Unit = writer.write(pretty(doc, textWidth).layout)
 
-  def write(decl: TlaOperDecl): Unit = {
-    writer.write(pretty(toDoc(decl), textWidth).layout)
-  }
+  def blockComment(commentStr: String): Doc = text(s"(* $commentStr *)") <> line
+  def lineComment(commentStr: String): Doc = text(s"\\* $commentStr") <> line
```
konnov, line 39: Why do we have two methods? If you need both, introduce an option on how to output it. But I would just always write comments in one style, e.g., as block comments.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
```diff
@@ -32,24 +33,42 @@ class PrettyWriter(writer: PrintWriter, textWidth: Int = 80, indent: Int = 2) ex
...
+  def moduleNameDoc(name: String, nDashes: Int = 5): Doc =
+    s"${List.fill(nDashes)("-").mkString} MODULE $name ${List.fill(nDashes)("-").mkString}" <> line <> line
+
+  def moduleTerminalDoc(nEquals: Int = 15): Doc = {
+    s"${List.fill(nEquals)("=").mkString}" <> line
   }
 
-  def toDoc(mod: TlaModule): Doc = {
-    s"----- MODULE ${mod.name} -----" <> line <>
+  def moduleExtendsDoc(moduleNamesString: String): Doc =
```
konnov, line 58: why are you passing a string here? It is the job of the writer to make a string. I would pass `List[String]`.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
```diff
@@ -32,24 +33,42 @@ class PrettyWriter(writer: PrintWriter, textWidth: Int = 80, indent: Int = 2) ex
...
+  def write(mod: TlaModule): Unit = prettyWriteDoc(toDoc(mod))
+
+  // Declarations have a trailing empty line
+  def write(decl: TlaOperDecl): Unit = prettyWriteDoc(toDoc(decl) <> line <> line)
 
-  def write(expr: TlaEx): Unit = {
-    writer.write(pretty(toDoc((0, 0), expr), textWidth).layout)
+  def write(expr: TlaEx): Unit = prettyWriteDoc(toDoc((0, 0), expr) <> line)
+
+  def moduleNameDoc(name: String, nDashes: Int = 5): Doc =
```
konnov, line 51: Why do you have to pass `nDashes` and who should come up with that? Why don't you compute `nDashes` from `textWidth` and `name.length()`?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
```diff
@@ -32,24 +33,42 @@ class PrettyWriter(writer: PrintWriter, textWidth: Int = 80, indent: Int = 2) ex
...
+    s"${List.fill(nDashes)("-").mkString} MODULE $name ${List.fill(nDashes)("-").mkString}" <> line <> line
+
+  def moduleTerminalDoc(nEquals: Int = 15): Doc = {
```
konnov, line 54: Why do you need `nEquals`? You could just use `textWidth`.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
```diff
@@ -32,24 +33,42 @@ class PrettyWriter(writer: PrintWriter, textWidth: Int = 80, indent: Int = 2) ex
...
+    s"${List.fill(nEquals)("=").mkString}" <> line
   }
 
-  def toDoc(mod: TlaModule): Doc = {
-    s"----- MODULE ${mod.name} -----" <> line <>
+  def moduleExtendsDoc(moduleNamesString: String): Doc =
+    s"EXTENDS $moduleNamesString" <> line <> line
+
+  def toDoc(mod: TlaModule, nDashes: Int = 5, nEquals: Int = 15): Doc = {
```
konnov, line 61: Again, no reason for `nDashes` and `nEquals`.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
```diff
@@ -32,24 +33,42 @@ class PrettyWriter(writer: PrintWriter, textWidth: Int = 80, indent: Int = 2) ex
   // the stack of lambda declarations
   private var lambdaStack: List[TlaOperDecl] = Nil
 
-  def write(mod: TlaModule): Unit = {
-    writer.write(pretty(toDoc(mod), textWidth).layout)
-  }
+  def prettyWriteDoc(doc: Doc): Unit = writer.write(pretty(doc, textWidth).layout)
```
konnov, line 36: The user of the `writeXXX` methods does not have to know about `Doc`, as you would see in the interface `TlaWriter`. Please revert this.

konnov (commented):
More comments on the new code. I am picky here, because pretty writers are hard to write.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
Kukovec: The short answer to this, plus all of the questions about `nDashes` and `nEquals` is that a lot of the tests compare output against hard-coded strings (file contents). It was either this, or rewriting all of the test files' formatting.

konnov: I don't think it is a good approach to create technical debt because of the bad tests ;-)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/PrettyWriter.scala
Kukovec: Not sure I understand your concern here, before, all three `write` methods were implemented as `
writer.write(pretty( [X] , textWidth).layout)`, with varying `X`. All I did here was implement this as a separate method that takes `X` as the parameter.

konnov: Check this new interface: https://github.com/informalsystems/apalache/blob/3dfa7a77e69762b0771a98d2ae591be9eab0be9d/tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/TlaWriter.scala

Kukovec: Yes, all of these methods still exist and are unit

Kukovec: Would you want `prettyWriteDoc` to be private?

konnov: yes, or at least protected

konnov (approved):
Looks good. Let's merge!

PR #562, opened 2021-02-08
Simplify some Python examples

Alexander-N:
Following the discussion in https://github.com/informalsystems/apalache/discussions/551 this is an example of how I would image the Python examples could be simplified. A lot of this is a matter of preference, but I did it like this:
- Used frozenset only where necessary, and with a comment, so that it is clear that this has nothing to do with TLA+.
- Used `range(1, 4)` instead of `{1, 2, 3}` since I found it closer in meaning to `1..3` as it was used in these examples.
- Tried to make the code look similar in general.
- Codestyle is [black](https://github.com/psf/black).

The format of the examples and the Python results is not 100% consistent,  the [doctest format](https://docs.python.org/3/library/doctest.html) could be used, but might not be necessary.


konnov:
@Alexander-N, thank you for the PR! Let me think a bit about the proposed changes.

Kukovec:
I have to say that I very much disagree with this line of thought. There are problems with both of your propositions: 
a) "use frozenset only where necessary" and 
b) "range(a, b+1) instead of {a,..., b} for a..b"

Both essentially boil down to the same thing: consistency. The idea here is to present, at least in the reader's mind, a mapping of the sort `T: TLA+ -> Python` (taking some liberties with function notation), that produces functional code. The problem with your idea (a) is that `T({1,2}) = {1,2}`, while `T({{1},{2}}) != {{1},{2}}`, but instead `{frozenset({1}), frozenset({2})} ` (or is it `frozenset({frozenset({1}), frozenset({2})})`? it's not clear).
The universal solution is to have a rule `T({A1, ..., An}) = frozenset({ T(A1), ..., T(An) })`, which works the same way for arbitrary set-nesting depth (i.e. when `A1, ..., An` are sets too), since it is always safe to use a frozenset instead of a mutable set in Python, but not the other way around.

As for (b), the translation might look nicer in a cherry-picked iteration example, but you mustn't forget that, in TLA+, `a..b` is a set, and therefore permits set operations, like for example `1..N \cup {3,4}`. In your approach, you cannot claim that 
`T(A \cup B) = T(A).intersection(T(B))`, because you have `T(1..3) = range(1,4)`, which does not admit set operations. 
However, if `T(a..b) = frozenset({ i for i in range(a,b+1) })`, the translation is compatible with all set operators.

The fact that some of these translations produce verbose code in Python, due to mutability concerns, is perhaps unfortunate, but at the end of the day, consistent translation rules are more important than minimality in the examples.

docs/src/lang/functions.md
```diff
@@ -231,15 +228,24 @@ that are computed under such a binding.
...
+xy = {(x, y) for x in range(1, 4) for y in range(4, 7)}
+{(x, y): x + y for (x, y) in xy}
+# TLA: [ n \in 1..3 |->
+#        [ i \in 1..n |-> n + i ]]
+# We have to use frozenset to use a set as key in a dict.
+{
+    n: {frozenset(range(1, i + 1)): n + 1}
+    for n in range(1, 4)
+    for i in range(1, n + 1)
+}
```
konnov, line 248: Are you sure that the python code reflects the TLA+ expression? From what I see in python repl, the result of evaluating the python expression is:

```
{1: {frozenset({1}): 2}, 2: {frozenset({1, 2}): 3}, 3: {frozenset({1, 2, 3}): 4}}
```

The TLA code is constructing a function of `n \in 1..3` that returns a function of `i \in 1..n` to `n+i`.

konnov:
@Alexander-N, I have made a pass over `functions.md`, added a discussion on python dictionaries and massaged the sections on python examples. How do you like it? I believe that in this context your examples are quite good.

Alexander-N (commented):
It's probably become clear already, but to expand my perspective a little, I think the Python examples are useful for engineers in that they put a probably strange and unfamiliar syntax next to run-of-the-mill Python code. This approach is also used in this [cheatsheet](https://github.com/parlarjb/tla_workshop/blob/master/tla_cheatsheet.pdf) and Hillel Wayne "pythonized" TLA+ in some [talks](https://hillelwayne.com/talks/distributed-systems-tlaplus/). On the other hand `frozenset`s without justification could seem even stranger than the TLA+ code :-).

docs/src/lang/functions.md
Alexander-N: Right! So it would be

```python
{
    n: {i: n + i for i in range(1, n + 1)}
    for n in range(1, 4)
}
```

That's nice because in this case there is no need for any `frozenset` or `frozendict` in this document.

docs/src/lang/functions.md
```diff
@@ -156,13 +158,83 @@ In this sense, the type restrictions of Apalache are similar to those for the
...
+**TLA+ functions and Python dictionaries**. As we mentioned before, TLA+
+functions are similar to maps and dictionaries in programming languages. To
+demonstrate this similarity, let us compare TLA+ functions with [Python
+dictionaries][].  Consider a TLA+ function `price` that is defined as follows:
+
+```tla
+  [ meal \in { "Schnitzel", "Gulash", "Cordon bleu" } |->
+                CASE meal = "Schnitzel" -> 18
+                  [] meal = "Gulash"    -> 11
+                  OTHER                 -> 12
```
shonfeder, line 170: I think the inclusion of `OTHER` here complicates the comparison. So I suggest:

```suggestion
                  "Cordon bleu"              -> 12
```

iiuc, this is also more faithful to the python encoding.

shonfeder (approved):
I think the added context here and incorporating this discussion into a dialectical understanding of how we can and cant' represent TLA+ in Python into the manual is a boon. So I'm adding my approval so that the change can be merged in whenever all other parties are happy with the state of things.

docs/src/lang/functions.md
konnov: fixed

docs/src/lang/functions.md
```diff
@@ -156,6 +158,79 @@ In this sense, the type restrictions of Apalache are similar to those for the
...
+```python
+py_price = { "Schnitzel": 18, "Gulash": 11, "Cordon bleu": 12 }
+```
+
+As long as we are using the variable `py_price` to access the dictionary, our
+approach works. For instance, we can type the following in the python shell:
+
+```python
+>>> # similar to DOMAIN price in TLA+
+>>> frozenset(py_price.keys())
```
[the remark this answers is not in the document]
konnov: Good point. I have added an explanation about frozen sets. It is up to the reader, how much in detail they would like to go.

docs/src/lang/functions.md
```diff
@@ -156,6 +158,79 @@ In this sense, the type restrictions of Apalache are similar to those for the
...
+```tla
+# TLA+: wrapping a function with a set
+{ price }
+```
+
+Unfortunately, this does not work as easy in Python:
+
+```python
+>>> # python expects hashable and immutable data structures inside sets
+>>> frozenset({py_price})
```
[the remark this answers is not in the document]
konnov: That's exactly the problem of mismatch between python sets and TLA sets:

> TypeError: unhashable type: 'dict'

konnov:
@Alexander-N, I have made another pass and incorporated your comments. Sorry for the delay with your PR. If you have other suggestions, let's discuss them in a separate PR.

PR #567, opened 2021-02-11
Generators for IR and tests for PrimePropagation and IncrementalRenaming

konnov:
This PR introduces Scalacheck generators for our intermediate representation (`TlaEx` and `TlaDecl`). It also adds two examples of using Scalacheck for testing transformations: `TestPrimePropagation` and `TestIncrementalRenaming`.

In the current form the generators may produce crazy expressions that could be probably rejected by the parser. I find this to be a good feature: We should make our transformations bullet-proof. That is, if a transformation finds an unexpected TLA structure, then it should throw `MalformedTlaError` instead of crashing somewhere. We have seen that our transformations need some form of preconditions, as otherwise they produce unexpected results when the order is changed.

In the future, we might also want to introduce pre and postconditions for the transformations that could be checked with Scalacheck. That would help us to understand, whether transformations compose.


shonfeder (approved):
I like it! 

LGTM

PR #577, opened 2021-02-13
Removed old, unused code

### LINKED ISSUE #572: [FEATURE] Rewrite TestIDAllocation and remove TlaSpec
The tests in `TestIDAllocation` were created in 2016. Instead of using `assert`, they are using print statements. Let's fix this zombie test.

Also, use `TlaModule` instead of `TlaSpec`. When it's done, remove `TlaSpec`.

Kukovec:
closes #572


konnov (approved):
Looks good

Issue #576, opened 2021-02-12
[BUG] Name clash in local instance


konnov:
The following example provokes an exception:

```tla
------------------ MODULE localTwice -----------------
------------------ MODULE inst -----------------------
Foo(x) ==
    LET msg == x IN
    msg
======================================================

LOCAL INSTANCE inst

Bar(x) ==
  LET msg == 3 IN
  Foo(x)

======================================================
```

The exception is as follows:

```
Unhandled exception                                               E@21:44:29.698
java.lang.IllegalStateException: Found two different declarations with the same name msg: DeclUnit(TlaOperDecl(msg,List(),3)) and DeclUnit(TlaOperDecl(msg,List(),x))
	at at.forsyte.apalache.tla.imp.Context$ContextImpl$$anonfun$push$1.applyOrElse(Context.scala:134)
	at at.forsyte.apalache.tla.imp.Context$ContextImpl$$anonfun$push$1.applyOrElse(Context.scala:131)
	at scala.PartialFunction$Lifted.apply(PartialFunction.scala:228)
	at scala.PartialFunction$Lifted.apply(PartialFunction.scala:224)
	at scala.Option.collect(Option.scala:432)
	at at.forsyte.apalache.tla.imp.Context$ContextImpl.push(Context.scala:131)
	at at.forsyte.apalache.tla.imp.ExprOrOpArgNodeTranslator.$anonfun$translateLetIn$1(ExprOrOpArgNodeTranslator.scala:113)
	at at.forsyte.apalache.tla.imp.ExprOrOpArgNodeTranslator.$anonfun$translateLetIn$1$adapted(ExprOrOpArgNodeTranslator.scala:107)
	at scala.collection.immutable.List.foreach(List.scala:431)
```

konnov:
This one works now

Issue #574, opened 2021-02-12
[FEATURE] Rename variables with `$` when expoting modules to files


Kukovec:
Internally, Apalache variable names contain `$` characters, due to renaming. Consequently, exporting internal representation as a `.tla` file produces files, which are not parseable, since variable names in TLA+  may not contain these characters (see #555).

We should solve this in the following way:
1. Introduce reserved variable names, e.g. those starting with an `APA_` prefix.
2. Implement a language watchdog pass immediately after parsing, which rejects specifications in which users introduce variable names belonging to the Apalache-reserved namespace (with a user-friendly warning)
3. Whenever variables are written to a file, they are stripped of illegal characters and given the Apalache-reserved prefix, to avoid the possibility of name collisions.

PR #581, opened 2021-02-15
Close 576

konnov:
This PR closes bug #576. I hope that this is the final fix in our long battle against `LOCAL` operators and `LOCAL INSTANCE`.


konnov:
I have accidentally created this branch starting from `igor/funasseq-223`. Check only the last three commits. The branch `igor/funasseq-223` has been merged already.

tla-import/src/main/scala/at/forsyte/apalache/tla/imp/Context.scala
```diff
@@ -100,7 +133,7 @@ object Context {
    * Create a new context, i.e., use Context().
    */
   def apply(): Context = {
-    new ContextImpl(List(), List(), Map())
+    new ContextImpl(List(), false, List(), Map())
```
Kukovec, line 136: Is there a way, or need, to ever pass `true` here w/o manually constructing `ContextImpl`? Not sure I like mutating after construction, if it's not necessary.
In general, it would be nice to have a line comment, explaining the default value.

Kukovec (approved):
I have one minor comment, looks fine aside from that. As a side remark, this issue seems to be related to #574, as both touch upon the issue of namespace collisions (in slightly different ways). There might be some unifying solution to both of these problems.

tla-import/src/main/scala/at/forsyte/apalache/tla/imp/Context.scala
konnov: Well, yes, they are related. However, I am trying to keep the imported IR close to the source. In case of LOCAL, we have no other choice but to introduce unique namespaces.

Issue #598, opened 2021-02-22
[FEATURE] New Json format


Kukovec:
Apalache's current Json output format is a) not well documented, b) fairly ad hoc, and c) incomplete (see #595).
We should implement a more standardized json output format, which mirrors the internal representation.

konnov:
The idea is to:

- use a standard JSON serializers of our IR, e.g., https://circe.github.io/circe/codec.html
- By documenting our IR once, we can simply map it to different formats, e.g., JSON, XML, S-expressions, without the need for documenting the format multiple times.

shonfeder:
As per discussion in slack, @andrey-kuprianov is OK with this change, and doesn't require any backward compatibility support, even for a limited time. He will just pin to the version they need until they have time to update their handling of the JSON format.

PR #599, opened 2021-02-22
New json format

Kukovec:
First step towards #598


konnov (commented):
This is definitely an improvement over the current JSON writer. I have several suggestions:

1. Can we use the type class pattern? See Chapter 1 in [Scala with Cats](https://underscore.io/books/scala-with-cats/). They give conversion to JSON as an example! We even don't have to use cats to follow this pattern. Then you would not have to use all those factories.
2. Can we convert integers, Booleans, and strings into the native JSON format? Having `{"type": "ValEx", {"Value": {"type": TlaInt, "value": 2}}}` is quite cumbersome.
3. Let's not use the name `"type"` for the expression case. Call it `"kind"`?
4. You could also output the `type` field of `TlaEx`. For the moment, it will be `untyped`, but very soon we will be able to output the expression types.

konnov:
What are "otherArgs" supposed to be and why do we have them at all?

konnov:
How about having a JSON representation of [S-expressions](https://en.wikipedia.org/wiki/S-expression)? This basically means:

1. Literals are just literals, e.g., `15`, `"Hello world"`, and `true`.
2. Every operator application is represented as an array, e.g., `OperEx(TlaArithOper.plus, ValEx(TlaInt(11)), ValEx(TlaInt22))` would be encoded in JSON as `["+", 11, 22 ]`. However, I don't know how to sneak the type tags here. Maybe we should not.
3. The few special sets such as `BOOLEAN`, `STRING`, `Int`, `Nat`, and `Real` can be represented by using the operator notation: `["BOOLEAN"]`, `["STRING"]`, etc. It was our decision to call them values. In SANY, they are actually operators.
4. Operator definitions could be represented as objects, e.g., `{ "operator": "Bar", "params": [ "x", ["Foo", 2] ], "body": ..., "recursive": false }`.  Here, `x` is a simple parameter, whereas `Foo` is an operator parameter of two arguments, that is, `Foo(_, _)`. Similarly, a variable definition could be `{ "variable": "velocity" }`

Let's discuss this alternative. I would like to have a format that is machine readable, but also not totally verbose like XML. Obviously, we should not introduce special syntax for every possible operator. Otherwise, we would have the same problem as the current JSON writer.

konnov:
@shonfeder, @andrey-kuprianov, do you like to comment on this discussion?

Kukovec:
> 
> 
> This is definitely an improvement over the current JSON writer. I have several suggestions:
> 
>     1. Can we use the type class pattern? See Chapter 1 in [Scala with Cats](https://underscore.io/books/scala-with-cats/). They give conversion to JSON as an example! We even don't have to use cats to follow this pattern. Then you would not have to use all those factories.
> 
>     2. Can we convert integers, Booleans, and strings into the native JSON format? Having `{"type": "ValEx", {"Value": {"type": TlaInt, "value": 2}}}` is quite cumbersome.
> 
>     3. Let's not use the name `"type"` for the expression case. Call it `"kind"`?
> 
>     4. You could also output the `type` field of `TlaEx`. For the moment, it will be `untyped`, but very soon we will be able to output the expression types.

(1) It's unclear to me what kind of typeclass you want to see. `Json`, as defined in that book is essentially `JsonRepresentation`. As for the factories (and parameterization), they serve a different purpose than the parameterization of `JsonWriter[A]` in the example. There, a json writer is parameterized by what type gets written to json. In our case, that's fixed: we always (currently) convert TLA+ (IR) to json. The parameterization, and the use of factories, in our case comes from the different external libraries that implement json handling. I currently use `ujson`, but `circe` is also mentioned in the issue as an option. Then, if you wanted to use `circe` instead of `ujson`, all you had to do was write a factory, and extend `TlaToJson` like `TlaToUjson` does.

(2) I think there is value in consistency. The current implementation creates a json that exactly matches the IR syntax tree, including wrapping integers in a `ValEx` and `TlaInt` layer. The actual value field is still a json-native integer. This approach also future-proofs implementing, e.g. model values, where you could have `{"type": "ValEx", {"value": {"type": TlaModelVal, "value": X}}}`. In addition, json just has "numbers", so integers and decimals would have the same representation. Lastly, while maybe more verbose, this representation makes is trivial to do reconstruction (json to TLA-IR). Verbosity doesn't even matter, at the end of the day, since json is for machines, not humans.

(3) Sure, I anticipated such comments and made the field name a parameter for exactly this reason.

(4) Also doable. 


> 
> 
> What are "otherArgs" supposed to be and why do we have them at all?

This is just a reference to the fact that some constructors have individually named arguments (e.g. `TlaOperDecl` has `body`), and some take an arbitrary list of arguments (e.g. `OperEx` has `oper` named and `args: TlaEx*`, a sequence of "other arguments"). Perhaps the description could be improved.



> 
> 
> How about having a JSON representation of [S-expressions](https://en.wikipedia.org/wiki/S-expression)? This basically means:
> 
>     1. Literals are just literals, e.g., `15`, `"Hello world"`, and `true`.
> 
>     2. Every operator application is represented as an array, e.g., `OperEx(TlaArithOper.plus, ValEx(TlaInt(11)), ValEx(TlaInt22))` would be encoded in JSON as `["+", 11, 22 ]`. However, I don't know how to sneak the type tags here. Maybe we should not.
> 
>     3. The few special sets such as `BOOLEAN`, `STRING`, `Int`, `Nat`, and `Real` can be represented by using the operator notation: `["BOOLEAN"]`, `["STRING"]`, etc. It was our decision to call them values. In SANY, they are actually operators.
> 
>     4. Operator definitions could be represented as objects, e.g., `{ "operator": "Bar", "params": [ "x", ["Foo", 2] ], "body": ..., "recursive": false }`.  Here, `x` is a simple parameter, whereas `Foo` is an operator parameter of two arguments, that is, `Foo(_, _)`. Similarly, a variable definition could be `{ "variable": "velocity" }`
> 
> 
> Let's discuss this alternative. I would like to have a format that is machine readable, but also not totally verbose like XML. Obviously, we should not introduce special syntax for every possible operator. Otherwise, we would have the same problem as the current JSON writer.

Honestly, I don't see what the benefit would be. Your encoding also completely erases the distinction between `NameEx(...)` and `ValEx(TlaString(...))`. What is `2 + x`? If it's `["+", 2, "x"]`, what is `2 + "x"`? Ultimately, verbosity is not an issue for machine-readable formats, especially not given the size of TLA+ specifications, the json files would still be tiny (for a machine to read).
The current implementation is both unambiguous and very intuitive for anyone who understands the IR.

shonfeder:
I'm inclined to agree with @Kukovec.  An unambiguous and easy to maintain serialization seems more valuable than catering to human readability in special cases. Humans can figure out this representation easily enough to code with it reliably, and imo humans shouldn't be looking at data serialization output except for to code consumers. And for the latter purpose, a simple and consistent format is much more valuable than certain special cases being easy to read.

I think we might take a queue from Pandoc, for which it is quite easy to write filters that work on the JSON encoding of its AST (confirmed by [the number of filters in different languages](https://github.com/search?p=1&q=pandoc+filters&type=Repositories)).

Here's a small example:

```json
sf@system76-pc:~/scratch$ echo "# Heading
> 
> *Foo* and \`bar\`." | pandoc -t json | jq
{
  "blocks": [
    {
      "t": "Header",
      "c": [
        1,
        [
          "heading",
          [],
          []
        ],
        [
          {
            "t": "Str",
            "c": "Heading"
          }
        ]
      ]
    },
    {
      "t": "Para",
      "c": [
        {
          "t": "Emph",
          "c": [
            {
              "t": "Str",
              "c": "Foo"
            }
          ]
        },
        {
          "t": "Space"
        },
        {
          "t": "Str",
          "c": "and"
        },
        {
          "t": "Space"
        },
        {
          "t": "Code",
          "c": [
            [
              "",
              [],
              []
            ],
            "bar"
          ]
        },
        {
          "t": "Str",
          "c": "."
        }
      ]
    }
  ],
  "pandoc-api-version": [
    1,
    20
  ],
  "meta": {}
}
```

This is a quite verbose, but easy to predict encoding of the haskell AST:

```haskell
sf@system76-pc:~/scratch$ echo "# Heading

*Foo* and \`bar\`." | pandoc -t native
[Header 1 ("heading",[],[]) [Str "Heading"]
,Para [Emph [Str "Foo"],Space,Str "and",Space,Code ("",[],[]) "bar",Str "."]]
```

And, modulo the names of the keys, I think this is approximately the approach taken in this PR currently.

tl;dr I'm pro :)

konnov:
I see your point @shonfeder and @Kukovec. IMO this one is somewhat in contradiction with #530. Perhaps, it is not a good idea to solve two problems at once.

I guess, if we want to pursue readability, we can introduce yet another format that is essentially S-expressions. JSON is almost like that but it also has unnecessary syntax.

@vitorenesduarte and @andrey-kuprianov, the json syntax is verbose, but it should be quite easy to parse. Why would you like to have a more concise syntax for automatic tooling?

konnov:
@Kukovec, can you introduce an ADR like this: https://apalache.informal.systems/docs/adr/004adr-annotations.html

It is hard to extract the logic from the scala docs and tests.

konnov:
> > What are "otherArgs" supposed to be and why do we have them at all?
> 
> This is just a reference to the fact that some constructors have individually named arguments (e.g. TlaOperDecl has body), and some take an arbitrary list of arguments (e.g. OperEx has oper named and args: TlaEx*, a sequence of "other arguments"). Perhaps the description could be improved.

Let's just use positioned arguments wherever possible, like we do in the IR. If we have clear documentation, positioned arguments will do the job. If find `otherArgs` quite confusing and not self-explanatory.

andrey-kuprianov:
> @vitorenesduarte and @andrey-kuprianov, the json syntax is verbose, but it should be quite easy to parse. Why would you like to have a more concise syntax for automatic tooling?

@konnov why do you say we want a more concise syntax? We don't! I completely agree the JSON format should be as straightforward as possible to parse, and not care at all about human readability.

konnov:
> @konnov why do you say we want a more concise syntax?

From #530, I had an impression that you like to have a JSON format for the counterexamples that maps well on JSON concepts. The mapping proposed in this PR is aiming to minimize the number of parsing rules rather than using a JSON-native representation.

vitorenesduarte:
> > @konnov why do you say we want a more concise syntax?
> 
> From #530, I had an impression that you like to have a JSON format for the counterexamples that maps well on JSON concepts. The mapping proposed in this PR is aiming to minimize the number of parsing rules rather than using a JSON-native representation.

One idea that came up is to instead generate that new JSON format for counterexamples by parsing the `counterexample.tla`. This should be feasible as we're already doing something like that for TLC.

andrey-kuprianov:
Let's distinguish between two formats:
 1. The general format for TLA+ syntax trees. This PR is about that format
 2. The format for counterexample. This is a _completely different_ one that's used in https://github.com/tlaplus/CommunityModules/blob/master/modules/Json.tla. And there is a particular reason for that format to be simpler, because the data it represents is much simpler (ground formulas)

And, as Vitor correctly pointed out, we don't even need to implement generation of this second format in Apalache, we can reuse the translator above.

konnov:
@Kukovec, I think it is all clear now. Ignore my comments on simpler JSON. But please have a look at 3-4, `otherArgs`, and the ADR.

Kukovec:
The latest commit is missing the ADR, since that takes a bit longer to write. Note (to self): don't merge until the ADR is finished too.

konnov:
If you like to split the work, as we discussed earlier, we can merge this PR first and then continue with an ADR

tla-import/src/main/scala/at/forsyte/apalache/io/json/TlaToJson.scala
```diff
@@ -2,138 +2,154 @@ package at.forsyte.apalache.io.json
...
 import at.forsyte.apalache.tla.lir.values.{TlaBool, TlaDecimal, TlaInt, TlaStr}
 import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.io.TypeTagPrinter
 
 /**
  * A semi-abstraction of a json encoder.
  * It is independent of the concrete JsonRepresentation, resp. JsonFactory implementation.
  *
- * Every internal representation object, Class( arg1 = v1, ... ,argN = vN, otherArgs = args) gets encoded as json in the following way:
+ * Every internal representation object, Class( arg1 = v1, ... ,argN = vN, variadicArgs : T* = args) gets encoded as json in the following way:
```
konnov, line 11: I don't think that renaming `otherArgs` to `variadicArgs` helps a lot. Does it really matter at the json level that some args are variadic?

Kukovec:
> 
> 
> If you like to split the work, as we discussed earlier, we can merge this PR first and then continue with an ADR

Makes sense, lets do that, since the ADR will likely spawn its own discussion/comments, that have nothing to do with code.

konnov (approved):
Looks good. Though I am not certain that renaming `otherArgs` to `variadicArgs` helps

tla-import/src/main/scala/at/forsyte/apalache/io/json/TlaToJson.scala
Kukovec: Yes, named args get named fields directly, variadic args get a list.

konnov: why? I think it is a feature of `TlaBuilder`, not the feature of the IR

Kukovec: I mean, take `OperEx`, concretely. It always has an `oper` field, so the json also has an `oper` field. On the other hand, it takes 0 or more actual arguments (an `args: TlaEx*` field), depending on the operator, so the only way you can consistently JSON-serialize it is if you turn `args` into a single JSON field, that holds a list (elements of which are recursively serialized TlaEx into JSON).
What's the alternative?

konnov: Why cannot we write both `oper` and `args` in a single array that is called `args`?

Kukovec: Well, it's not so much that we can't, but I don't think we should. For several reasons, first, it diverges from the internal representation, second, it makes deserialization (which we aren't doing now, but probably want to in the future) harder and third, this is conceptually type-inconsistent (JSON isn't really typed, but our intuition is), as this array would mix objects and strings, which I think is bad form.
I'm going to merge the PR as it is, since it's been sitting for a while (and approved), if you feel strongly about this please open an issue.

PR #602, opened 2021-02-24
close #601: type signatures for the TLC module

### LINKED ISSUE #601: [FEATURE] Add type support for TLC!RandomElement
I find myself using the RandomElement operator from the TLC module in several specs. It would be great to have Snowcat support typechecking when I use the TLC module in a spec.

konnov:
This PR adds type signatures for the operators that are defined in the module [TLC](https://github.com/jameshfisher/tlaplus/blob/master/org.lamport.tla.toolbox/StandardModules/TLC.tla). Some of these signatures are pretty useless. For instance, `TLC!Any` has the type `() => a`, which does not help the type checker. Also, `TLC!Assert` does not have a well-defined type, so we label it with `(a, Str) => Bool`.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,3 +16,4 @@
 ### Bug fixes
 
 * Parser: propagating type annotations in INSTANCES, see #592 and #596
+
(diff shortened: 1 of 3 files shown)


shonfeder (approved):
LGTM

Issue #521, opened 2021-01-31
[FEATURE] Integrate the type checker Snowcat into the model checker


konnov:
The model checker should run Snowcat as a preprocessing pass and use the inferred types.

konnov:
The integration is in unstable now

PR #607, opened 2021-02-25
Ground work for #521

konnov:
More ground work towards integration of the new type checker:

1. `DeepCopy` is now copying types.
2. Add `TestDeepCopy` to test that `DeepCopy` does a complete copy, including types, but does not copy the identifiers.
3. `IrGenerators` are now generating tag types: either `Untyped()` or `Typed[Int](i)` for a random number `i`.
3. Move the test (yep, the only one!) for `InlinerOfUserOper` to its own class.
4. Add missing annotation `@RunWith(classOf[JUnitRunner])` to several tests. They were not running automatically!

### DIFF
--- a/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
+++ b/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
@@ -197,7 +197,8 @@ trait IrGenerators {
                   genLetInEx(genTlaEx(builtInOpers))(ctx), const(OperEx(oper, args: _*).withType(tt)))
             } else {
               // as above but no user-defined operators
-              oneOf(genValEx, genNameEx, const(OperEx(oper, args: _*).withType(tt)))
+              oneOf(genValEx, genNameEx, genLetInEx(genTlaEx(builtInOpers))(ctx),
+                  const(OperEx(oper, args: _*).withType(tt)))
             }
         } yield result
       }
--- a/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
+++ b/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
@@ -160,7 +160,9 @@ trait IrGenerators {
         ndefs <- choose(1, maxDefsPerLetIn)
         defs <- listOfN(ndefs, resize(size - 1, genTlaOperDecl(exGen)(ctx))) suchThat { ds =>
           // no new name is present in the context
-          ds.map(_.name).toSet.intersect(ctx.map(_.name).toSet).isEmpty
+          ds.map(_.name).toSet.intersect(ctx.map(_.name).toSet).isEmpty &&
+          // and all new names are mutually unique
+          ds.map(_.name).toSet.size == ds.size
         }
         body <- resize(size - 1, exGen(ctx ++ defs.map(d => UserOperSig(d.name, d.formalParams.length))))
         i <- arbitrary[Int]
--- a/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
+++ b/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
@@ -14,13 +14,16 @@ import org.scalacheck.Gen.{choose, const, identifier, listOf, listOfN, lzy, oneO
  * solutions: (1) By adding preconditions to the constructors of the IR operators, or (2) by throwing
  * `MalformedTlaError`.</p>
  *
- * <p>Limitations:</p>
+ * <p>Assumptions and limitations:</p>
  *
  * <ol>
  * <li>The current implementation of the generators works best for the code that is unaware of the semantics
  * of TLA+ operators.</li>
  *
  * <li>Our generators neither produce nor apply higher-order operators.</li>
+ *
+ * <li>The generators tag the produced expressions with either Untyped() or Typed[Int](i) for a random integer value,
+ * which should be sufficient for checking that the types are correctly propagated.</li>
  * </ol>
  *
  * @author Igor Konnov
@@ -32,6 +35,11 @@ trait IrGenerators {
 
   // an internal context of the generated operator definitions
   type UserContext = Seq[UserOperSig]
+
+  /**
+   * An empty user context.
+   */
+  val emptyContext: UserContext = Seq[UserOperSig]()
 
   /**
    * The maximal number of arguments in user-defined operators and built-in operators.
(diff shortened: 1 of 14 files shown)


shonfeder:
Thank you for staging these well-sized, reviewable change sets! <3 <3 <3

tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
```diff
@@ -85,7 +94,9 @@ trait IrGenerators {
    */
   def genBool: Gen[ValEx] = for {
     b <- arbitrary[Boolean]
-  } yield ValEx(TlaBool(b))
+    i <- arbitrary[Int]
+    tt <- oneOf(Untyped(), Typed[Int](i))
```
shonfeder, line 98: Should this maybe be factored out into an `arbitraryType` constructor or something?

tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
```diff
@@ -76,7 +84,8 @@ trait IrGenerators {
    */
   def genInt: Gen[ValEx] = for {
     i <- arbitrary[Int]
-  } yield ValEx(TlaInt(BigInt(i)))
+    tt <- oneOf(Untyped(), Typed[Int](i))
+  } yield ValEx(TlaInt(BigInt(i))).withType(tt)
```
shonfeder, line 88: I think we should have different arbitrary generators for the integer value and the type ID, right? Otherwise it seems like this would constraint the tests to a weird invariant.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestDeepCopy.scala
```diff
@@ -0,0 +1,84 @@
...
+        lop == rop &&
+        l.typeTag == r.typeTag &&
+        largs.length == rargs.length &&
+        largs.zip(rargs).forall(equalExCopies.tupled)
+
+    case (l @ LetInEx(lbody, ldefs @ _*), r @ LetInEx(rbody, rdefs @ _*)) =>
+      equalExCopies(lbody, rbody) &&
+        l.ID != r.ID &&
+        l.typeTag == r.typeTag &&
+        ldefs.length != rdefs.length &&
```
shonfeder, line 33: Its surprising to me that the declaration lengths should **not** be equal here. Could you explain?

shonfeder (changes requested):
Would like clarification on one point, and iiuc, the suggestion around arbitrary int generation is important.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
konnov: That makes sense

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestDeepCopy.scala
konnov: Oh yes, that's a bug, which was not discovered by Scalacheck :(

konnov: Well, Scalacheck does not hit it. Most likely, this expression is not getting generated. It would be great to have test coverage for Scalacheck tests.

shonfeder: Is it possible your generators are not generating `LetIn` exprs?

shonfeder: Seems to be the case?

https://github.com/informalsystems/apalache/blob/unstable/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala#L237-L249

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestDeepCopy.scala
```diff
@@ -0,0 +1,84 @@
...
+    case _ => false
+  }
+
+  test("deep copy replicates expressions and declarations but not their identifiers") {
+    val gens = new IrGenerators {
+      override val maxArgs: Int = 10
+    }
+
+    val prop = {
+      def exGen(ctx: gens.UserContext) = gens.genTlaEx(gens.simpleOperators ++ gens.setOperators)(ctx)
```
shonfeder, line 74: @konnov I think you are never hitting the `LetInEx` branch because these generator sets don't include letIn expressions: https://github.com/informalsystems/apalache/blob/unstable/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala#L57-L71

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestDeepCopy.scala
shonfeder: I'd say this function is complex enough that it should be probably be tested in its own right (on generators for any arbitrary tla expression, not just a subset).

I wonder if this predicate isn't also useful and general enough to be included in a library (not just the tests)? Seems there could be other cases where we want to test the "structural" equality of two terms without reference to their unique ids (and "full equality" could be defined just as the conjunction an id equality check and this structural equality check. That would also ensure we use the same logic for equality testing in the code base and in the tests.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestDeepCopy.scala
konnov: True. I have added `LetInEx`. Still this case is not covered.

konnov: OK. Now I have fixed the generator in the right place. The test is failing. However, we should introduce coverage of our tests.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestDeepCopy.scala
konnov: It probably could be refactored for future reuse in the tests. But I am not sure how to do it at the moment.

konnov:
@shonfeder could you have a look at this PR one more time?

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestDeepCopy.scala
shonfeder: Agree!

shonfeder (approved):

Issue #617, opened 2021-03-03
[BUG] Type checker is processing the chains of EXCEPTs incorrectly


konnov:
The new type checker does not account for the advanced syntax of `EXCEPT`, e.g.:

```tla
smokers' = [smokers EXCEPT ![r].smoking = FALSE]
```

This syntax is used in multiple specs, e.g., in `CigaretteSmokers.tla`.

PR #618, opened 2021-03-03
Recover types from the type checker

### LINKED ISSUE #608: [FEATURE] Tag expressions with the types that are computed by the type checker
The type checker reports about computed types by calling a listener. We should record the computed types and tag the expressions.

konnov:
This PR closes #608. The types that are found by the type checker are stored in a map. All expressions and declarations are tagged with the computed types. This seemingly simple feature required some debugging and refactoring of the type checker. Also, I have found a bug with chains of excepts that is postponed for #617.


tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
```diff
@@ -25,38 +26,202 @@ class TestTypeCheckerTool extends FunSuite with BeforeAndAfterEach with EasyMock
...
-        |A(x, y) == x + y
-        |
-        |Init ==
-        |  msgs = {}
-        |
-        |Next ==
-        |  \E i \in 1..10:
-        |    Send([type |-> "1a", val |-> i])
-        |================================
+  // We push plenty of TLA+ constructs in this spec, to make sure that:
```
shonfeder, line 38: I wonder if it wouldn't be more maintainable to just read this from a file?

shonfeder (approved):
LGTM

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
konnov: Good point. So far I did want to mess with java class paths. But having this test in an actual file could improve the code.

PR #620, opened 2021-03-04
inlining nullary operators

konnov:
This PR forces the inliner to inline nullary operators too. Although I like nullary operators a lot, they are in conflict with the preprocessing code. It often happens that the optimizers cannot rewrite expressions because they are partially hidden in a LET-definition.


Kukovec (approved):
Seems fine, and I'm not going to block on this, but I feel like this should be a flag option in the future.

konnov:
@Kukovec please have a look at the breaking tests

konnov:
It looks like we have to think more about it.

PR #632, opened 2021-03-10
One more step to #521

konnov:
This PR makes one more step towards the integration #521:

- The command `check` has a new option `--with-snowcat`, which enables the type checker in the beginning and the end of the pipeline
- Some of the model checker's tests are adapted to Snowcat. Some are still failing, so they are running without the option `--with-snowcat`.
- The types are still not propagated to the model checker itself

### DIFF
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
@@ -782,6 +782,34 @@ class ToEtcExpr(annotationStore: AnnotationStore, varPool: TypeVarPool) extends 
         val opsig = OperT1(Seq(FunT1(IntT1(), a), IntT1()), SeqT1(a))
         mkExRefApp(opsig, Seq(fun, len))
 
+      case OperEx(BmcOper.assign, lhs, rhs) =>
+        val a = varPool.fresh
+        // (a, a) => Bool
+        val opsig = OperT1(Seq(a, a), BoolT1())
+        mkExRefApp(opsig, Seq(lhs, rhs))
+
+      case OperEx(BmcOper.expand, set) =>
+        val a = varPool.fresh
+        // a => Bool
+        val opsig = OperT1(Seq(a), a)
+        mkExRefApp(opsig, Seq(set))
+
+      case OperEx(BmcOper.skolem, predicate) =>
+        // Bool => Bool
+        val opsig = OperT1(Seq(BoolT1()), BoolT1())
+        mkExRefApp(opsig, Seq(predicate))
+
+      case OperEx(BmcOper.constCard, predicate) =>
+        // Bool => Bool
+        val opsig = OperT1(Seq(BoolT1()), BoolT1())
+        mkExRefApp(opsig, Seq(predicate))
+
+      case OperEx(BmcOper.distinct, args @ _*) =>
+        val a = varPool.fresh
+        // (a, ..., a) => Bool
+        val opsig = OperT1(args.map(_ => a), BoolT1())
+        mkExRefApp(opsig, args)
+
       //******************************************** MISC **************************************************
       case OperEx(BmcOper.withType, lhs, _) =>
         // Met an old type annotation. Warn the user and ignore the annotation.
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
@@ -24,11 +24,11 @@ class ToEtcExpr(annotationStore: AnnotationStore, varPool: TypeVarPool) extends 
     decl match {
       case d: TlaConstDecl =>
         // CONSTANT N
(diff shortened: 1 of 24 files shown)


tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -152,6 +150,22 @@ class ToEtcExpr(annotationStore: AnnotationStore, varPool: TypeVarPool) extends
...
   def apply(ex: TlaEx): EtcExpr = {
+    val tex = transform(ex)
+
+    ex.typeTag match {
+      case Typed(typeInTag: TlaType1) =>
+        // the expression has a type tag, so we use an identity operator
+        val a = varPool.fresh
+        val identity = OperT1(Seq(a, a), a)
+        val blameRef = BlameRef(ex.ID)
+        mkApp(blameRef, Seq(identity), mkConst(blameRef, typeInTag), tex)
```
shonfeder, line 161: I don't understand why we introduce an identity operator here. Should you perhaps add a line or two in the comments  to motivate this?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -782,6 +796,34 @@ class ToEtcExpr(annotationStore: AnnotationStore, varPool: TypeVarPool) extends
         val opsig = OperT1(Seq(FunT1(IntT1(), a), IntT1()), SeqT1(a))
         mkExRefApp(opsig, Seq(fun, len))
 
+      case OperEx(BmcOper.assign, lhs, rhs) =>
```
shonfeder, line 799: So are these expressions we missed in previous implementations? Are we sure we're supporting them all know?

Any tests we could do to ensure we have complete coverage?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/passes/EtcTypeCheckerPassImpl.scala
```diff
@@ -3,32 +3,42 @@ package at.forsyte.apalache.tla.typecheck.passes
...
    *
    * @return true, if the pass was successful
    */
   override def execute(): Boolean = {
-    if (tlaModule.isDefined) {
+    if (tlaModule.isEmpty) {
+      logger.info(" > no input for type checker")
+      false
+    } else if (!options.getOrElse("typechecker", "snowcatOn", false)) {
+      logger.info(" > Snowcat is disabled. Use --with-snowcat to enable it")
```
shonfeder, line 38: These seems like noise to me.

The option is documented in the help. I don't think we want to tell the user about every option they're *not* using in the logs.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
```diff
@@ -94,4 +94,33 @@ class TestTypeCheckerTool extends FunSuite with BeforeAndAfterEach with EasyMock
       }
     }
   }
+
+  test("the tools runs consumes its output") {
```
shonfeder, line 98: Is this mean to be something like?

```suggestion
  test("the tools consumes its output") {
```

I think there's one too many `s`s for me to parse this out.

shonfeder (approved):
LGTM

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: Good point. I am now receiving complaints by the type checker about this. It is quite confusing. Basically, I want to enforce that the expression has the same type as the one in the tag.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: Now we are supporting them. There are tests for the newly added operators. As for coverage, they popped up in the integration tests.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/passes/EtcTypeCheckerPassImpl.scala
konnov: Yeah. This message is only for the short transition period.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
konnov: Yes, you are right :))

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: I have fixed that in #641, as it was popping up there too

PR #641, opened 2021-03-11
step+2 to #521

konnov:
Another PR for the integration of types in the model checker. I am pushing it before adding too much code.

- Disabled the preprocessing step that introduced `CALL_*`. This does not change the behavior that is visible to the user.
- Fixed the type checker, to make the diagnostic messages comprehensible. Also fixed a few bugs.
- The results of type checking are now dumped in JSON, which makes debugging much easier.
- Refactored `ExprOptimizer` and `ConstSimplifier` to make them type-aware.


tla-types/src/main/scala/at/forsyte/apalache/tla/types/TlaType1TagPrinter.scala
```diff
@@ -0,0 +1,20 @@
...
+      case Typed(tt: TlaType1) => tt.toString
+      // unexpected type, output as much as we can
+      case Typed(_) => "Typed[%s](%s)".format(tag.getClass.getSimpleName, tag)
+    }
+  }
+}
+
+object TlaType1PrinterPredefs {
+  implicit val printer: TypeTagPrinter = new TlaType1TagPrinter
+}
```
Kukovec, line 20: Did you not see ` tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TlaType1Printer.scala` ? It was written as a placeholder, so please either delete it or change it to look like this new file, so we don't have duplicate printers in the codebase.

tla-types/src/main/scala/at/forsyte/apalache/tla/types/TlaType1TagPrinter.scala
```diff
@@ -0,0 +1,20 @@
...
+import at.forsyte.apalache.tla.lir.io.TypeTagPrinter
+import at.forsyte.apalache.tla.typecheck.TlaType1
+
+class TlaType1TagPrinter extends TypeTagPrinter {
+  def apply(tag: TypeTag): String = {
+    tag match {
+      case Untyped()           => "Untyped"
+      case Typed(tt: TlaType1) => tt.toString
+      // unexpected type, output as much as we can
+      case Typed(_) => "Typed[%s](%s)".format(tag.getClass.getSimpleName, tag)
```
Kukovec, line 13: Missing `case somethingElse` ? Make it throw explicitly (or write something along the lines of `UNDEFINED`), so we don't get surprised by crashes on `TlaType7` tags in the future.

Kukovec (commented):
LGTM for the most part, apart from the one issue with `ReplaceFixed`. I do have an additional comment/question regarding tags. Looking at some of these changes, I have to ask myself if it wouldn't have been better to add type-tags to the builder directly. That way, we wouldn't need to manually write things like `ValEx(TlaInt(...))(intTag)`

konnov:
> LGTM for the most part, apart from the one issue with ReplaceFixed. I do have an additional comment/question regarding tags. Looking at some of these changes, I have to ask myself if it wouldn't have been better to add type-tags to the builder directly. That way, we wouldn't need to manually write things like ValEx(TlaInt(...))(intTag)

Yes. This looks ugly. I don't have a perfect solution yet. One workaround is to use the `.withTag` method of `TlaEx`. Then you can write `tla.int(2).withType(Typed(IntT1()))`. Still, when we have many nested expressions, you have to sprinkle all of them with types. Such code looks ugly.

Maybe we should add an expression builder in the real sense of it. That is build chained expressions, add types by IDs and then produce a `TlaEx`. Similar to `StringBuilder`, which collects pieces and then produces a resulting expression.

tla-types/src/main/scala/at/forsyte/apalache/tla/types/TlaType1TagPrinter.scala
konnov: No missing case, as `Typed` and `Untyped` are the only case classes. `TypeTag` is sealed.

tla-types/src/main/scala/at/forsyte/apalache/tla/types/TlaType1TagPrinter.scala
konnov: I would prefer to remove `TlaType1Printer`, as it is more ad-hoc. We should in general avoid objects like this one.

tla-types/src/main/scala/at/forsyte/apalache/tla/types/TlaType1TagPrinter.scala
Kukovec: Yeah, for whatever reason my brain just ignored l.13

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
```diff
@@ -64,81 +67,66 @@ class ExprOptimizer(nameGen: UniqueNameGenerator, tracker: TransformationTracker
...
         if intVal == BigInt(2) =>
+      // Cardinality(Set) >= 2.
       // We can find this pattern in real TLA+ benchmarks more often than one would think.
       // Rewrite into LET T3 = set IN \E t1 \in T3: \E t2 \in T3: t1 /= t2.
       // We use let to save on complex occurrences of set.
+      val elemType = set.typeTag match {
+        case Typed(SetT1(et)) => et
+        case _                =>
+          // this should never happen as the type checker ensures that types are correct
+          throw new TypingException("Unexpected type %s in expression %s".format(set.typeTag, set))
```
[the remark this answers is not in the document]
konnov: Then it deserves an exception, as it is happening now. The type checker is running right after `ConfigurationPass`. All transformations must either propagate types without understanding them, or be type aware, like `ExprOptimizer`.

I was considering having purely untyped transformations in the loop and resolving the missing types with the type checker running once again before the model checker. This is too error-prone, as some transformations may introduce ambiguous expressions, like tuples.

konnov:
> I think this is wrong, but it's very subtle. The 2nd argument of ReplaceFixed(tracker).apply takes a => TlaEx, not a TlaEx, so it can construct fresh copies (with unique IDs). The way you're passing it by name (ye), does not achieve this, but instead replaces them with exact copies of ye, which all have the same UID.

I was thinking about that. Thanks for confirming! How do we avoid that mistake in the future? Why don't we produce a unique copy on each replacement?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
```diff
@@ -147,19 +134,18 @@ class ExprOptimizer(nameGen: UniqueNameGenerator, tracker: TransformationTracker
...
 
-    case OperEx(TlaBoolOper.exists, NameEx(boundVar), OperEx(TlaSetOper.map, mapEx, varsAndSets @ _*), pred) =>
-      // e.g., \E x \in {e: y \in S}: g becomes \E y \in S: LET tmp == e IN g [x replaced with e]
-      // LET x == e IN g in the example above
-      val letName = nameGen.newName()
-      val newPred = ReplaceFixed(tracker)(NameEx(boundVar), tla.appOp(NameEx(letName))).apply(pred)
+    case OperEx(TlaBoolOper.exists, xe @ NameEx(_), OperEx(TlaSetOper.map, mapEx, varsAndSets @ _*), pred) =>
+      // e.g., \E x \in {e: y \in S}: g becomes \E y \in S: g[x replaced with e]
+      // g[x replaced with e] in the example above
+      val newPred = ReplaceFixed(tracker)(replacedEx = xe, newEx = DeepCopy(tracker).deepCopyEx(mapEx)).apply(pred)
```
konnov, line 147: @Kukovec does it look better now? I think `deepCopy` should be the default behavior on every replacement. Are there any scenarios where this should not be the case?

Kukovec:
> 
> 
> > I think this is wrong, but it's very subtle. The 2nd argument of ReplaceFixed(tracker).apply takes a => TlaEx, not a TlaEx, so it can construct fresh copies (with unique IDs). The way you're passing it by name (ye), does not achieve this, but instead replaces them with exact copies of ye, which all have the same UID.
> 
> I was thinking about that. Thanks for confirming! How do we avoid that mistake in the future? Why don't we produce a unique copy on each replacement?

If you want fresh copies just call it as `ReplaceFixed(a,b.copy())`. The reason `ReplaceFixed` doesn't copy by default is because the`=> TlaEx` argument allows it to be much more flexible in what it's replacing the argument with (and because it's easy to just write the `.copy()`).

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
Kukovec: Ah, missed this by 42s apparently. See my other comment.

konnov:
> If you want fresh copies just call it as ReplaceFixed(a,b.copy()). The reason ReplaceFixed doesn't copy by default is because the=> TlaEx argument allows it to be much more flexible in what it's replacing the argument with (and because it's easy to just write the .copy()).

Should it be `copy` or `deepCopy`. The problem is that the interface user (in this case me) does not know about all these subtleties. It is better to make copy the default behavior.

Kukovec:
> 
> 
> > If you want fresh copies just call it as ReplaceFixed(a,b.copy()). The reason ReplaceFixed doesn't copy by default is because the=> TlaEx argument allows it to be much more flexible in what it's replacing the argument with (and because it's easy to just write the .copy()).
> 
> Should it be `copy` or `deepCopy`. The problem is that the interface user (in this case me) does not know about all these subtleties. It is better to make copy the default behavior.

Well, `DeepCopy`, if you want to preserve tracking.

konnov:
> > > If you want fresh copies just call it as ReplaceFixed(a,b.copy()). The reason ReplaceFixed doesn't copy by default is because the=> TlaEx argument allows it to be much more flexible in what it's replacing the argument with (and because it's easy to just write the .copy()).
> > 
> > 
> > Should it be `copy` or `deepCopy`. The problem is that the interface user (in this case me) does not know about all these subtleties. It is better to make copy the default behavior.
> 
> Well, `DeepCopy`, if you want to preserve tracking.

`ReplaceFixed` should take of tracking too, should not it?

Kukovec:
> 
> 
> > > > If you want fresh copies just call it as ReplaceFixed(a,b.copy()). The reason ReplaceFixed doesn't copy by default is because the=> TlaEx argument allows it to be much more flexible in what it's replacing the argument with (and because it's easy to just write the .copy()).
> > > 
> > > 
> > > Should it be `copy` or `deepCopy`. The problem is that the interface user (in this case me) does not know about all these subtleties. It is better to make copy the default behavior.
> > 
> > 
> > Well, `DeepCopy`, if you want to preserve tracking.
> 
> `ReplaceFixed` should take of tracking too, should not it?

It does, in fact, so `.copy()` is actually enough. But if you want this to be default behavior, you can open an issue.

tla-types/src/main/scala/at/forsyte/apalache/tla/types/TlaType1TagPrinter.scala
Kukovec: Feel free to delete it then.

Kukovec (approved):

PR #642, opened 2021-03-11
Hotfix: TlaOperDecl.isRecursive propagation

### LINKED ISSUE #635: [BUG] Hotfix: TlaOperDecl.isRecursive propagation
Despite the copy-constructor introduced in #347, there are several instances where `TlaOperDecl` are constructed as derivatives  of existing declarations, but without using the copy constructor. Copying this way does not reference the original declaration's `.isRecursive` field, resulting in the flag defaulting to false. For instance, in `IncrementalRenaming`, any operator that is both a) recursive and b) defined inside a LET expression becomes flagged as nonrecursive, triggering a stack overflow in unrolling. 
All such cases need to be found and replaced with copy-constructors. This also future-proofs for #345.

Kukovec:
Closes  #635

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Cacher.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Cacher.scala
@@ -5,6 +5,8 @@ import at.forsyte.apalache.tla.lir._
 import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModuleTransformation, TransformationTracker}
 import at.forsyte.apalache.tla.lir.UntypedPredefs._
 
+// TODO: Remove?
+
 class Cacher(nameGenerator: UniqueNameGenerator, tracker: TransformationTracker) extends TlaModuleTransformation {
 
   def prepareAppMap(operNames: Set[String], init: Map[TlaEx, String] = Map.empty): TlaEx => Map[TlaEx, String] = {
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
@@ -81,9 +81,10 @@ class Unroller(nameGenerator: UniqueNameGenerator, tracker: TransformationTracke
    * with the default value.
    */
   private def replaceWithDefaults(defaultsMap: Map[String, TlaEx]): TlaExTransformation = tracker.trackEx {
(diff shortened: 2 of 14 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Cacher.scala
```diff
@@ -5,6 +5,8 @@ import at.forsyte.apalache.tla.lir._
 import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TlaModuleTransformation, TransformationTracker}
 import at.forsyte.apalache.tla.lir.UntypedPredefs._
 
+// TODO: Remove?
+
```
konnov, line 9: what is the role of this `Cacher`? Do we still need it? If not, let's remove it.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
```diff
@@ -157,31 +158,64 @@ class Unroller(nameGenerator: UniqueNameGenerator, tracker: TransformationTracke
...
+  /**
+   * Finds the names of all recursive operators defined under decl.
+   * This includes recursiveoperators defined in LET-IN subexpressions,
+   * as well as the name of `decl` itself, if it is recursive.
+   */
+  private def allRecursiveOperatorsInDecl(decl: TlaOperDecl): Set[String] = {
+    // If this declaration is recursive, we take its base name (for LET-IN defined)
+    val selfOpt = if (decl.isRecursive) Some(IncrementalRenaming.getBase(decl.name)) else None
+    allRecursiveLetInOperatorNames(decl.body) ++ selfOpt
+  }
```
konnov, line 205: Does it mean that we do not rely on SANY on identifying recursive operators?

konnov (approved):
Looks great! I've left a few comments

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Unroller.scala
Kukovec: We do, but LET-IN defined recursive operators are not toplevel, so you need to scan the entire syntax tree for them.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Cacher.scala
Kukovec: I think it was something we disabled a long time ago, I just found it while scanning for instances of OperDecl ctors. If we do remove it, however, that should be its own PR.

PR #652, opened 2021-03-15
Integrating types 4: migrated tests to Snowcat

konnov:
This PR fixes type annotations in the integration tests. Several tests are still failing. This PR is created for the purpose of seeing a small diff.


test/tla/Paxos.tla
```diff
@@ -124,8 +129,13 @@ TypeOK == /\ maxBal \in [Acceptor -> Ballot \cup {-1}]
 (* the array `votes' describing the votes cast by the acceptors is defined *)
 (* as follows.                                                             *)
 (***************************************************************************)
+
+\* the original specification does not have this definition, we need it for types
```
shonfeder, line 133: Why do we need this? Is this kind of thing a short term work around, or a permanent requirement?

test/tla/Paxos.tla
```diff
@@ -151,7 +161,7 @@ Send(m) == /\ msgs' = msgs \cup {m <: MT}
 (* m with m.type = "1a") that begins ballot b.                             *)
 (***************************************************************************)
 Phase1a(b) == /\ Send([type |-> "1a", bal |-> b])
-              /\ UNCHANGED <<maxBal, maxVBal, maxVal>>
+              /\ UNCHANGED <<maxBal, maxVBal, maxVal, 0>>
```
shonfeder, line 164: Why are we adding a 0 here?

shonfeder (approved):
LGTM, my questions notwithstanding.

shonfeder:
But, iiuc, we should hold of merging this into the feature branch until that has been reviewed first.

test/tla/Paxos.tla
konnov: That's a hack that I don't know how to avoid. Basically, if all variables have the same type, the type checker does not know, whether the argument to UNCHANGED is a tuple or a sequence. It's kind of a stupid behavior, but there is no good type we can enforce, as the user can pass literally anything to UNCHANGED. By adding `0`, I made the types in `<<...>>` non-uniform, so the type checker is not complaining.

konnov: It's obviously a hack. I hope we find a better solution when we used type annotations for some time.

shonfeder: Would it be possible to add a special annotation for `UNCHANGED` declarations for this purpose?

konnov: We can add a special rule in `ToExpr`, as this idiom is literally everywhere.

test/tla/Paxos.tla
konnov: I think it is a permanent construct. It is often too cumbersome to distinguish between tuples and sequences. I was even thinking of introducing operators like in Scala: `tuple0()`, `tuple1(x)`, `tuple2(x, y)`, ...

test/tla/Paxos.tla
konnov: See https://github.com/informalsystems/apalache/issues/660

konnov:
> But, iiuc, we should hold of merging this into the feature branch until that has been reviewed first.

Yes, we are doing matryoshka PRs now :)

test/tla/Paxos.tla
Kukovec: I would actually be more in favor of fixing the signature of UNCHANGED to always expect tuples. There's really no reason to have sequences in UC, you never get the variable vectors from `Seq(...)`-type sets.

Kukovec (approved):
LGTM

konnov:
> I would actually be more in favor of fixing the signature of UNCHANGED to always expect tuples. There's really no reason to have sequences in UC, you never get the variable vectors from Seq(...)-type sets.

You can write `UNCHANGED x`. If `x` is a sequence, you have a sequence ;-)

PR #663, opened 2021-03-19
Integrating types 7: victory over enthropy

konnov:
This PR fixes a bunch of preprocessing code and TLC configuration parser. We have a green bar!

That means that all preprocessing steps are producing typed expressions. In theory. No integration tests are failing and the type watchdog is not barking.

Plenty of files are marked as changed, as I had to move `TlaType1`, `TypedPredef`, and `TypingException` to the IR module (tlair). This was necessary to break a circular dependency between the type checker and the `tla-import`.

The final step is to feed the types to the SMT translator and then we are done.


Kukovec (approved):
A lot of files changed, but from what I can tell, the non-trivial fixes look fine.

PR #664, opened 2021-03-19
Doctests and minor fixes in functions.md

Alexander-N:
When the doctest format is used then the results can be checked. Downside is the output has to match exactly what the Python interpreter returns, which is inconvenient in case of sets where the order is undefined. Also already defined variables like `py_price` have to be redefined for every doctest.


konnov (approved):
@Alexander-N, thanks a lot for your PR. The proposed changes look great!

Issue #656, opened 2021-03-17
[FEATURE] Simplify FormalParam


konnov:
We have the following three definitions to express operator parameters:

```scala
/**
 * A formal parameter of an operator.
 */
sealed abstract class FormalParam extends Serializable {
  def name: String

  def arity: Int

}

/** An ordinary formal parameter, e.g., x used in A(x) == ... */
case class SimpleFormalParam(name: String) extends FormalParam with Serializable {
  override def arity: Int = 0
}

/** A function signature, e.g., f(_, _) used in A(f(_, _), x, y) */
case class OperFormalParam(name: String, arity: Int) extends FormalParam with Serializable {}
```

It is not clear to me why the parameters are so complex. We can have one class instead:

```scala
class OperParam(name: String, arity: Int) extends Serializable {
  def isHigherOrder: Bool = arity > 0

  /* override equals and hashCode */
}
```

This refactoring will touch a lot of code. But we should do it after having integrated the type checker.

Issue #634, opened 2021-03-10
[FEATURE] Standardize operator names in the IR


konnov:
Related to #610. We have to go over all operator names and change them to concise alphanumeric names. Some names are pretty ad-hoc in the current version like `_.._`, `{...}`, and `_()`. Since the operators names are used in the JSON format, we have to bring some order there.

konnov:
We should fast track it for an internal project

PR #670, opened 2021-03-22
Json ADR

### LINKED ISSUE #610: [FEATURE] ADR for new JSON format
Splitting from #599, the ADR itself deserves its own PR and comment process.

Kukovec:
Closes #610


shonfeder (approved):
LGTM

shonfeder:
You might want to also add an for this in the mdBook's `SUMMARY.md`, as this is probably of general interest.

vitorenesduarte:
Is it possible to tag each "entry" with its source file/line?

In modelator, we currently do some ad-hoc parsing to fetch the names of the TLA+ assertions representing tests: https://github.com/informalsystems/modelator/blob/99f656fa8b3cf46a2aa0b6513e4e140d1778c4bd/modelator/src/module/tla/mod.rs#L70-L98
Ideally, we should be using Apalache for this, and having the source file on each "entry" would make that process more reliable (i.e. not requiring another hack to check from which file the assertion comes from) .

konnov:
Can we add a version number to the format. So the user code will be able to see that they have to deal with an incompatible version?

Kukovec:
> 
> 
> Is it possible to tag each "entry" with its source file/line?
> 
> In modelator, we currently do some ad-hoc parsing to fetch the names of the TLA+ assertions representing tests: https://github.com/informalsystems/modelator/blob/99f656fa8b3cf46a2aa0b6513e4e140d1778c4bd/modelator/src/module/tla/mod.rs#L70-L98
> Ideally, we should be using Apalache for this, and having the source file on each "entry" would make that process more reliable (i.e. not requiring another hack to check from which file the assertion comes from) .

In principle yes, that shouldn't be too difficult to add, but you'd have to keep in mind that the "location" field would have to be optional, because these JSON representations are intended to bidirectionally connect Apalache with 3rd-party applications, which should be able to produce JSON files, without necessarily starting from a TLA spec. That said, output generated by Apalache could certainly include source information. 

> 
> 
> Can we add a version number to the format. So the user code will be able to see that they have to deal with an incompatible version?

Yes.

konnov:
This issue is related: #656. I think we should really go for simple formal parameters. This will make both json and our code easier to read.

konnov:
This one is also related: #634. I wonder, whether we should include the list of operator names in this ADR? Otherwise, the readers will have to figure out the operator names themselves.

docs/src/adr/005adr-json.md
```diff
@@ -0,0 +1,121 @@
...
+    kind: "TlaInt",
+    value: 1
+  }
+}
+```
+
+Note that in the above example, the outer `value` field contains a JSON serialization of `TlaInt(1)`, and not just the JSON literal `1`. This is because of deserialization:
+Apalache encodes its TLA+ integers as `BigInt`, which means that it permits values, for which `.isValidInt` does not hold. If we chose to encode values directly as JSON numbers, we would need a) a sensible encoding of `BigInt` values, which are not valid integers and b) a way to distinguish both variants of `BigInt`, as well as decimals, when deserializing (since JSON is not typed).
+Alternatively, we could encode all values as strings, which would be similarly indistinguishable when deserializing. Importantly, the `type` field is not guaranteed to contain a hint, as it could be `Untyped`.
+For the above reasons, we choose to instead serialize `TlaValue` as a JSON object, which is more verbose, but trivial to deserialize. It has the following shape
```
konnov, line 85: Can you add a bit more structure to this text? For instance, describe the general rules in Section "1. General rules" and the rules for Int in Section "2. Rules for Int". Then it will be clear that Int is a special case, and the rest follows the general principle.

docs/src/adr/005adr-json.md
```diff
@@ -0,0 +1,121 @@
...
+Serializations of typed entities will have an additional field named `type`, containing the type of the expression (see [ADR-002][], [ADR-004][] for a description of our type system and the syntax for types-as-string-annotations respectively). For example, the integer literal `1` is represented by a `ValEx`, which has type `Int` and is serialized as follows:
+```
+{
+  kind: "NameEx"
+  type: "Int"
+  ...
+}
+```
+
+## 4. Serialization rules
```
konnov, line 51: Can you add examples for all kinds of declarations? This should not be hard. This will help the reader to see the format without running the tool. Perhaps, you would want to use the simplified format for the parameters, which I mentioned in a comment to this PR.

konnov (commented):
I like it. Several comments to improve the text. Since this ADR will be most likely used by all external tooling, let's add clarifications.

konnov:
> In principle yes, that shouldn't be too difficult to add, but you'd have to keep in mind that the "location" field would have to be optional, because these JSON representations are intended to bidirectionally connect Apalache with 3rd-party applications, which should be able to produce JSON files, without necessarily starting from a TLA spec. That said, output generated by Apalache could certainly include source information.

Yeah, an optional location field makes sense. We have to decide, whether we want it structured or just text, that is, either `MySpec.tla:10:1-12:10` or `{ filename: "MySpec.tla", from: { line: 10, col: 1 }, to: { line: 12, col: 10} }`. I guess, @vitorenesduarte you would prefer the latter?

I thought of adding IDs to the expressions and collecting the source locations in a separate file. This may be a better solution in the long run, if we want to import/export additional data (e.g., annotations). Or perhaps, we could add a separate section in the json file that contains additional data like source info, annotations, etc.

vitorenesduarte:
> Yeah, an optional location field makes sense. We have to decide, whether we want it structured or just text, that is, either `MySpec.tla:10:1-12:10` or `{ filename: "MySpec.tla", from: { line: 10, col: 1 }, to: { line: 12, col: 10} }`. I guess, @vitorenesduarte you would prefer the latter?

Both would work for modelator but indeed I prefer the latter as it is explicit.

konnov:
For the table of operators, we don't care about the removed ones: #687, #686, #685, #684

docs/src/adr/005adr-json.md
shonfeder: That kind of documentation will surely be useful for readers, but I don't think manually generating it for the ADR is a very good use of time.

Once it's implemented, we can just automatically generate it for the docs (and even ensure that the docs stay up to date).

shonfeder:
> This one is also related: #634. I wonder, whether we should include the list of operator names in this ADR? Otherwise, the readers will have to figure out the operator names themselves.

This also seems to me like something we should automatically once the serialization in place and put in the documentation, rather than putting it in the ADR>

docs/src/adr/005adr-json.md
```diff
@@ -0,0 +1,121 @@
...
+```
+{
+  kind: "NameEx"
+  ...
+}
+```
+
+## 3. Serializing typed entities
+
+Serializations of typed entities will have an additional field named `type`, containing the type of the expression (see [ADR-002][], [ADR-004][] for a description of our type system and the syntax for types-as-string-annotations respectively). For example, the integer literal `1` is represented by a `ValEx`, which has type `Int` and is serialized as follows:
```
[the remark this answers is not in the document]
Kukovec: This is not about typed vs untyped, everything but `TlaModule` is typed, in the sense that it has _some_ type tag (possibly `Untyped`). The language should probably be clearer and say "tagged", not "typed".

docs/src/adr/005adr-json.md
konnov: That would be even better

docs/src/adr/005adr-json.md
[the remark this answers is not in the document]
konnov: Yes. We should be also clear about how we represent untyped expressions

konnov:
> > This one is also related: #634. I wonder, whether we should include the list of operator names in this ADR? Otherwise, the readers will have to figure out the operator names themselves.
> 
> This also seems to me like something we should do automatically once the serialization in place and put in the documentation, rather than putting it in the ADR.

My motivation here is that the names we have right now are quite ad-hoc. If people start integrating with our code via json, it will be much hard to change the names. Of course, I can just go over the list and pick the names I like :)

docs/src/adr/005adr-json.md
```diff
@@ -70,23 +140,20 @@ OperEx(
...
-Apalache encodes its TLA+ integers as `BigInt`, which means that it permits values, for which `.isValidInt` does not hold. If we chose to encode values directly as JSON numbers, we would need a) a sensible encoding of `BigInt` values, which are not valid integers and b) a way to distinguish both variants of `BigInt`, as well as decimals, when deserializing (since JSON is not typed).
-Alternatively, we could encode all values as strings, which would be similarly indistinguishable when deserializing. Importantly, the `type` field is not guaranteed to contain a hint, as it could be `Untyped`.
-For the above reasons, we choose to instead serialize `TlaValue` as a JSON object, which is more verbose, but trivial to deserialize. It has the following shape
+Observe that we choose to serialize `TlaValue` as a JSON object, which is more verbose, but trivial to deserialize. It has the following shape
 ```
 {
-  kind: _
-  value: _
+  "kind": _ ,
+  "value": _
```
konnov, line 156: what are the underscores? Is this standard JSON semantics? Shall we use the syntax for comments like `/* the kind */` and `/* the value /*`.

docs/src/adr/005adr-json.md
```diff
@@ -95,21 +162,27 @@ The `value` field depends on the kind of `TlaValue`:
 1. For `TlaBool`: a JSON Boolean
 1. For `TlaInt(bigIntValue)`: 
   1. If `bigIntValue.isValidInt`: a JSON number
-  2. Otherwise: `{ bigInt: bigIntValue.toString() }`
+  2. Otherwise: `{ "bigInt: bigIntValue.toString() }`
```
konnov, line 165:
```suggestion
  2. Otherwise: `{ "bigInt": bigIntValue.toString() }`
```

docs/src/adr/005adr-json.md
```diff
@@ -95,21 +162,27 @@ The `value` field depends on the kind of `TlaValue`:
...
 1. For `TlaDecimal(decValue)`: a JSON string `decValue.toString`
 
+The reason for the non-uniform treatment of integers, is that Apalache encodes its TLA+ integers as `BigInt`, which means that it permits values, for which `.isValidInt` does not hold. 
+
+While it might seem more natural to encode the entire `TlaValue` as a JSON primitive, without the added object layer, we would have a much tougher time deserializing.
+We would need a) a sensible encoding of `BigInt` values, which are not valid integers, and b) a way to distinguish both variants of `BigInt`, as well as decimals, when deserializing (since JSON is not typed).
+We could encode all values as strings, but they would be similarly indistinguishable when deserializing. Importantly, the `type` field of the `ValEx` expression is not guaranteed to contain a hint, as it could be `Untyped`
```
konnov, line 172: You could cite [RFC7159](https://tools.ietf.org/html/rfc7159#page-6) to justify why we cannot just use unbounded numbers.

docs/src/adr/005adr-json.md
```diff
@@ -95,21 +162,27 @@ The `value` field depends on the kind of `TlaValue`:
...
 Take `jsonOf1` to be the serialization of `ValEx( TlaInt( 1 ) )( typeTag = Typed( IntT1() ) )` shown above. The serialization of `1 + 1` is then equal to
 ```
 {
-  kind: "OperEx",
-  type: "Int",
-  oper: "(+)",
-  args: [jsonOf1, jsonOf1]
+  "kind": "OperEx",
+  "type": "Int",
+  "oper": "(+)",
```
konnov, line 179:
```suggestion
  "oper": "+",
```

`(+)` is a different operator in TLA+.

tla-import/src/main/scala/at/forsyte/apalache/io/json/TlaToJson.scala
```diff
@@ -53,146 +54,187 @@ import at.forsyte.apalache.tla.lir.io.TypeTagPrinter
...
-class TlaToJson[T <: JsonRepresentation](factory: JsonFactory[T])(implicit typTagPrinter: TypeTagPrinter)
+class TlaToJson[T <: JsonRepresentation](
+    factory: JsonFactory[T], locatorOpt: Option[SourceLocator] = None
+)(implicit typeTagPrinter: TypeTagPrinter)
     extends JsonEncoder[T] {
-  val kindFieldName: String = "kind"
-  val typeFieldName: String = "type"
+  import TlaToJson._
 
   implicit def liftString: String => T = factory.fromStr
```
konnov, line 63: Is there a good reason for using implicit conversions here? AFAIK, this may stop working in Scala 3. In general, it's hard to refactor code that is using implicits, as we have seen recently.

docs/src/adr/005adr-json.md
```diff
@@ -32,23 +32,93 @@ The following classes are serializable:
...
+## 5. Version information
+JSON serializations of `TlaModule`s contain a filed named `version`, holding a string representation of the current JSON encoding version, shaped `{major}.{minor}`.
+This document defines JSON Version 1.0. If and when a different JSON version is defined, this document will be updated accordingly.
+Apalache may refuse to import, or trigger warnings for, JSON objects with obsolete versions of the encoding in the future.
+Example:
+```
+{
+  "kind": "TlaModule",
+  "name": "MyModule",
+  "version": "1.0"
```
konnov, line 116: I think we should introduce a different root object for the json format. We may want to add meta information to it later. So I would suggest adding another object on top of `TlaModule`, e.g.:

```tla
{
  "name": "ApalacheIR",
  "version": "1.0",
  "modules": [{
    "kind": "TlaModule",
    "name": "MyModule",
    ...
  }]
}
```

konnov (commented):
I like it a lot, but have a few concerns with the new version. By the way, are we consistent with https://tools.ietf.org/html/rfc7159 ?

docs/src/adr/005adr-json.md
```diff
@@ -32,23 +32,93 @@ The following classes are serializable:
...
+}
+```
+or
+```
+{
+  "source": "UNKNOWN"
+}
+```
+if no source information is available (e.g. for expressions generated purely by Apalache).
+Serializations generated by Apalache are always guaranteed to contain a `source` field entry.
```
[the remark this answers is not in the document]
Kukovec: I mean, I very explicitly write "or [UNKNOWN] if no source information is available." The next line asserts that a `source` field exists, not that the contents of said field are not "UNKNOWN".

docs/src/adr/005adr-json.md
Kukovec: The underscores are, at least to me, very obvious meta-language placeholders. No less formal than `/* */`, since JSON doesn't support comments anyway.

docs/src/adr/005adr-json.md
Kukovec: Perhaps, but that is the value of `TlaArithOper.plus.name`

tla-import/src/main/scala/at/forsyte/apalache/io/json/TlaToJson.scala
Kukovec: Yes, being able to write `s -> "abc"` instead of `s -> factory.fromStr("abc")` in like 20 spots. This is far from the kind of implicits used in (un)typed predefs, since it's declared in the same spot as it is used, so it's more of a convenience shorthand.

docs/src/adr/005adr-json.md
Kukovec: I don't think it's about the capabilities of JSON numbers, it has more to do with the unnecessary complexity of interpreting those JSON values as the correct Scala type, which we can avoid pretty much entirely by just having a `kind` field hint.

Kukovec:
> I like it a lot, but have a few concerns with the new version. By the way, are we consistent with https://tools.ietf.org/html/rfc7159 ?

In what way do you think we are not?

konnov:
> > I like it a lot, but have a few concerns with the new version. By the way, are we consistent with https://tools.ietf.org/html/rfc7159 ?
> 
> In what way do you think we are not?

That was a suggestion to quickly check RFC 7159 :)

docs/src/adr/005adr-json.md
konnov: Since it is an ADR and we have #726, let's change it to `PLUS` right away.

docs/src/adr/005adr-json.md
konnov: The reason for not using JSON number is that they have a bounded range, something like `[-(2**53)+1, (2**53)-1]`, which I am sure a few people know about.

docs/src/adr/005adr-json.md
konnov: Well, I surely understand what you mean by `_`, but we are writing an ADR. So we should not assume too much about the reader. I have not seen the underscore notation outside of OCaml, Scala, and maybe Perl.

docs/src/adr/005adr-json.md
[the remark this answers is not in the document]
konnov: I am concerned about this sentence:

> Serializations generated by Apalache are always guaranteed to contain a `source` field entry.

Sometimes, we are missing source information. So "always guaranteed" is way to strong.

shonfeder:
I think the CI failure here is due to a bintray outage: https://status.bintray.com/

I'll look into reducing the dependency chain in the meanwhile, but I'll make the mac-os integration tests optional for the time being.

shonfeder:
CI Failure should be fixed, just need to rebase/merge on top of the trunk.

konnov:
Is there much to do, in order to get it merged? We might get some integration blocked on this JSON export/import.

PR #679, opened 2021-03-24
close #660: UNCHANGED <<...>> treats the brackets as a tuple

### LINKED ISSUE #660: [FEATURE] Add a special type-checking rule for UNCHANGED <<...>>
Currently, `UNCHANGED` and `<<...>>` are translated in `EtcExpr` independently. As a result, an expression like `UNCHANGED <<x, y>>` is not considered well-typed, if `x` and `y` have the same type. The reason is that the type checker is not sure about whether `<<x, y>>` is a sequence or a tuple. As this idiom is met so often in the specs, we should add a separate translation rule for `UNCHANGED <<...>>` that treats the inner `<<...>>` as a tuple.

konnov:
A fix for the annoying complaint by the type checker. It was really easy. I believe that this the effect of designing the type checker right.


shonfeder (approved):
LGTM

PR #681, opened 2021-03-24
Clarify instructions on issue template

shonfeder:
It wasn't clear from the template that the apalache version should be
included. This aims to make that clear, and a few other points, while
also improving the formatting of the template and making it easier for
reporters to use (hopefully) by putting the instructions in comments.

-------


konnov (approved):
Awesome!

PR #704, opened 2021-03-31
Adding type aliases

### LINKED ISSUE #703: [FEATURE] Introduce type aliases in Snowcat
As the example of [Raft](https://github.com/informalsystems/apalache-tests/blob/master/performance/raft/APAraft.tla) shows, the type annotations may be quite large. To this end, we should introduce type aliases. It is not obvious, where to place such aliases. The easiest solution is to allow the user to write a type alias in the place, where they can write `@type`. For instance:

```tla
VARIABLE
  @typeAlias("entry", "[term: Int, contents: Str]")
  @type: Set([seqno: Int, entry: entry]);
  messages
```

To introduce a type alias, we would have to use the general syntax of Java-like annotations, which is a bit harder to read than the single-argument syntax. The above type alias can be used in the type annotations that are below the alias, in the syntactic order of the specification.

konnov:
Per popular request this PR adds type aliases of the form:

```
@typeAlias FOO = [a: Int, b: Bool];
```

This closes #703.


docs/src/adr/002adr-types.md
```diff
@@ -57,30 +98,49 @@ Z3](https://github.com/Z3Prover/z3), a type constant can be understood as an
 uninterpreted sort in SMT. Essentially, values of an uninterpreted type can
 be only checked for equality.
 
-Assume that notation `e <: T` means that an expression `e` has type `T`.
-(More precisely, `T` is a supertype of the type of `e`.)
-The following examples demonstrate the use of the type grammar:
+Like in programming languages, a type may contain a reference to a type alias.
+This is purely a convenience feature to make type annotations more concise and
+easier to maintain. We expect only the user to write type aliases, whereas
+tools should always exchange with types in the alias-free form.
```
shonfeder, line 104:
```suggestion
easier to maintain. We expect that only users will write type aliases:
tools should always exchange data with types in the alias-free form.
```

docs/src/adr/002adr-types.md
```diff
@@ -57,30 +98,49 @@ Z3](https://github.com/Z3Prover/z3), a type constant can be understood as an
...
+   Its type is  `((Int, Str) => Int) => Bool`.
 * `Baz` is a polymorphic operator that takes two arguments of the same type
-    and returns a value of the type equal to the types of its arguments:
-    `Baz <: "(a, a) => a"`.
-* `Proc` and `Faulty` are sets of the same type:
-    `Proc <: "Set(PID)"` and `Faulty <: "Set(PID)"`.
+    and returns a value of the type equal to the types of its arguments.
+   Its type is `(a, a) => a`.
+* `Proc` and `Faulty` are sets of the same type.
+   Their type is `Set(PID)`.
```
shonfeder, line 125: where does PID come from? What does it mean?

docs/src/adr/002adr-types.md
```diff
@@ -57,30 +98,49 @@ Z3](https://github.com/Z3Prover/z3), a type constant can be understood as an
...
+
+### 1.2. Type aliases
 
-### 1.2. Discussion
+We extend the type grammar of `T` with one more rule that introduces a type
+alias:
```
shonfeder, line 130:
```suggestion
We extend the type grammar of `T` with one more rule that introduces type
aliases:
```

shonfeder: This intro to the section doesn't make sense to me, because the type alias is specified in the grammar to begin with, not added as an extensions here (the way the current document is written). Perhaps just say "kthe grammer of `T` includes a rule" rather than "we extend"?

docs/src/adr/002adr-types.md
```diff
@@ -248,6 +308,34 @@ auxiliary LET-definition to specify the type of the empty collection:
...
+<a id="typeAlias"></a>
+
+### 2.4. Introducing and using type aliases
+
+Follow the four simple rules:
+
+1. Type aliases can be only introduced at top-level declarations. That is, type aliases next to nested LET-IN
+   definitions are ignored.
+
+1. Assuming Rule 1, you can define a type alias with `@typeAlias` in the same position, where you could define `@type`.
```
shonfeder, line 320:
```suggestion
1. Assuming Rule 1, you can define a type alias with `@typeAlias` anywhere you can define a `@type`.
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/CachingType1Parser.scala
```diff
@@ -1,44 +0,0 @@
-package at.forsyte.apalache.tla.typecheck
```
shonfeder, line 1: Is this change related to the feature added in this PR?

konnov: Yes. Caching is not consistent with type aliases. If we want to cache parsing results, we have to take the aliases into account. That seems to be too difficult. There is no good use case for caching anyways.

konnov:
The first solution to aliases was not working well:
 - It was introducing a new level of complexity in the type grammar.
 - It was sensitive to the order of declarations, which could change with the topological sort (done by the importer).

Here is a new solution. We just let to redefine a constant type as an alias:

```tla
\* @typeAlias: ENTRY = Str;
\* @type: Int -> ENTRY;
```

I find this approach much more elegant. It also gives us another use for the type constants.

konnov:
@shonfeder, if you have time, please have a look at it.

shonfeder (approved):

PR #725, opened 2021-04-10
Logic chapter in manual: Some fixes and use doctest format

Alexander-N:
Made some fixes, removed `frozenset` where possible and converted the Python examples to the doctest format.

### DIFF
--- a/docs/src/lang/logic.md
+++ b/docs/src/lang/logic.md
@@ -83,11 +83,15 @@ simply syntax sugar for the form with nested quantifiers: `\A x \in S: \A y
 **Example in Python:** Python conveniently offers us a concise syntax:
 
 ```python
-  S = {1, 2, 3, 4}
-  all(x > 0 for x in S)         # True
-  all(x > 2 for x in S)         # False
-  T2 = {(x, y) for x in [1, 2] for y in [3, 4]}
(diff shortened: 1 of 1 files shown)


shonfeder:
Thanks, @Alexander-N!

We discussed internally today, and came to the agreement that these Python examples in the manual are only to be considered as pedagogical illustrations. Since they are no longer meant to prototype or illustrate automated or formally correct translations, the inconsistency between the uses of sets and frozensets is not longer an issue. And we all agree that, when considered from a purely pedagogical perspective, the simplification is worth the inconsistency.

You persistent contributions here have been vital to helping us reach this shared understanding, @Alexander-N. So thank you very much!

Would you be interested in adding a followup with some documentation on how we can run the python tooling on these examples? A good place for that might be in the [CONTRIBUTING.md](https://github.com/informalsystems/apalache/blob/shon/716/no-version-suffixed-release/CONTRIBUTING.md#L183). That would let us integrate verification of the correctness of these snippets into our CI.

shonfeder (approved):
Looks great!

Alexander-N:
Great, I'll convert the other Python examples as well then. I opened https://github.com/informalsystems/apalache/pull/732 to document how to run them.

PR #731, opened 2021-04-12
Fix CI

shonfeder:
Due either to a [homebrew update](https://brew.sh/2021/04/12/homebrew-3.1.0/) or a [bintray outage](https://status.bintray.com/), the part of our macos integration tests that depends on homebrew broke. Adding a step to update homebrew fixes this.


shonfeder:
This also goes to show that we'd benefit from getting the dependencies for our CI pipeline under tighter control. I've some ideas on this, but we should get this change in to fix the pipeline.

konnov (approved):
Totally for it!

PR #733, opened 2021-04-12
Add more doctests

Alexander-N:


docs/src/lang/sets.md
```diff
@@ -106,12 +106,12 @@ If this is not the case, the type checker flags an error.
...
-**Example in Python:** TLA+ sets are immutable, so we are using `frozenset`:
+**Example in Python:**
 
 ```python
-  frozenset({1, 2, 3})
-  frozenset({frozenset({1, 2}), frozenset({2, 3})})
-  frozenset({False, 1})
+{1, 2, 3}
+{frozenset({1, 2}), frozenset({2, 3})}
+{False, 1}
```
shonfeder, line 114: To be consistent, I guess these should also be runnable, like on lines 373-379?

docs/src/lang/sets.md
```diff
@@ -612,7 +649,11 @@ set of sets.
 in Python is quite simple:
 
 ```python
-    functools.reduce(lambda x, y: x | y, s, frozenset())
+>>> from functools import reduce
+>>> s = { frozenset({0, 1}), frozenset({1, 2}), frozenset({3}) }
+>>> reduce(lambda x, y: x | y, s, set())
```
shonfeder, line 654:
```suggestion
>>> reduce((lambda x, y: x | y), s, set())
```

otherwise I'm having trouble parsing out the scope of the commas.

shonfeder (changes requested):
To minor changes, please. Looks great otherwise!

docs/src/lang/sets.md
Alexander-N: I think doctests only make sense when the output is of interest, as in lines 373-379 where some expressions get evaluated. Here it would simply be
```python
>>> {1, 2, 3}
{1, 2, 3}
>>> {frozenset({2, 3}), frozenset({1, 2})}
{frozenset({2, 3}), frozenset({1, 2})}
>>> {False, 1}
{False, 1}
```
Seems strange doesn't it?

shonfeder: I prefer the consistency at a price of a bit of redundancy, but I'm ok with leaving it this way if you prefer. I can see your point.

Alexander-N: Sure, I think it doesn't make much of a difference.

shonfeder (approved):

shonfeder:
Thanks!

Issue #705, opened 2021-03-31
[FEATURE] Warn about unmatched type annotations


lemmy:
Beginner problem but I just wasted 10-15 minutes banging my head:

```shell
Checker options: filename=EWD840, init=, next=, inv=              I@09:21:26.881
Tuning:                                                           I@09:21:27.338
PASS #0: SanyParser                                               I@09:21:27.339
Parsing file /home/markus/src/TLA/_specs/models/tutorials/ewd840/EWD840.tla
Parsing file /tmp/Naturals.tla
PASS #1: TypeCheckerSnowcat                                       I@09:21:27.976
 > Running Snowcat .::.                                           I@09:21:27.976
[EWD840.tla:9:14-9:14]: Undefined name N. Introduce a type annotation. E@09:21:28.124
 > Snowcat asks you to fix the types. Meow.                       I@09:21:28.124
Checker has found an error                                        I@09:21:28.124
It took me 0 days  0 hours  0 min  1 sec                          I@09:21:28.125
Total time: 1.247 sec                                             I@09:21:28.125
EXITCODE: OK
```

 ```tla
------------------------------ MODULE EWD840 -------------------------------
EXTENDS Naturals

\* @type: Int;
CONSTANT  N

ASSUME NAssumption == N \in (Nat \ {0})

Node == 0 .. N-1

black == "black"
white == "white"
Color == {black, white}

...
```

It would be nice if snowcat would warn about the unmatched type annotation on line four.

**Comment by lemmy** [comment]
Btw. the error message should probably quote the identifier to avoid warnings such as the one below (for a spec with a variable `name`):

```
[EWD840.tla:9:14-9:14]: Undefined name name. Introduce a type annotation. E@09:21:28.124
```

konnov:
> It would be nice if snowcat would warn about the unmatched type annotation on line four.

Do you mean that Snowcat should complain as soon as it meets a constant or a variable without an annotation? That makes sense to me.

lemmy:
Maybe I'm mistaken, but a while ago you mention upcoming type inference that would handle the basic if not most constructs.  Have you dropped that goal?

konnov:
We still have a type inference prototype by Jure. However, it requires more work to: decode type errors from SMT and compute principle types (the most general ones). The type checker does not have these problems, as the user has to help it.

[...]

PR #735, opened 2021-04-13
Closes 705: error on missing annotations

konnov:
This PR closes issue #705. The type checker complains about a missing annotation for a constant or a variable, instead of showing an error later.

### DIFF
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
@@ -23,13 +23,16 @@ class ToEtcExpr(annotationStore: AnnotationStore, aliasSubstitution: ConstSubsti
 
   def apply(decl: TlaDecl, inScopeEx: EtcExpr): EtcExpr = {
     decl match {
-      case d: TlaConstDecl =>
-        // CONSTANT N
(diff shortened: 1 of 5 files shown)


tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -23,13 +23,16 @@ class ToEtcExpr(annotationStore: AnnotationStore, aliasSubstitution: ConstSubsti
...
-        // VARIABLE x
-        findTypeFromTagOrAnnotation(d).map(tt => mkTypeDecl(ExactRef(d.ID), d.name, tt, inScopeEx)).getOrElse(inScopeEx)
+      case TlaConstDecl(_) | TlaVarDecl(_) =>
+        // CONSTANT N or VARIABLE x
+        findTypeFromTagOrAnnotation(decl) match {
+          case Some(tt) =>
+            mkTypeDecl(ExactRef(decl.ID), decl.name, tt, inScopeEx)
+
+          case None =>
+            val kind = if (decl.isInstanceOf[TlaConstDecl]) "CONSTANT" else "VARIABLE"
```
shonfeder, line 33: Is there a `toString` function or `name` attribute we can use here instead?

konnov: The name is also printed (below in the code). The standard `toString` of the case classes is not very TLAish.

PR #741, opened 2021-04-15
RFC006: proposal on unit-testing and property-based testing for TLA+

konnov:
This is an RFC to discuss unit testing and PBT for TLA+. Once we agreed on that internally, we should get the feedback from a wider audience.

**Compiled version in [pdf](https://github.com/informalsystems/apalache/blob/4a2474f49026aa3941844d814966b13936599ef3/docs/src/adr/006rfc.pdf).**

To see the code in the RFC, run the following:

```sh
git clone https://github.com/informalsystems/apalache -b igor/rfc-unit && cargo install mdbook \
  && cd apalache/docs && mdbook serve
```


docs/src/adr/006rfc-unit-testing.md
```diff
@@ -0,0 +1,242 @@
...
+
+```tla
+{{#include ../../../test/tla/ChangRobertsTyped_Test.tla:91:101}}
+```
+
+The operator `Action_n0` carries several annotations:
+
+ - The annotation `@testAction` indicates that `Action_n0` should be tested
+    as an action that is an operator over unprimed and primed variable.
+ - The annotation `@requre("Prepare_n0")` tells the framework that
```
vitorenesduarte, line 197:
```suggestion
 - The annotation `@require("Prepare_n0")` tells the framework that
```

docs/src/adr/006rfc-unit-testing.md
```diff
@@ -0,0 +1,242 @@
...
+
+It is often nice to see examples of test inputs that pass the
+test. Apalache has all the ingredients to do that that. We should be able
+to run a command like that:
+
+```sh
+apalache example --cinit=ConstInit --include=TestAction_n0 ChangRobertsTyped_Test.tla
+```
+
+The above call would produce `exampla.tla`, a TLA+ description of two states
```
vitorenesduarte, line 227:
```suggestion
The above call would produce `example.tla`, a TLA+ description of two states
```

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,103 @@
...
+    /\ TypeOK
+
+\* Assertion that we expect to hold true after firing Action_n0.
+Assert_n0 ==
+    \E n, m \in Node:
+        msgs'[n] = msgs[n] \union {m}
+
+\* Execute the action under test.
+\* Note that we decouple Assert_n0 from Action_n0.
+\* The reason is that we always assume that Action_n0 always holds,
```
vitorenesduarte, line 93:
```suggestion
\* Note that we decouple Assert_n0 from TestAction_n0.
\* The reason is that we always assume that TestAction_n0 always holds,
```

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,103 @@
...
+    /\ Assumptions
+    \* let the solver pick some data structures within the bounds
+    \* up to 15 messages
+    /\ msgs = Gen(3 * MAX_N)
+    /\ pc = Gen(MAX_N)
+    /\ initiator = Gen(MAX_N)
+    /\ state = Gen(MAX_N)
+    \* restrict the contents with TypeOK,
+    \* so we don't generate useless data structures
+    /\ TypeOK
```
vitorenesduarte, line 84: Would it make sense to have a way to define generators that are useful and then invoke them here?

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,103 @@
...
+
+InitAndAssumptions ==
+    Init /\ Assumptions
+
+(********************** STATELESS OPERATOR TESTS ******************************)
+
+\* Note that succ(n) is not referring to state variables,
+\* so we can test it in isolation.
+\*
+\* @testStateless
```
vitorenesduarte, line 53: Could we go with `@test` instead of `@testStateless`? Or do we predict that `@test` could be used in the future for something else?

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,103 @@
+--------------------- MODULE ChangRobertsTyped_Test -------------------------
+(*
+ * A test setup for ChangRobertsTyped.
+ *)
+EXTENDS Integers, Apalache
```
vitorenesduarte, line 5: Why do we need to import `Apalache`?

vitorenesduarte (commented):
It's great to see this! I'll do another pass later, but here are some initial comments. I really like the idea of `apalache-mc example`!!

test/tla/ChangRobertsTyped_Test.tla
konnov: I don't know. I think this is actually a rare case, when we have a stateless operator. When I was writing this text, it was not even clear to me, why cannot we directly tag `succ` with `@require` and `@ensure`.

test/tla/ChangRobertsTyped_Test.tla
konnov: We are using `Gen`, which is defined in `Apalache`. If you are not using `Gen`, you would not need it.

test/tla/ChangRobertsTyped_Test.tla
konnov: Can you elaborate? Do you want to have generators in `TypeOK`?

konnov:
I have added a section on testing executions

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,125 @@
...
+\*
+\* @require("ConstInit")
+\* @require("Prepare_n0")
+\* @invariant("Correctness")
+\* @ensure("Assert_noWinner")
+\* @testExecution(5)
+TestExec_n0_n1 ==
+    \* in this test, we only execute actions by processes 1 and 2
+    \E self \in { 1, 2 }:
+        n0(self) \/ n1(self)
```
vitorenesduarte, line 123: I really like this idea!

I wonder if it would be possible to somehow write the following:
```tla
IfExec(trace) ==
   /\ trace[0].action.name = "create_client"
   /\ trace[0].action.id = 1
   /\ trace[0].action.height = 3
   /\ trace[1].action.name = "update_client"
   /\ trace[1].action.id = 1
   /\ trace[1].action.height = 2

ThenExec(trace) ==
   /\ trace[0].outcome = "ok" 
   /\ trace[1].outcome = "invalid_height"

\* @require("IfExec")
\* @ensure("ThenExec")
\* @testExecution(2, "Init", "Next")
\* the execution doesn't have to start at `Init`;
\* instead, the execution can start from any state reachable from `Init` by applying `Next`
```

I don't think I'm using `@require` and `@ensure` correctly (i.e. as in the other examples), but I believe that it conveys the idea.

If something like this was supported, it could probably also be used to generate MBT tests in a much nicer way.

konnov: Do you like to refer to the intermediate states in a trace? The idea was the you can talk about the variables of the first state and the last state. Otherwise, we are quickly approaching temporal logic :)

konnov: @andrey-kuprianov what is your take on that? How far is it from the tests you need in MBT?

vitorenesduarte: > Do you like to refer to the intermediate states in a trace?

Yes. My example is bad in that sense. I should have used a trace with more than 2 states.

docs/src/adr/006rfc-unit-testing.md
```diff
@@ -0,0 +1,300 @@
...
+```sh
+apalache example --include=TestExec_n0_n1 ChangRobertsTyped_Test.tla
+```
+
+
+[Unit testing]: https://en.wikipedia.org/wiki/Unit_testing
+[Property-based testing]: https://en.wikipedia.org/wiki/QuickCheck
+[TLA+ examples]: https://github.com/tlaplus/Examples/
+[LCR]: https://github.com/tlaplus/Examples/tree/master/specifications/chang_roberts
+[ChangRobertsTyped_Test.tla]: ../../../test/tla/ChangRobertsTyped_Test.tla
```
shonfeder, line 293: I think this should be a link to the file hosted by github, otherwise the link will be broken when this is served from the generated site.

test/tla/ChangRobertsTyped_Test.tla
vitorenesduarte: In the MBT for IBC, we have a single module, [IBC](https://github.com/informalsystems/ibc-rs/blob/master/modules/tests/support/model_based/IBC.tla), changing the state.
The remaining modules ([ICS02](https://github.com/informalsystems/ibc-rs/blob/master/modules/tests/support/model_based/ICS02.tla) and [ICS03](https://github.com/informalsystems/ibc-rs/blob/master/modules/tests/support/model_based/ICS03.tla)) only contain stateless operators that are then invoked by the IBC one.

I don't know if this is a common idiom, however.

shonfeder (commented):
Exciting stuff!  Our discussion with @istoilkovska today seems to really confirm the utility of this stuff.

I posed some questions, just to help us consider different angles.

test/tla/ChangRobertsTyped_Test.tla
konnov: I think @shonfeder and you are asking about more or less the same thing. I don't like the idea of having `trace` explicitly. I guess what you are asking for is a temporal property. Instead of giving an explicit index, we could write `A /\ []B /\ <>C`. This would be the syntax for advanced users :-)

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,125 @@
...
+        state[n] /= "won"
+
+(*************************** EXECUTION TESTS **********************************)
+\* Execute a sequence of 5 actions, similar to TestAction_n0.
+\* We test a final state with Assert_n0.
+\* Additionally, every state in an execution is tested for Correctness.
+\*
+\* @require("ConstInit")
+\* @require("Prepare_n0")
+\* @invariant("Correctness")
```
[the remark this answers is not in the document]
konnov: I will add a discussion on that in the document. The problem is that all of these annotations can be written directly in TLA by writing temporal formulas. However, this often leads to confusion. Moreover, people start writing very powerful formulas right away. What we need in a test: (1) How to initialize the system, (2) what to execute, (3) what to check. Decomposing a formula into these three components is surprisingly hard, unless people are following a very rigid syntax, which nobody wants to follow.

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,125 @@
...
+\* @ensure("Assert_noWinner")
+\* @testExecution(5)
+TestExec_n0_n1 ==
+    \* in this test, we only execute actions by processes 1 and 2
+    \E self \in { 1, 2 }:
```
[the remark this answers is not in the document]
konnov: `Gen` is a bit of a hack for testing. I just realized that it is not explained in the document at all.

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,125 @@
...
+\* The reason is that we always assume that TestAction_n0 always holds,
+\* whereas we may want to see Assert_n0 violated.
+\*
+\* @require("ConstInit")
+\* @require("Prepare_n0")
+\* @ensure("Assert_n0")
+\* @testAction
+TestAction_n0 ==
+    \E self \in Node:
+        n0(self)
```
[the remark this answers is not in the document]
konnov: > If it is essential, we should maybe work out a coherent syntax for the annotations. E.g., the annotations here don't really fit with the type annotations: here we're missing a trailing ; and we use a quoted argument, rather than a colon an unquoted term.

We have two kinds of syntax for single-argument string annotations: https://apalache.informal.systems/docs/adr/004adr-annotations.html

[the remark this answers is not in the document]
konnov: > I ask because, imo, it's generally not great when a parallel system of annotations starts to emerge, and preferable to express things within the language as much as possible. Of course, if it's really essential, then worth pursuing.

I know what you mean. Several times we tried to express a DSL in TLA+ itself, and it happened to be a disaster. I think there are two reasons for that: (1) TLA+ is not an imperative language, (2) formulas have a very flexible structure, so it is tempting to combine things in unexpected ways.

Another reason for having `@require` and `@ensure` in annotations is that you can see them like that in contracts for verification tools, e.g., see Dafny, Ivy, VeriFast. So I am just copying what works well :-)

test/tla/ChangRobertsTyped_Test.tla
```diff
@@ -0,0 +1,125 @@
...
+(*************************** EXECUTION TESTS **********************************)
+\* Execute a sequence of 5 actions, similar to TestAction_n0.
+\* We test a final state with Assert_n0.
+\* Additionally, every state in an execution is tested for Correctness.
+\*
+\* @require("ConstInit")
+\* @require("Prepare_n0")
+\* @invariant("Correctness")
+\* @ensure("Assert_noWinner")
+\* @testExecution(5)
```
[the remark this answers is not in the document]
konnov: True, annotations are another layer. Some languages support contracts right in the language, e.g., Scala has `requires` as part of the language. Maybe one day TLA+ will have it too, but for the moment annotations would work too.

test/tla/ChangRobertsTyped_Test.tla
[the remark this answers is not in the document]
shonfeder: I see! Thanks for explaining. This seems well motivated to me.

I'd be in favor of adding a section to the RFC to record the motivation for putting this in annotations instead of as a DSL.

test/tla/ChangRobertsTyped_Test.tla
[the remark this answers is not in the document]
shonfeder: I think this reply may have meant for a different comment?

test/tla/ChangRobertsTyped_Test.tla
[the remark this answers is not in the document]
shonfeder: This is a good reason. Let's do record these motivations, along with https://github.com/informalsystems/apalache/pull/741#discussion_r615377671 into the RFC>

konnov:
I have pushed more explanations. The discussions section is not finished yet. Also, added a pdf version of the file, so it is easy to read

konnov:
Added discussions on the issues raised by @shonfeder and I guess also by @vitorenesduarte

docs/src/adr/006rfc-unit-testing.md
```diff
@@ -0,0 +1,607 @@
...
+
+As you can see, we clearly decompose a test in three parts:
+
+ - preparing the states (like a small version of `Init`),
+ - executing the action (like a small version of `Next`),
+ - testing the next states against the previous states (like an action invariant).
+
+In the rest of this section, we comment on the alternative approaches.
+
+#### 3.5.1. But I can do all of that in TLA+
```
shonfeder, line 323: These sections make the particular feature set and approach very well motivated, IMO. Thanks for adding them!

shonfeder (approved):

docs/src/adr/006rfc-unit-testing.md
konnov: I was actually surprised how hard it was to express such a simple test in a dedicated operator, instead of creating a test in a separate file.

test/tla/ChangRobertsTyped_Test.tla
konnov: It's reasonable to expect that people would write larger specs like this. So far, I have not seen it often. It may also be a difference between people who come from imperative and functional languages.

docs/src/adr/006rfc-unit-testing.md
konnov: Good point. I will have to update the link, once this PR is merged.

konnov:
@andrey-kuprianov do you like to comment on this RFC? My plan was to post it on reddit, to see if TLA+ users would like to see a testing framework like that.

Isaac-DeFrain:
@konnov you'll probably have better luck emailing the google group, I have not found many TLA+ people on Reddit

konnov:
@Isaac-DeFrain, yeah, the tla mailing list too. I just want to do it in stages, so we don't get a storm of replies :)

test/tla/ChangRobertsTyped_Test.tla
lemmy: Related: `Trace` operator in CommunityModules https://github.com/tlaplus/CommunityModules/blob/4337afb74653cedcc2a24107759bf1613812db34/modules/TLCExt.tla#L42-L56

konnov: Yep. I don't like the idea of exposing the trace explicitly. Once you do that, the users will write arbitrary code involving traces, which makes it much harder to analyze for symbolic techniques.

konnov: I am also not sure what to do about the primed and unprimed variables in this test. It would be nice to make it less of a hack.

lemmy: > Yep. I don't like the idea of exposing the trace explicitly. Once you do that, the users will write arbitrary code involving traces, which makes it much harder to analyze for symbolic techniques.

Power to the ~People~ tools. ✊

konnov:
Let's continue the discussion in #817

PR #742, opened 2021-04-16
Doctests and removal of frozendict

Alexander-N:
Converted the rest of the Python examples.

I was also wondering if we should switch set operations like 
```python
{0, 1, 2} & {1, 2, 3}
```
to its equivalents which are closer to TLA+ and maybe a bit easier to read: 
```python
# TLA: {0, 1, 2} \intersect {1, 2, 3}
{0, 1, 2}.intersection({1, 2, 3})
```

What do you think?


shonfeder:
I'm for the proposed change to make the relation between the TLA and Python more explicit :+1

docs/src/lang/records.md
```diff
@@ -69,9 +69,6 @@ and `[type |-> "B", b |-> TRUE]` have the joint type:
 
 ## Operators
 
-In the Python examples, we are using the package [frozendict], to produce an
```
shonfeder, line 72: I think we should still use frozendict where appropriate. And it is needed in parts of this section.

docs/src/lang/records.md
```diff
@@ -148,17 +145,18 @@ is picked with `\E r \in [ field_1: S_1, ..., field_n: S_n]`.
...
-**Example in Python:** TLA+ functions are immutable, so we are using [frozendict]:
+**Example in Python:** We use a list instead of a set, since Python dictionaries are not hashable.
 
 ```python
-  frozenset({ frozendict({ "name": n, "port": p })
-                for n in { "A", "B", "C" } for p in range(1, 65535 + 1) })
+>>> [ { "name": n, "port": p } for n in [ "A", "B", "C" ] for p in range(65533, 65535 + 1) ]
+[{'name': 'A', 'port': 65533}, {'name': 'A', 'port': 65534}, {'name': 'A', 'port': 65535}, {'name': 'B', 'port': 65533}, {'name': 'B', 'port': 65534}, {'name': 'B', 'port': 65535}, {'name': 'C', 'port': 65533}, {'name': 'C', 'port': 65534}, {'name': 'C', 'port': 65535}]
+
 ```
```
shonfeder, line 160: Let's just use frozen dict where it's needed (as in this case). Where we can omit frozen dict for simplicity, it is worth the gains for clearer exposition, but we shouldn't use different, incorrect data structures to get around using the correct data structure.

docs/src/lang/tuples.md
```diff
@@ -158,21 +158,23 @@ is picked with `\E t \in S_1 \X S_2`.
...
-  # the pythonic way: a set of python tuples (indexed with 0, 1, ...)
-  frozenset({ (n, p)
-                for n in { "A", "B", "C" } for p in range(1, 65535 + 1) })
-  # the TLA+ way
-  frozenset({ frozendict({ 1: n, 2: p })
-                for n in { "A", "B", "C" } for p in range(1, 65535 + 1) })
+>>> # the pythonic way: a set of Python tuples (indexed with 0, 1, ...)
+>>> tuples = { (n, p) for n in { "A", "B", "C" } for p in range(65533, 65535 + 1) }
+>>> tuples == {('C', 65533), ('B', 65535), ('A', 65533), ('A', 65534), ('B', 65534), ('C', 65535), ('A', 65535), ('C', 65534), ('B', 65533)}
+True
```
shonfeder, line 173: Why switch to assigning the value to `tuples` here and testing for equality, instead of just using the output as in all other cases?

docs/src/lang/tuples.md
```diff
@@ -158,21 +158,23 @@ is picked with `\E t \in S_1 \X S_2`.
...
+>>> # the TLA+ way, but with a list instead of a set, since dictionaries are not hashable
+>>> [ { 1: n, 2: p } for n in [ "A", "B", "C" ] for p in range(65533, 65535 + 1) ]
+[{1: 'A', 2: 65533}, {1: 'A', 2: 65534}, {1: 'A', 2: 65535}, {1: 'B', 2: 65533}, {1: 'B', 2: 65534}, {1: 'B', 2: 65535}, {1: 'C', 2: 65533}, {1: 'C', 2: 65534}, {1: 'C', 2: 65535}]
+
```
shonfeder, line 177: Again, let's just use the right data structure here, and not get rid of frozen.

docs/src/lang/tuples.md
```diff
@@ -188,7 +190,6 @@ a_n>>, Int) => a_i`, for some types `a_1, ..., a_n`.
 
 [Control Flow and Non-determinism]: ./control-and-nondeterminism.md
 [Specifying Systems]: http://lamport.azurewebsites.net/tla/book.html?back-link=learning.html
-[frozendict]: https://pypi.org/project/frozendict/
```
shonfeder, line 191: This should stay.

docs/src/lang/records.md
```diff
@@ -148,17 +145,18 @@ is picked with `\E r \in [ field_1: S_1, ..., field_n: S_n]`.
...
-  [ name: { "A", "B", "C" }, port: 1..65535 ]
+  [ name: { "A", "B", "C" }, port: 65533..65535 ]
     \* A set of records. Each has two fields:
     \* field "name" that has the value from the set { "A", "B", "C" }, and
-    \* field "port" that has the value from the set 1..65535.
+    \* field "port" that has the value from the set 65533..65535.
 ```
 
-**Example in Python:** TLA+ functions are immutable, so we are using [frozendict]:
+**Example in Python:** We use a list instead of a set, since Python dictionaries are not hashable.
```
shonfeder, line 154: This can be changed to saying we need `frozendict` because of the need for hashability.

shonfeder (changes requested):
A couple small changes requested. Thanks for pushing this change all the way through!

docs/src/lang/records.md
Alexander-N: I think we should only use data structures which are part of the language. frozendict is an external package which does not work with recent Python versions.

docs/src/lang/tuples.md
Alexander-N: Because the order is not defined in sets and the doctest will randomly fail since the output is not fixed.

docs/src/lang/records.md
Alexander-N: I should have put this information in the PR message, it was a bit hidden in the commit message. Maybe there are ways to get a more faithful analogy within standard Python but personally I don't see the need to make this more complicated. There is surely some other package which could be used instead of frozendict, but I feel strongly that this should be avoided.

konnov:
I think we should decide, whether we want to make examples native to python, as Alexander is suggesting. The downside of this approach is that we cannot freely compose expressions (e.g., sets of functions), as we do in TLA+.

docs/src/lang/tuples.md
shonfeder: Ah yes! This is a good solution. Thanks :)

konnov:
I believe that this discussion (as well as the previous ones) show that python dictionaries are not the data structure that could adequately represent TLA+ data functions (and records). Actually, I could not see all these issues, when I was writing the manual. I see two possible directions:

 1. Make the code even more pythonic. That is the approach proposed in this PR. However, we start losing the connection with the original TLA+ data structures. For instance, there is a conceptual difference between lists and sets.
 2. Choosing another data structure. As the examples show we need a hashable (or even an immutable) data structure.

I would prefer option 2, even if the produced code is less optimal than it could be. After all, we want to see the relation between TLA+ and a programming language, not write the most optimal program. Why cannot we use [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple)? It is a standard collection in python 3 and seems to be immutable.

docs/src/lang/records.md
shonfeder: I agree with igor's remarks here: https://github.com/informalsystems/apalache/pull/742#issuecomment-822511588

If you'd want to change the use of dictionaries to instead use namedtuples consistently, and add the additional explanatory passages needed to help motivate and explicate this for readers, we'd accept that in this PR.

Otherwise, let's just stick with `frozendicts` here.

We all agree that using a totally different data structure (lists instead of sets) is not compatible with our aims for this documentation, and we don't mind the dependency on the frozendict library if it allows a more accurate encoding.

Let me know if you  have any questions!

Alexander-N:
My goal would not be to make the examples pythonic, but to keep them simple. Of course the analogy should be as close as possible, but what good is an example if it's more complicated than the TLA+ code it should explain?

Let's have a look on one specific example:
```python
>>> # the TLA+ way, but with a list instead of a set, since dictionaries are not hashable
>>> [ { 1: n, 2: p } for n in [ "A", "B", "C" ] for p in range(65533, 65535 + 1) ]
```

How would you use namedtuples here, I fail to see a way which is not super weird :-). e.g.
```python
>>> from collections import namedtuple
>>> t = namedtuple("TlaTuple", ['i1', 'i2'])                                
>>> { t(n,p) for n in [ "A", "B", "C" ] for p in range(65533, 65535 + 1)} 
{TlaTuple(i1='A', i2=65533),
 TlaTuple(i1='A', i2=65534),
 TlaTuple(i1='A', i2=65535),
 TlaTuple(i1='B', i2=65533),
 TlaTuple(i1='B', i2=65534),
 TlaTuple(i1='B', i2=65535),
 TlaTuple(i1='C', i2=65533),
 TlaTuple(i1='C', i2=65534),
 TlaTuple(i1='C', i2=65535)}

```

Of course I can revert the commit and leave the frozendict, but it's broken from Python 3.8 on, making the doctests a bit pointless.

docs/src/lang/records.md
Alexander-N: The change is now unrelated to frozendict.

shonfeder:
> what good is an example if it's more complicated than the TLA+ code it should explain

Our hypothesis is that it can be useful for people who understand Python to see how you'd encode *the same* data structures and properties in the language they are familiar with. This *doesn't* mean that the python will always be (or must always be) simpler. In cases like this, where python dictionaries don't really work, I think it is actually very valuable to draw attention to and represent this difference. This helps drive home the fact that specifying systems using temporal logic propositions over sets and immutable records is actually quite different than imperative programming with mutable records etc. By translating into the accurate, but slightly more complicated Python, we are able to show both the similarity and the differences in a way that is instructive.

> How would you use namedtuples here, I fail to see a way which is not super weird :-). e.g.

I agree that this probably isn't a great place for named tuples. Tho I do think your example with a named tuple makes perfect sense. Perhaps a natural alternative for this case would be to use use a normal Python tuple?

I'm happy with the changes you've added here tho! We can think over whether it is worth using named tuples in some places in the future.

Thank you for the contribution!

PR #767, opened 2021-04-26
Adding exit codes in Tool

### LINKED ISSUE #249: [FEATURE] Return a non-zero exit code when model checking fails
Currently users have to inspect stdout in order to determine whether `check` succeeds or fails. It would help with command-line usability to also signal this result with a meaningful error code.

konnov:
Closes #249. This PR introduces non-zero exit codes upon a counterexample or error. In contrast to TLC, I prefer staying lightweight on exit codes. For the moment, we have only three exit codes:
 - 0 for no error,
 - 12 for any kind of a counterexample found by the model checker, and
 - 255 for any kind of an error, including a type error.

These error codes are compatible with TLC, but they are not as detailed as the error codes of TLC.


mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/ExitCodes.scala
```diff
@@ -0,0 +1,29 @@
...
+object ExitCodes {
+
+  /**
+   * This exit code should be used when the tool is reporting no error in a specification.
+   */
+  val NO_ERROR: Int = EC.ExitStatus.SUCCESS
+
+  /**
+   * This exit code should be used when the model checker has found a counterexample.
+   * We do not go in distinction between different kinds of counterexamples and always report VIOLATION_SAFETY.
```
shonfeder, line 21:
```suggestion
   * We do not distinguish between different kinds of counterexamples and always report VIOLATION_SAFETY.
```

shonfeder (approved):

Issue #373, opened 2020-12-10
[FEATURE] Add test coverage metrics to CI


shonfeder:

thpani:
We now have coverage via codecov:

https://github.com/informalsystems/apalache/blob/9d7a7604477aeadf118fd2e53f4009d5f6d39564/.github/workflows/main.yml#L103-L109

PR #770, opened 2021-04-27
remove scoverage from maven builds

konnov:
I think I have enough watching maven building things 3 times to get scoverage reports, which we have not been reading for years. We should integrate coverage into CI, which is planned for #373, but we should not impede builds on user machines.

Here is what I have on my machine:

```
# with scoverage
mvn clean && mvn package  -Dmaven.test.skip=true
# 4 min 25 sec
# without scoverage
mvn clean && mvn package  -Dmaven.test.skip=true
# 2 min 03 sec
```

Looks like a good improvement :P

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,3 +13,7 @@
 ### Bug fixes
 
  * The command `config --enable-stats=true` creates `$HOME/.tlaplus` if needed, see #762
+
+### Changes
+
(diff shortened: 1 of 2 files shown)


shonfeder (approved):
YAY

konnov:
That was a fast response. I hear a lot of pain :D

PR #791, opened 2021-04-30
Allow supplying the JVM args when running Apalache

### LINKED ISSUE #790: [FEATURE] Parse more apalache command line args
**Is your feature request related to a problem? Please describe.**
I'm encountering `java.lang.OutOfMemoryError: Java heap space` errors when I check larger models (~10 nodes in an asynchronous message passing system; I'm not sure what the state space size is, but it's certainly astronimical). I am only checking `--length=1` to verify an inductive invariant.

**Describe the solution you'd like**
It would be great to be able to pass the `JVM_ARGS` a different value from the command line when doing `apalache check` and have it default to the current value `-Xmx4096m` when no arg is passed.

This is the only arg that I have an immediate use for, but I'm sure someone can think of others they would like.

**Describe alternatives you've considered**
Building from source and making the change manually, but I'd like to continue using the docker image and have more flexibility than that.

shonfeder:
Closes #790

This change alters the runner script to allow users to tune the JVM when running
Apalache. The runner script now appends its args whatever is already in the
JVM_ARGS environment variable. If a max heapsize is set by the user, it lets
that take precedence. 

-------

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,12 +12,12 @@
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
 ### Documentation
 
- * RFC006 on unit testing: see #741
+* RFC006 on unit testing: see #741
 
 ### Features
 
- * apalache quits with a non-zero exit code on counterexample or error, see #249
- * type checker: supporting one-line comments in types, see #773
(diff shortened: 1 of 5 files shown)


konnov (approved):
Looks great!

PR #803, opened 2021-05-04
Add missing colons in ADR2

shonfeder:
Added two missing colons in type annotations.


konnov (approved):
Ugh. I thought I fixed that before, after talking to Ilina. Apparently, I did not :)

PR #807, opened 2021-05-04
Use the latest openjdk version for the runtime

shonfeder:
Prior to this change were were only using JDK 9 for the runtime. Afaict, there's no reason for us to so such an old version.


konnov (approved):
Yes. Good idea!

PR #809, opened 2021-05-04
Remove homebrew update

shonfeder:
This should be resolved in our github action setup now, and removing
the update will shave 50s or so off our macos builds.


konnov (approved):
Nice!

Issue #810, opened 2021-05-06
[FEATURE] Prototype of RFC006


konnov:
Implement a prototype of the approach to unit testing that is described in https://apalache.informal.systems/docs/adr/006rfc-unit-testing.html.

This feature is a refinement of #191

konnov:
Not relevant anymore

PR #811, opened 2021-05-06
Level finder

konnov:
This PR implements computation of TLA+ levels (that are described in Specifying Systems). In a nutshell, a TLA+ expression is assigned one of the four levels:

 - Level 0: the expression is only using constants.
 - Level 1: the expression is using constants, state variables, but no action or temporal operators.
 - Level 2: the expression is using action operators, but no temporal operators.
 - Level 3: the expression using temporal operators.

The concept of levels may be reused in other code. That's why it is placed in tlair. The code in this PR is required in #810.

### DIFF
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
@@ -1,0 +1,55 @@
+package at.forsyte.apalache.tla.lir
+
+/**
+ * This class computes the level of a TLA+ declaration by using TlaExLevelFinder. See Lamport. Specifying Systems, p. 322.
+ *
+ * @param module a TLA module that contains the declarations to query
+ * @author Igor Konnov
+ */
+class TlaDeclLevelFinder(module: TlaModule) {
+  private val consts = Set(module.constDeclarations.map(_.name): _*)
+  private val vars = Set(module.varDeclarations.map(_.name): _*)
+  private val defs = Map(module.operDeclarations map { d => d.name -> d }: _*)
+  private var levelCacheMap: Map[String, TlaLevel] = Map()
+
+  def apply(decl: TlaDecl): TlaLevel = {
+    decl match {
+      case TlaConstDecl(_) =>
+        TlaLevelConst
+
+      case TlaVarDecl(_) =>
+        TlaLevelState
+
+      case TlaAssumeDecl(_) =>
+        TlaLevelConst
+
+      case TlaOperDecl(name, _, _) =>
+        cacheLevel(Set(name), name)
+        levelCacheMap(name)
+
+      case TlaTheoremDecl(_, _) =>
+        TlaLevelConst
+    }
+  }
+
+  private def cacheLevel(callers: Set[String], name: String): Unit = {
+    def levelOfName(name: String): TlaLevel = {
+      if (consts.contains(name)) {
+        TlaLevelConst
+      } else if (vars.contains(name)) {
+        TlaLevelState
+      } else {
+        if (!callers.contains(name) && defs.contains(name)) {
+          // as the module comes from the parser, we assume that defs contains a definition for the name `name`
+          cacheLevel(callers + name, name)
+          levelCacheMap(name)
+        } else {
+          TlaLevelConst
+        }
+      }
+    }
+
+    val level = new TlaExLevelFinder(levelOfName)(defs(name).body)
+    levelCacheMap += name -> level
+  }
+}
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
@@ -32,7 +32,7 @@ class TlaDeclLevelFinder(module: TlaModule) {
     }
   }
 
-  private def cacheLevel(callers: Set[String], name: String): Unit = {
+  private def cacheLevel(callers: Set[String], name: String): TlaLevel = {
     def levelOfName(name: String): TlaLevel = {
       if (consts.contains(name)) {
         TlaLevelConst
@@ -41,8 +41,11 @@ class TlaDeclLevelFinder(module: TlaModule) {
       } else {
         if (!callers.contains(name) && defs.contains(name)) {
           // as the module comes from the parser, we assume that defs contains a definition for the name `name`
-          cacheLevel(callers + name, name)
-          levelCacheMap(name)
+          levelCacheMap.get(name) match {
+            case Some(level) => level
+
+            case None => cacheLevel(callers + name, name)
+          }
         } else {
           TlaLevelConst
         }
@@ -51,5 +54,6 @@ class TlaDeclLevelFinder(module: TlaModule) {
 
     val level = new TlaExLevelFinder(levelOfName)(defs(name).body)
     levelCacheMap += name -> level
+    level
   }
 }
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaExLevelFinder.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaExLevelFinder.scala
@@ -15,20 +15,9 @@ class TlaExLevelFinder(nameLevel: String => TlaLevel) {
     case NameEx(name) =>
       nameLevel(name)
 
-    case OperEx(op, args @ _*)
-        if op == TlaActionOper.prime || op == TlaActionOper.enabled
-          || op == TlaActionOper.unchanged || op == TlaActionOper.stutter
-          || op == TlaActionOper.nostutter || op == TlaActionOper.composition =>
-      TlaLevelAction.join(args.map(find))
-
-    case OperEx(op, args @ _*)
-        if op == TlaTempOper.box || op == TlaTempOper.diamond || op == TlaTempOper.leadsTo
-          || op == TlaTempOper.weakFairness || op == TlaTempOper.strongFairness
-          || op == TlaTempOper.guarantees || op == TlaTempOper.AA || op == TlaTempOper.EE =>
-      TlaLevelTemporal.join(args.map(find))
-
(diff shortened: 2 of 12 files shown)


tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaExLevelFinder.scala
```diff
@@ -0,0 +1,39 @@
...
+      TlaLevelAction.join(args.map(find))
+
+    case OperEx(op, args @ _*)
+        if op == TlaTempOper.box || op == TlaTempOper.diamond || op == TlaTempOper.leadsTo
+          || op == TlaTempOper.weakFairness || op == TlaTempOper.strongFairness
+          || op == TlaTempOper.guarantees || op == TlaTempOper.AA || op == TlaTempOper.EE =>
+      TlaLevelTemporal.join(args.map(find))
+
+    case OperEx(_, args @ _*) =>
+      TlaLevelConst.join(args.map(find))
```
shonfeder, line 31: It could just be a matter of taste, but IMO this approach is a bit easier to read (assuming it's feasible):

```suggestion
case OperEx(op, args @ _*) =>
    val ctor = op match {
        case TlaActionOper.prime | TlaActionOper.enabled
            | TlaActionOper.unchanged | TlaActionOper.stutter
            | TlaActionOper.nostutter | TlaActionOper.composition => TlaLevelAction
        case TlaTempOper.box | TlaTempOper.diamond
            | TlaTempOper.leadsTo | TlaTempOper.weakFairness
            | TlaTempOper.strongFairness | TlaTempOper.guarantees
            | TlaTempOper.AA | TlaTempOper.EE => TlaLevelTemporal
        case _ => TlaLevelConst
    }
    ctor.join(args.map(find))
```

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaExLevelFinder.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    case LetInEx(body, defs @ _*) =>
+      find(body).join(defs.map(d => find(d.body)))
+
+    case _ =>
+      TlaLevelConst
```
shonfeder, line 37: Could this catchall come back to bite us if new expressions are introduced? Would it be worth while to explicitly enumerate the other constructors here instead of having a wildcard?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
```diff
@@ -0,0 +1,55 @@
...
+
+      case TlaVarDecl(_) =>
+        TlaLevelState
+
+      case TlaAssumeDecl(_) =>
+        TlaLevelConst
+
+      case TlaOperDecl(name, _, _) =>
+        cacheLevel(Set(name), name)
+        levelCacheMap(name)
```
shonfeder, line 28: Given that we are caching results, why don't we first check whether the `name` is cached before we try to add it to the cache?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
```diff
@@ -0,0 +1,55 @@
...
+    decl match {
+      case TlaConstDecl(_) =>
+        TlaLevelConst
+
+      case TlaVarDecl(_) =>
+        TlaLevelState
+
+      case TlaAssumeDecl(_) =>
+        TlaLevelConst
+
```
shonfeder, line 25: unless I missunderstand, it seems like where checking for this and then determining the level in two parallel ways, Once here by pattern matching, and once when we construct the level finder, when we cache to the various sets. Couldn't we handle all of this logic with the `_.contains` checks and the caching lookup?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevel.scala
```diff
@@ -0,0 +1,92 @@
...
+   * @param that the level to compare
+   * @return 0, if the levels are equal; a negative number if `this` is smaller than `that`, and a positive number
+   *         if `this` is larger than `that`.
+   */
+  override def compare(that: TlaLevel): Int = {
+    level - that.level
+  }
+
+  /**
+   * Join the levels by computing the smallest level that covers both of `this` and `that`.
```
shonfeder, line 30: I'm surprised that joining gives the least, rather than greatest level. Is this because its actually more difficult to handle , e.g., constants than handling temporal formulas?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevel.scala
```diff
@@ -0,0 +1,92 @@
...
+object TlaLevel {
+  private val INT_TO_LEVEL = Seq(TlaLevelConst, TlaLevelState, TlaLevelAction, TlaLevelTemporal)
+
+  def fromInt(level: Int): TlaLevel = {
+    if (level < 0 || level >= INT_TO_LEVEL.length) {
+      throw new IllegalArgumentException(s"Unexpected level of TlaValue: $level")
+    } else {
+      INT_TO_LEVEL(level)
+    }
+  }
```
shonfeder, line 59: Since we already have the `Int` values inhe `Const` objects, why not just have `fromInt` give `level.level`? It seems like this duplicates the logic: we record the int once as a position in the sequence and another time explicitly in the `level` attribute.

shonfeder (approved):
A few suggestions, but looks good to me overall.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaExLevelFinder.scala
konnov: I doubt that, as the types of expressions were stable for long time.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaExLevelFinder.scala
konnov: Good point. I think I will just introduce a map.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevel.scala
konnov: no-no, it's actually the maximum of the two levels, just badly formulated

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
konnov: yep, added that

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
konnov: The code in `apply` is also handling the cases of `ASSUME` and `THEOREM`. Also, this code avoids an extra lookup in `defs`, `consts` and `vars` (but that's not the main point).

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevel.scala
konnov: Good point. Fixed that.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
Kukovec: What happens if I write an assume declaration with primes?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevel.scala
```diff
@@ -0,0 +1,93 @@
...
+  def join(otherLevels: Seq[TlaLevel]): TlaLevel = {
+    otherLevels.foldLeft(this) { case (l, r) => l.join(r) }
+  }
+}
+
+object TlaLevel {
+  private val INT_TO_LEVEL = Seq(TlaLevelConst, TlaLevelState, TlaLevelAction, TlaLevelTemporal)
+
+  def fromInt(levelNo: Int): TlaLevel = {
+    INT_TO_LEVEL.find(_.level == levelNo) match {
```
Kukovec, line 54: Minor gripe,I would use an indexed collection and do boundary tests here, instead of find.

Kukovec (approved):
LGTM

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
konnov: SANY will tell you that it does not like your spec. I found that out a couple of months ago :D

konnov: It will even complain if you use state variables in `ASSUME`.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevel.scala
konnov: I've replaced it with find, which should be ok for a list of 4

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDeclLevelFinder.scala
Kukovec: Ok, what about the expression `(1 + 1)'` outside of assume? Just because it has a prime operator, does not mean its level isn't 0.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevel.scala
konnov: Ha-ha. My solution was exactly as you suggest, and @shonfeder did not like it :P

PR #820, opened 2021-05-10
Make target dir in release script

shonfeder:
We've had a couple release pipeline failures because this directory
didn't exist. This change ensures it does exist.


konnov (approved):
Good point!

PR #824, opened 2021-05-13
Add type information to overriding error message

### LINKED ISSUE #823: [FEATURE] Add type description to error message when overriding type doesn't match
**Is your feature request related to a problem? Please describe.**
The error message for when an overriding type doesn't match the original type currently does not give any information on how the types are different. E.g. `The types of Acceptor and OVERRIDE_Acceptor do not match.`

**Describe the solution you'd like**
The error message should ideally display the mismatched types. E.g. `The types of Acceptor (Set(ACCEPTOR)) and OVERRIDE_Acceptor (Set(Str)) do not match.`

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

klinvill:
This PR fixes #823 by adding type information into the overriding error message in the event of a type mismatch.

- [ ] ~~Tests added for any new code~~
- [ ] ~~Documentation added for any new functionality~~

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstAndDefRewriter.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstAndDefRewriter.scala
@@ -28,7 +28,16 @@ class ConstAndDefRewriter(tracker: TransformationTracker) extends TlaModuleTrans
           logger.error("  > If you need support for n-ary CONSTANTS, write a feature request.")
           throw new OverridingError(msg, overridingDef.body)
         } else if (d.typeTag != overridingDef.body.typeTag) {
-          val msg = s"The types of ${d.name} and ${overridingDef.name} do not match."
+          val dTypeString = d.typeTag match {
+            case Typed(tt) => tt.toString
+            case Untyped() => d.typeTag.toString
+          };
+          val overridingDefTypeString = overridingDef.body.typeTag match {
+            case Typed(tt) => tt.toString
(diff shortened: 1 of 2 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstAndDefRewriter.scala
```diff
@@ -28,7 +28,16 @@ class ConstAndDefRewriter(tracker: TransformationTracker) extends TlaModuleTrans
           logger.error("  > If you need support for n-ary CONSTANTS, write a feature request.")
           throw new OverridingError(msg, overridingDef.body)
         } else if (d.typeTag != overridingDef.body.typeTag) {
-          val msg = s"The types of ${d.name} and ${overridingDef.name} do not match."
+          val dTypeString = d.typeTag match {
+            case Typed(tt) => tt.toString
+            case Untyped() => d.typeTag.toString
+          };
+          val overridingDefTypeString = overridingDef.body.typeTag match {
```
klinvill, line 35: I ended up using match statements here to unwrap the types in the type tag (otherwise the output would look like `Typed(Set(Str))` which is more confusing to me than `Set(Str)`. Is there a better way of doing this? I saw that `asTlaType1()` is used elsewhere unwrap type tags, but I'm not confident that `transformDef()` will only ever be called with TlaType1 types.

konnov (approved):
Thank you for the contribution! It will definitely improve the error message.

We were planning to refactor expression printers, but that will not happen soon.

PR #833, opened 2021-05-18
Enumerating counterexamples

### LINKED ISSUE #542: Exhaustive search via transition enumeration
The ability to enumere counterexamples is crucial in such usage scenarios as model-based testing, where counterexamples are used as test cases. As the first step towards #79, we need to implement the simplest exhaustive search for counterexamples: via enumerating sequences of transitions.

* the tool should support `--continue` option;
* every new counterexample returned with `-continue` should use a different sequence of transitions leading to an invariant violation.

In case of a lot of non-determinism this can of course blow up, because it means enumerating the tree branches. But exhaustiveness on the base of transition sequences is the simplest to implement, and offers an intuitive mental picture for the end user. This strategy can later be complemented in the scope of #79 with different ones (based on the values of state variables for example).

konnov:
This PR closes #542, it implements the approach discussed in #827


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/trex/ConstrainedTransitionExecutor.scala
```diff
@@ -0,0 +1,111 @@
...
+  override def pickTransition(): Oracle = {
+    val oracle = trex.pickTransition()
+    assertPathConstraints()
+    oracle
+  }
+
+  private def assertPathOrConstraint(execution: EncodedExecution, pathConstraint: PathConstraint): Unit = {
+    if (execution.path.length >= pathConstraint.length) {
+      val constraints =
+        execution.path.tail.zip(pathConstraint.tail).map { case ((binding, _), cons) =>
```
Kukovec, line 62: A comment explaining the `.tail`s would be helpful.

Kukovec (approved):
Looks mostly fine, two general comments would be to include more (or more verbose) comments, especially for nested function defs, and to split your PRs into smaller chunks in the future.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/trex/ConstrainedTransitionExecutor.scala
konnov: Good point. I have added a comment.

Issue #846, opened 2021-06-09
[FEATURE] Make assignment finding independent of preprocessing


Kukovec:
Currently, the assignment solver relies on unrolling and inlining, more out of convenience than out of necessity.
We should make the pass more independent, in two steps:

1. Implement considerations in assignment solving for non-nullary and recursive operators
2. Potentially reorder passes

PR #847, opened 2021-06-09
Adapt assignment solving for non-nullary and recursive operators

Kukovec:
- ~[ ] Tests added for any new code~
- ~[ ] Documentation added for any new functionality~
- ~[ ] Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality~

This PR, on its own, does not add or remove visible functionality. It addresses point (1) of #846. Due to the order of passes, the generalizations cover scenarios which are currently impossible, but will be relevant in phase (2) of #846.

### DIFF
--- a/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
+++ b/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SmtFreeSymbolicTransitionExtractor.scala
@@ -23,13 +23,11 @@ class SmtFreeSymbolicTransitionExtractor(
     /** Manual assignments at such locations throw exceptions */
     case ex @ OperEx(ApalacheOper.assign, OperEx(TlaActionOper.prime, NameEx(_)), _) =>
       val locString = getLocString(ex)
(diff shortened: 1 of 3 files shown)


konnov:
Nice! It would be good to have some integration tests though.

Kukovec:
> Nice! It would be good to have some integration tests though.

That would be pointless until phase (2), because inlining and unrolling erase all the test cases where this change would differ from the current implementation.

konnov:
Makes sense. Then unit tests?

konnov (approved):
Looks good. Does it mean that we could replace `x' = e` with `x' := e` after this pass? I would prefer doing that right after running the type checker.

Kukovec:
> Looks good. Does it mean that we could replace `x' = e` with `x' := e` after this pass? I would prefer doing that right after running the type checker.

`TransitionPassImpl`, already does `:=` introduction, though I suppose you could rework the passes so that this independently happens earlier.

konnov:
> `TransitionPassImpl`, already does `:=` introduction, though I suppose you could rework the passes so that this independently happens earlier.

Sure. But `TransitionPassImpl` is run very late in the pipeline.

PR #858, opened 2021-06-14
Package ADR

### LINKED ISSUE #857: [FEATURE] Redesign and formaliize Apalache package structure
Looking back, the way package boundaries were defined during development was entirely ad-hoc, which often caused trouble with refactoring, when the desired design would have introduced cyclical package dependencies (e.g. with IO).

To this end, we should write up a design document, providing guidelines and criteria for package boundaries.

Kukovec:

Closes #857


konnov (commented):
Can you also add the source file for the architecture pic, if you did it in a diagram editor? It would be useful to do changes later.

konnov (commented):
In general, I like it. But some names are not explanatory, e.g., `tla-base` and `tla-aux`. Also, how do the level numbers corresponds to the layers in the diagram?

Kukovec:
> 
> 
> In general, I like it. But some names are not explanatory, e.g., `tla-base` and `tla-aux`. Also, how do the level numbers corresponds to the layers in the diagram?

Any names in particular you'd suggest for those? Those are easy to change.

About the level numbers, as per our Zoom discussion, I removed the level numbers from the diagram, since you can draw diagrams as general graphs, not just linked lists. The levels are one linearization of the package order (every DAG has one).

docs/src/adr/007adr-restructuring.md
```diff
@@ -0,0 +1,161 @@
...
+When introducing new classes, use the guidelines defined below to determine which package to place them in.
+
+## 1. Level Structure
+
+We define Apalache architecture in terms of enumerated levels (L0, L1, L2, etc.). Each level may only hold dependencies belonging to a lower level.
+
+Levels are split into two categories:
+
+  1. Interface core: Classes within the interface core relate to general TLA+ concepts, and are intended to be usable by 3rd party developers.
+  2. Apalalache core: Classes in the Apalache core relate to specific Apalache- or model-checking- functionality
```
shonfeder, line 17: To make clear why these are called "core" I suggest adding something like:

```suggestion
Levels are split into two categories:

  1. Interface core: Classes within the interface core relate to general TLA+ concepts, and are intended to be usable by 3rd party developers.
  2. Apalalache core: Classes in the Apalache core relate to specific Apalache- or model-checking- functionality
  
Staking the "core" strata together gives us the complete stack implementing Apalache. 
```

docs/src/adr/007adr-restructuring.md
```diff
@@ -0,0 +1,161 @@
...
+| Jure Kukovec |        1 |
+
+This ADR documents the design policies guiding the package and dependency structure of Apalache.
+When introducing new classes, use the guidelines defined below to determine which package to place them in.
+
+## 1. Level Structure
+
+We define Apalache architecture in terms of enumerated levels (L0, L1, L2, etc.). Each level may only hold dependencies belonging to a lower level.
+
+Levels are split into two categories:
```
shonfeder, line 14: Because, to my mind, "categories" suggested that we could have two parallel sets of levels, or that the categories were otherwise independent from the levels, but in fact they are just grouping of levels:

```suggestion
Levels are split into two strata:
```

docs/src/adr/007adr-restructuring.md
```diff
@@ -0,0 +1,161 @@
...
+
+  1. Interface core: Classes within the interface core relate to general TLA+ concepts, and are intended to be usable by 3rd party developers.
+  2. Apalalache core: Classes in the Apalache core relate to specific Apalache- or model-checking- functionality
+
+## 2. Interface core
+
+Together, L0-L6 make up the interface core. Notably, these levels are Apalache-agnostic. The concepts of these levels are packaged as follows:
```
shonfeder, line 21: What does it mean to "package the concepts" of these levels?

shonfeder: What is the difference between a "level" and a package?Are the "levels" just a conceptual framing? Do they correspond to anything in the structure of the code or packaging?

docs/src/adr/007adr-restructuring.md
```diff
@@ -0,0 +1,161 @@
...
+
+# L13: GUI
+
+In L13, we have implementations of the following concepts:
+
+ - Graphical interface, if/when one exists
+
+## 4. Aliases, factories and exceptions
+
+Any alias or factory belongs to the lowest possible level required to define it. For example, `type uidToExMap = Map[UID, TlaEx]` belongs to L0, since `UID` and `TlaEx` are both L0 concepts, so it should be defined in an L0 alias package, even if it is only being used in a package of a higher level.
```
shonfeder, line 152: I'm not sure about this guide line.

The point of aliases and such is to abbreviate the names of something for use *in a particular context*. This principle would seem to make that intended use disallowed.

shonfeder (approved):
LGTM, tho I have a few small suggestions, and questions.

docs/src/adr/007adr-restructuring.md
Kukovec: The way I originally approached this, levels were the linearization of the dependency graph, which should be a DAG and thus admit a topological sort. They don't really correspond to anything at the code level.

docs/src/adr/007adr-restructuring.md
Kukovec: Good point, I was thinking more of global aliases (e.g. `type TlaExTransformation = TlaEx => TlaEx`) when I wrote this bit, but you're right, it should be more clear w.r.t. local aliasing.

PR #891, opened 2021-06-28
ADR on exceptions

Kukovec:

First version of the ADR outlining the approach to exceptions.


docs/src/adr/008adr-exceptions.md
```diff
@@ -0,0 +1,122 @@
+# ADR-008: Apalache Exceptions
+
+| author     | revision |
+| ------------ | --------:|
+| Jure Kukovec |    1 |
+
+This ADR documents the various exception thrown in Apalache, and the circumstances that trigger them.
+
+## 1. User input exceptions
+Exceptions in this family are caused by incorrect input from the user. All of these exceptions should exit cleanly and should NOT report a stack trace.
```
shonfeder, line 10: Should we use [functional error handling](https://docs.scala-lang.org/overviews/scala-book/functional-error-handling.html) for these kinds of errors, or perhaps `Either` types for these kinds of errors?

This can ensure we don't every leave a non-fatal exception unhandled.

docs/src/adr/008adr-exceptions.md
```diff
@@ -0,0 +1,122 @@
...
+Exceptions thrown in response to these issues should extend 
+
+```
+UnsupportedFeatureException
+```
+
+The exceptions should report the following details:
+
+  * The location of the unsupported expression(s)
+
```
shonfeder, line 94: Seems this doesn't include user errors at the CLI level.

docs/src/adr/008adr-exceptions.md
```diff
@@ -0,0 +1,122 @@
...
+should be thrown. It should report the following details:
+
+  * The assumption being violated
+  * The pass/class/method in which the assumption is made
+
+### 2.2. Pass/Transformation-specific exceptions
+Depending on the pass/transformation, specialized exceptions may be thrown, to indicate some problem in either the pipeline, malformed input, missing or incomplete metadata or any other issue that cannot be circumvented. The exceptions should include a reasonable (concise) explanation and, whenever possible, source information for relevant expressions. 
+
+## 3. Exception explanations
+On their own, exceptions should include concise messages with all the relevant information components, outlined above. In addition to that, we should implement an advanced variant of `ExceptionAdapter`, called `ExceptionExplainer`, that is optionally enabled, if Apalache is invoked with the flag `--explain-exceptions` (alternatively, it could be on by default and disabled with a similar flag).
```
shonfeder, line 114: I suggest on by default. It's easy for more advanced users to disable things, but not so for beginners.

Also, many advanced users get on very well with more verbose exceptions.

docs/src/adr/008adr-exceptions.md
Kukovec: Sure, but that's more of an implementation detail than design, I'd say.

docs/src/adr/008adr-exceptions.md
Kukovec: Those fall under 1.2, no? `Problems with `--init/next/cinit/...`

docs/src/adr/008adr-exceptions.md
Kukovec: Makes sense.

docs/src/adr/008adr-exceptions.md
shonfeder: Is this more of an implementation detail than what is proposed in Section 3? Perhaps the design-level constraint to add here is: "We should be able to statically enforce that none of these exceptions can be left unhandled"?

docs/src/adr/008adr-exceptions.md
shonfeder: Ah, I missed that. I guess it does fall under inputs.

docs/src/adr/008adr-exceptions.md
shonfeder:
```suggestion
## 3. Exception explanations
On their own, exceptions should include concise messages with all the relevant information components, outlined above. In addition to that, we should implement an advanced variant of `ExceptionAdapter`, called `ExceptionExplainer`, that is enabled by default, but can be quieted if Apalache is invoked with the flag `--quiet-exceptions` . 
```

docs/src/adr/008adr-exceptions.md
Kukovec: SG, I'll add that.

shonfeder (approved):

PR #899, opened 2021-07-01
WIP: New approach to rewriting rules

Kukovec:

This PR outlines a proof-of-concept alternative implementation of rewriting rules, using `scalaz.State`. It should facilitate discussion regarding the new approach to rewriting, and is not intended to be integrated (in its current form) in the near future.

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/BasicOps.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/BasicOps.scala
@@ -1,0 +1,44 @@
+package at.forsyte.apalache.tla.bmcmt.rewriter2
+
+import at.forsyte.apalache.tla.bmcmt._
+import at.forsyte.apalache.tla.bmcmt.types.CellT
+import at.forsyte.apalache.tla.lir.TlaEx
+import at.forsyte.apalache.tla.lir.smt.SmtTools.False
+import scalaz._
+import scalaz.Scalaz._
+
+/**
+ * A collection of common methods for manipulating the internal state
+ */
+object BasicOps {
+  private def safeBind(key: String, value: ArenaCell): RewritingComputation[Unit] = modify { s =>
+    s.copy(binding = Binding(s.binding.toMap + (key -> value)))
+  }
+
+  // Extends the internal binding with key->value
+  def bind(key: String, value: ArenaCell): StateCompWithExceptions[Unit] =
+    EitherT.rightT(safeBind(key, value))
+
+  private def safeAssertGroundExpr(ex: TlaEx): RewritingComputation[Unit] = modify { s =>
+    // xform ex into BoolFormula
+    // proof of concept: every formula is False
+    s.copy(constraints = False() +: s.constraints)
+  }
+
+  // Extends the collection of constraints with the SMT formula represented by ex
+  // Note: A real implementation should not use the BoolT1 fragment of TlaEx as a proxy for
+  // representing SMT constraints.
+  def assertGroundExpr(ex: TlaEx): StateCompWithExceptions[Unit] = EitherT.rightT(safeAssertGroundExpr(ex))
+
+  def safeNewCell(cellType: CellT): RewritingComputation[ArenaCell] = symbState { state =>
+    val newArena = state.arena.appendCell(cellType)
+    (state.copy(arena = newArena), newArena.topCell)
+  }
+
+  // Adds a new cell of the given type to the arena and returns it
+  def newCell(cellType: CellT): StateCompWithExceptions[ArenaCell] = EitherT.rightT(safeNewCell(cellType))
+
+  // Shorthand for State[SymbStateInternal, T] { fn }
+  def symbState[T](fn: RewritingState => (RewritingState, T)): RewritingComputation[T] =
+    State[RewritingState, T] { fn }
+}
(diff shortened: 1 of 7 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/NegationRule.scala
```diff
@@ -0,0 +1,49 @@
...
+    // that value is propagated to the actual final result (i.e. the first exception encountered is "thrown")
+  } yield predCellName
+
+  // Static evaluation of ex as a representation of a TlaOper.not expression.
+  // If ex is equal to (- y), returns Right(y), otherwise returns Left(RewritereException)
+  private def exOrException(ex: TlaEx): Either[Exception, TlaEx] = ex match {
+    case OperEx(TlaBoolOper.not, negEx) => Right(negEx)
+    case badEx =>
+      Left(new RewriterException("%s is not applicable".format(getClass.getSimpleName), badEx))
+  }
```
shonfeder, line 48: I think that either `isApplicable` should be defined in terms of this (using `isRight`) or you could add a default method to the `Rule` class that derives this method from the `isApplicable`.

shonfeder (commented):
This approach looks great to me. I think there'll be lots of gains by moving the rewriting into pure computations. :+1: 

I left two incidental suggestions, but they are just superficial.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/BasicOps.scala
```diff
@@ -0,0 +1,44 @@
...
+import scalaz._
+import scalaz.Scalaz._
+
+/**
+ * A collection of common methods for manipulating the internal state
+ */
+object BasicOps {
+  private def safeBind(key: String, value: ArenaCell): RewritingComputation[Unit] = modify { s =>
+    s.copy(binding = Binding(s.binding.toMap + (key -> value)))
+  }
```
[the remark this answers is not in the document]
Kukovec: Fair enough, the idea here was mostly to demonstrate the difference (and the relative ease of lifting to a "computation with possible exceptions"), but I agree that the "safe" methods don't add much.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/NegationRule.scala
Kukovec: You could define it that way, but that's a roundabout way to define `isApplicable`, which should be a very simple structural check. It's also much easier to read and understand the rule if it's written explicitly like this, I think.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/package.scala
```diff
@@ -0,0 +1,11 @@
+package at.forsyte.apalache.tla.bmcmt
+
+import scalaz._
+import scalaz.Scalaz._
+
+package object rewriter2 {
+
+  type RewritingComputation[T] = State[RewritingState, T]
+
+  type StateCompWithExceptions[T] = EitherT[Exception, RewritingComputation, T]
```
shonfeder, line 10: How about something like this:

```suggestion
  type RewritingComputationExn[T] = EitherT[Exception, RewritingComputation, T]
```

Otherwise its not clear from the name that this pertains to the `RewritingComputation` or that `Comp` is short for `Computation` rather than `Comparison`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/BasicOps.scala
```diff
@@ -0,0 +1,47 @@
...
+  def newCell(cellType: CellT): StateCompWithExceptions[ArenaCell] = cmpIsRight {
+    cmp { state =>
+      val newArena = state.arena.appendCell(cellType)
+      (state.copy(arena = newArena), newArena.topCell)
+    }
+  }
+
+  // Shorthand for State[SymbStateInternal, T] { fn }
+  def cmp[T](fn: RewritingState => (RewritingState, T)): RewritingComputation[T] =
+    State[RewritingState, T] { fn }
```
shonfeder, line 42: If `cmp` is short for `Computation` than maybe use `Comp` since that's already appearing in the `StateComptWithException` type alias?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/BasicOps.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  // Adds a new cell of the given type to the arena and returns it
+  def newCell(cellType: CellT): StateCompWithExceptions[ArenaCell] = cmpIsRight {
+    cmp { state =>
+      val newArena = state.arena.appendCell(cellType)
+      (state.copy(arena = newArena), newArena.topCell)
+    }
+  }
+
+  // Shorthand for State[SymbStateInternal, T] { fn }
```
shonfeder, line 40:
```suggestion
  // Shorthand for State[RewritingState, T] { fn }
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rewriter2/Rule.scala
```diff
@@ -0,0 +1,14 @@
+package at.forsyte.apalache.tla.bmcmt.rewriter2
+
+import at.forsyte.apalache.tla.lir.TlaEx
+
+/**
+ * A rule in the new rewriting system. It can be tested, or applied. In the latter case, it either returns
+ * a TlaEx (rewritten expression) or throws a RewritingException (represented by computation)
```
shonfeder, line 7: I don't understand this part: `throws a RewritingException (represented by computation)`.

shonfeder (changes requested):
This all looks good to me, with just one minor naming change suggestion. I wonder if you should get input from Rodrigo or Igor before starting to migrate over the other rewriting stuff?

PR #903, opened 2021-07-05
Fold tutorial

### LINKED ISSUE #901: [FEATURE] Prepare a tutorial on `FoldSet` and `FoldSeq`
As an alternative to recursive operators.

Kukovec:

closes #901


docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+                       THEN tailMax
+                       ELSE Head(seq)
+
+MaxFold(seq) == LET Max(p,q) == IF p > q THEN p ELSE q
+                IN FoldSeq( Max, negInf, seq )
+```
+
+The first advantage of the fold implementation, we feel, is that it is much more clear and concise. It also does not require a termination condition, unlike the recursive case.
+One inherent problem of using recursive operators with a symbolic encoding, is the inability to estimate termination.
+While it may be immediately obvious to a human, that `MaxRec` terminates after no more than `Len(seq)` steps, automatic termination analysis is, in general, a rather complex form of static analysis.
```
konnov, line 86:
```suggestion
While it may be immediately obvious to a human, that `MaxRec` terminates after no more than `Len(seq)` steps, automatic termination analysis is, in general, a rather complex and incomplete form of static analysis.
```

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+                    THEN 0
+                    ELSE LET x == CHOOSE y \in set: TRUE
+                         IN 1 + Cardinality( set \ {x} )
+
+CardinalityFold(set) == LET Count(p,q) == p + 1 \* the value of q, the set element, is irrelevant
+                        IN FoldSet( Count, 0, set )
+```
+
+Notice that, in the case of sets, picking an arbitrary element `x`, to remove from the set at each step, utilizes the `CHOOSE` operator. This is a common trait shared by many operators that implement recursion over sets. 
+Since the introduction of folds, the use of `CHOOSE` in Apalache is heavily discouraged as it is both inefficient, as well as nondeterministic (unlike how `CHOOSE` is defined in TLA+ literature).
```
konnov, line 115:
```suggestion
Since the introduction of folds, the use of `CHOOSE` in Apalache is heavily discouraged as it is both inefficient, as well as nondeterministic (unlike how `CHOOSE` is defined in TLA+ literature). For details, see the discussion in [issue 841](https://github.com/informalsystems/apalache/issues/841).
```

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+So the third advantage of using folds is the  ability to, almost always, avoid using the `CHOOSE` operator.
+
+The downside of folding, compared to general recursion, is the inability to express non-primitively recursive functions.
+For instance, one cannot define the Ackermann function, as a fold. We find that in most specifications, this is not something the users would want to implement anyway, so in practice, we believe it is almost always better to use fold over recursive functions.
+
+### Examples: The versatility of folds
+
+Here we give some examples of common operators, implemented using folds:
+
+```
```
konnov, line 126:
```suggestion
```tla
```

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+\* Returns TRUE iff fn is injective
+IsInjective(fn) == 
+  LET SeenBefore( seenAndResult, e ) == 
+    IF fn[e] \in seenAndResult[1]
+    THEN [ seenAndResult EXCEPT ![2] = FALSE ]
+    ELSE [ seenAndResult EXCEPT ![1] = seenAndResult[1] \union {fn[e]} ]
+  IN FoldSet( SeenBefore, << {}, TRUE >>, DOMAIN fn )[2]
+```
+
+For the sake of comparison, here are the same operators, using recursion, `CHOOSE` or quantification:
```
konnov, line 165:
```suggestion
For the sake of comparison, we rewrite the above operators using recursion, `CHOOSE` or quantification:
```

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+\* Finds the the value that appears most often in a sequence. Returns elIfEmpty for empty sequences
+Mode(seq, elIfEmpty) == LET ExtRange == Range(seq) \union {elIfEmpty}
+                        IN LET CountElem(countersAndCurrentMode, e) ==
+                             LET counters == countersAndCurrentMode[1]
+                                 currentMode == countersAndCurrentMode[2]
+                             IN LET newCounters == [ counters EXCEPT ![e] == counters[e] + 1 ]
+                                IN IF newCounters[e] > newCounters[currentMode]
+                                   THEN << newCounters, e >>
+                                   ELSE << newCounters, currentMode >>
+                           IN FoldSeq( CountElem, <<[ x \in ExtRange |-> 0 ], elIfEmpty >>, seq )[2]
```
konnov, line 154: I guess, one can write this operator in a much nicer way by using `Range`, `SelectSeq`, and `Len`. Perhaps, this example does not demonstrate the power of `FoldSeq` :-)

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+                     THEN 0
+                     ELSE LET x == CHOOSE y \in S : TRUE
+                          IN (IF P(x) THEN 1 ELSE 0) + Quantify(S \ {x}, P) 
+
+RECURSIVE Range(_)
+Range(seq) == IF seq = <<>>
+              THEN {}
+              ELSE {Head(seq)} \union Range(Tail(seq))
+
+Mode(seq, elIfEmpty) == IF seq = <<>>
```
konnov, line 199: This is definitely the case, where `CHOOSE` gives us a cleaner solution. I guess, the above version is just "too efficient".

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+                        THEN elIfEmpty 
+                        ELSE LET numOf(p) == Quantify( DOMAIN seq, LAMBDA q: q = p )
+                             IN CHOOSE x \in Range(seq): \A y \in Range(seq) : numOf(x) >= numOf(y)
+
+IsInjective(fn) == \A a,b \in DOMAIN fn : fn[a] = fn[b] => a = b
+```
+
+In most cases, recursive operators are much more verbose, and the operators using `CHOOSE` and/or quantification mask double iteration (and thus have quadratic complexity). 
+For instance, the evaluation of the fold-less `IsInjective` operator actually requires the traversal of all domain pairs, instead of the single domain traversal with fold.
+In particular, `Mode`, the most verbose among the fold-defined operators, is still very readable (most LET-IN operators are introduced to improve readability, at the cost of verbosity) and quite efficient, as its complexity is linear w.r.t. the length of the sequence (the mode could also be computed directly, without a sub-call to `Range`, but the example would be more difficult to read), unlike the variant with `CHOOSE` and `\forall`, which is quadratic.
```
konnov, line 209:
```suggestion
In particular, `Mode`, the most verbose among the fold-defined operators, is still very readable (most LET-IN operators are introduced to improve readability, at the cost of verbosity) and quite efficient, as its complexity is linear w.r.t. the length of the sequence (the mode could also be computed directly, without a sub-call to `Range`, but the example would be more difficult to read), unlike the variant with `CHOOSE` and `\A`, which is quadratic.
```

konnov: Why don't you give two versions with folds and explain which one would be better for Apalache? I believe that people are not writing TLA+ specs for efficiency. So you should label the more efficient version as efficient from the SMT encoding point of view, as there is no built-in notion of complexity in TLA+.

konnov (commented):
Very nice! Plenty of comments because I like this tutorial :)

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+  1. Examples of common operators defined with folds
+
+### Semantics of fold operators
+
+Folding refers to iterative application of a binary operator over a collection. Given an operator `Op`, a base value `b` and a collection of values `C`, folding `Op` over `C` starting with `b` is defined as follows:
+
+  1. If `C` is empty, the resulting value is `b`
+  1. If `C` is nonempty, we select an element `x` from `C`. If `C` is a sequence, `x` is the head value, and if `C` is a set, `x` is an arbitrary member. Then, we compute `b2 = Op(b, x)` and take `C2 = C without x` (`Tail(C)` or `C \ {x}` respectively). We continue the process, by computing the fold of `Op` over `C2` starting with `b2`.
+
+Note that this is the definition of a _left fold_ in the literature. Apalache does not implement a right fold.
```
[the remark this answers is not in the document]
Kukovec: Not really, `FoldRight(Op, b, C) = FoldLeft(  Op*, b*, C reversed)` where `Op(a,b) = Op*(b,a)` and `b*` depends on the details (but likely `b`)

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+
+For example, if `C` is the sequence `<<x,y,z>>`, the result is equal to `Op( Op( Op(b, x), y), z)`. If `C = {x,y}`, the result is either `Op( Op(b, x), y)` or `Op( Op(b, y), x)`. Because the order of elements selected is arbitrary in the case of sets, users should be careful, as the result is only uniquely defined in the case that the operator is both associative (`Op(Op(a,b),c) = Op(a,Op(b,c))`) and commutative (`Op(a,b) = Op(b,a)`).
```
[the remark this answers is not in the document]
Kukovec: What's wrong with "... the result is only uniquely defined in the case that the operator is both associative (...) and commutative (...).")?

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+
+```
+\* @type: ( (a, b) => a, a, Set(b) ) => a
+FoldSet( operator, base, set )
+
+\* @type: ( (a, b) => a, a, Seq(b) ) => a
+FoldSeq( operator, base, seq )
+```
+
+As shown by the type signature, Apalache permits a very general form of folding, where the types of the collection elements and the type of the base element/return-type of the operator do not have to match. Again, we urge users to exercise caution when using `FoldSet` with an operator, for which the types `a` and `b` are different, as such operators cannot be commutative or associative, and therefore the result is not guaranteed to be unique and predictable.
```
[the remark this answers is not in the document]
Kukovec: Cannot, e.g. in `A: (Int, Str) => X` you can't just flip the arguments from `A(0, "a")` to `A("a",0)`, since that's a type violation.

docs/src/apalache/fold.md
Kukovec: `ExtRange`, `counters`, `currentMode` are all just convenience variables. This looks, at least to me, like a linear mode-calculation algorithm in a programming language.

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+\* Define a LET-IN operator Plus(p,q) == p + q instead
+X == FoldSet( + , 0, {1,2,3} )
+```
+
+Local LET definitions can also be used as closures:
+
+```
+A(x) == LET PlusX(p,q) == p + q + x IN FoldSeq( PlusX, 0, <<1,2,3>> )
+
+X == A(1) \* X = 9
```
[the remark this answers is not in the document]
Kukovec: No? `v0 = 0, v1 = (v0 + 1) + 1 = 2, v2 = (v1 + 2) + 1 = 5, v3 = (v2 + 3) + 1 = 9`

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+### Folding VS recursion
+
+While TLA+ allows users to write arbitrary recursive operators, they are, in our experience, mostly used to implement collection traversals. Consider the following implementations of a `Max` operator, which returns the largest element of a sequence:
+
+```
+CONSTANT negInf \* Max(<<>>) = -inf, but TLA+ has no literal for ±inf, so we use it as a constant
```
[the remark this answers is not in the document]
Kukovec: It's not defined for integers, and also not supported by Apalache.

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,213 @@
...
+
+RECURSIVE MaxRec(_)
+MaxRec(seq) == IF seq = <<>>
```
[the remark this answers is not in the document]
Kukovec: Yes, but we're explicitly comparing fold to recursion here.

docs/src/apalache/fold.md
Kukovec: Makes sense. They will care about efficiency, if they want to use automatic tools though.

konnov:
It is actually a good idea to turn all of your examples in TLA+ files and then include them via:

```
{{#include ../../../test/tla/y2k_override.tla::}}
```

See how it is done in `./docs/src/apalache/principles.md`. You should also add your examples in the integration tests. By doing so we will immediately see if they are breaking.

docs/src/apalache/fold.md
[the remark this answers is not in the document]
konnov: Sure, you don't need both for expressivity. But I think that reversing a sequence would produce `O(n^2)` constraints in our encoding. Btw, the community modules have both left and right folds for sequences: https://github.com/tlaplus/CommunityModules/blob/master/modules/SequencesExt.tla

I would prefer having both left and right folds for sequences in `Apalache.tla`, even if the right fold is `foldl(rev(l))`. We could find a better encoding for it in the future.

docs/src/apalache/fold.md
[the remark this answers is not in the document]
konnov: Just explain a bit what you mean by this sentence. That `FoldSet` may yield different results depending on the order of iteration, unless the operator is commutative and associative. The tricky part here is that if we define `FoldSet` by using `CHOOSE`, the order of iteration is unknown but fixed. But Apalache does not give you such a guarantee. Give an example of what happens if the operator is either non-commutative or non-associative. It's a tutorial ;-)

docs/src/apalache/fold.md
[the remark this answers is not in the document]
konnov: Oh, that was a comment. It looked like somebody has put a slash in front of multiplication :)

docs/src/apalache/fold.md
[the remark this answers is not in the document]
konnov: I mean this sentence "TLA+ has no literal for +/- inf" is not precise. You could say that integers are unbounded in TLA+, so there is no natural minimum or maximum like `MIN_INT` and `MAX_INT` in programming languages.

docs/src/apalache/fold.md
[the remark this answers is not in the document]
konnov: Right. But it would be good to have pros and cons of the both approaches. You can write this recursive operator without `CHOOSE`, but it is not how TLC users would do it.

docs/src/apalache/fold.md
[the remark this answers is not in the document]
Kukovec: We could have both, in which case it would make more sense to have separate rewriting rules and operators at the IR level. However, this is definitely outside the scope of this tutorial (at least v1), since we don't have a right fold atm.

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,188 @@
...
+  1. How do I use these operators in Apalache?
+  1. Should I use folding or recursion?
+  1. Examples of common operators defined with folds
+
+### Syntax
+
+The syntax of the fold operators is as follows:
+
+```tla
+\* @type: ( (a, b) => a, a, Set(b) ) => a
```
konnov, line 16: the closing `;` is missing

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,188 @@
...
+FoldSet( operator, base, set )
+
+\* @type: ( (a, b) => a, a, Seq(b) ) => a
```
konnov, line 19: the closing `;` is missing

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,188 @@
...
+#### Semantics of `FoldSet`
+
+In the case of folding over sequences, `C` is a set `{a_1, ..., a_n}`. Then, `FoldSet( Op, b, C )` is defined as follows:
+
+  1. If `C` is empty, then `FoldSeq( Op, b, {} ) = b`, regardless of `Op`
+  1. If `C` is nonempty, we establish a recursive relation between folding over `C` and folding over some subset of `C` in the following way: `FoldSeq( Op, b, C ) = FoldSeq( Op, Op(b, x), C \ {x} )`, where `x = CHOOSE y \in C: TRUE`. Note that Apalache selects `x` nondeterministically.
+
+Note that the above are definitions of a _left fold_ in the literature. Apalache does not implement a right fold.
+
+For example, if `C` is the sequence `<<x,y,z>>`, the result is equal to `Op( Op( Op(b, x), y), z)`. If `C = {x,y}`, the result is either `Op( Op(b, x), y)` or `Op( Op(b, y), x)`. Because the order of elements selected is nondeterministic in the case of sets, users should be careful, as the result is only uniquely defined in the case that the operator is both associative (`Op(Op(a,b),c) = Op(a,Op(b,c))`) and commutative (`Op(a,b) = Op(b,a)`).
```
konnov, line 43:
```suggestion
For example, if `C` is the sequence `<<x,y,z>>`, the result is equal to `Op( Op( Op(b, x), y), z)`. If `C = {x,y}`, the result is either `Op( Op(b, x), y)` or `Op( Op(b, y), x)`. Because the order of elements selected from a set is not predefined, users should be careful, as the result is only uniquely defined in the case that the operator is both associative (`Op(Op(a,b),c) = Op(a,Op(b,c))`) and commutative (`Op(a,b) = Op(b,a)`). 
```

konnov: Let's not use the word "non-deterministic", as the users might want to exploit non-determinism. We do not give guarantees of non-determinism here.

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,188 @@
...
+| Order | FoldSet value |
+| --- | --- |
+| 1 -> 2 -> 3 | 11 | 
+| 1 -> 3 -> 2 | 12 | 
+| 2 -> 1 -> 3 | 13 | 
+| 2 -> 3 -> 1 | 15 | 
+| 3 -> 1 -> 2 | 16 | 
+| 3 -> 2 -> 1 | 17 | 
+
+Because Apalache chooses nondeterministically, all of the above results are possible outcomes, unlike in the case of the fixed-but-unknown order that would arise from a deterministic choice.
```
konnov, line 57: Is this actually true? I thought that we implement `FoldSet` for some order that depends on the arena representation of a set. So we cannot guarantee that all possible orders will be explored, right?

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,188 @@
...
+                       ELSE Head(seq)
+
+MaxFold(seq) == LET Max(p,q) == IF p > q THEN p ELSE q
+                IN FoldSeq( Max, negInf, seq )
+```
+
+The first advantage of the fold implementation, we feel, is that it is much more clear and concise. It also does not require a termination condition, unlike the recursive case.
+One inherent problem of using recursive operators with a symbolic encoding, is the inability to estimate termination.
+While it may be immediately obvious to a human, that `MaxRec` terminates after no more than `Len(seq)` steps, automatic termination analysis is, in general, a rather complex and incomplete form of static analysis.
+Apalache addresses this by finitely unrolling recursive operators and requires users to provide unroll limits (`UNROLL_LIMIT_MaxRec == ...`), which serve as a static upper bound to the number of recursive re-entries, because in general, recursive operators may take an unpredictable number of steps (e.g. computing the Collatz sequence) or never terminate at all.
```
konnov, line 116: Give a link to the Collatz sequence? https://en.wikipedia.org/wiki/Collatz_conjecture

docs/src/apalache/fold.md
```diff
@@ -0,0 +1,188 @@
...
+                        IN FoldSet( Count, 0, set )
+```
+
+Notice that, in the case of sets, picking an arbitrary element `x`, to remove from the set at each step, utilizes the `CHOOSE` operator. This is a common trait shared by many operators that implement recursion over sets. 
+Since the introduction of folds, the use of `CHOOSE` in Apalache is heavily discouraged as it is both inefficient, as well as nondeterministic (unlike how `CHOOSE` is defined in TLA+ literature). For details, see the discussion in [issue 841](https://github.com/informalsystems/apalache/issues/841).
+
+So the third advantage of using folds is the  ability to, almost always, avoid using the `CHOOSE` operator.
+
+The downside of folding, compared to general recursion, is the inability to express non-primitively recursive functions.
+For instance, one cannot define the Ackermann function, as a fold. We find that in most specifications, this is not something the users would want to implement anyway, so in practice, we believe it is almost always better to use fold over recursive functions.
```
konnov, line 149: reference to https://en.wikipedia.org/wiki/Ackermann_function ?

konnov (approved):
I think we are converging. Added several comments.

docs/src/apalache/fold.md
Kukovec: Only one order will be explored, but the point is that the user has no way of determining _which_ order, so they must consider all of them as possibilities.

konnov: Right. The above explanation can be understood as there is a way to produce all possible iteration orders. This is the ambiguity of the word "non-determinism". So we should probably say "fixed but unknown order of iteration".

Kukovec: "fixed but unknown order of iteration" is what happens with deterministic CHOOSE

Kukovec: I've reworded the section to instead say something to the effect of "makes no guarantee of determinism". This way, we're not claiming nondeterminism, but are also in the clear if we choose to go with full determinism.

konnov (approved):
LGTM

PR #909, opened 2021-07-09
SelectSeq rewritten as fold

### LINKED ISSUE #873: [FEATURE] implement Sequences.SelectSeq?
Having #866, it should be trivial. Most likely, we can just define `SelectSeq` via fold over sequences.

Kukovec:

closes #873

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
@@ -1,0 +1,87 @@
+package at.forsyte.apalache.tla.pp
+
+import at.forsyte.apalache.tla.lir.oper.{ApalacheOper, TlaActionOper, TlaControlOper, TlaFunOper, TlaOper, TlaSeqOper}
+import at.forsyte.apalache.tla.lir.{
+  BoolT1, LetInEx, NameEx, OperEx, OperParam, OperT1, SeqT1, TlaEx, TlaOperDecl, Typed, TypingException
+}
+import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TransformationTracker}
+
+/**
+ * Replaces instances of SelectSeq, using the following definition:
+ *
+ * SelectSeq(seq, Test(_)) == LET CondAppend(s,e) == IF Test(e) THEN Append(s, e) ELSE s
+ *                            IN FoldSeq( CondAppend, <<>>, seq )
+ */
+class SelectSeqAsFold(nameGenerator: UniqueNameGenerator, tracker: TransformationTracker) extends TlaExTransformation {
+
+  override def apply(expr: TlaEx): TlaEx = {
+    transform(expr)
+  }
+
+  def transform: TlaExTransformation = tracker.trackEx {
+    case OperEx(TlaSeqOper.selectseq, seqEx, testNameEx @ NameEx(operName)) =>
+      // Sanity check on types
+      val seqElemType = seqEx.typeTag match {
+        case Typed(SeqT1(a)) => a
+        case badType         => throw new TypingException(s"Argument $seqEx expected to have type Seq(a), found $badType.")
+      }
+
+      testNameEx.typeTag match {
+        case Typed(OperT1(Seq(`seqElemType`), BoolT1())) => () // all good
+        case badType =>
+          throw new TypingException(s"$operName expected to have type ($seqElemType) => Bool, found $badType.")
+      }
+
+      // Prep fresh names for the operator and its params
+      val condAppOperName = nameGenerator.newName()
+      val sParamName = nameGenerator.newName()
+      val eParamName = nameGenerator.newName()
+
+      def e(): NameEx = NameEx(eParamName)(Typed(seqElemType))
+      def s(): NameEx = NameEx(sParamName)(seqEx.typeTag)
+
+      // IF Test(e) THEN Append(s,e) ELSE s
+      val condAppBody =
+        OperEx(
+            TlaControlOper.ifThenElse,
+            OperEx(TlaOper.apply, testNameEx, e())(Typed(BoolT1())),
+            OperEx(TlaSeqOper.append, s(), e())(seqEx.typeTag),
+            s()
+        )(seqEx.typeTag)
+
+      val condAppDecl =
+        TlaOperDecl(
+            condAppOperName,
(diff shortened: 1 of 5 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
```diff
@@ -0,0 +1,87 @@
...
+  override def apply(expr: TlaEx): TlaEx = {
+    transform(expr)
+  }
+
+  def transform: TlaExTransformation = tracker.trackEx {
+    case OperEx(TlaSeqOper.selectseq, seqEx, testNameEx @ NameEx(operName)) =>
+      // Sanity check on types
+      val seqElemType = seqEx.typeTag match {
+        case Typed(SeqT1(a)) => a
+        case badType         => throw new TypingException(s"Argument $seqEx expected to have type Seq(a), found $badType.")
```
shonfeder, line 26: Is type checking not done by this point in the rewriting?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
```diff
@@ -0,0 +1,87 @@
...
+
+      LetInEx(foldEx, condAppDecl)(foldEx.typeTag)
+
+    case ex @ OperEx(op, args @ _*) =>
+      val newArgs = args map transform
+      if (args == newArgs) ex else OperEx(op, newArgs: _*)(ex.typeTag)
+
+    case ex @ LetInEx(body, defs @ _*) =>
+      // Transform bodies of all op.defs
+      val newDefs = defs map tracker.trackOperDecl { d => d.copy(body = transform(d.body)) }
```
shonfeder, line 75: how come we need to map using the `tracker.trackOperDecl` and `copy` etc here, but don't need any of that when mapping over the operator arguments on line 70?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
```diff
@@ -0,0 +1,87 @@
+package at.forsyte.apalache.tla.pp
```
shonfeder, line 1: I'd suggest adding some unit tests for this.

shonfeder (changes requested):
Just one change request, which is the addition of unit tests for the transformation. Let me know if you that's inappropriate for some reason.

Other comments are just questions for my own edification.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
Kukovec: It should be, but I have to do a match to read the inner `a` value anyway, so the other case is just a safeguard.

shonfeder: If it should be type checked, I'd suggest not repeating the checks here. That can lead to maintenance issues and weird divergence of behavior. I think it's better to just ensure we do the checking properly in the appropriate phase.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
Kukovec: `transform` calls `tracker.trackEx`, but `trackEx` operates over `TlaEx` not `TlaDecl`. L75 is a `Decl => Decl` transformation (which transforms the declaration by modifying its body) , whereas  L70 is an `Ex => Ex` transformation.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
Kukovec: This is somewhat related to discussions we've had with Igor on this topic, i.e. whether to do unit tests or integration tests. The short version is, that it's very tedious to unit test processing passes, because you'd have to write tests that check if your transformed value has a very specific syntactic shape. OTOH, the tests in `SelectSeqTests` implicitly require that the transformation produces the result we expect, but are significantly easier to write. Unless there's some particular property of the transformation you'd like to verify in unit testing, integration testing suffices, IMO.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
shonfeder: Ah, I see.Thanks for explaining.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
Kukovec: Counterpoint: writing such sanity checks (which are cheap anyway, because I need to pattern match with the first case anyway) actually frees us from the implicit assumption and dependency, and it makes this pass movable in the pass order. Pass dependencies have historically been a big pain point.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
shonfeder: Ok. Make sense.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SelectSeqAsFold.scala
shonfeder: My concern is not for the computational cost, but for the maintenance. 

If we end up moving this above the type checking level, then, wouldn't it be incorrect for it to be doing it's own type checking? To my mind, this actually makes the transformation less portable and less general. (E.g., it cannot be used on untyped code.)

If we end up imposing an improved system for giving errors on type checking fails, and we've peppered these kinds of adhoc sanity checks through the code, we're going to have to track down every case do duplicate work. Or else maybe we miss them, and we end up with two different kinds of errors reported. Ditto regarding unnecessary re-work if we change the name or representation of the type.

I won't block on this, but I don't see how it's a value add, and do foresee allowing this kind of sanity check to lead to maintenance issues and needless code bloat.

shonfeder (approved):

PR #933, opened 2021-07-27
RFC-010: Transition Exploration Server

### LINKED ISSUE #815: Write RFC to propose architecture for server mode
Prerequisite for #730

As per discussion in planning, I'll have a first crack at getting this in shape. It should propose (at least):

- [ ] Basic architecture for running in server mode
- [ ] Approach to an RPC interface to the transition executor module's API
- [ ] Some considerations around extensibility, so we can later implement LSP support or similar in a natural way

shonfeder:
Closes #815

This PR presents an RFC proposing the preliminary design for a Transition
Exploration Server, as described in more detail in the document.

I should give this one more read through, to catch typos etc., before asking for
review


shonfeder:
Thanks for the review!

> The passes up to the BMC pass essentially implement load() and the incrementality should be mostly supported by SeqModelChecker/TransitionExecutor

Yep, this is what I meant to indicate. I've tweaked the language to make this more explicit.

> I would have liked for the RFC to specify the CLI in more detail though, since that's what I imagine to be the biggest reason for having such a server (e.g. for use in MBT automation).

Note that the current proposal is not scoped to include the **client** CLI at all. The current assumption is that MBT will communicate with the server via RPC calls, not vial shelling out the command line. As indicated in the section on the protocol, we can implement a CLI client as a special case. But I've currently left that for followup work.

In any case, I would expect the CLI client to be a very direct and translation of the RPC calls, so, e.g.,

```sh
$ apalache-mc client configure <URL:PORT>
$ apalache-mc client reset
$ apalache-mc client load-model Spec.tla
# etc..
```

However if you (or the MBT team) think the CLI client should be part of this proposal, I'm happy to expand the scope. What do you think?

If you were talking about the CLI for starting the server, I imagine that to be as simple as `apalache-mc server start` (with maybe some utility flags that will be added along the way, but none occur to me). Do you think more will be involved here?

docs/internal/rfc/002rfc-transition-explorer.md
```diff
@@ -0,0 +1,323 @@
...
+   *
+   * Since the number of computable initial states can be infinite, an upper
+   * limit must be set.
+   *
+   * @params max the maximum number of initial states to return (default to 100)
+   * @params start the nth state to begin fetching from (defaults to 0)
+   * @return `Some(exprs)` where `exprs` are `n` computed initial states, with
+   *         `n` <= `max`, or `None` if no model is loaded.
+   */
+  def initialStates(max: Int = 100, start: Int = 0): Option[TlaEx]
```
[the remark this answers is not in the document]
shonfeder: Ah yeah. I think this should be a list of states, actually. Thanks!

docs/internal/rfc/002rfc-transition-explorer.md
```diff
@@ -0,0 +1,323 @@
...
+
+  /**  The initial state predicates given in the spec  */
+  def initialStatePredicates: Option[List[TlaEx]]
```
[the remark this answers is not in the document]
shonfeder: I think i just had the plurality in mind from the previous method. I think this should be singular!

docs/internal/rfc/002rfc-transition-explorer.md
```diff
@@ -0,0 +1,323 @@
...
+
+  /** The action predicates defining "next" transitions given in the spec
+   *
+   */
+  def actions: Option[List[TlaEx]]
+
+  /** The invariants defined by the spec */
+  def invariants: Option[CheckerInputVC]
```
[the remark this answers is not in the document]
shonfeder: Cause I thought the segregation of the invariants that appears in the `ChevkerInputVC` (considered as a struct) is quite helpful. My idea is that we will return a dictionary representing this data and the associated fields. What do you think?

docs/internal/rfc/002rfc-transition-explorer.md
```diff
@@ -0,0 +1,323 @@
...
+
+  /** Setting constants will also reset the explorer, if an exploration is 
+   * underway
+   */
+  def initializeConstants(Map[TlaExp, TlaExp]): Either[LoadErr, Unit]
+
+  /** Gives a map of constants to their current values */
+  def constants: Option[Map[TlaEx, Option[TlaEx]]]
```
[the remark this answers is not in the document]
shonfeder: My logic was that this would let the method serve both as a way of listing the current values of the constants as a way of listing which constants there are (in case none have been set). I see what you mean about the implication that some be left absent tho.

Do you think I should add two different methods like `constants: Option[List[TlaEx]]` and `constantValues: Option[Map[TlaEx, TlaEx]]`?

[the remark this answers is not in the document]
Kukovec: Not sure you need a method for that, `constants` would just be `constantValues map { _.keys }`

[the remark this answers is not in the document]
shonfeder: Might be a slight misunderstanding. Of course if we are in our code base we don’t need a method for it. But the api here will be exposed via gRPC, and the clients will not have access to constantValues if we don’t explicitly then a method.

How this RFC is meant to be read is that the input types of the method signatures given here specify the requests the server will handle and the output types specify the replies the server will send in response.

I’ll think about what to add to make that clearer.

but the question here remains: do we want one endpoint for fetching the the constants and their values, or two: one to fetch just the constants (in case they aren’t set) and another to fetch their values if they have been set.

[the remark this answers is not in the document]
Kukovec: I see what you mean. In my mind, the server should never be in a partially configured state, where it has received the spec, but the constants remain uninitialized. It should be possible to swap constant values, sure, but I would still require some constants be a part of the initialization. In that sense, there is only one option, fetching both the constants and the values.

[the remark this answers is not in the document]
shonfeder: In that case, we would have to change the `loadSpec` endpoint to receive the constants values as well.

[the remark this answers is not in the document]
Kukovec: Yes, but I think it's ultimately for the better.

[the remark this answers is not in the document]
shonfeder: Cool. I've made that change, and I've also added a short, concrete example sketching out how the actual implementation will look using gRPC.

shonfeder:
Just had a long meeting in which the aim and scope of the API we mean to expose has been corrected and clarified. The correction and clarification is thanks to patient explanation and insight from @konnov and @Kukovec, and to illuminating contributions from @danwt.

I’ll update the issue with details clarifying the new aim tomorrow.

shonfeder:
TODOs suggested by @Kukovec and @konnov in our planning meeting yesterday:

- [ ] Rework RFC to expose an API grounded in the symbolic model checking perspective, since this is the unique strength of Apalache.
- [ ] Add a couple example usecases (Igor)
- [ ] Make available transitions returned as a map from operator names to expressions (Jure)
- [ ] Add consideration of more interesting use cases like:
  - Implementing a repl (Igor)
  - Grafing new constraints into spec (Igor, @danwt)
  - Include discussion of mutable vs. immutable aspects of spec (Jure)

konnov:
by the way, can we sync the RFC numbers with ADR numbers, so we don't have the same number for RFC and ADR?

shonfeder:
> by the way, can we sync the RFC numbers with ADR numbers, so we don't have the same number for RFC and ADR?

Given that they are two very different kinds of documents, shouldn't they have different numbers?

konnov:
I don't know. We don't have too many of them, so I would prefer to have unique numbers for all of them.

shonfeder:
I've tweaked the RFC to reflect our discussion, at least the best of my abilities at the moment. I think it'd be good to get another review here, and either work off of concrete suggestions given in the review or begin moving forward with the implementation work.

konnov:
What I just realized is that we never discussed that this server is supposed to be stateful. One can see that from the API. Shall we introduce a session id to all queries? For the moment, it would be just a safety measure for the case, when two clients connect concurrently and mess up the state of the model checker. In the future, we could use the session id to balance the load between multiple servers.

konnov:
Also, for me the basic litmus test of this API would be about whether we could implement https://github.com/informalsystems/apalache/blob/unstable/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala as a client.

docs/internal/rfc/009rfc-transition-explorer.md
```diff
@@ -0,0 +1,459 @@
...
+ *
+ * E.g., an execution from `s1` to `sn` with transitions `ti` for each `i` in `1..n-1`:
+ * 
+ *     List((s1, Some(t1)), (s2, Some(t2)), ..., (sn, None))
+ */
+type Execution = List[(State, Option[Transition])]
+
+trait UninitializedModel = {
+  val spec: TlaModule
+  val transitions: Set(Transition)
```
konnov, line 214: Shall we have a sequence of transitions instead of a set? It's often important to know the order of the transitions.

docs/internal/rfc/009rfc-transition-explorer.md
```diff
@@ -0,0 +1,459 @@
...
+type Model = Either[UninitializedModel, InitializedModel]
+
+/**  The type of errors that can result from failures when loading a spec. */
+type LoadErr
+
+/**  The type of errors that can result from failures to assert a constraint. */
+type AssertErr
+
+/**  The type of errors that can result from checking satisfiability of a frame. */
+type SatError
```
konnov, line 239: Just curious, are you going to use case classes for these errors? It's not clear to me why we want to define error types like that.

docs/internal/rfc/009rfc-transition-explorer.md
```diff
@@ -0,0 +1,459 @@
...
+
+  /** Used internally */
+  private def connection(): Connection
+
+  /**  Create a transition explorer
+   *
+   * Establishes the connection and a running session
+   *
+   * The channel is managed by the gRPC framework. */
+  def apply(channel: Channel): TransEx
```
konnov, line 254: that method should be in a companion singleton, right?

docs/internal/rfc/009rfc-transition-explorer.md
```diff
@@ -0,0 +1,459 @@
...
+  def applyTransition(transition: Transition): Either[AssertErr, Unit]
+
+  /** Assert a constraint into the current frame
+   *
+   *  Interface to `assertState` */
+  def assert(assertion: TlaEx): Either[AssertErr, Unit]
+
+  /** The example of an execution from the an initial state up to the current symbolic state
+   *
+   * TODO: I don't know how it is expected to obtain multile different executions. */
```
konnov, line 313: We will be able to enumerate multiple transitions in a loop like:
  - `sat`
  - `execution`
  - `assert(<the execution is different from the last one>)`
  - `sat`
  - `execution`
  - ...

There are many ways to exclude executions. One of them is by using the view mapping.

docs/internal/rfc/009rfc-transition-explorer.md
```diff
@@ -0,0 +1,459 @@
...
+   * @param aux auxiliary modules that may be required by the root module
+   * @return `Left(LoadErr)` if parsing or loading the model from `spec` goes
+   *          wrong, or `Right(UninitializedModel)` if the model is loaded successfully.
+   */
+  def loadModel(spec: String, aux: List[String] = List()): Either[LoadErr, UninitializedModel]
+
+  /** Initialize the constants of a model
+   *
+   * This will always also reset an exploration back to its initial frame. */
+  def initializeConstants(constants: Map[String, TlaEx]): Either[AssertErr, InitializedModel]
```
shonfeder, line 285: This part probably requires special note:

- @Kukovec previously asked that it be impossible to load a module that was in an invalid state, suggesting instead that the constants be initialized when the model is loaded.
- @konnov subsequently requested that we *not* initialize the constants when the model is created, and that we instead just have a single method to assert constraints, and use that also for initializing constants. We all discussed that it seems reasonable to want to load a model even when it cannot be run, as there as the model object returned has lots of useful information that users may want (including, e.g., a list of constants, which they may not know before parsing the spec).

I'm here suggesting a mix:

- Use types that will give us static guarantees internally about when a model can be run (and maybe return this data also to users?)
- Allow initializing constants after load, but use a higher-level method rather than making users go through `assertConstraint`. I opted for this given that the `TransitionExecutor` also has an `initializeConstants` method, and it seems unkind to not give our users the same helper methods we give our programmers.

But I'm of course open to changes here.

konnov (approved):
Looks great!

docs/internal/rfc/009rfc-transition-explorer.md
shonfeder: Yes, I think so. Only it's not totally clear yet to me whether we'll want different error types (allow to create distinct parts of the computation, narrowing the kind of error handling required in those segments) or if they should all just be sub-types of a single error type.

docs/internal/rfc/009rfc-transition-explorer.md
konnov: Yes, this makes sense. I have checked what `TransitionExecutorImpl.initializeConstants` is doing. Not only it uploads an assertion, but it also checks that all constants have been assigned symbolic values. However, `TransitionExecutorImpl.initializeConstants` accepts a single constraint, not a map from names to `TlaEx`. The reason is that we could add additional constraints in the expression. These additional constraints could be posed with `assert` too.

docs/internal/rfc/009rfc-transition-explorer.md
shonfeder: Ah yeah. I'll move that.

docs/internal/rfc/009rfc-transition-explorer.md
shonfeder: How will the user form the assertion `<the execution is different from the last one>`?

konnov: For instance, by adding an assertion that `tx.type` is different from the types we have in the last execution.

docs/internal/rfc/009rfc-transition-explorer.md
shonfeder: So I guess the question is whether we want the user to have to form a constraint (using, what, and EDSL that constructs TLA+ expression? Just strings that the server will parse? I'm still not clear on that bit.) in order to initialize their constants, or if we let me do the high-level of thing of just setting these particular constraint values, and then they can use `assert` if they want to add additional constraints. 

I honestly don't know which users would prefer, by my general sense is that it's nice to make simple things simple, and support more advanced usage, but not require users to do advanced things to accomplish the simple ones.

docs/internal/rfc/009rfc-transition-explorer.md
shonfeder: This poses the question, which we havne't talked about yet, of how users will form constraints. Do you imagine that to be in the form of an EDSL to construct our IR, or as just raw strings that we'll parse server side?

docs/internal/rfc/009rfc-transition-explorer.md
konnov: Yeah. I am also for keeping things simple. TLA+ specs have enough complexity already

docs/internal/rfc/009rfc-transition-explorer.md
konnov: I expect the users to use Apalache IR in the json format. This part is well documented, so it makes sense to use it.

shonfeder: Very good. So then the libraries we provide for users will probably include a module for constructing the IR. I'll clarify this TODO, and then I think we're good to merge?

konnov: I also think we are good to merge. I am not sure how we should go about IR. It is a very simple tree-like representation, which makes JSON (or s-expressions) quite natural there. Do we want to mirror this simple representation with gRPC objects?

shonfeder: Added a note on this. Agree a JSON ser/de (and s-expers for the better languages ;P) will be a good fit. I think we won't need to mirror this gRPC, we'll just delegate to the ser/de for the AST representation in the target language in the libs we provide. I've added a note on this.

PR #942, opened 2021-08-02
Include LICENSE in release builds

rnbguy:


shonfeder (approved):
LGTM

Thanks for the fix!

PR #949, opened 2021-08-06
Update features.md to clarify semantics of `@@` operator

danwt:
The fact that `f@@g` prefers `f` to `g` is very unlike similar operators in other languages, and can be a cause of headaches for beginners.


shonfeder (approved):
I find this clarification very helpful!

PR #966, opened 2021-08-28
LetPoly2: inlining polymorphic operators

### LINKED ISSUE #953: [FEATURE] Fix operator inlining to support polymorphism
Following #864, #869, and #952, we have to fix operator inlining, to correctly instantiate type variables at call sites. The type checker is now computing polymorphic types for user-defined operators, see #952.

konnov:
Closes #953: This PR fixes operator inlining of polymorphic operators. When a call to a polymorphic operator is replaced with the operator's body, we now do type unification and apply the substitution to the operator body. This happens in `InlinerOfUserOper`. Although the idea is quite simple, I had to fix several tests that previously worked with untyped IR. Additionally, this fix has uncovered a few bugs in the tests and preprocessing.


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/TypeSubstitutor.scala
```diff
@@ -0,0 +1,52 @@
...
+        if (newArgs.map(_.ID) != args.map(_.ID)) {
+          // Introduce a new operator only if the arguments have changed.
+          // Otherwise, we would introduce lots of redundant chains in ChangeListener.
+          tracker.hold(oper, OperEx(op, newArgs: _*)(oper.typeTag)) // fixes #41
+        } else {
+          oper
+        }
+
+      val reducedType = sub.subRec(oper.typeTag.asTlaType1())
+      newOper.withTag(Typed(reducedType))
```
Kukovec, line 40: Like in the line above, this also often introduces chains in ChangeListener, whenever `sub(x) = x`. You might want to consider a post-check on whether the substitution is trivial. Same for other `case` branches.

Kukovec (approved):
LGTM

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/TypeSubstitutor.scala
konnov: Good point. Fixed.

PR #971, opened 2021-08-31
Fix location of example files in record-set idiom

shonfeder:
Followup to #938

I think the writeup was moved from HOWTO into the idioms, but the example files
were left behind.

-------


konnov (approved):
Thanks for the fix!

PR #989, opened 2021-09-16
closes #985: do not skolemize let-definitions that are used as values

### LINKED ISSUE #985: [BUG] Unsound skolemization
A piece of spec that triggers an error. An MWE follows later:

```tla
  LET fail ==
     \/ ~IsValidCtrSpec(cspec) 
     \/ \E cs \in contractSpecs: cs.id = cspec.id
  IN
  /\ tx_fail' = (tx_fail \/ fail)
```

We should not skolemize `cs` in `fail`, but we do.

konnov:
Closes #985: do not skolemize let-definitions that are used as values. The solution to introduce two versions of let-definitions: one that is skolemized and another that is not skolemized. The downside is that we have redundant definitions. We will have to introduce another pass that removes unused let-definitions.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,4 +12,5 @@
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
 ### Bug fixes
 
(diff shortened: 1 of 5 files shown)


Kukovec (approved):
LGTM

PR #997, opened 2021-09-21
Remove unused let definitions

### LINKED ISSUE #995: [FEATURE] Remove unused let-definitions
The recent fix of Skolemization #985 introduces two let-definitions for each let-definition: a skolemized one and an unskolemized one. While the produced TLA+ specification is equivalent to an original one, it sometimes contains unused let-definitions. This may lead to performance issues, as both definitions are still translated to SMT.

konnov:
closes #995: remove unused let-definitions. This is the main contribution of this PR. It is done in `LetInOptimizer`. There is only a unit test, `TestLetInOptimizer`, not integration tests. It's hard to test that some let-in definitions are removed. The PR visibly improves performance on some of our benchmarks.

Another contribution is `TlaExTouchTransformation`, as a better version of `TlaExTransformation`. Check how it is used in `LetInOptimizer`.


Kukovec (approved):
Looks fine, but we might want to consider a less ad-hoc implementation of checking for no-op transformations in the future. There are many places this information could be used.

PR #1032, opened 2021-10-10
Add nix-shell support

JonathanLorimer:
This PR adds support for a development nix shell. Instructions on how to install nix can be found [here](https://github.com/informalsystems/cosmos.nix#setup), although there is no cache support yet. This should allow a developer to have a complete dev environment by running `nix develop` from the project root.

NOTE: All the build and test steps still use maven, you can just run those commands from within the dev shell with all the dependencies (jdk8, scala 2.12, etc) pinned. The dev shell does come with a `scalafmt` pre-commit hook though, this will only be present after you enter the dev shell, and it is git ignored.


shonfeder:
Thanks, @JonathanLorimer! This is great :)

I see that the documentation TODO is checked, but I don't actually see any documentation associated. Were you planning on adding something? If not, I can add whatever I think is appropriate (probably just a few lines in the `CONTRIBUTING.md`).

shonfeder:
No worries! For now, I don't think we need much more in the way of documentation than a link to the `cosmos.nix` `README.md` and instructions and the invocation to build and run the shell. Oh, and probably the most important think is noting how to take advantage of the dev env when using IntelliJ, which most of the team uses.

Once I figure out out how to get a nice, seamless integration with emacs working, I can add that myself.

PR #1044, opened 2021-10-14
First version of the ADR describing the SMT encoding with arrays

### LINKED ISSUE #1041: [FEATURE] Document the handling of sets in the SMT encoding with arrays
Add an ADR to describe the new encoding. Initially it will only describe the handling of sets.

rodrigo7491:

First version of this ADR, the encoding of operators other than sets will be added in the future. Closes #1041.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
         <scalaBinaryVersion>2.12</scalaBinaryVersion>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <logback.version>1.2.6</logback.version>
-        <spotless.version>2.17.1</spotless.version>
+        <spotless.version>2.17.2</spotless.version>
     </properties>
 
     <repositories>


rodrigo7491:
> That's great! Let's start small and produce relatively small pull requests, as soon as you have unit tests working for the smallest unit of work possible. There is no problem merging your code in the main trunk, as long as it does not break any tests. I would also focus on the fragment that supports Booleans, integers, maybe string constants, and sets. For the first three you can just use the existing rules. For the sets, you have to define new rules. For starters, remove all other rewriting rules in `SymbStateRewriterImplWithArrays`, so you don't have to worry about them.

My idea was to keep all the rules as they are, to begin with, and gradually substitute them by new ones. I don't think all rules will need to be replaced, and this way we avoid code duplication. Does this make sense @konnov ?

> This version of ADR is good enough to start. Go ahead, merge it, and iterate. We can think about the other parts of the encoding later.

Will just make some minor changes based on the feedback and then merge it, thanks!

Issue #1069, opened 2021-11-02
[FEATURE] Route all configuration management through a single component


shonfeder:
This ticket records a plan hatched in conversation with @Kukovec today.

The plans for the `OutputManager` introduced in #1025, and implemented in followups, calls for a "cascade loading" configuration system. The configuration is derived from reading

1. CLI arguments
2. environment variables
3. A local configuration file (perhaps with the location overridden by a CLI flag)
4. A global configuration file 

The final configuration is computed based on the priority of the source.

Currently, we are doing this in a somewhat ad hoc way. But gathering all of this logic into a configuration management sub-system will help reduce the complexity of the `OutputManager` and contribute to the current plan to move it away from a singleton, into something we can put in DI, that gets all of its configurations from the `PassOptions`.

With the server mode coming down the pipe, we'll have more need for a configuration system that is easy to maintain.

Scope of work expected to resolve this ticket:

- [x] Briefly research existing configuration management systems used in the Scala ecosystem.
- [x] Decide whether to use one of the existing solutions or roll our own.
- [x] Write a short ADR recording the decisions.
- [ ] Move the configuration parsing out of the `OutputManager` and into the new subsystem.

PR #1068, opened 2021-11-02
Fix 3 compiler warnings

shonfeder:
Towards #1064

See the individual commits for details on each fix.

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] Documentation added for any new functionality
- [x] Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality

PR #1073, opened 2021-11-03
Fix regression #1072 - parse output file not being created

### LINKED ISSUE #1072: [BUG] cli `parse` command output is written to intermediate directory but not copied to output directory
Appears in _APALACHE version 0.17.0 build f62f370_

## Description

The command `apalache-mc parse --output=<filename> <tlafile>` does not actually write the output to $filename in the output directory. In fact, it actually writes the output to the intermediate directory, which is only visible if passing `--write-intermediate`.

`apalache-mc parse --output=<filename> <tlafile>` 

## Expected behavior

A file <filename> should be created in the output directory.

## System information

- Apalache version [`apalache-mc version`]: APALACHE version 0.17.0 build f62f370
- OS [e.g. Ubuntu Linux or Mac OS, and the current version]: OSX 11.5
- JDK version [e.g. OpenJDK 0.8.0]: OpenJDK Runtime Environment Zulu16.32+15-CA (build 16.0.2+7)

## Additional context

Reproduce with file main.tla

```
------------------------------------ MODULE main -----------------------------

\* @type: () => Bool;
Init ==
    /\ TRUE

Next == TRUE

===============================================================================
```

```
tree 3
.
├── apalache-pkg-0.17.0-full.jar
└── main.tla

java -jar apalache-pkg-0.17.0-full.jar parse --output=foobar.tla main.tla

tree 3
.
├── _apalache-out
│  └── main.tla_2021-11-03T19-20-28_10048578539153344840
│     ├── detailed.log
│     └── run.txt
├── apalache-pkg-0.17.0-full.jar
└── main.tla


java -jar apalache-pkg-0.17.0-full.jar --write-intermediate=true parse --output=foobar.tla main.tla

tree 4

.
├── _apalache-out
│  └── main.tla_2021-11-03T19-22-08_5579138356771785926
│     ├── detailed.log
│     ├── intermediate
│     │  ├── 00_OutParser.json
│     │  ├── 00_OutParser.tla
│     │  └── foobar.tla
│     └── run.txt
├── apalache-pkg-0.17.0-full.jar
└── main.tla
```

shonfeder:
Closes #1072

Follow up to #1025, which caused a regression in the handling of the `--output` flag. Instead of writing the specified file, as it used (and as it seems it should), it was only writing to the `intermediate` subdirectory of the run directory, and then only when the `--write-intermediate` flag was also given.

This restores the previous behavior. I think the guiding principle for this kind of UI decision re: output files might be something like:

- Files that we generate automatically during runs should only ever be written to the rundir
- But if we are allowing users to get specific information out of apalche written to a file, then we should allow them to specify any file that for this output that is convenient for their process.

This PR also introduces a few utilities to the OutputManager to support the fix.

-------

### DIFF
--- a/test/tla/cli-integration-tests.md
+++ b/test/tla/cli-integration-tests.md
@@ -2081,5 +2081,7 @@ $ rm -rf ./test-out-dir
 
 ```sh
 $ apalache-mc check --out-dir=./test-out-dir --profiling=true --length=0 Counter.tla | sed 's/[IEW]@.*//' && test -s ./test-out-dir/*/profile-rules.txt
+...
+EXITCODE: OK
 $ rm -rf ./test-out-dir
 ```
--- a/test/tla/cli-integration-tests.md
+++ b/test/tla/cli-integration-tests.md
@@ -181,8 +181,8 @@ $ rm output.tla
 $ apalache-mc parse --output=output.json Annotations.tla | sed 's/I@.*//'
 ...
 EXITCODE: OK
-$ test -s output.tla
-$ rm output.tla
+$ test -s output.json
+$ rm output.json
 ...
 ```
 
@@ -2002,7 +2002,9 @@ EXITCODE: OK
 ### set out-dir by CLI flag
 
 ```sh
-$ apalache-mc check --out-dir=./test-out-dir --length=0 Counter.tla &> /dev/null && find ./test-out-dir -type f -exec basename {} \; | ./sort.sh
+$ apalache-mc check --out-dir=./test-out-dir --length=0 Counter.tla | sed 's/[IEW]@.*//' && find ./test-out-dir -type f -exec basename {} \; | ./sort.sh
+...
+EXITCODE: OK
 detailed.log
 log0.smt
 run.txt
@@ -2012,7 +2014,9 @@ $ rm -rf ./test-out-dir
 ### set out-dir by envvar
 
 ```sh
-$ OUT_DIR=./test-out-dir apalache-mc check --length=0 Counter.tla &> /dev/null && find ./test-out-dir -type f -exec basename {} \; | ./sort.sh
+$ OUT_DIR=./test-out-dir apalache-mc check --length=0 Counter.tla | sed 's/[IEW]@.*//' && find ./test-out-dir -type f -exec basename {} \; | ./sort.sh
+...
+EXITCODE: OK
 detailed.log
 log0.smt
 run.txt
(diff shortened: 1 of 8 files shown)


shonfeder:
(The PR is currently based on #1068, just because it touches one overlapping line and I want to avoid a merge conflict. Hopefully we can get that dependency merged in, then I'll rebase this before review.)

test/tla/cli-integration-tests.md
```diff
@@ -1998,7 +2002,9 @@ EXITCODE: OK
 ### set out-dir by CLI flag
 
 ```sh
-$ apalache-mc check --out-dir=./test-out-dir --length=0 Counter.tla &> /dev/null && find ./test-out-dir -type f -exec basename {} \; | ./sort.sh
+$ apalache-mc check --out-dir=./test-out-dir --length=0 Counter.tla | sed 's/[IEW]@.*//' && find ./test-out-dir -type f -exec basename {} \; | ./sort.sh
+...
+EXITCODE: OK
```
shonfeder, line 2007: These changes just let us get more helpful feedback when the outputs fail due to a crash or something.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputManager.scala
```diff
@@ -150,4 +151,68 @@ object OutputManager {
...
+   * @param f a function that will be applied to the the PrintWriter, if the `IntermediateFlag` is set.
+   * @returns `true` if the `IntermediateFlag` is set the output is written, otherwise `false`
+   * If the IntermediateFlag is true, then this applies `f` to the PrintWriter
+   * created by appending the `parts` to the intermediate output dir, and returns `true`. Otherwise, it is false
+   */
+  def withWriterRelativeToIntermediateDir(parts: String*)(f: PrintWriter => Unit): Boolean = {
+    if (flags(Names.IntermediateFlag)) {
+      val intermediateDir = new File(runDirOpt.get.toFile(), Names.IntermediateFoldername)
+      if (!intermediateDir.exists()) {
+        intermediateDir.mkdir()
```
konnov, line 209: `mkdir` may return `false`, if it is unable to create a directory. Do we have to do anything about that? I guess, the line below will fail with an IO exception in this case, which may be sufficient.

konnov (approved):
Looks great! Thanks for fixing this issue. It is an important feature for the integration with MBT and other tools.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputManager.scala
shonfeder: Good call. Would be nicer to give a nice user error in this case.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputManager.scala
```diff
@@ -150,4 +168,72 @@ object OutputManager {
...
+        case Some(dir) => {
+          withWriter(f)(printWriter(dir, parts: _*))
+          true
+        }
+        case None => {
+          val dirName = intermediateDirPathOpt.getOrElse("")
+          logger.error(
+              s"Unable to find or create intermediate output directory ${dirName}. Intermediate output will not be written.")
+          false
+        }
```
shonfeder, line 235: @konnov I've opted for logging an error rather than crashing in the event of failure. The rationale being that if you're in the middle of a long running checking session, you may not want to lose all your progress and results just because this file cannot be created for some reason.

Does that seem sensible, or should it rather raise an exception?

It just occurs to me, it would be better if we just ensure all the directories needed for the configured output exist at startup, then error out early if they don't. However, I think that reworking of the initialization of the `OutputManager` is probably out of scope for this PR, and should instead be included in #1069. How does that sound?

konnov: I think it's a good compromise.

PR #1074, opened 2021-11-04
rm -rf TlaEx.toSimpleString

### LINKED ISSUE #437: [FEATURE] Refactor toString and remove toSimpleString
`TlaEx` has two methods for printing expressions:

```scala
  sealed abstract class TlaEx extends Identifiable with Serializable {
    // TODO: there must be a nice way of defining default printers in scala, so we do not have to make a choice here
    override def toString: String =  UTFPrinter( this )

    def toSimpleString: String = ""
    ...
  }
```

I have always been finding this problematic. Moreover, we have multiple printers now. We should have a canonical `toString` method that is only used in logging, debugging, etc. When we want to pretty print an expression, we should use `PrettyWriter` or `UTFPrinter`.

Shall we do the following:
   - let `toString` output the canonical (ASCII) representation,
   - remove `toSimpleString`

@shonfeder , what is your opinion?

Kukovec:

Removes the unused `toSimpleString` method from `TlaEx`. While #437 also suggests 
>let toString output the canonical (ASCII) representation,

I find this to be pointless as everything in our technology stack is perfectly capable of decoding non-ASCII.

Closes #437


konnov (approved):
I agree that we should remove `toSimpleString`. The remaining question is what `toString` should print. The UTF printer is our own invention, it is not compatible with the rest of the TLA+ tooling. Since these strings may propagate in error messages, perhaps, we should output ASCII. @shonfeder what is your opinion?

Kukovec:
> 
> 
> I agree that we should remove `toSimpleString`. The remaining question is what `toString` should print. The UTF printer is our own invention, it is not compatible with the rest of the TLA+ tooling. Since these strings may propagate in error messages, perhaps, we should output ASCII. @shonfeder what is your opinion?

For any outputs (e.g. TLA files) we use a specialized writer anyway, so that shouldn't be an issue w.r.t. tooling. Especially because these strings show up in error messages, to be read by end users (humans), there's value in non-ASCII.

shonfeder:
I'm kind of the fence on this one. I've also gotten value out of unicode characters in various places, but I have also encountered numerous subtle issues with unicode support still poping up in various editors or different OSs or terminals, and there was just that big exploit using unicode to disguise malicious code. So... yeah, not sure what to prioritize here :)

shonfeder:
I can add: the concern around supporting the many people who read and write languages with scripts outside of the unicode set is arguably the most important consideration imo. So if we do allow unicode in the parsed specs, then I guess we should always support it in our outputs to be sure users who might use other scripts in their specs don't lose info?

konnov:
I think we should merge this one and open an issue on whether we want to use ASCII or unicode as the standard output format in Apalache.

PR #1078, opened 2021-11-05
Separate sorts for uninterpreted types/model values

### LINKED ISSUE #570: [FEATURE] Add support for different uninterpreted types in Rewriter
The type checker allows us to use uninterpreted types such as `Set(RM)` or `Data -> Data`. The model checker currently supports only strings as the only uninterpreted type. We have to add the support for different uninterpreted types in the model checker, which would translate to different uninterpreted sorts in SMT.

Kukovec:

Closes #570


docs/src/HOWTOs/uninterpretedTypes.md
```diff
@@ -0,0 +1,66 @@
...
+```tla
+\* @type: UTNAME;
+```
+
+where `UTNAME` matches the pattern `[A-Z_][A-Z0-9_]*`, and is not a type alias defined elsewhere.
+
+## How to introduce values belonging to an uninterpreted type
+Apalache uses the following convention-based naming scheme for values of uninterpreted types:
+```tla
+"PREFIX_TYPENAME_IDENTIFIER"
```
konnov, line 33:
```suggestion
"uval_TYPENAME_IDENTIFIER"
```

docs/src/HOWTOs/uninterpretedTypes.md
```diff
@@ -0,0 +1,66 @@
...
+```
+where:
+  - `PREFIX` is a prefix shared by all values holding uninterpreted types: `uval`
```
konnov, line 36:
```suggestion
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateDecoder.scala
```diff
@@ -75,21 +76,40 @@ class SymbStateDecoder(solverContext: SolverContext, rewriter: SymbStateRewriter
...
-            tla.str(cell.toString).typed()
-        }
+    case ConstT(utype) =>
+      utype match {
+        case "" =>
+          val found = rewriter.strValueCache.findKey(cell)
+          if (found.isDefined) {
+            tla.str(found.get).typed()
+          } else {
+            findCellInSet(arena, rewriter.strValueCache.values().toSeq, cell.toNameEx) match {
```
konnov, line 86: What is the difference between `rewriter.strValueCache.findKey` and `findCellInSet`? A comment would help.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/caches/ModelValueCache.scala
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A cache for model values that are translated as uninterpreted constants, with a unique sort per uniterpreted type.
+ * Since two values are equal iff they are literally the same string, we force
+ * inequality between all the respective SMT constants.
+ *
+ * @author Jure Kukovec
+ */
+class ModelValueCache(solverContext: SolverContext)
+    extends AbstractCache[Arena, (String, String), ArenaCell] with Serializable {
```
konnov, line 18: Why do we need both `ModelValueCache` and `StrValueCache`? For strings, we can use `ModelValueCache` with the type `Str`. There will be no collision on type names, so this should work. We would have to maintain one class instead of two.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/caches/ModelValueCache.scala
```diff
@@ -0,0 +1,35 @@
...
+    val (utype, _) = typeAndIndex
+    val newArena = arena.appendCell(ConstT(utype))
+    val newCell = newArena.topCell
+    // the freshly created cell should differ from the others
+    for (
+        other <- values().withFilter { v =>
+          v.cellType == ConstT(utype) //compare just with same sorted cells
+        }
+    ) {
+      solverContext.assertGroundExpr(tla.neql(newCell.toNameEx, other.toNameEx))
```
konnov, line 30: We should use `distinct`. There is `Apalache.distinct` for that reason.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/types/package.scala
```diff
@@ -225,14 +228,15 @@ package object types {
   /**
    * A cell constant, that is, just a name that expresses string constants in TLA+.
    */
-  sealed case class ConstT() extends CellT with Serializable {
-    override val signature: String = "str"
+  sealed case class ConstT(utype: String = "") extends CellT with Serializable {
```
konnov, line 231: I would prefer using `utype = "Str"` for the generic strings. This would give you a direct way of using `ModelValueCache`. And you would not need all these if-then-else tests below.

konnov (approved):
That's a great feature, which should improve performance. There are several ways to improve this PR, see the comments.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateDecoder.scala
Kukovec: This is literally unchanged code for the string case, you wrote this part.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/caches/ModelValueCache.scala
Kukovec: Yeah, that makes sense. I started out with a different idea for the model cache, but I guess in the latest iteration it can basically replace strings, since they are just 1 standard "uninterpreted type"

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/types/package.scala
Kukovec: I agree, though I'd go with lowercase `str`

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateDecoder.scala
konnov: I have no idea about what it is doing. Worth writing a comment, if you understood it? :))

PR #1081, opened 2021-11-05
Write output to stable locations

### LINKED ISSUE #1077: [FEATURE] Produce latest output in a known location
I have tried to upgrade [modelator](https://github.com/informalsystems/modelator) to the recent version of Apalache, but it turned out to be not as simple as I hoped for :smile:  

For some reason the Apalache output behavior has changed: now, instead of just producing an output in the current directory, as it was before, it produces it in a subdirectory like `_apalache-out/NumbersAMaxBMinTest.tla_2021-11-04T14-07-40_6241382755432900060/` , i.e. a combination of `_apalache-out`, the name of the file, plus a timestamp. The problem is that this is not printed on the output of the tool, it still prints

```
Check the counterexample in: counterexample1.tla, MC1.out, counterexample1.json
```

This makes it very difficult to interact with the tool, both for the command-line user, and automatically. E.g. if a user runs the tool multiple times, the `_apalache-out`  dir will be populated with multiple directories, and it is difficult to navigate there. Conversely, from `modelator` side, we parse the tool output, but there is no way to deduce from the output where the counterexample file is actually located. I understand that this change comes from a desire not to clog the current directory, but this just goes to the other extreme. 

Would it be possible to provide an intermediate solution? I propose the following:
 - produce the output in the directory structure as it is now
 - _additionally_, produce the _latest_ output in the directory `_apalache-out`, without subdirectories
 - print in the tool output `Check the counterexample in: _apalache-out/counterexample1.tla, _apalache-out/MC1.out, _apalache-out/counterexample1.json`

This will make it easy and clear both for the user and for other tools to deduce where the latest output is located.

P.S. I am aware of the `--out-dir` option, but this doesn't solve the problem, as the artifacts are still produced in a subdirectory with a timestamp in the name.

shonfeder:
Closes #1077

- Adds namespaces for each specification/command within the output directory
- Adds a `--run-dir` flag that lets users specify a stable directory into which all run outputs are written

See the integration tests and ADR update for details.
 

### DIFF
--- a/docs/src/adr/009adr-outputs.md
+++ b/docs/src/adr/009adr-outputs.md
@@ -72,19 +72,3 @@ In other words, parameter values are determined in the following way, by order o
   1. If `--<flag>=<value>` is given, use `<value>`, otherwise
   2. If `apalache-global-config` specifies `<flag>=<CFGvalue>` use `<CFGvalue>`, otherwise 
   3. Use Apalache defaults.
-
-## 5. Output files
-
-Some commands write specific files that are used to communicate with other
-programs. E.g., 
-
-- `parse` writes files with the results of parsing the input spec
-- `check` writes counterexamples used by MBT
-
-To facilitate programmatic use of this data, such commands also take an
-`--output=<FILE>` flag, and then write the relevant output to the given file.
-These output files are usually not in the `out-dir`.
-
-The format for the output files is determined by the extension of the specified
-file. E.g., `--output=foo.json` should result in JSON being written to file
-`foo.json` (if supported for the command).
--- a/docs/src/adr/009adr-outputs.md
+++ b/docs/src/adr/009adr-outputs.md
@@ -21,20 +21,14 @@ All such optional flags should default to `false`.
 ## 2. Output directory and run directories
 Apalache should define an `out-dir` parameter, which defines the location of all outputs produced by Apalache. If unspecified, this value should default to the working directory, during each run, but it should be possible to designate a fixed location, e.g. `<HOME>/apalache-out/`.
 
-Each run looks a subdirectory inside of the `out-dir` with the same name as the
-principle file provided as input or, for commands that do not input a file,
-named after the subcommand executed. This subdirectory is called the
-specification's (resp.  file's, command's) *namespace* within the `out-dir`. All
-outputs originating from that input file (resp. command) will be written in its
+Each run looks for a subdirectory inside of the `out-dir` with the same name as
+the principle file provided as input (or, for commands that do not read input
+from a file, named after the executed subcommand). This subdirectory is called
+the specification's (resp. command's) *namespace* within the `out-dir`.  All
+outputs originating from that input file (resp. command) will be written to this
 namespace.
 
-Within each namespace is a subdirectory `latests`. This directory is cleaned on
-each run, and holds the output of the most recent run. This provides a stable
-location from which users can, e.g., open the counterexample files in their
-editor and have them updated on each run (thanks to Andrey for this idea:
-[#1077](https://github.com/informalsystems/apalache/issues/1077)).
-
-Finally, Each run produces a subdirectory in its namespace, with the following name:
+Each run produces a subdirectory in its namespace, with the following name:
 
 ```
 <DATE>T<TIME>_<UNIQUEID>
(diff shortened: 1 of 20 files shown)


shonfeder:
Sorry for the size here.

Fortunately much of the bulk are documentation updates and integration test output lines tho.

shonfeder:
@konnov and @Kukovec: I'm off next week, so won't be available to reply to change requests. In the event that you're generally OK with the basic idea but want to fix or adjust any parts of the implementation, please feel free to do so. My aim was to get this in to unblock https://github.com/informalsystems/modelator/issues/119, but I obviously didn't quite make it :/

shonfeder:
bah. the test failure is trivial, due to different paths output between all the other tests and the ones output from the docker container. 

As a workaround for this, might just remove the file output locations from the integration tests.

shonfeder:
Ah, I've just realized that we cannot simply write a single counterexample file, because we output `n` of these files depending on the circumstance. So I guess the `--output` argument should be removed from the `check` command in this PR or else reworked to specify a directory into which all counterexamples are written.

andrey-kuprianov:
> Ah, I've just realized that we cannot simply write a single counterexample file, because we output `n` of these files depending on the circumstance. So I guess the `--output` argument should be removed from the `check` command in this PR or else reworked to specify a directory into which all counterexamples are written.

I think both options are fine. In our current usage of Apalache it is only important that we can find the files in the location which is permanent wrt. the cwd. This works with both having the files in the `_apalache-out/latest`, or supplying cwd into the `--output`, so it's up to you which option is easier to implement.

konnov:
> I think both options are fine. In our current usage of Apalache it is only important that we can find the files in the location which is permanent wrt. the cwd. This works with both having the files in the _apalache-out/latest, or supplying cwd into the --output, so it's up to you which option is easier to implement.

Would not `--out-dir` solve your problem? You can directly give the name of the output directory there. This option was designed for scripting.

docs/src/adr/009adr-outputs.md
```diff
@@ -19,38 +19,72 @@ Each optional category is associated with a flag: `--write-intermediate` for int
...
+named after the subcommand executed. This subdirectory is called the
+specification's (resp.  file's, command's) *namespace* within the `out-dir`. All
+outputs originating from that input file (resp. command) will be written in its
+namespace.
+
+Within each namespace is a subdirectory `latests`. This directory is cleaned on
+each run, and holds the output of the most recent run. This provides a stable
+location from which users can, e.g., open the counterexample files in their
+editor and have them updated on each run (thanks to Andrey for this idea:
+[#1077](https://github.com/informalsystems/apalache/issues/1077)).
```
Kukovec, line 35: How is this meant to work with parallel runs of Apalache?

shonfeder:
In meeting today, we decided: 

Remove `latest` directory, and replace `--output` flag introduced here with `--rundir` flag, that will allow specifying a run-directory for all outputs.

shonfeder:
@Kukovec I've reworked this to just take a `--run-dir` command line argument, as you suggested. The `latest` directory has been dropped. Please take a look when you're ready!

shonfeder:
(I've set this PR to merge on approval, so please don't mark it as approved unless you're OK with it being merged in directly!)

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
```diff
@@ -37,7 +40,8 @@ class LogbackConfigurator(runDirOpt: Option[Path]) extends ContextAwareBase with
     val consoleAppender = mkConsoleAppender(loggerContext)
     // only warnings at the root level
     rootLogger.setLevel(Level.WARN)
-    rootLogger.addAppender(mkFileAppender(loggerContext))
+    runDir.foreach(d => rootLogger.addAppender(mkFileAppender(loggerContext, d.resolve("detailed.log").toFile())))
+    customRunDir.foreach(d => rootLogger.addAppender(mkFileAppender(loggerContext, d.resolve("detailed.log").toFile())))
```
Kukovec, line 44: You can simplify this with option magic:
```suggestion
    (runDir ++ customRunDir).foreach(d => rootLogger.addAppender(mkFileAppender(loggerContext, d.resolve("detailed.log").toFile())))
```

Kukovec (commented):
LGTM, I have one suggestion, and the only reason I'm not immediately setting it as approved is the auto-merge. Feel free to ignore the suggestion and ping me to mark it as approved.

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
shonfeder: That is cleaner. Thanks!

shonfeder:
Change added! Please do green check if you are happy here, at your convenience. Thanks :)

Kukovec (approved):

PR #1093, opened 2021-11-12
Add support for sets to the arrays encoding

### LINKED ISSUE #1092: [FEATURE] Add support for sets to the SMT encoding with arrays
Add the core of the SMT encoding with arrays, the handling of sets. The changes to be implemented are described in the ADR11 (see #1041).

rodrigo7491:

This PR contains the initial version of the arrays encoding. It adds support for basic types, which are unchanged from the original encoding, as well as sets. A number of encoding rules was changed, either explicitly, through new `RewritingRule` classes, or implicitly, through changes to `Z3SolverContext` that make the SMT constraints produced be different from those of the original rules. A number of existing unit and integration tests are now running with the arrays encoding. Below is an overview of the changes. Closes #1092.

- **SolverConfig**. A new field was added to the solver configuration, to record the encoding being used. This field is set based on the `check` command options. 
- **SolverContext**. The core of the changes is in `Z3SolverContext`, as this is where the SMT constraints are made. See the updated ADR for details.
- **SymbStateRewriter**. Three new explicit rules were added to `ruleLookupTable` in `SymbStateRewriterImplWithArrays`: `EqRuleWithArrays`, `SetInRuleWithArrays`, and `SetInclusionRuleWithArrays`.
- **Unit tests**. The unit tests were refactored to account for the two possible encodings in the solver configuration. The most relevant refactoring change was in `RewriterBase`. To see which unit tests were enabled to run with the arrays encoding, check `TestArenaWithArrays`, `TestRewriterWithArrays`, and `TestRecordingSolverContextWithArrays`.
- **Integration tests**. Some tests were enabled to run with the arrays encoding, by tagging them with `array-encoding` in `cli-integration-tests`. The CI setup to run the tagged tests is not part of this PR.
- **Documentation**. ADR11 was updated to reflect the implementation changes made. Its high level description was not changed.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
```diff
@@ -0,0 +1,83 @@
...
+        assert(expectedType == actualType)
+    }
+    // double check that the type finder did its job
+    checkType(powsetCell.cellType, elemCell.cellType)
+
+    val leftElems = state.arena.getHas(elemCell)
+    val rightDomain = state.arena.getDom(powsetCell)
+    val rightDomainElems = state.arena.getHas(rightDomain)
+    // EqConstraints need to be generated, since missing in-relations, e.g. in sets of tuples, will lead to errors.
+    // TODO: Inlining this method is pointless. We should consider handling tuples and other structures natively in SMT.
```
konnov, line 26: Good point. We should figure out how to deal with records that can be used as discriminated unions. Perhaps, ADTs should help us.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
```diff
@@ -0,0 +1,83 @@
...
+    def checkType: PartialFunction[(CellT, CellT), Unit] = {
+      case (PowSetT(FinSetT(expectedType)), FinSetT(actualType)) =>
+        assert(expectedType == actualType)
+    }
+    // double check that the type finder did its job
+    checkType(powsetCell.cellType, elemCell.cellType)
+
+    val leftElems = state.arena.getHas(elemCell)
+    val rightDomain = state.arena.getDom(powsetCell)
+    val rightDomainElems = state.arena.getHas(rightDomain)
```
konnov, line 24: I find the names "left" and "right" confusing, especially that it is in the order opposite to the order of arguments. Shall we call them `setDomain`, `setDomainElems`, and`powsetElems`?

konnov (commented):
The first batch of comments. I will continue today.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterAutoWithArrays.scala
```diff
@@ -2,4 +2,6 @@ package at.forsyte.apalache.tla.bmcmt
 
 import at.forsyte.apalache.tla.bmcmt.smt.SolverContext
 
-class SymbStateRewriterAutoWithArrays(_solverContext: SolverContext) extends SymbStateRewriterAuto(_solverContext) {}
+class SymbStateRewriterAutoWithArrays(_solverContext: SolverContext) extends SymbStateRewriterAuto(_solverContext) {
+  override protected val impl = new SymbStateRewriterImplWithArrays(solverContext, exprGradeStore)
+}
```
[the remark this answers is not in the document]
rodrigo7491: I'm not sure if we strictly need it, I simply saw that it uses a `SymbStateRewriter` and I extended it. It might break `create` in `RewriterBase` if there is a mismatch between the state rewriter and the solver context.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
rodrigo7491: I meant "left" and "right" here as the order in the `\in` operator. I think you suggestion does make things more clear. Thanks!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
```diff
@@ -0,0 +1,83 @@
...
+      val pred = newState.arena.topCell
+
+      def isInAndEqSetElem(powsetDomainElem: ArenaCell) = {
+        simplifier.simplifyShallow(tla.and(tla.in(powsetDomainElem.toNameEx, powsetDomain.toNameEx),
+                tla.eql(powsetDomainElem.toNameEx, setElem.toNameEx)))
+      }
+
+      val elemsInAndEqLeftElem = powsetDomainElems.map(isInAndEqSetElem)
+      rewriter.solverContext.assertGroundExpr(simplifier.simplifyShallow(tla.equiv(pred.toNameEx,
+                  tla.and(tla.in(setElem.toNameEx, elemCell.toNameEx), tla.or(elemsInAndEqLeftElem: _*)))))
```
konnov, line 40: Why is it `and` and not an implication?

konnov: If it's correct, a comment that describes the formula, e.g, as a TLA+ formula, would help.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/SolverConfig.scala
```diff
@@ -7,15 +7,16 @@ package at.forsyte.apalache.tla.bmcmt.smt
  * @param debug Enable the debug mode (activated with --debug). Write the log file to log%d.smt.
  * @param profile Enable the profile mode (activated with --profile). Report the metrics.
  * @param randomSeed The random seed to be passed to z3 as :random-seed.
+ * @param smtEncoding The SMT encoding to be used.
  *
  * @author Igor Konnov
```
konnov, line 12:
```suggestion
 * @author Igor Konnov, Rodrigo Otoni
```
```suggestion
 * @author Igor Konnov
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -129,9 +146,18 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext {
     val cellSort = getOrMkCellSort(cell.cellType)
     log(s"(declare-const $cell $cellSort)")
     val cellName = cell.toString
-    z3context.mkConstDecl(cellName, cellSort)
```
konnov, line 132: why is it removed? Was it redundant?

konnov (commented):
More comments. More yet to come :)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
rodrigo7491: The original rule uses implication (see `LazyEquality.subsetEq`), but I decided to use conjunction here because I think there is no case in which `tla.in(setElem.toNameEx, elemCell.toNameEx)` evaluates to false. In the beginning of the method we explicitly have `setElems = state.arena.getHas(elemCell)`.

I'm not sure what a TLA+ formula of this would be, since this rule is applied for, e.g., {1} \in SUBSET {1,2}. Is this what you mean?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/SolverConfig.scala
rodrigo7491: I didn't know if making such a change to a file warranted being added as an author.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
rodrigo7491: I believe so, `mkConstDecl` returns a declaration (see line 535 in the [Z3 API](https://github.com/Z3Prover/z3/blob/master/src/api/java/Context.java)) that is not asserted, so this line does not do anything. If you look at the original code of `declareInPredIfNeeded`, it also uses `mkConst`, but without `mkConstDecl`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
konnov: Well, how about the result of `{1} \ {1}`? The arena would point to one cell (the removed `1`), but that cell would not be in the set (difference). I think you need an implication.

konnov: > I'm not sure what a TLA+ formula of this would be, since this rule is applied for, e.g., {1} \in SUBSET {1,2}. Is this what you mean?

What I mean is that sometimes it is useful to write in comments the constraints that you are constructing. TLA+ is still more readable than the expression in `assertGroundExpr`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
konnov: Most likely, it is a left-over from the outdated code then

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/EncodingBase.scala
```diff
@@ -0,0 +1,6 @@
+package at.forsyte.apalache.tla.bmcmt
+
+trait EncodingBase {
+  protected val oopsla19EncodingType = "oopsla19"
+  protected val arraysEncodingType = "arrays"
+}
```
Kukovec, line 6: Any particular reason why the fields are protected in a trait and not public in an object?
```suggestion
object EncodingBase {
  val oopsla19EncodingType = "oopsla19"
  val arraysEncodingType = "arrays"
}
```

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/TestRewriterWithOOPSLA19.scala
```diff
@@ -17,6 +18,11 @@ class TestRewriterWithOOPSLA19
     with TestSymbStateRewriterTlc with TestSymbStateRewriterTuple with TestPropositionalOracle with TestSparseOracle
     with TestUninterpretedConstOracle {
   override protected def withFixture(test: OneArgTest): Outcome = {
-    test(oopsla19RewriterType)
+    solverContext = new PreproSolverContext(new Z3SolverContext(SolverConfig.default.copy(debug = true,
+                smtEncoding = oopsla19EncodingType)))
+    arena = Arena.create(solverContext)
+    val result = test(oopsla19EncodingType)
+    solverContext.dispose()
+    result
```
konnov, line 26: This code will probably appear in many tests. Does it make sense to refactor this method into a trait?

konnov (commented):
This looks very good. See the other comments. As far as I understand, this PR contains the rules for set equality, set membership, and subseteq. Are other set operators planned for a separate PR?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/SolverConfig.scala
konnov: It's a tiny file. We don't have precise rules about authorship. It's like if you make substantial changes to the code, which in this case applies :)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
rodrigo7491: > Well, how about the result of `{1} \ {1}`? The arena would point to one cell (the removed `1`), but that cell would not be in the set (difference). I think you need an implication.

I believe in this case the arena would point to a set cell with no edges, i.e. the set cell does not point to any other cells, no? Before `powSetIn` is called, `apply` calls `rewriter.rewriteUntilDone` twice, passing the lhs and rhs arguments, so this should not be an issue if the other rules are working as intended. We would have `{1} \in SUBSET ({1} \ {1})` rewritten as `{1} \in SUBSET ({ })`, since the `SUBSET` expression is not evaluated, but its argument is.

I tried a simple example here and adding `{1} \in SUBSET ({1} \ {1})` as a constraint to the spec leads to a deadlock, while adding `{1} \in SUBSET ({1} \ {2})` is fine, as expected. If I'm missing something, let me know.

> What I mean is that sometimes it is useful to write in comments the constraints that you are constructing. TLA+ is still more readable than the expression in `assertGroundExpr`.

I do find the `assertGroundExpr` quite readable, but I see your point. I did not add comments regarding the constraints being constructed because I didn't see any on the existing rules and simply followed that.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImplWithArrays.scala
```diff
@@ -3,7 +3,120 @@ package at.forsyte.apalache.tla.bmcmt
...
+        key(tla.funDef(tla.name("e"), tla.name("x"), tla.name("S")))
+          -> List(new FunCtorRule(this)),
+        key(tla.appFun(tla.name("f"), tla.name("x")))
+          -> List(new FunAppRule(this)),
+        // tuples, records, and sequences
+        key(tla.tuple(tla.name("x"), tla.int(2)))
+          -> List(new TupleOrSeqCtorRule(this)),
+        key(tla.enumFun(tla.str("a"), tla.int(2)))
+          -> List(new RecCtorRule(this))
+    )
```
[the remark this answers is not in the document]
rodrigo7491: The end goal is to do the overwriting as you described, once all rules are accounted for. I've done this because I wanted to only use a subset of all rules for now.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInclusionRuleWithArrays.scala
```diff
@@ -0,0 +1,79 @@
...
+          case (FinSetT(FinSetT(t1)), PowSetT(FinSetT(t2))) =>
+            if (t1 != t2) {
+              throw new RewriterException(
+                  "Unexpected set types: %s and %s in %s"
+                    .format(t1, t2, state.ex), state.ex)
+            } else {
+              subset(rightState, leftCell, rightCell, true)
+            }
+
+          case _ =>
```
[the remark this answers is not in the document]
rodrigo7491: This is equal to the original `SetInclusionRule` rule. I didn't think much about it since an exception is already thrown there. If we make a change it should be in both encodings I think. Maybe in a separate PR?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImplWithArrays.scala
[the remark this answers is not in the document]
Kukovec: In what scenario would you want to delete an existing rule and not replace it with anything (= not copying it over in this implementation)?

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/EncodingBase.scala
rodrigo7491: I imagined `EncodingBase` only as a means to inherit the encoding types, thus I used `trait`. Is there any advantage in using `object`?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInclusionRuleWithArrays.scala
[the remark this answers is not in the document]
Kukovec: @konnov thoughts?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImplWithArrays.scala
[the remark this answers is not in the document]
rodrigo7491: The current one, in which we are still developing the encoding. Basically I don't want to add rules to the encoding before going over them, even if they remain unchanged. Getting a "noRuleApplicable" exception (I think it is written like this) is what I want for the rules we haven't gone over yet.

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/EncodingBase.scala
Kukovec: I mean, this allows access to the encoding type from anywhere. `oopsla19EncodingType` is essentially a global constant this way.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImplWithArrays.scala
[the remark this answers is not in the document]
Kukovec: In that case, I'd still use copy and then manually delete the rules you don't want AND write a comment above explaining how this is temporary.

rodrigo7491:
> This looks very good. See the other comments. As far as I understand, this PR contains the rules for set equality, set membership, and subseteq. Are other set operators planned for a separate PR?

The PR contains rules for all set operators in the map, excluding the ones flagged at the bottom. Most rules, like set ctr, set union, etc., were added implicitly by changing the constraints they generated, as mentioned in the top comment.

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/EncodingBase.scala
rodrigo7491: I see, I just didn't think we would need this anywhere else.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImplWithArrays.scala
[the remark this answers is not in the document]
rodrigo7491: I see, I will try to implement it as you suggested.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
```diff
@@ -0,0 +1,83 @@
...
+      val pred = newState.arena.topCell
+
+      def isInAndEqSetElem(powsetDomainElem: ArenaCell) = {
+        simplifier.simplifyShallow(tla.and(tla.in(powsetDomainElem.toNameEx, powsetDomain.toNameEx),
+                tla.eql(powsetDomainElem.toNameEx, setElem.toNameEx)))
+      }
+
+      val elemsInAndEqSetElem = powsetDomainElems.map(isInAndEqSetElem)
+      rewriter.solverContext.assertGroundExpr(simplifier.simplifyShallow(tla.equiv(pred.toNameEx,
+                  tla.and(tla.in(setElem.toNameEx, elemCell.toNameEx), tla.or(elemsInAndEqSetElem: _*)))))
```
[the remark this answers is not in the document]
rodrigo7491: `SetInRuleWithArrays` and `SetInclusionRuleWithArrays` are similar, as I mentioned in the comments in the code. See the comments on the previous thread (the one about `SetInclusionRuleWithArrays`), once we reach an agreement there I will also implement it here as well.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -156,29 +183,77 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext {
...
+        val updatedSetName = set.toString + "_" + name
+        val setSort = getOrMkCellSort(set.cellType)
+        log(s";; declare edge inclusion $name")
+        log(s"(declare-const $updatedSetName $setSort)")
+        val updatedSet = z3context.mkConst(updatedSetName, setSort)
+        // The updated set is cached here, it needs to be added to a predicate later, via "getInPred".
+        // The inStored map is used to ensure that the first call to "getInPred" related to the declared inPred will
+        // return a "store", and the subsequent calls will return a "select".
+        inCache += ((set.id, elem.id) -> (updatedSet, level))
+        inStored += ((set.id, elem.id) -> List((false, level)))
```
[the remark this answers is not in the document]
rodrigo7491: The point about using the cache wrongly could be made about any of the caches in the solver context. `inStored` is currently used only in `declareInPredIfNeeded` and `getInPred`.

I like the idea of adding `store` and `select`, I haven't though of adding new `ApalacheOper` before.

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/TestRewriterWithOOPSLA19.scala
rodrigo7491: I don't think so, every class with `JUnitRunner` has a similar, yet different, setup. See `TestArenaWithOOPSLA19`, `TestSeqModelCheckerWithOOPSLA19`, `TestRecordingSolverContextWithOOPSLA19`, and `TestTransitionExecutorWithOfflineAndOOPSLA19`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInclusionRuleWithArrays.scala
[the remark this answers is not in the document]
konnov: There is `case _ =>` that throws an exception below. So it will be reported as an error. Perhaps, we should throw a better exception there that indicates that this case is not supported.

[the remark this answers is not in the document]
Kukovec: Yeah, but shouldn't 
```
A == {1,2}
B == SUBSET A \subseteq { x \in SUBSET A: x < 5 } \* <=> SUBSET A \subseteq SUBSET A <=> TRUE
```
translate into exactly the case where the LHS is a `PowSetT` and the RHS is a standard `FinSetT` because of the filter?
You can't throw on this example.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
konnov: In this simple case, you are right, the expression `{1} \ {1}` will be probably simplified. But there are less certain cases. For instance, consider the expression `{x} \ {y}`, where `x` and `y` are state variables. Then the arena of the set will point to a single cell. However, the set is empty when `x = y`.

konnov:
> > This looks very good. See the other comments. As far as I understand, this PR contains the rules for set equality, set membership, and subseteq. Are other set operators planned for a separate PR?
> 
> The PR contains rules for all set operators in the map, excluding the ones flagged at the bottom. Most rules, like set ctr, set union, etc., were added implicitly by changing the constraints they generated, as mentioned in the top comment.

So it means they did not require any change?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInclusionRuleWithArrays.scala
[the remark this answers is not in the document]
konnov: I am not sure. You can try this example and see what happens. Maybe the left-hand side will get expanded. If it does not, we should fix `ExpansionMarker` to do that.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
rodrigo7491: I tried the example with state variable that you suggested and it still works as intended. The set minus operation needs to be evaluated at some point, and from what I saw in the smt constraints generated it is rewritten as set filter. If `{x} \ {y}` points to `x` in the case where `x = y` then there is a problem with how the arena is being updated, no?

rodrigo7491:
> So it means they did not require any change?

The `RewritingRule` classes themselves no, but the "conceptual" rules did change, since they now generate different smt constraints and thus lead to a different 4-tuple state, as described in the paper.

konnov:
> > So it means they did not require any change?
> 
> The `RewritingRule` classes themselves no, but the "conceptual" rules did change, since they now generate different smt constraints and thus lead to a different 4-tuple state, as described in the paper.

That's interesting. If we introduce `store` and `select`, you would have to rewrite those rules too, right?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
konnov: The arena is an overapproximation of the set contents. In case like `{x} \ {y}`, the rewriting rules cannot statically figure out that the set is empty in case of `x = y`. So the arena will still have one edge, even though the SMT model will have this edge evaluated to false. If you look at how set filter is implemented, it does not remove edges from the arena, but disables them in SMT when the predicate evaluates to false. If we could statically figure out which edges are disabled, we would not need set-membership predicates (or arrays) in SMT.

rodrigo7491: I see. This does explain the behavior of filter, which I found peculiar. For some reason I got the impression that this computation was done by the rules, but as you said we would essentially be doing the solver's job in Apalache if it was the case. I will make the change from conjunction to implication.

rodrigo7491:
> That's interesting. If we introduce `store` and `select`, you would have to rewrite those rules too, right?

Probably. It depends on how `store` and `select` end up being implemented. Since this is quite a fundamental change, should I do this here or in a follow up PR?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImplWithArrays.scala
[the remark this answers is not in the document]
rodrigo7491: I'm trying to override `ruleLookupTable` with `(super.ruleLookupTable ++ newRules) -- unsupportedRules.keys`, but end up getting `super may not be used on <none>`. Is there a correct way to access the map in the parent class?

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/EncodingBase.scala
rodrigo7491: Committing this change as is will break the unit tests, I will do it in a separate commit.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -46,6 +46,10 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext {
   //      logWriter.println(";; %s = %s".format(p, Global.getParameter(p)))
   }
 
+  private def useArrays(): Boolean = {
+    config.smtEncoding == "arrays"
+  }
```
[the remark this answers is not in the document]
rodrigo7491: Will make a separate commit also updating the calls to `useArrays`.

konnov:
> > That's interesting. If we introduce `store` and `select`, you would have to rewrite those rules too, right?
> 
> Probably. It depends on how `store` and `select` end up being implemented. Since this is quite a fundamental change, should I do this here or in a follow up PR?

Yes, I think we can postpone it for another PR. This one is not going to break anything.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImplWithArrays.scala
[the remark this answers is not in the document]
rodrigo7491: Scala handles inheritance of `val` in an [unexpected way,](https://docs.scala-lang.org/tutorials/FAQ/initialization-order.html) at least to me. My fix was to make a `defaultRuleLookupTable` that could be accessed by the child class. If there is a better way, please let me know.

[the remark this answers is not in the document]
Kukovec: This solution looks  fine to me.

Kukovec (approved):

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInclusionRuleWithArrays.scala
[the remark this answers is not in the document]
Kukovec: See #1103

Issue #1088, opened 2021-11-11
[BUG] Locally defined polymorphic operators cannot reference parent operator's type variables


Kukovec:
## Description

Whenever a polymorphic operator defines a local LET-IN operator, which is also polymorphic, the inner operator's type annotations are treated as separately quantified and not as references to the parent operator's type variables . 

## Input specification
```
---------- MODULE test ----------

EXTENDS Integers, Apalache

\* @type: (a, a) => a;
F(x,y) ==
  \* @type: Set(a);    
  LET A == {x}
  IN IF y \in A THEN x ELSE y

Init == TRUE

Next == TRUE

Inv == F(1,1) = 1
====================
```

## The command line parameters used to run the tool

`--inv=Inv --length=1`

## Additional context

Tool will not throw, because the operator in question gets removed by inlining, but combining such operators with e.g. folds would result in the same behavior as in #1085.

konnov:
Your understanding is consistent with e.g. Ocaml. This code compiles in OCaml:

```ml
let f: 'a -> 'a -> 'a = fun x -> fun y ->
    let g: 'a -> 'a list = fun z ->
        [z]
    in
    if List.mem y (g x) then x else y
```

Will fix Snowcat.

shonfeder:
Reporting a summary of a discussion on our internal slack.

Iiuc, the OCaml code snippet given by @konnov, does not given an example of a type variable inside a function body referencing the type variables in the function head. Rather, this code snippet compiles because `g` is (correctly) annotated as a function which will take a term of any type '`a` to list of terms of type `'a`. Since the function works for arguments of any type, it will of course also work when applied to `x`, regardless of what type it is. We can make this clear by making the universal quantification over the type variables in `g`'s annotation explicit:

```ocaml
# let f: 'a -> 'a -> 'a = fun x -> fun y ->
    let g: type b. b -> b list = fun z ->
        [z]
    in
    if List.mem y (g x) then x else y
;;
val f : 'a -> 'a -> 'a = <fun>
```

I am unfamiliar with any convention in ML-family languages that allows "normal" (i.e., universally quantified *implicitly*) type variables used in annotations of subexpressions to refer to universally quantified type variables annotating their parent expression. But after, a bit of research, I now know of two examples where special syntax is introduced to enable this kind of thing:


- The "scoped typed variables" extension of GHC:  https://wiki.haskell.org/Scoped_type_variables
- Ocaml's explicit universal type annotation (probably implemented following the previous):

```
[...]

Issue #973, opened 2021-09-01
[FEATURE] Use disjoint sets in the type checker


konnov:
Following the discussion in #952

konnov:
Somehow, I messed up the title several times. @Kukovec would you like to implement that in October?

konnov:
The chapter 10 of [Advanced Topics...](https://dl.acm.org/doi/10.5555/1076265) by Pottier and Remy contains the unification algorithm that uses union-find. Shall we check that one before re-inventing anything?

konnov:
We will hopefully have a fast type checker soon, so there is no point of improving the old one.

PR #1104, opened 2021-11-18
closes #1084: fix computation of principal types

### LINKED ISSUE #1084: [BUG] Polymorphic operator types are not instantiated correctly
## Description

When using a polymorphic operator twice in the same context, the type of the 1st application instance is considered when computing the 2nd application instance, even though they should be instantiated independently.

## Input specification

```
---------- MODULE test ----------

EXTENDS Integers

Init == TRUE

Next == TRUE

\* @type: (a -> Int) => Set(a);
X(f) == DOMAIN f

f1 == [a \in {1} |-> 1]
fa == [a \in {"a"} |-> 1]

Inv1 ==
  /\ X(fa) = {"a"}
  /\ X(f1) = {1}

Inv2 ==
  /\ X(f1) = {1}
  /\ X(fa) = {"a"}

Inv3 == 
  LET
    A == X(fa)
  IN LET
    B == X(f1)
  IN 
  /\ A = {"a"}
  /\ B = {1}
====================
```

`--inv=Inv[N] --length=1`

## Expected behavior

No type error for any of the invariants

## Additional context

Returns
```
[test.tla:41:6-41:10]: No match between operator signature (((Str -> Int)) => Set(Str)) and argument (Int -> Int) E@11:32:49.829
```
When checking `Inv1`,
```
[test.tla:41:6-41:10]: No match between operator signature (((Int -> Int)) => Set(Int)) and argument (Str -> Int) E@11:35:24.173
```
When checking `Inv2`, and no error when checking the workaround `Inv3`

### LINKED ISSUE #1085: [BUG] Types are not propagated correctly when instantiating polymorphic operators that use folds
## Description

After instantiating a polymorphic operator, which uses a fold, the annotations that are propagated for the fold operator body are not correctly unified with the types derived from the operator arguments at the call site, resulting in the following error:
```
[test.tla:9:32-9:32]: Found a polymorphic type: (a79 -> Int)      E@18:24:28.109
Probable causes: an empty set { } needs a type annotation or an incorrect record field is used E@18:24:28.109
Typing input error: [test.tla:9:32-9:32]: Found a polymorphic type: (a79 -> Int) E@18:24:28.111
```

## Input specification

```
---------- MODULE test ----------

EXTENDS Integers, Apalache

\* @type: (Set(a)) => a -> Int;
PolyOper(S) == 
  LET
    \* @type: (a -> Int, a) => a -> Int;
    Add(f,e) == [ x \in DOMAIN f \union {e} |-> 0 ]
  IN FoldSet( Add, [x \in {} |-> 0], S )

Init == TRUE

Next == TRUE

Inv == PolyOper({1}) = [x \in {1} |-> 0]

====================
```

`--inv=Inv --length=1`

## Expected behavior

No type error.

## Additional context

This is what the last pass before the 2nd snowcat pass outputs:
```
---------------------------- MODULE 11_OutAnalysis ----------------------------

EXTENDS Integers, Sequences, FiniteSets, TLC, Apalache

(*
  @type: (() => Bool);
*)
Init_si_0000 == TRUE

(*
  @type: (() => Bool);
*)
Next_si_0000 == TRUE

(*
  @type: Bool;
*)
VCInv_si_0 ==
  FoldSet(LET (*
    @type: (((a45 -> Int), a45) => (a45 -> Int));
  *)
  t_2_si_1(f_si_1, e_si_1) == [ x$1 \in DOMAIN f$1 \union {e$1} |-> 0 ]
  IN
  t_2$1, [ x$2 \in {} |-> 0 ], {1})
    = [ x$3 \in {1} |-> 0 ]

(*
  @type: Bool;
*)
VCNotInv_si_0 ==
  ~(FoldSet(LET (*
    @type: (((a45 -> Int), a45) => (a45 -> Int));
  *)
  t_2_si_2(f_si_2, e_si_2) == [ x$4 \in DOMAIN f$2 \union {e$2} |-> 0 ]
  IN
  t_2$2, [ x$5 \in {} |-> 0 ], {1})
    = [ x$6 \in {1} |-> 0 ])

```
Inside `FoldSet`, the LET-defined operator is still marked as having a polymorphic type (`    @type: (((a45 -> Int), a45) => (a45 -> Int));`, however, the unification at the call site should have determined `a45 = Int`.

konnov:
This PR closes #1084. Instead of tracking the maximal variable index and using it to compute quantified variables in the principal types, compare against the variables in the constraint solver.

It looks like it closes #1085 and #1088 too.


tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeContext.scala
```diff
@@ -30,16 +30,16 @@ class TypeContext(val largestVarIndex: Int, val types: Map[String, (TlaType1, Se
 object TypeContext {
 
   /**
-   * The value of the variable index that means that all variables can be quantified.
+   * A type scheme that carries a type and a set of variables that are universally quantified.
    */
-  val NIL_VAR_INDEX: Int = -1
+  type TypeScheme = (TlaType1, Set[Int])
```
shonfeder, line 35: Maybe move this type alias to the top of the file (or at least above where it is first used?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeContext.scala
```diff
@@ -30,16 +30,16 @@ class TypeContext(val largestVarIndex: Int, val types: Map[String, (TlaType1, Se
 object TypeContext {
 
   /**
-   * The value of the variable index that means that all variables can be quantified.
+   * A type scheme that carries a type and a set of variables that are universally quantified.
```
shonfeder, line 33: What exactly is the relationship between between the `TlaType1` and the set of variables? Is it that the set of variables are all bound to that type?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ConstraintSolver.scala
```diff
@@ -77,6 +77,16 @@ class ConstraintSolver(approximateSolution: Substitution = Substitution.empty) {
...
+  /**
+   * Test whether a variable is free in the context that is induced by the solved constraints.
+   *
+   * @param varNo a variable number
+   * @return true if the variable occurs in the partial solution of the solver
+   */
+  def isFreeVar(varNo: Int): Boolean = {
+    approximateSolution.mapping.keySet.forall(cls => !cls.typeVars.contains(varNo))
+  }
+
```
shonfeder, line 89: Any chance that we'll need to hit this frequently enough that it'll effect performance having to search through every item each time we check for free vars? If so, perhaps (as a followup?) we should keep a cash of free variables, and then just remove vars from the cache as they are bound?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EqClass.scala
```diff
@@ -18,15 +18,43 @@ class EqClass {
   /**
    * Type variables that are associated with the equivalence class.
    */
-  var typeVars: Set[Int] = Set.empty
+  private var _typeVars: Set[Int] = Set.empty
+
+  /**
+   * The representative of an equivalence class.
+   */
+  private var _reprVar: Int = -1
```
shonfeder, line 26: Could this either be `Option[Int]` have the initial value set via an an argument to the class constructor? It seems to me that either would be preferable to having an invalid value for initialization.

shonfeder (changes requested):
Looks generally good to me! Tho I've picked a couple little nits.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeContext.scala
konnov: For some reason, this does not work in Scala. I guess, type definitions work only in packages. Anyhow, it is a good idea to move it closer to `TlaType1`.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeContext.scala
konnov: It is a subset of type variables used in the type. Some of the type variables may be bound in outer scopes, so they should not be quantified.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ConstraintSolver.scala
konnov: I hope there is no big drawback. This check is done only for a few variables per definition, and TLA+ specs are below 2-3 KLOC. There is another trick in Chapter 10 of Advanced Types. I did not get to it, but it may be a good idea to read about it :)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeContext.scala
konnov: I have refactored it to `TlaType1Schema` next to `TlaType1`.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ConstraintSolver.scala
konnov: In any case, @Kukovec  wanted to rewrite the unifier with union-find in #973. So this code will most likely change.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/EqClass.scala
konnov: Refactored `EqClass` too. This will also change in #973.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ConstraintSolver.scala
Kukovec: Shall I bump that up in priority then?

Kukovec (approved):
LGTM

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ConstraintSolver.scala
konnov: I think so.

konnov:
I have addressed all comments and jumped forward to merging it.

PR #1107, opened 2021-11-19
Fixes type synchronization after inlining

### LINKED ISSUE #1088: [BUG] Locally defined polymorphic operators cannot reference parent operator's type variables

Kukovec:

Inlining was performing type substitution only on the body of a declaration, which caused the type tag of a polymorphic declaration to potentially desync from the types used inside the body, whenever that operator was instantiated with another polymorphic operator.

Followup to #1104. 
Closes #1088

### DIFF
--- a/test/tla/Poly1107.tla
+++ b/test/tla/Poly1107.tla
@@ -1,0 +1,37 @@
+---------- MODULE Poly1107 ----------
+
+EXTENDS Integers, Apalache
+
+\* @type: (a) => a -> Int;
+MkFun(p) == [ x \in {p} |-> 1 ]
+
+\* @type: (a -> Int, a -> Int) => a -> Int;
+F1 (+) F2  ==
+  LET 
+    D1 == DOMAIN F1
+    D2 == DOMAIN F2
+  IN [e \in D1 \union D2 |->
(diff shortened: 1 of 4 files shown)


konnov (approved):
So the solution is to apply substitutions recursively when doing inlining? I think it is sound.

Kukovec:
> So the solution is to apply substitutions recursively when doing inlining? I think it is sound.

I'm not 100% happy with this solution, since it's more or less a bandaid. That said, I also don't want to be blocked on this indefinitely, so as a compromise, I'll look into this in more detail when I get to #1108.

Issue #675, opened 2021-03-24
[FEATURE] A better version of Builder


konnov:
As @Kukovec and @shonfeder found it hard to use the new type-aware version of `Builder`, we are going to explore the possible alternatives. This issue is purely on the code clarity and ease of code maintenance, not affecting any user-facing features.

konnov:
I have an idea how to make it simple without writing too much code. Taking away the token :)

konnov:
also #1020 and #690

konnov:
Shall we just close this issue, since we have #1655?

PR #1109, opened 2021-11-21
Sketch of a type-level expression builder

### LINKED ISSUE #1102: Draft POC of type-level expression builder approach
This was an idea suggested by Igor. The aim is to run a short experiment to see how much we can push the typing of `TlaEx` into the type level, to be able to offload the type system reasoning to Scala's static analysis.

shonfeder:
Closes #1102
Supports  #675

This presents a very small (very rough) example of how we might push
guaranteed type-correct TLA expressions into the type level. It is meant
to demo the general technique, so we can get a rough sense of whether
it's worth pursuing.

I suspect there may be many nonobvious complications that would arise in the
process of trying to apply such an approach, and that porting the existing code
would involve a *lot* of work.

The benefit to hope for is that we might eliminate large sections of
code and supporting tests.

For the most part, the key mechanism enabling the encodings is Shapeless'
heterogenous list type
[HList](https://books.underscore.io/shapeless-guide/shapeless-guide.html#generic-product-encodings),
which gives us an encoding for generic products.

### DIFF
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
@@ -1,0 +1,186 @@
+package at.forsyte.apalache.tla.lir
+
+import at.forsyte.apalache.tla.lir.oper.{TlaArithOper, TlaBoolOper, TlaFunOper, TlaOper}
+
+// We use shapeless for facilitating the type-level encodings of TlaExpressions
+import shapeless._
+import poly._
+
+// A (very) rough sketch at what TLA Expressions might look like when
+// we represent out typing discipline inside of Scala's type system.
+object Tla {
+  // The type of any TLA expression
+  sealed trait Exp[T] {
+    def toString: String
+  }
+
+  // Just an alias
+  type IntT = BigInt
+
+  // A value constructs a TLA expression tagged with the Scala type it carries as a payload.
+  class Value[T](value: T) extends Exp[T]
+  case class Int(value: IntT) extends Value(value)
+  case class Str(value: String) extends Value(value)
+  case class Bool(value: Boolean) extends Value(value)
+
+  // Each compound expression is reprsented by a new type
+
+  sealed trait SeqType[T]
+  case class TlaSeq[T](seq: Seq[Exp[T]]) extends Exp[SeqType[T]]
+
+  sealed trait TupleType[A, B]
+  case class Empty() extends Exp[TupleType[Unit, Unit]]
+  case class Single[A](fst: A) extends Exp[TupleType[A, Unit]]
+  case class Tuple[A, B](fst: A, snd:B) extends Exp[TupleType[A, B]]
+
+  sealed trait RecType[Fields]
+  case class Rec[Fields <: HList](fields: Fields) extends Exp[RecType[Fields]]
+
+  sealed trait OperType[Args <: HList, Ret]
+
+  // Names are expressions
+  // The idea here is that names also record the type of the expression
+  // they refer to. When we don't know what that type is, we might use
+  // `Name[Exp[Any]]`.
+  case class Name[T](name: String) extends Exp[T]
+
+  // An unapplied operator
+  case class Op[Args <: HList, Ret](oper: TlaOper)
+      extends Exp[OperType[Args, Ret]]
+
+  // Application of an operator
+  case class App[Args <: HList, Ret](op: Exp[OperType[Args, Ret]], args: Args)
+      extends Exp[Ret]
+
+  // Binary operators of type (T, T) => Ret
+  type BinOpApp[T, Ret] = App[T :: T :: HNil, Ret]
+  type Tup[A, B] = Exp[TupleType[A, B]]
+}
+
+// Utility functions for constructing expressions. Most of these don't actually do much,
+// but a few are useful.
+object TypeLevelBuilder {
+  import Tla._
+
+  def binOp[T, Ret](op: TlaOper)(a: T, b: T): Exp[Ret] = App(Op(op), a :: b :: HNil)
+
+  def plus: (Exp[IntT], Exp[IntT]) => Exp[Tla.IntT] =
+    binOp(TlaArithOper.plus)
+
+  // A polymorphic operator
+  def equal[T]: (Exp[T], Exp[T]) => Exp[Boolean] =
+    binOp(TlaOper.eq)
+
+  def and(a: Exp[Boolean], b: Exp[Boolean]): Exp[Boolean] =
+    binOp(TlaBoolOper.and)(a, b)
+
+  // Variadic conjunction
+  def and(args: Exp[Boolean]*): Exp[Boolean] = args match {
+    case Seq() => Bool(true)
+    case arg +: args => and(arg, and(args: _*))
+  }
+
+  // Polymorphic sequences
+  // def seq[T](): Seq[T] = Nil()
+  // def seq[T](args: Exp[T]*): Seq[T] = args match {
+  //   case Seq() => Nil()
+  //   case arg +: args => Cons(arg, seq(args: _*))
+  // }
+
+  def seq[T](args: Exp[T]*): TlaSeq[T] = {
+    TlaSeq(args)
+  }
+
+  // def tup[A, B](a: A, b: B): A => B => Tup[A, B] = a => b => Tuple(a, b)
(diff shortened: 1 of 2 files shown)


shonfeder:
FYI @konnov and @Kukovec

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
```diff
@@ -0,0 +1,186 @@
...
+  sealed trait RecType[Fields]
+  case class Rec[Fields <: HList](fields: Fields) extends Exp[RecType[Fields]]
+
+  sealed trait OperType[Args <: HList, Ret]
+
+  // Names are expressions
+  // The idea here is that names also record the type of the expression
+  // they refer to. When we don't know what that type is, we might use
+  // `Name[Exp[Any]]`.
+  case class Name[T](name: String) extends Exp[T]
```
konnov, line 45: Is there a way to refer to an identifier like `:foo` in lisp?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
```diff
@@ -0,0 +1,186 @@
...
+
+  // Each compound expression is reprsented by a new type
+
+  sealed trait SeqType[T]
+  case class TlaSeq[T](seq: Seq[Exp[T]]) extends Exp[SeqType[T]]
+
+  sealed trait TupleType[A, B]
+  case class Empty() extends Exp[TupleType[Unit, Unit]]
+  case class Single[A](fst: A) extends Exp[TupleType[A, Unit]]
+  case class Tuple[A, B](fst: A, snd:B) extends Exp[TupleType[A, B]]
```
konnov, line 34: Interesting. We would have to define tuples for all sizes, similar to Scala, right?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
```diff
@@ -0,0 +1,186 @@
...
+
+  // A polymorphic operator
+  def equal[T]: (Exp[T], Exp[T]) => Exp[Boolean] =
+    binOp(TlaOper.eq)
+
+  def and(a: Exp[Boolean], b: Exp[Boolean]): Exp[Boolean] =
+    binOp(TlaBoolOper.and)(a, b)
+
+  // Variadic conjunction
+  def and(args: Exp[Boolean]*): Exp[Boolean] = args match {
```
konnov, line 78: Variadic conjunctions are quite common in TLA+. So it would be good to have it as a normal form, instead of unwinding them into binary conjunctions.

konnov (commented):
That's an interesting development. Why do you think it would not be possible to capture the whole IR in this representation? This language-native representation could be useful for writing tests and improving type-correctness of our transformations.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
shonfeder: I'm not sure exactly how that would factor in. Perhaps give an example?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
shonfeder: I think we could actually just define tuples as

```suggestion
  case class Tuple[Items <: HList](items: Items) extends Exp[TupleType[Items]]
```

That is, just as a thin wrapper around `HList`s, since those are generalized product types already :)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
shonfeder: I think that should be pretty feasible! I'd be interested to learn the stakes here. On the face of it, I'd assume variadic conjunction were just syntax suggar for chained binary conjunctions, but does it affect performance or the encoding to SMT somehow?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
konnov: I am not sure either. Actually, TLA+ identifiers may have the form `Foo!Bar`, when they are instantiated from modules. So maybe this does not matter. It would probably matter when we have to describe the shape of records.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
konnov: This is much nicer!

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
konnov: It's just harder to write some transformations when you have hanging trees of and-s and or-s. I would suspect that SMT solvers group conjunctions and disjunctions into lists during the preprocessing stage. But that might be harder to do, since our encoding would produce multiple constraints for non-variadic conjunctions and disjunctions.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
```diff
@@ -0,0 +1,186 @@
...
+  import record._
+  import syntax.singleton._
+
+  val exRecord = Tla.Rec(("msg" ->> Tla.Str("hello")) :: ("value" ->> Tla.Int(3)) :: HNil)
+  val exRecordAccess1 = plus(exRecord.fields("value"), Tla.Int(3))
+  val exRecordAccess2 = equal(Tla.Str("hello"), exRecord.fields("msg"))
+
+  // NOTE: Statically fails with type error, due to invalid record field type
+  // I.e., field "msg" holds a string, so we cannot use it's value in `plus`.
+  // val exInvalidRecordAccess = plus(exRecord.fields("msg"), Tla.Int(3))
```
konnov, line 168: Wow. That's magic! Do you know how it works?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
konnov: But I see that you are using `Fields` for the records. I am not sure how it works.

shonfeder: I'm not clear what the result you're wanting from identifiers/fields here. What would you like to do in relation to to `Name`s?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TypeLevelBuilder.scala
shonfeder: The constructor generates a singleton field-types for the label, that encode the value as a parameter! :D

https://github.com/milessabin/shapeless/blob/36089a7c0eb009dd9dc313bed44df822fea8a96e/core/src/main/scala/shapeless/labelled.scala#L80-L85

shonfeder:
> Why do you think it would not be possible to capture the whole IR in this representation?

I think it *should* be possible! I only wanted to flag that there is some risk involved. Since this makes the construction of expressions dependent on what we can represent at the type level, it's *possible* that some expected complications will arise in trying to deal with input at runtime, and we'll need a bit of careful typeclass work to provide combinators that work smoothly over all `Exp`. 

Of course, mitigating these risks is the fact that:

- We can always fall back to just dealing with `Exp[Any]`
- The use of shapless for generic programming is well documented and there is lots of prior art, so we won't need to reinvent these techniques, we'll just need to apply them correctly.

If we think this is a promising direction, I'd be happy to continue the work on it, or collaborate on sketching out its continuation.

shonfeder:
I think this demo as served its purpose! I'm gonna close the PR but I'll leave the branch in case we want to pursue this line at some point.

shonfeder:
Opening again, since this came under discussion and is to be reviewed this week.

shonfeder:
This served its purpose and we’re not currently pursuing this approach. Gonna close it to reduce clutter, but will keep the branch around.

PR #1133, opened 2021-12-02
Closes #1130:   Improve the format of uninterpreted constants to name_OF_TYPE

### LINKED ISSUE #1130: [FEATURE] Change the format of uninterpreted constants
In the current format, we have to write uninterpreted constants as `uval_TYPE_name`, e.g.:

```tla
  /\ tmState = "uval_SORT_STATE_init"
```

See [TwoPhaseUFO](https://github.com/informalsystems/apalache/blob/unstable/test/tla/TwoPhaseUFO.tla) for more examples.

I find the current notation to be verbose and hard to read. I am proposing to change it to the more readable format:

```
name_OF_TYPE
```

So the above example would look like follows:

```tla
  /\ tmState = "init_OF_SORT_STATE"
```

The benefit of this new format is that the most important part of the constant appears on the left, that is, `init`. Moreover, the type and the infix `_OF_` are both capitalized, and they appear on the right, which is more convenient to type and read. It is also more consistent with our type annotations.

shonfeder:

### DIFF
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/ModelValueHandler.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/ModelValueHandler.scala
@@ -9,37 +9,49 @@ import scala.util.matching.Regex
  * types (and SMT sorts) are unique and they are incomparable with regular strings of with other
  * model values of a different type.
  *
- * Each model value follows the pattern
- *  [PREFIX]_[TYPE]_[INDEX]
- * where:
- *  - PREFIX is a prefix shared by, and identifying, all model values.
- *  - TYPE is the type of this particular model value. e.g., a value [PREFIX]_XYZ_[INDEX] has the uninterpreted type XYZ
- *  - INDEX is a unique identifier within the model value type. Its actual value is unimportant, however,
- *  model values with different indices are unequal by definition
+ * Each model value follows the pattern: `[NAME]_OF_[TYPE]`, where:
+ *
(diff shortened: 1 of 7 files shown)


tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/ModelValueHandler.scala
```diff
@@ -9,37 +9,49 @@ import scala.util.matching.Regex
...
  * model values of a different type.
  *
- * Each model value follows the pattern
- *  [PREFIX]_[TYPE]_[INDEX]
- * where:
- *  - PREFIX is a prefix shared by, and identifying, all model values.
- *  - TYPE is the type of this particular model value. e.g., a value [PREFIX]_XYZ_[INDEX] has the uninterpreted type XYZ
- *  - INDEX is a unique identifier within the model value type. Its actual value is unimportant, however,
- *  model values with different indices are unequal by definition
+ * Each model value follows the pattern: `[NAME]_OF_[TYPE]`, where:
```
Kukovec, line 12: You know, since we're dealing with strings, why not make the syntax literally `[NAME] : [TYPE]`? `:` is a character too.

Kukovec (approved):
LGTM

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/ModelValueHandler.scala
konnov: I like your out-of-the-box thinking :))

PR #1138, opened 2021-12-03
Update the manual

konnov:
Following the [issue by Markus](https://github.com/tlaplus/CommunityModules/issues/56), I have restructured and updated the manual. Plenty of files were moved, so it is probably easier to read the manual with `mdbook server`, rather than looking at diffs.


docs/src/apalache/principles/apalache-mod.md
```diff
@@ -6,12 +6,13 @@ be found in
 Most of the operators in that modules are introduced internally by Apalache,
 when it is rewriting a TLA+ specification.  It is useful to read the comments
 to the operators defined in `Apalache.tla`, as they will help you in
-understanding the [detailed output](./running.md#detailed) produced by the tool, see.
-Perhaps, the most interesting operator in `Apalache` is the type assignment
-operator that is defined as follows:
+understanding the [detailed output](./running.md#detailed) produced by the
+tool, see.  Perhaps, the most interesting operator in `Apalache` is the type
+assignment operator that is defined as follows:
```
Kukovec, line 11: 1. See what?
2. 
```suggestion
tool, see.  Perhaps, the most interesting operator in `Apalache` is the
assignment operator that is defined as follows:
```

docs/src/apalache/principles/assignments.md
```diff
@@ -0,0 +1,84 @@
...
+and run `apalache` against
+[`test/tla/y2k_override.tla`](https://github.com/informalsystems/apalache/blob/unstable/test/tla/y2k_override.tla):
+
+```console
+$ apalache check y2k_override.tla
+```
+
+ We can check the detailed output of the `TransitionFinderPass` in the file
+`x/<timestamp>/out-transition.tla`, where `<timestamp>` looks like
+`09.03-10.03.2020-508266549191958257`:
```
Kukovec, line 16: Old output folder shape and file name.

docs/src/apalache/principles/assignments.md
```diff
@@ -0,0 +1,84 @@
...
+```console
+...
+PASS #9: TransitionFinderPass
+To understand the error, check the manual:
+[https://apalache.informal.systems/docs/apalache/assignments.html]
+Assignment error: No assignments found for: a
+It took me 0 days  0 hours  0 min  1 sec
+Total time: 1.88 sec
+EXITCODE: ERROR (99)
+```
```
Kukovec, line 78: I think this is from when we still used SMT in assignments. Now we have line numbers for errors.

docs/src/apalache/principles/index.md
```diff
@@ -0,0 +1,19 @@
+# Principles of Symbolic Model Checking with Apalache
+
+In order to take advantage of Apalache's symbolic model checking, there are a
+few principles one must bear in mind when writing TLA.
+
+Note that Apalache requires type annotations.  Check the [Snowcat
+tutorial](../tutorials/snowcat-tutorial.md) and [HOWTO on
+annotations](../HOWTOs/howto-write-type-annotations.md).
```
Kukovec, line 8: Feels like it could use more binding text (low priority though).

Kukovec (approved):
LGTM, with minor nits.

docs/src/apalache/principles/assignments.md
konnov: Actually, it is exactly the same output. So you might want to check the output by the assignment solver ;-)

docs/src/apalache/principles/index.md
konnov: yep, maybe later

docs/src/apalache/principles/assignments.md
Kukovec: Looking at it,obviously there are no line numbers to display, since something is _missing_. Not a bug

konnov: ah, that's true. But where is it missing?

Kukovec: Hard to say, in general, because of disjunctions. It's not necessarily a range either:
```
A == 
  \/ x' = 1
  \/ TRUE
  \/ x' = 2
  \/ TRUE
```

konnov: You could point to the location of `A`

Kukovec: You could, but then you have the same problem at
```
X == 
  \/ A \* missing assignments.
  \/ B
  \/ C \* Also missing assignments
  \/ D
```
This is more of a design question: When multiple things are wrong, what do you report? If you want to go down the path of "show the first issue and show the 2nd one next time they run" you just output the first `\/ TRUE` line #. Otherwise the best you can do is just say "x is missing".

Issue #359, opened 2020-12-06
Rewire operators on Bags


konnov:
This should be done similar to #1484. Though it may require more work to redefine the operators.

Kukovec:
Blocked by #1084, making a note here for reference.

konnov:
Solved by @Kukovec. See the definitions in https://github.com/informalsystems/apalache/blob/unstable/src/tla/__rewire_bags_in_apalache.tla

PR #1141, opened 2021-12-03
TypeSub ignored OperDecl tags

Kukovec:

MWE was difficult to determine, #359 should come with a test suite for this eventually.


konnov (approved):
Yep, we did the fix together, but could not identify an MWE. The definition inliner was failing on a really large spec. We have identified a fix, but no unit/integration test. Let's make an exception to the rule :(

PR #1148, opened 2021-12-06
Move all RFCs into same directory

shonfeder:
Keeping our rfcs/adrs in separate directories causes me confusion, and I
imagine would do so for others too. Moreover, I think it's not useful.
When we initially separated them out, it was to protect users from
encountering design documents that wouldn't be useful to them. Now we
can do this by just omitting the docs we wish to hide from the listing
in `SUMMARY.md`.

-------


Kukovec (approved):
LGTM, but won't this break links to these files?

shonfeder:
@Kukovec good call! I fixed the one link that would have been broken. Thanks :D

PR #1150, opened 2021-12-06
Add ADR template

shonfeder:
This PR proposes a template for use when drafting ADRs. The motivation and
rationale is explained in an accompanying ADR, drafted primarily as an
illustration of what the proposed template would look like in action.

The idea to clarify our understanding of the purpose of ADRs, and to agree on a
template to use as a basis was discussed in a sync meeting some weeks back.


Kukovec (approved):
LGTM

shonfeder:
Here's another example of the template:  #1156

konnov (approved):
Looks good. I agree that we should keep the process lightweight.

PR #1153, opened 2021-12-06
Fix the use of oracles in the arrays encoding

### LINKED ISSUE #1152: [BUG] Sets becoming unconstrained after set minus operation
## Description

In the arrays encoding, when set minus is used it leaves the updated set unconstrained. This is due to a break in the chain of SSA constraints.

## Input specification

This bug was found when testing the encoding with [SetSndRcv.tla](https://github.com/informalsystems/apalache-tests/blob/master/performance/array-encoding/SetSndRcv.tla).

`check --inv=Inv --smt-encoding=arrays --length=4 --cinit=CInit4 SetSndRcv.tla`

## Expected behavior

No invariant violation.

## Additional context

Counter-example generated:

```

EXTENDS SetSndRcv

ConstInit == Values = { 0, 1, 2, 3, 4 }

(* Initial state *)
State0 ==
  Values = { 0, 1, 2, 3, 4 }
    /\ medium = {}
    /\ receiver = {}
    /\ sender = { 0, 1, 2, 3, 4 }

(* Transition 0 to State1 *)
State1 ==
  Values = { 0, 1, 2, 3, 4 }
    /\ medium = {0}
    /\ receiver = {}
    /\ sender = { 1, 2, 3, 4 }

(* Transition 1 to State2 *)
State2 ==
  Values = { 0, 1, 2, 3, 4 }
    /\ medium = {0}
    /\ receiver = {}
    /\ sender = {}

(* Transition 1 to State3 *)
State3 == Values = { 0, 1, 2, 3, 4 }/\ medium = {}/\ receiver = {}/\ sender = {}

InvariantViolation == sender = {} /\ medium = {} /\ receiver = {}

```

rodrigo7491:

Oracles currently produce conjunctions of implications. In the arrays encoding, if the lhs of an implication does not hold and its rhs contains an array update we need to explicitly state that the array stays unchanged. This is done implicitly in the current encoding. Closes #1152.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
```diff
@@ -441,22 +448,30 @@ class CherryPick(rewriter: SymbStateRewriter) {
...
         }
       }
 
-      val assertions = toPickFrom.zip(memberSets).zipWithIndex map (nthIn _).tupled
+      val assertions = (toPickFrom.zip(memberSets).zipWithIndex map (nthIn _).tupled).unzip
       // add the cell to the arena
       nextState = nextState.updateArena(_.appendHas(resultCell, picked))
       // (chosen = 1 /\ in(z_i, R) = in(c_i, S_1)) \/ (chosen = 2 /\ in(z_i, R) = in(d_i, S_2))
-      solverAssert(oracle.caseAssertions(nextState, assertions :+ elseAssert))
+      solverAssert(oracle.caseAssertions(nextState, assertions._1 :+ elseAssert, assertions._2 :+ ValEx(TlaBool(true))))
```
konnov, line 474: Extract `assertions._1` and `assertions._2` to `membershipAssertions` and `nonMembershipAssertions`? Just for documentation purposes.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/IntOracle.scala
```diff
@@ -33,21 +34,25 @@ class IntOracle(val intCell: ArenaCell, nvalues: Int) extends Oracle {
...
-   * @param assertions a sequence of assertions, one per oracle value
+   * @param state          a symbolic state
+   * @param assertions     a sequence of assertions, one per oracle value
+   * @param elseAssertions an optional sequence of assertions, one per oracle value
    * @return an expression ite(oracle = 0, ite(oracle = 1, ...))
    */
-  override def caseAssertions(state: SymbState, assertions: Seq[TlaEx]): TlaEx = {
+  override def caseAssertions(state: SymbState, assertions: Seq[TlaEx], elseAssertions: Seq[TlaEx] = Seq()): TlaEx = {
+    if (elseAssertions.nonEmpty & assertions.size != elseAssertions.size) {
+      throw new IllegalStateException(s"Invailid call to Oracle, malformed elseAssertions")
```
konnov, line 44:
```suggestion
      throw new IllegalStateException(s"Invalid call to Oracle, malformed elseAssertions")
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/IntOracle.scala
```diff
@@ -33,21 +34,25 @@ class IntOracle(val intCell: ArenaCell, nvalues: Int) extends Oracle {
...
       case 0 => state.arena.cellTrue().toNameEx
 
       case 1 => assertions.head
 
       case _ =>
-        val es = assertions
-          .slice(0, nvalues)
-          .zipWithIndex
-          .map { case (e, i) => tla.or(e, tla.not(whenEqualTo(state, i))) }
+        val iteCases = assertions.zipAll(elseAssertions, ValEx(TlaBool(true)), ValEx(TlaBool(true)))
```
konnov, line 53: We could use `zip` instead of `zipAll`, since there is an assertion above.

konnov (approved):
Looks good! I can see how it was failing on expanding a set instead of keeping it tight.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
rodrigo7491: Will do, thanks!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/IntOracle.scala
rodrigo7491: Can we? If `elseAssertions` is empty we have to use `zipAll`, no?

rodrigo7491: I don't think it is necessary. If `elseAssertions` is not empty `zipAll` does nothing, if it is empty it makes the second element of each tuple in `iteCases` be `ValEx(TlaBool(true))`. If you think it would be good to explicitly populate `elseAssertions` with `assertions.size` copies of `ValEx(TlaBool(true))` an then use `zip` to make clearer what is happening I could do it.

rodrigo7491: Maybe adding a comment would be enough.

PR #1157, opened 2021-12-07
ADR section describing the encoding of TLA+ functions and sequences using SMT arrays

### LINKED ISSUE #1154: [FEATURE] Document the handling of functions and sequences in the SMT encoding with arrays
Update [ADR11](https://apalache.informal.systems/docs/adr/011adr-smt-arrays.html) to describe how TLA+ functions and sequences will be encoded as SMT arrays.

rodrigo7491:

Update to ADR11 describing how TLA+ functions and sequences are to be encoded using SMT arrays. The high level description is the main point of this PR. The `Code changes` subsection will be further detailed as the implementation progresses. Closes #1154.

### DIFF
--- a/docs/src/adr/011adr-smt-arrays.md
+++ b/docs/src/adr/011adr-smt-arrays.md
@@ -265,6 +265,11 @@ attempting to update an entry outside the function's domain will lead to no chan
 (assert (= (select finSucc_5 tuple_with_4) 16)) ; SAT
 ```
 
+In contrast to the current encoding, which produces a number of constraints that is linear in the
+size of the set approximating the function when encoding both function application and update,
+the new encoding will produce a single constraint for each operation. This will potentially lead
+to a significant increase in solving performance.
+
 ### Code changes
 
 The following changes will be made to implement the new encoding of functions:
--- a/docs/src/adr/011adr-smt-arrays.md
+++ b/docs/src/adr/011adr-smt-arrays.md
@@ -180,12 +180,12 @@ The new encoding of functions will thus encompass sequences, as their tuple repr
 intended to be kept.
 
 The new encoding will, like the current one, also map tuples of arguments to return values, but
-will do so natively  instead of simply relying on sets. A function will be represented by two SMT
+will do so natively instead of simply relying on sets. A function will be represented by two SMT
 arrays. The first array will store the domain of the function and will be encoded as a standard 
 TLA+ set. The second array will store the mappings, having sort `<S1,...,Sn>` as its domain, with
 `Si` being the sort of argument `i`, and the sort of the function's codomain as its range. 
 The sorts of the array domain and range can be infinite, but the domain of the function itself,
-and by implication the number of mappings, will always be finite.
+and by implication the number of mappings tuples, will always be finite.
 
 To encode the TLA+ function `finSucc = [x \in {1,2,3} |-> x + 1 ]`, which computes the successors
 of integers from `1` to `3`, we first have to declare its domain, as shown below; tuples are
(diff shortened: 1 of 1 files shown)


docs/src/adr/011adr-smt-arrays.md
```diff
@@ -169,11 +169,112 @@ The following changes will be made to implement the new encoding of sets:
...
+The new encoding of functions will thus encompass sequences, as their tuple representations is
+intended to be kept.
+
+The new encoding will, like the current one, also map tuples of arguments to return values, but
+will do so natively  instead of simply relying on sets. A function will be represented by two SMT
+arrays. The first array will store the domain of the function and will be encoded as a standard 
+TLA+ set. The second array will store the mappings, having sort `<S1,...,Sn>` as its domain, with
+`Si` being the sort of argument `i`, and the sort of the function's codomain as its range. 
+The sorts of the array domain and range can be infinite, but the domain of the function itself,
+and by implication the number of mappings, will always be finite.
```
konnov, line 188: I guess, the number of mappings may be infinite, if a function maps to integers

docs/src/adr/011adr-smt-arrays.md
```diff
@@ -169,11 +169,112 @@ The following changes will be made to implement the new encoding of sets:
...
+             (= finSucc_2 finSucc_1)))
+(assert (ite (select finSucc_domain_3 tuple_with_3)
+             (= finSucc_3 (store finSucc_2 tuple_with_3 4))
+             (= finSucc_3 finSucc_2)))
+```
+
+Note that, unlike with the new encoding for sets, we do not use constant arrays. The reason is that
+the function's domain cannot be altered, so the array has to constrain only the values in said domain.
+Function application can be done by simply accessing the array at the index of the passed arguments.
+A function application with arguments outside the function's domain leads to an unspecified result in
```
konnov, line 235: Just a note: If a function is applied outside of its domain, though the argument belongs to the domain in the arena, it is still fine.

docs/src/adr/011adr-smt-arrays.md
```diff
@@ -169,11 +169,112 @@ The following changes will be made to implement the new encoding of sets:
...
+TLA+, which is perfectly captured by unconstrained entries in the SMT array. Below are some examples
+of function application.
+
+```
+(assert (= (select finSucc_3 tuple_with_1) 2)) ; SAT
+(assert (= (select finSucc_3 tuple_with_2) 3)) ; SAT
+(assert (= (select finSucc_3 tuple_with_3) 4)) ; SAT
```
konnov, line 242: It is worth adding a note that this is in sharp contrast with the OOPSLA encoding, which has to produce a number of constraints that is linear in the size of the set approximation.

docs/src/adr/011adr-smt-arrays.md
```diff
@@ -169,11 +169,112 @@ The following changes will be made to implement the new encoding of sets:
...
+update operator. The update will be encoded as a guarded array update, as illustrated below;
+attempting to update an entry outside the function's domain will lead to no change happening.
+
+```
+(declare-const finSucc_4 (Array Tuple_Int Int))
+(declare-const finSucc_5 (Array Tuple_Int Int))
+
+(assert (ite (select finSucc_domain_3 tuple_with_1) ; [finSucc EXCEPT ![1] = 9]
+             (= finSucc_4 (store finSucc_3 tuple_with_1 9))
+             (= finSucc_4 finSucc_3)))
```
konnov, line 258: This also produces a linear number of constraints, in contrast to the linear number of constraints in the OOPSLA encoding.

konnov (approved):
Looks great!

docs/src/adr/011adr-smt-arrays.md
rodrigo7491: I don't know how this can be possible. Doesn't the number of mappings equal the cardinality of the function's domain? If so, and if we are not mapping from an infinite set, the number of mappings should be finite.

PR #1158, opened 2021-12-07
Add cache for empty set declarations

### LINKED ISSUE #1100: [FEATURE] Remove duplicate empty set representations in arrays encoding
In the initial version of the arrays encoding, every set declaration is followed by an equality to a new empty set, as shown below.
```
(declare-const set1 (Array some_sort Bool))
(assert (= set1 ((as const (Array some_sort Bool)) false)))
```
As suggested by [Nikolaj Bjorner](https://github.com/NikolajBjorner) in an e-mail exchange, we should define constant empty sets for every type used in the spec and refer to them instead, as shown below for integers.
```
(define-const emptyIntSet (Array Int Bool) ((as const (Array Int Bool)) false))
(declare-const set1 (Array Int Bool))
(assert (= set1 emptyIntSet))
```

rodrigo7491:

To avoid duplicate declarations of empty sets, they are now cached and reused. Instead of the initially intended `define-const` macro, the already used `declare-const` is used, since Z3''s Java API does not (apparently) expose `define-const`. Closes #1100.


konnov (approved):
LGTM

PR #1156, opened 2021-12-06
ADR013: Configuration management system

shonfeder:
This ADR proposes adoption of the `PureConfig` library for configuration
management. Please see the ADR for details on the proposal.

Towards #1069

- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] Documentation added for any new functionality

PR #1160, opened 2021-12-08
Add ConfigManager

### LINKED ISSUE #1069: [FEATURE] Route all configuration management through a single component

shonfeder:
Closes #1069

This adds the configuration management component, based around `PureConfig`,
decided for in #1156:

## Changes:

- [x] Add dependency on `PureConfig`
- [x] Specify `ApalacheConfigs` trait (extending `OutputManagerConfig`) to serve
      as the collection of loadable application configurations.
- [x] Add new `ConfigManager(cli: CliArgs)` class as a sibling of `OutputManager`
- [x] Look for a local config file `.apalache.cfg`, starting from the cwd, and climbing up through parent dirs.
- [x] Factor out all configuration from `OutputManager`, removing the
      `syncFromCli` method.

## Routine:

### DIFF
--- a/docs/src/adr/009adr-outputs.md
+++ b/docs/src/adr/009adr-outputs.md
@@ -1,8 +1,9 @@
-# ADR-008: Apalache Outputs
+---
+authors: Jure Kukovec, Shon Feder
+last revised: 2021-12-14
+---
 
-| author                   | revision |
-|--------------------------|---------:|
-| Jure Kukovec, Shon Feder |        2.1 |
+# ADR-008: Apalache Outputs
 
 This ADR documents the various files produced by Apalache, and where they get written to.
 
@@ -67,9 +68,18 @@ following:
 
 ## 4. Global Configuration File
 
-Apalache should define a global configuration file `apalache-global-config`, e.g. in the `<HOME>/.tlaplus` directory, in which users can define the default values of all parameters, including all flags listed in section 1, as well as `out-dir`. The format of the configuration file is an implementation detail and will not be specified here.
+Apalache should define a global configuration file `apalache.cfg`, e.g. in the `<HOME>/.tlaplus` directory, in which users can define the default values of all parameters, including all flags listed in section 1, as well as `out-dir`. The format of the configuration file is an implementation detail and will not be specified here.
+
+Apalache should also look for a local configuration file `.apalache.cfg`, within
+current working directory or its parents. If it finds such file, any configured
+parameters therein will override the parameters from the global config file.
+
 If a parameter is specified in the configuration file, it replaces the default value, but specifying a parameter manually always overrides config defaults.
 In other words, parameter values are determined in the following way, by order of priority:
-  1. If `--<flag>=<value>` is given, use `<value>`, otherwise
-  2. If `apalache-global-config` specifies `<flag>=<CFGvalue>` use `<CFGvalue>`, otherwise 
-  3. Use Apalache defaults.
+
+1. If `--<flag>=<value>` is given, use `<value>`, otherwise
+2. if a local `.apalache.cfg` file is found containing `<flag>: <value>`, then
+    use `<value>`, otherwise
+3. if the global `apalache.cfg` specifies `<flag>=: <value>` use `<value>`,
+    otherwise
+4. Use the defaults specified in the `ApalacheConfig` class.
--- a/docs/src/adr/009adr-outputs.md
+++ b/docs/src/adr/009adr-outputs.md
@@ -78,8 +78,8 @@ If a parameter is specified in the configuration file, it replaces the default v
 In other words, parameter values are determined in the following way, by order of priority:
 
 1. If `--<flag>=<value>` is given, use `<value>`, otherwise
-2. if a local `.apalache.cfg` file is found containing `<flag>: <value>`, then
+2. if a local `.apalache.cfg` file is found (or is specified with the `--config-file` argument) containing `<flag>: <value>`, then
     use `<value>`, otherwise
-3. if the global `apalache.cfg` specifies `<flag>=: <value>` use `<value>`,
+3. if the global `apalache.cfg` specifies `<flag>: <value>` use `<value>`,
     otherwise
 4. Use the defaults specified in the `ApalacheConfig` class.
--- a/docs/src/apalache/config.md
+++ b/docs/src/apalache/config.md
@@ -16,10 +16,10 @@ a later numbered source.
 ## Command line arguments and environment variables
 
 To view the available command line arguments, run Apalache with the `--help`
-flag and consult the section on [Running the Too](./running.md) for more
+flag and consult the section on [Running the Tool](./running.md) for more
 details.
 
-*Some* parameters configurable the command are also configurable via environment
+*Some* parameters configurable via the command line are also configurable via environment
 variables. These parameters are noted in the CLI's inline help. If a parameter
 is configured both through a CLI argument and an environment variable, then the
 CLI argument always takes precedence. 
--- a/docs/src/apalache/config.md
+++ b/docs/src/apalache/config.md
@@ -1,23 +1,72 @@
-# Apalache global configuration file
-Apalache allows you to specify certain parameters, governing outputs produced by it, as described in [ADR-009](../adr/009adr-outputs.md).
+# Apalache configuration
 
-You can create a global configuration file named `apalache-global-config.yml` in `$HOME/.tlaplus/` and populate it with values for the following flags, in YAML format:
-  - `out-dir` (String): The value of `out-dir` defines a path to the directory,
-  in which all Apalache runs write their outputs. Each run will produce a unique
-  subdirectory inside `out-dir` using the following convention:
-  `<SPECNAME>_yyyy-MM-dd_HH-mm-ss_<UNIQUEID>`. The use of `~` in the path
-  specification is permitted. The directory path need not already exist.
-  Example value: `'~/apalache-out'`.
-  If this value is not defined in `apalache-global-config.yml`, Apalache will, for each individual run, define it to be equal to `CWD/_apalache-out/`, where `CWD` is the current working directory.
-  - `profiling` (Bool): This flag governs the creation of `profile-rules.txt` used in [profiling](profiling.md). The file is only created if `profiling` is set to `True`.  Setting `profiling` to `False` is incompatible with the `--smtprof` flag.
-  The default is `False`.
-  - `write-intermediate` (Bool): This Boolean flag governs the creation of intermediate outputs. If set to `True`, apalache will produce an `intermediate` subdirectory in the run directory and output the state of the module after each pass.
-  The default is `False`.
+Apalache supports configuration of some parameters governing its behavior.
 
+Application configuration is loaded from the following four sources:
 
-Example of a valid `apalache-global-config.yml`:
+1. Command line arguments
+2. Environment variables
+3. A local configuration file
+4. The global configuration file
+
+The order of precedence of the sources follows their numbering: i.e., and any
+configuration set in an earlier numbered source overrides a configuration set in
+a later numbered source.
+
+## Command line arguments and environment variables
+
+To view the available command line arguments, run Apalache with the `--help`
+flag and consult the section on [Running the Too](./running.md) for more
+details.
+
+*Some* parameters configurable the command are also configurable via environment
+variables. These parameters are noted in the CLI's inline help. If a parameter
+is configured both through a CLI argument and an environment variable, then the
+CLI argument always takes precedence. 
+
+## Configuration files
+
+### File format and supported parameters
+
+Local configuration files support JSON and the JSON superset
+[HOCON](https://github.com/lightbend/config/blob/master/HOCON.md).
+
+Here's an example of a valid configuration for the currently 
+supported parameters, along with their default values: 
+
+```yaml
+# Directory in which to write all log files and records of each run
+out-dir: "${PWD}/_apalache-out"
+
+# Whether or not to write additional files, that report on intermediate
+# processing steps
+write-intermediate: false
+
+# Whether or not to write the SMT profiling file
+profiling: false
+
+# Fixed directory into which generated files are written (absent by default)
+# run-dir: ~/my-run-dir
 ```
-out-dir: '~/apalache-out'
-profiling: True
-write-intermediate: True
-```
+
+A `~` found at the beginning of a file path will expanded into the value set for
+the user's home directory.
+
+Details on the effect of these parameters can be found in [Running the
+Too](./running.md).
+
+### Local configuration file
+
+You can specify a local configuration file explicitly via the `config-file`
+command line argument. If this is not provided, then Apalache will look for the
+nearest `.apalache.cfg` file, beginning in the current working directory and
+searching up through its parents.
+
+Parameters configured in the local configuration file will be overridden by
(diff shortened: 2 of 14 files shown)


docs/src/adr/009adr-outputs.md
```diff
@@ -67,9 +68,18 @@ following:
 
 ## 4. Global Configuration File
 
-Apalache should define a global configuration file `apalache-global-config`, e.g. in the `<HOME>/.tlaplus` directory, in which users can define the default values of all parameters, including all flags listed in section 1, as well as `out-dir`. The format of the configuration file is an implementation detail and will not be specified here.
+Apalache should define a global configuration file `apalache.cfg`, e.g. in the `<HOME>/.tlaplus` directory, in which users can define the default values of all parameters, including all flags listed in section 1, as well as `out-dir`. The format of the configuration file is an implementation detail and will not be specified here.
+
+Apalache should also look for a local configuration file `.apalache.cfg`, within
+current working directory or its parents. If it finds such file, any configured
+parameters therein will override the parameters from the global config file.
```
konnov, line 75: I wonder, whether we want to merge the TLC config in our configuration settings too. Certainly, we have way more options than the TLC config.

docs/src/apalache/config.md
```diff
@@ -1,23 +1,72 @@
...
+variables. These parameters are noted in the CLI's inline help. If a parameter
+is configured both through a CLI argument and an environment variable, then the
+CLI argument always takes precedence. 
+
+## Configuration files
+
+### File format and supported parameters
+
+Local configuration files support JSON and the JSON superset
+[HOCON](https://github.com/lightbend/config/blob/master/HOCON.md).
```
konnov, line 32: JSON has a weird limit on the length of integer literals. Does it apply to HOCON too?

docs/src/apalache/config.md
```diff
@@ -1,23 +1,72 @@
...
+Here's an example of a valid configuration for the currently 
+supported parameters, along with their default values: 
+
+```yaml
+# Directory in which to write all log files and records of each run
+out-dir: "${PWD}/_apalache-out"
+
+# Whether or not to write additional files, that report on intermediate
+# processing steps
+write-intermediate: false
```
konnov, line 43: This example looks like yaml, not json, which contradicts the preceding sentence.

docs/src/apalache/running.md
```diff
@@ -11,10 +11,13 @@ The model checker can be run as follows:
 ```bash
 $ apalache check [--config=filename] [--init=Init] [--cinit=ConstInit] \
     [--next=Next] [--inv=Inv] [--length=10] [--algo=(incremental|offline)] \
-    [--discard-disabled] [--no-deadlock] [--tuning=filename] [--tune-here=options] \
+    [--discard-disabled] [--no-deadlock] \
+    [--tuning=filename] [--tune-here=options] \
```
konnov, line 15: Also, I would like to integrate the tuning options in this file.

docs/src/adr/009adr-outputs.md
```diff
@@ -67,9 +68,18 @@ following:
...
 In other words, parameter values are determined in the following way, by order of priority:
-  1. If `--<flag>=<value>` is given, use `<value>`, otherwise
-  2. If `apalache-global-config` specifies `<flag>=<CFGvalue>` use `<CFGvalue>`, otherwise 
-  3. Use Apalache defaults.
+
+1. If `--<flag>=<value>` is given, use `<value>`, otherwise
+2. if a local `.apalache.cfg` file is found containing `<flag>: <value>`, then
+    use `<value>`, otherwise
+3. if the global `apalache.cfg` specifies `<flag>=: <value>` use `<value>`,
+    otherwise
```
konnov, line 84: Shall we add a CLI option to specify the location of a config file? I am thinking about the case of having several configuration files for experiments in the same directory.

konnov: I see that you have the CLI option `--config-file` below. If it specifies the config file, shall we add it on the list?

docs/src/apalache/running.md
```diff
@@ -242,12 +252,38 @@ the [TLA+ Community Modules](https://github.com/tlaplus/CommunityModules).
...
+For an example, consider using the default location of the `run-dir` for a run of
+Apalache on a spec named `test.tla`. This will create a directory structuring matching following pattern:
+
+```
+./_apalache-out/
+└── test.tla
+    └── 2021-11-05T22-54-55_810261790529975561
+```
+
+The default value for the `out-dir` is the `_apalahce-out` directory in the
```
konnov, line 272:
```suggestion
The default value for the `out-dir` is the `_apalache-out` directory in the
```

konnov (approved):
Looks great! See several comments.

docs/src/adr/009adr-outputs.md
```diff
@@ -67,9 +68,18 @@ following:
 
 ## 4. Global Configuration File
 
-Apalache should define a global configuration file `apalache-global-config`, e.g. in the `<HOME>/.tlaplus` directory, in which users can define the default values of all parameters, including all flags listed in section 1, as well as `out-dir`. The format of the configuration file is an implementation detail and will not be specified here.
+Apalache should define a global configuration file `apalache.cfg`, e.g. in the `<HOME>/.tlaplus` directory, in which users can define the default values of all parameters, including all flags listed in section 1, as well as `out-dir`. The format of the configuration file is an implementation detail and will not be specified here.
```
Kukovec, line 71: I recall Igor complaining about the name `apalache.cfg` last time I suggested it
https://github.com/informalsystems/apalache/pull/1025#discussion_r728696302

docs/src/adr/009adr-outputs.md
```diff
@@ -67,9 +68,18 @@ following:
...
 If a parameter is specified in the configuration file, it replaces the default value, but specifying a parameter manually always overrides config defaults.
 In other words, parameter values are determined in the following way, by order of priority:
-  1. If `--<flag>=<value>` is given, use `<value>`, otherwise
-  2. If `apalache-global-config` specifies `<flag>=<CFGvalue>` use `<CFGvalue>`, otherwise 
-  3. Use Apalache defaults.
+
+1. If `--<flag>=<value>` is given, use `<value>`, otherwise
+2. if a local `.apalache.cfg` file is found containing `<flag>: <value>`, then
+    use `<value>`, otherwise
+3. if the global `apalache.cfg` specifies `<flag>=: <value>` use `<value>`,
```
Kukovec, line 83: Why is one `: ` and the other `=:` ?

docs/src/apalache/config.md
```diff
@@ -1,23 +1,72 @@
...
+4. The global configuration file
+
+The order of precedence of the sources follows their numbering: i.e., and any
+configuration set in an earlier numbered source overrides a configuration set in
+a later numbered source.
+
+## Command line arguments and environment variables
+
+To view the available command line arguments, run Apalache with the `--help`
+flag and consult the section on [Running the Too](./running.md) for more
```
Kukovec, line 19:
```suggestion
flag and consult the section on [Running the Tool](./running.md) for more
```

docs/src/apalache/config.md
```diff
@@ -1,23 +1,72 @@
...
+details.
+
+*Some* parameters configurable the command are also configurable via environment
```
Kukovec, line 22:
```suggestion
*Some* parameters configurable via the command line are also configurable via environment
```

docs/src/apalache/config.md
```diff
@@ -1,23 +1,72 @@
...
+variables. These parameters are noted in the CLI's inline help. If a parameter
+is configured both through a CLI argument and an environment variable, then the
+CLI argument always takes precedence.
```
Kukovec, line 25: This is a bit redundant, you explain precedence ~10 lines above.

docs/src/apalache/config.md
```diff
@@ -1,23 +1,72 @@
...
+
+Details on the effect of these parameters can be found in [Running the
+Too](./running.md).
+
+### Local configuration file
+
+You can specify a local configuration file explicitly via the `config-file`
+command line argument. If this is not provided, then Apalache will look for the
+nearest `.apalache.cfg` file, beginning in the current working directory and
+searching up through its parents.
```
Kukovec, line 63: I wonder if crawling up the directory tree makes sense. As a user, I know I'd hate for the tool to do unexpected things because there's a `.cfg` file 2 folders above.

docs/src/apalache/running.md
```diff
@@ -242,12 +252,38 @@ the [TLA+ Community Modules](https://github.com/tlaplus/CommunityModules).
...
+The directory also includes a file `run.txt`, reporting the command line
+arguments used during the run.
 
-Additionally, if the flag `--write-intermediate=true` is set or this option is
-enabled in the `apalache-global-config.yml` (see [configuration
-instructions](config.md)) each pass of the model checker produces an
-intermediate TLA+ file in the run-specific directory:
+Additionally, if the parameter `write-intermediate` is set to `true` (see
+[configuration instructions](config.md)) each pass of the model checker produces
+an intermediate TLA+ file in the run-specific subdirectory:
```
Kukovec, line 286:
```suggestion
intermediate TLA+ files in the run-specific subdirectory:
```

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
```diff
@@ -0,0 +1,126 @@
...
+case class ApalacheConfig(
+    file: Option[File] = None, outDir: File = new File(System.getProperty("user.dir"), "_apalache-out"),
+    runDir: Option[File] = None, configFile: Option[File] = None, writeIntermediate: Boolean = false,
+    profiling: Boolean = false
+)
+
+case class ConfigManager(cmd: CliConfig) {
+  val TLA_PLUS_DIR = ".tlaplus"
+  val APALACHE_CFG = "apalache.cfg"
+  val DOT_APALACHE_CFG = "." + APALACHE_CFG
```
Kukovec, line 64: Wouldn't these be better suited for a companion object?

Kukovec (approved):
LGTM. I'm just happy `General` finally defines a `file` field.

docs/src/apalache/config.md
shonfeder: I don't see a limit on integer sizes in the JSON spec: https://www.json.org/json-en.html

Is this an implementation detail of certain deserialization libs, or is there a different spec where you've seen this?

I'm not sure whether the implementation of HOCON has such a limit. The docs seem to indicate a limit of 64 bit integers: https://github.com/lightbend/config/blob/main/HOCON.md

I don't see how this could be a problem for any of the current configurations we support tho.

docs/src/apalache/config.md
shonfeder: Which sentence does it contradict? This is valid HOCON.

docs/src/apalache/config.md
shonfeder: This is very common behavior. Tools you already use follow this behavior. E.g., git. Git looks for the closest `.git`, descending up parents, and uses the data and configuration from that to determine its behavior when you invoke it.

Basically creating a `.apalache.cfg` file in a directory will make all subdirectories part of a "project" that share the same config. For another example of this behavior, see the `sbt` documentation, where the scope of a project is determined by finding the closest parent directory containing a `build.sbt` file: https://www.scala-sbt.org/1.x/docs/Directories.html#Base+directory

docs/src/apalache/running.md
shonfeder: Thanks!

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
shonfeder: I don't know :)

What's the advantage of putting it it in the companion object?

I do think these should private tho. I'll at least change that.

I can't find anything in the Scala style guide that says where fields like this should be declared. Could you point me towards something that explains why putting such things in companion object might be preferable?

My own feeling is that -- in this case -- moving these into the companion object it would negatively impact readability: these  values are only accessed in the class methods, and don't show up anywhere in the one method in the companion object. Moving them would then require readers to jump back and forth between the class and the companion object to see what means what.

shonfeder, line 64:
```suggestion
  private val TLA_PLUS_DIR = ".tlaplus"
  private val APALACHE_CFG = "apalache.cfg"
  private val DOT_APALACHE_CFG = "." + APALACHE_CFG
```

docs/src/apalache/config.md
Kukovec: I suppose, but from my experience TLA+ repositories are far less branching and project-based than code repositories (typically, it's 1-4 specs in a single folder). I'm not opposed to this behavior per se, just that my assumption as a user would be that configs are either in the same folder or global. It's fine, as long as it's clearly documented, like in the paragraph above.

shonfeder: If it ends up causing any confusion, we can remove this behavior. But I expect it might be helpful for many of our emerging use cases, where we are integrating the specs alongside the implementation, and as a result need to cope with all the complicated folder structures of a standard project.

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
Kukovec: I think we generally have the convention across the codebase that any `public static X`-type value goes in a companion object, so it can be accessed without creating a fresh instance. 
This page quotes the "Scala Cookbook": https://alvinalexander.com/scala/how-to-static-members-in-scala-companion-objects-fields-methods/

shonfeder: Does the fact that these are (now) private and *not* meant to be accessed outside of the class at all impact your feelings on this? (I.e., they are never supposed to be accessed at all, whether by creating instance or not).

I still don't see there anything that explains a benefit in this case, and the downside on readability (and abstraction) seem pretty evident to me. I don't want to buck our conventions, but I'd rather there be a good reason if I'm going to negatively impact readability.

shonfeder: A bit more concretely, that recipe is for the case where

> You want to create a Scala class that has both instance methods and static methods, but unlike Java, Scala does not have a static keyword.

afaict, that's not what we're actually after here.

PR #1161, opened 2021-12-08
Refactor SMT encoding flag use

### LINKED ISSUE #1159: [FEATURE] Refactor SMT encoding option flag use
The new SMT encoding option flag is currently hardcoded in many places, making it hard to maintain it. We should have it as a global constant, as suggested by @Kukovec in #1093, and as an enum type instead of plain string, as suggested by @shonfeder in a discussion. A check for the correct encoding use by different symbolic state rewriter implementations also should be added, as suggested by @konnov in #1121.

rodrigo7491:

This PR makes the the SMT encoding options global constants (see `SMTEncodings`). The bulk of the changes relates to enabling access to the new global constants. Regarding the use of the correct encoding for symbolic state rewriter instantiations, all such uses are already guarded; every instantiation, current and future, needs to check the encoding flag to select the appropriate rewriter class, so this is a non-issue. Closes #1159.

### DIFF
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
@@ -1,7 +1,9 @@
 package at.forsyte.apalache.tla.tooling.opt
 
-import java.io.File
+import at.forsyte.apalache.tla.bmcmt.{SMTEncoding, arraysEncoding, oopsla19Encoding}
+import org.backuity.clist.util.Read
 
+import java.io.File
 import org.backuity.clist.{Command, _}
 
 /**
@@ -11,12 +13,20 @@ import org.backuity.clist.{Command, _}
  */
 class CheckCmd extends Command(name = "check", description = "Check a TLA+ specification") with General {
 
+  // Parses the smtEncoding option
+  implicit val smtEncodingRead: Read[SMTEncoding] =
+    Read.reads[SMTEncoding]("a SMT encoding, either oopsla19 or arrays") {
+      case "arrays"        => arraysEncoding
+      case "oopsla19"      => oopsla19Encoding
+      case oddEncodingType => throw new IllegalArgumentException(s"Unexpected SMT encoding type $oddEncodingType")
+    }
+
   var file: File = arg[File](description = "a file containing a TLA+ specification (.tla or .json)")
   var nworkers: Int = opt[Int](name = "nworkers", default = 1,
       description = "the number of workers for the parallel checker (soon), default: 1")
   var algo: String = opt[String](name = "algo", default = "incremental",
       description = "the search algorithm: offline, incremental, parallel (soon), default: incremental")
-  var smtEncoding: Option[String] = opt[Option[String]](name = "smt-encoding", useEnv = true,
+  var smtEncoding: SMTEncoding = opt[SMTEncoding](name = "smt-encoding", useEnv = true, default = oopsla19Encoding,
       description =
         "the SMT encoding: oopsla19, arrays (experimental), default: oopsla19 (overrides envvar SMT_ENCODING)")
   var config: String = opt[String](name = "config", default = "",
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SMTEncodings.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SMTEncodings.scala
@@ -1,6 +1,6 @@
 package at.forsyte.apalache.tla.bmcmt
 
-object SMTEncodings extends Enumeration {
-  type SMTEncoding = Value
-  val oopsla19Encoding, arraysEncoding, invalidEncoding = Value
-}
+sealed trait SMTEncoding
+
+case object oopsla19Encoding extends SMTEncoding
+case object arraysEncoding extends SMTEncoding
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SMTEncodings.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SMTEncodings.scala
@@ -1,0 +1,6 @@
+package at.forsyte.apalache.tla.bmcmt
+
+object SMTEncodings extends Enumeration {
+  type SMTEncoding = Value
+  val oopsla19Encoding, arraysEncoding, invalidEncoding = Value
+}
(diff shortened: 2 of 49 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
```diff
@@ -83,7 +81,11 @@ class BoundedCheckerPassImpl @Inject() (val options: PassOptions, hintsStore: Fo
     val tuning = options.getOrElse[Map[String, String]]("general", "tuning", Map[String, String]())
     val debug = options.getOrElse[Boolean]("general", "debug", false)
     val saveDir = options.getOrError[Path]("io", "outdir").toFile
-    val smtEncoding = options.get[Option[String]]("checker", "smt-encoding").flatten.getOrElse(oopsla19Encoding)
+    val smtEncoding = options.get[Option[String]]("checker", "smt-encoding").flatten.getOrElse("oopsla19") match {
+      case "oopsla19"      => oopsla19Encoding
+      case "arrays"        => arraysEncoding
+      case oddEncodingType => throw new IllegalArgumentException(s"Unexpected SMT encoding type $oddEncodingType")
+    }
```
shonfeder, line 88: My expectation would be that instead of saving this as a string in `options`, we would save it the enum type, so that we could here just do `get[Option[SMTEncodings]]`. So, ideally, I think we should push the conversion from string into the cli parser. Documentation on extending the CLI parser to read these arguments into the target type are here: https://github.com/backuity/clist#parsing

I'd be be happy to sync up to discuss how to apply this approach!

shonfeder (changes requested):
I suggest pushing this farther out, in the outermost edge of the program, so that we only ever the `SMTEncoding`, and never a string representation to deal with, once we've gotten beyond initial IO required for configuring startup of the program. I've left one comment trying to indicate how this can be done.

The principle that guides this suggestion is well articulated in Alexi King's awesome post, "Parse, Don't Validate": https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

shonfeder:
I should add, other than this small suggested change to where we introduce the mapping from strings, the changes here generally look good to me! But I'll take another quick look after that change is made (or we have decided it is not warranted).

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
rodrigo7491: Thanks for the suggestion! You did mention the "Parse, Don't Validate" post to me before, but I guess haven't internalized it yet. I moved the check from `BoundedCheckerPassImpl` to `Tool`.

I wanted to have it in `CheckCmd`, but it seems that the CLI parses doesn't support this. I found the following comment on the [clist repo](https://github.com/backuity/clist/blob/master/core/src/main/scala/org/backuity/clist/util/Read.scala): "Note that we do not support scala Enumeration as this would require having a runtime dependency on the reflection API. On the other hand Java enums are supported.".

rodrigo7491:
I realized that the encoding option is also relevant for the `test` command, since it is a special case of `check`. Maybe it is better to keep things as they are for now, since the arrays encoding option is still experimental, but I will make a mental note of that (in addition to the very concrete note you just read). @shonfeder, that is the reason why I had to add a default value for the encoding in `BoundedCheckerPassImpl`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
shonfeder: Ok. That's not too bad. A solution might be to make this a case class, but this is already pretty far out in the edge of the program :)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SMTEncodings.scala
```diff
@@ -0,0 +1,6 @@
+package at.forsyte.apalache.tla.bmcmt
+
+object SMTEncodings extends Enumeration {
+  type SMTEncoding = Value
+  val oopsla19Encoding, arraysEncoding, invalidEncoding = Value
+}
```
shonfeder, line 6: Sorry for the churn here, but your remark about the lack of support for enums in the CLI library led me to look up info about enums in Scala, only to find that, for enums:

> There’s no exhaustive matching check during compile

This defeats about 50% of the advantage from using some kind of disjoint union here. This is due to lack of clarity in my own expressions, tho I also blame it partially on Scala having multiple,  variously defective, ways of achieving the same thing.

E.g., in the [scalabook docs on "Case Objecs"](https://docs.scala-lang.org/overviews/scala-book/case-objects.html) they write:

> Creating enumerations with case objects
>
> As we showed earlier in this book, you create enumerations in Scala like this:
>
> ```
> sealed trait Topping
> case object Cheese extends Topping
> case object Pepperoni extends Topping
> case object Sausage extends Topping
> case object Mushrooms extends Topping
> case object Onions extends Topping
> ```

That is, according to this resource, "you create enumerations" not by using `Enumeration` (as one might very well think!) but by defining case classes/case objects :/

So, again, sorry for the churn, but what we really want here is a (sealed) case object (or case class, if we end up needing to carry around any additional data).

This will give us 

- exhaustiveness checking, meaning there shouldn't be any need for the `invalidEncoding` value any more
- ability to parse the CLI argument directly into the proper type, so we don't have to do the conversion twice.

shonfeder (changes requested):

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SMTEncodings.scala
rodrigo7491: No need to apologize, it is a good thing that you caught this! I will try to make the change to case objects as suggested. Thanks!

konnov:
The PR looks OK to me. It would be great to take @shonfeder 's suggestions into account.

rodrigo7491:
Thanks for the review @konnov! We seem to be almost done. @shonfeder, can you have a quick look at the changes to `SMTEncodings`, `CheckCmd`, and `Tool`?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
```diff
@@ -11,12 +13,20 @@ import org.backuity.clist.{Command, _}
...
 class CheckCmd extends Command(name = "check", description = "Check a TLA+ specification") with General {
 
+  // Parses the smtEncoding option
+  implicit val smtEncodingRead: Read[SMTEncoding] =
+    Read.reads[SMTEncoding]("a SMT encoding, either oopsla19 or arrays") {
+      case "arrays"        => arraysEncoding
+      case "oopsla19"      => oopsla19Encoding
+      case oddEncodingType => throw new IllegalArgumentException(s"Unexpected SMT encoding type $oddEncodingType")
+    }
+
```
shonfeder, line 23: This is actually really lovely. It introduces a clear example of the right way to do all such flags I think, and sets a great precedent. Thanks for doing the work to get this in place!

shonfeder (approved):
LGTM

Thanks!

PR #1163, opened 2021-12-10
add Lamport's mutex

konnov:
Add Lamport's mutex, as it is an interesting example for typechecking records.


Kukovec (approved):
LGTM

PR #1164, opened 2021-12-10
Replace implactions by ITE in array updates in set union rule

### LINKED ISSUE #1162: [BUG] Unconstrained sets due to incorrect use of set membership implications
## Description

The OOPSLA19 encoding uses the structure `tla.equiv(inOldSet, inNewSet)` extensively. This causes problems to the arrays encoding, since the equivalence does not constrain the SSA representation of `newSet` if `inOldSet` does not hold. This was the root cause of #1152, and is also happening with non-trivial runs of `SetAddDel.tla` ([report](https://github.com/informalsystems/apalache-tests/pull/100/files)).

## Input specification

[SetAddDel.tla](https://github.com/informalsystems/apalache-tests/blob/master/performance/array-encoding/SetAddDel.tla)

`check --smt-encoding=arrays --length=4 --cinit=CInit4 --init=Init --next=Next --inv=Inv`

## Expected behavior

No error.

## Additional context

For `SetAddDel.tla`, the implications in the class `SetCupRule` seem to be the problem. To avoid this kind of issue in the future, the rules currently supported in the arrays encoding should be checked for occurrences of implications as described above.

rodrigo7491:

It seems that the only remaining incorrect use of implications when dealing with SMT arrays is in set union rule, which is fixed by this PR. Occurrences of implications in rules not yet supported by the arrays encoding will be looked at when these rules are reworked. Closes #1162.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,3 +13,4 @@
 ### Bug fixes
 
 * Fix the use of set minus in the array encoding, see #1152
+* Fix the use of set union in the array encoding, see #1162
--- a/test/tla/SetAddDel.tla
+++ b/test/tla/SetAddDel.tla
@@ -1,0 +1,40 @@
+---------------------------- MODULE SetAddDel --------------------------
+\* A regression test for the use of set union, see:
+\* https://github.com/informalsystems/apalache/issues/1162
+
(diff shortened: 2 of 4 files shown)


konnov (approved):
LGTM

Issue #649, opened 2021-03-15
[FEATURE] Add unit tests for Normalizer


konnov:
We need a complete coverage for the `Normalizer` transformation. All expressions are passing through this transformation.

PR #1165, opened 2021-12-10
New test scenarios for Normalizer

gabrielamafra:
Hello :octocat: 

Here are some new tests for the normalizer, which currently has poor coverage as stated in https://github.com/informalsystems/apalache/issues/649. Scenarios that have some change in the expression are covered with individual tests, while expressions that should be left unchanged are all covered in a single test.

This also fixes a bug were temporal expressions with negated `\box` operators were being incorrectly normalized (see [discussion](https://github.com/informalsystems/apalache/pull/1165#discussion_r767738234)), and a bug where declarations inside negated `LET` expressions were not being normalized at all.

Fixes https://github.com/informalsystems/apalache/issues/649

### DIFF
--- a/tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
+++ b/tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
@@ -237,18 +237,39 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
 
   test("""unchanged cases""") {
     val expressions = List(
-        // x ~> y
-        tla.leadsTo(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1()
-        // TODO: Add remaining unchanged scenarios
+        // x ~> y and negation
+        tla.leadsTo(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1(),
+        tla.not(tla.leadsTo(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1()) as BoolT1(),
+        // x -+-> y and negation
+        tla.guarantees(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1(),
+        tla.not(tla.guarantees(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1()) as BoolT1(),
+        // WF_<x>(y) and negation
+        tla.WF(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1(),
+        tla.not(tla.WF(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1()) as BoolT1(),
+        // SF_<x>(y) and negation
+        tla.SF(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1(),
+        tla.not(tla.SF(tla.name("x") as BoolT1(), tla.name("y") as BoolT1()) as BoolT1()) as BoolT1(),
+        // \E x \in s . y
+        tla.exists(tla.name("x") as BoolT1(), tla.name("s") as SetT1(IntT1()), tla.name("y") as BoolT1()) as BoolT1(),
+        // \A x \in s . y
+        tla.forall(tla.name("x") as BoolT1(), tla.name("s") as SetT1(IntT1()), tla.name("y") as BoolT1()) as BoolT1(),
+        // x < y
+        tla.lt(tla.name("x") as IntT1(), tla.name("y") as IntT1()) as BoolT1(),
+        // x <= y
+        tla.le(tla.name("x") as IntT1(), tla.name("y") as IntT1()) as BoolT1(),
+        // x > y
+        tla.gt(tla.name("x") as IntT1(), tla.name("y") as IntT1()) as BoolT1(),
+        // x >= y
+        tla.ge(tla.name("x") as IntT1(), tla.name("y") as IntT1()) as BoolT1(),
+        // x = y
+        tla.eql(tla.name("x") as IntT1(), tla.name("y") as IntT1()) as BoolT1(),
+        // x \in s
+        tla.in(tla.name("x") as BoolT1(), tla.name("s") as SetT1(BoolT1())) as BoolT1()
     )
 
     expressions.foreach({ expression =>
       val output = normalizer.apply(expression)
       assert(expression == output)
-
-      val negatedExpression = tla.not(expression) as BoolT1()
-      val negatedOutput = normalizer.apply(negatedExpression)
-      assert(negatedExpression == negatedOutput)
     })
   }
 }
--- a/tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
+++ b/tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
@@ -246,13 +246,12 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
     assert(expected == output)
   }
 
-  test("""~ (LET x = ~FALSE IN ~x ~~> LET x = ~FALSE IN x)""") {
+  test("""~ (LET x = ~FALSE IN ~x ~~> LET x = TRUE IN x)""") {
     val decl = tla.declOp("x", tla.not(tla.bool(false)) as BoolT1()) as BoolT1()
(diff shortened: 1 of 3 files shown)


Kukovec:
> ... its not important to test scenarios where the recursion is called without change (i.e. for operators such as leads to `~>`). 
> 

I disagree, in the sense that it _is_ important to test that expected no-ops are actual no-ops. There's value in `assert(input == output)` tests. 

As as side note, PRs labeled as [WIP] usually don't get reviewed (by convention) until the tag is removed (or the author explicitly asks).

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
```diff
@@ -38,6 +38,30 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
     assert(expected == output)
   }
 
+  test("""~ IF x THEN y ELSE z ~~> IF x THEN ~y ELSE ~z""") {
+    val input = tla
+      .not(
+          tla.ite(
+              tla.name("x") ? "b",
+              tla.name("y") ? "b",
+              tla.name("z") ? "b"
```
konnov, line 47: Let's produce types with the newer API, that is, `tla.name("x") as BoolT1()`. The benefit is that we do not have to carry a map of types around. We have not switched to the new API in all tests yet, so `? "foo"` appears in a lot of places.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
```diff
@@ -38,6 +38,30 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
...
+          ) ? "b"
+      )
+      .typed(types, "b")
```
konnov, line 50:
```suggestion
      as BoolT1()
```

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
```diff
@@ -63,4 +87,32 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
...
     assert(expected == output)
   }
+
+  test("""x => y ~~> ~x \/ y""") {}
+
+  test("""~ (x => y) ~~> x /\ ~y""") {}
+
+  test("""(~FALSE <=> x) ~~> TRUE <=> x""") {}
+
+  test("""~ (~FALSE <=> x) ~~> TRUE \neq x""") {}
```
konnov, line 97: Why do we preserve `<=>` in one case and produce `\neq` in another?

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
```diff
@@ -63,4 +87,32 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
...
+
+  test("""~ (x < y) ~~> x >= y""") {}
+
+  test("""~ (x <= y) ~~> x > y""") {}
+
+  test("""~ (x > y) ~~> x <= y""") {}
+
+  test("""~ (x >= y) ~~> x < y""") {}
+
+  test("""~(x \neq y) ~~> x <=> y""") {}
```
konnov, line 111: Shall we produce `=` instead of `<=>` in this case?

konnov (approved):
Looks good! See a few comments.

I agree with Jure that we should in principle cover the cases when the Normalizer is not changing anything. Perhaps, introducing a lot of tests for that is not a good idea. Maybe we can introduce one test that iterates over a collection of expressions and asserts that none of them is changed via the normalizer?

Also, there is no test for `LetInEx`. Its normalization is quite non-trivial.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
gabrielamafra: I believe I've made a confusion because I didn't notice that `TlaBoolOper.equiv` and `TlaOper.eq` are two different things. However, I still don't think I understand this scenario of normalization. Why are we transforming `equiv` to `eq` or `neq`?  Are they different operators before this normalization and the same operator after?
```scala
    case equiv @ OperEx(TlaBoolOper.equiv, left, right) =>
      val nnfNonNegated = nnf(neg = false)
      // we do not negate the arguments but recurse to deal with the negations below the tree
      if (!neg) {
        OperEx(TlaOper.eq, nnfNonNegated(left), nnfNonNegated(right))(equiv.typeTag)
      } else {
        OperEx(TlaOper.ne, nnfNonNegated(left), nnfNonNegated(right))(equiv.typeTag)
      }
```
https://github.com/informalsystems/apalache/blob/normalizer-tests/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Normalizer.scala#L70

Kukovec: `<=>` is literally just the `Bool`-typed variant of `=`. Since we support `=` in its full generality, rewriting `<=>` to `=` shrinks the language subset that needs top be supported.

konnov: Right. We can see `<=>` as a convenient form of `=` on Booleans. Since normalization happens after type checking, it does not really matter that we use `=` instead of `<=>`. Also, I was out of context here. Negating `A <=> B` into `A /= B` is the most optimal solution. So let's use `=` and `/=`.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
```diff
@@ -63,4 +87,32 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
...
+
+  test("""~(~FALSE \in s) ~~> ~(TRUE \in s)""") {}
+
+  test("""~<>x ~~> []~x""") {}
+
+  test("""~[]x ~~> <>x""") {}
```
gabrielamafra, line 117: I believe there is a problem in the implementation of this normalization scenario. I wrote this description `~[]x ~~> <>x` description based on implementation, but it is not a valid equivalence. It should be `~[]x ~~> <>~x`. The implementation is currently:
```scala
    case ex @ OperEx(TlaTempOper.box, args @ _*) =>
      if (!neg) {
        OperEx(TlaTempOper.box, args map nnf(true): _*)(ex.typeTag)
      } else {
        OperEx(TlaTempOper.diamond, args map transform: _*)(ex.typeTag)
      }
```
and I believe it was meant to be:
```scala
    case ex @ OperEx(TlaTempOper.box, args @ _*) =>
      if (!neg) {
        OperEx(TlaTempOper.box, args map transform: _*)(ex.typeTag)
      } else {
        OperEx(TlaTempOper.diamond, args map nnf(true): _*)(ex.typeTag)
      }
```
https://github.com/informalsystems/apalache/blob/normalizer-tests/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Normalizer.scala#L160

@konnov @Kukovec @shonfeder Can you verify this, please?

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestNormalizer.scala
```diff
@@ -31,36 +31,268 @@ class TestNormalizer extends FunSuite with BeforeAndAfterEach {
...
+    val input = tla.letIn(tla.not(tla.not(tla.name("x") as BoolT1()) as BoolT1()) as BoolT1(), decl) as BoolT1()
+    val output = normalizer.apply(input)
+
+    val expectedDecl = tla.declOp("x", tla.bool(true)) as BoolT1()
+    val expected =
+      tla.letIn(tla.name("x") as BoolT1(), expectedDecl) as BoolT1()
+    assert(expected == output)
+  }
+
+  test("""~ (LET x = ~FALSE IN ~x ~~> LET x = ~FALSE IN x)""") {
```
gabrielamafra, line 249: I noticed that we are not normalizing declarations on negated LET IN expressions, but I think we should. Is this intended somehow?

konnov: Yes! Especially that the comment is saying that we are normalizing the declaration body.

konnov (approved):
Amazing! Please fix the issue with LET-IN not being normalized and merge.

PR #1170, opened 2021-12-13
closes #1151: ADR-014 on precise type inference for records and variants

### LINKED ISSUE #1151: [FEATURE] Write an ADR about an extension of Type System 1
As discussed in #401 and #789, we want to go ahead and make type checking for records precise. This would require changes at multiple levels. The first step is to write an ADR that outlines the potential solution and the steps required for implementing it.

konnov:
Here is the ADR-014. It is a longer document for a reason. In the end, I had to introduce new operators for variants. But having done that, I could finally produce a consistent set of inference rules.


docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,670 @@
...
+`FieldAccess` is ill-typed.
+
+Hence, we formulate the type inference rule for record access in our type
+system as follows:
+
+```
+r: Rec(RowCons("f", t_1), t_2)
+-------------------------------- [rec_acc]
+r.f: t_1
+```
```
Kukovec, line 329: You'd also need a rule
```
r: RowCons("f", t_1, t_2)
-------------------------------- [xyz]
r: t2
```

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,670 @@
...
+
+```
+Rec(RowCons("a", Int, z))
+```
+
+In this example, `RowCons("a", Int, z)` is a type term indicating that the type
+has the field called `a`, and its type is `Int`. On top of that, this row
+extends a parametric type `z`, which either contains a non-empty sequence of
+rows, or is an empty sequence, that is, `RowNil`. Importantly, `RecCons` is
+wrapped with the term `Rec`, so no additional fields can be added to the type.
```
Kukovec, line 278: This feels like a lot of info crammed into 1 paragraph. I'd suggest making separate, detailed entries for both operators. It's easy to miss the purpose of `Rec`, in that it acts as a guarantee of the absence of unspecified fields.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,670 @@
...
+We introduce a special type inference rule for `FilterByTag`:
+
+```
+set: Set(Variant(RowCons("<TAG>", r, z)))
+------------------------------------------- [variant_filter]
+FilterByTag(set, "<TAG>"): Set(r)
+```
+
+Importantly, `FilterByTag` returns a set of records that carry the tag `<TAG>`,
+so we can access record fields of every individual record in the set.
```
Kukovec, line 508: One important thing, which holds in practice (but not necessarily for an arbitrarily constructed type-term), but is not immediately obvious from this paragraph, is to emphasize that `r` is always of `Rec` shape, that is, it is always a complete record and not a generic row type. I'd make a special note on that point.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,670 @@
...
+ not present in the actual record type.
+
+ 1. Implement the
+ [suggestion by Shon Feder](https://github.com/informalsystems/apalache/discussions/789#discussioncomment-1592118).
+
+Although Option 2 looks nicer, we prefer keeping Option 1. The reason is that
+the current implementation introduces the minimal number of constraints
+by mashing all possible fields into a super-record. The alternative solution
+(option 2) would introduce additional constraints, when the spec requires us
+to extract an element from a set.
```
Kukovec, line 601: Notably, with the precise record types from Snowcat, it's trivial to trim records in post-processing.

Kukovec (approved):
It's good that we finally have a writeup on this.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,670 @@
...
+
+In this case, we would only allow to mix records that have exactly the same
+shape. As a result, when we use variants, we would have to add extra
+fields to all of them. This solution is not very different from the current
+implementation of record type checking, though it would allow us to quickly
+detect spelling errors.
+
+**Blocker.** This does not look like a real solution, as it would immediately
+render the existing examples invalid. Moreover, they would be no obvious way to
+repair these examples.
```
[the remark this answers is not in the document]
konnov: Oh yes, when we are talking about sets. I have no idea how you would do that with sequences, e.g., check `TestVariants.tla`.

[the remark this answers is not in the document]
Kukovec: I imagine a collection of sequences of pairs `(entry, syncedId)` where `syncedId` is the index the entry of a particular type would have in a unified sequence.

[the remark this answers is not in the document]
konnov: Well, sure, there is probably a way, but it definitely requires more book-keeping

docs/src/adr/014adr-precise-records.md
konnov: Why? You would not. The trick is that unification is allowed to swap pairs in rows.

docs/src/adr/014adr-precise-records.md
konnov: I am not sure, what you mean. The hardest case is `\E x \in S: ...`, where `S` is a set of variants. Since it is impossible to statically compute the exact type of `x`, you would have to deal with tuples in their generality.

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
Kukovec: Oh, for sure.

docs/src/adr/014adr-precise-records.md
Kukovec: >The trick is that unification is allowed to swap pairs in rows.

Where is this mentioned?

docs/src/adr/014adr-precise-records.md
Kukovec: What I mean is, if you have the variant operators guiding type-checking, you can use those types to reconstruct the smt model values of record type.

UNRELEASED.md
```diff
@@ -10,6 +10,9 @@
          * Some bug fix, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
+### Design
```
shonfeder, line 13: Not a blocker or anything, but I don’t think we need to include design documents in the release notes.

min my view, the audience for release notes are users who might experience changes or new features in the api. Anyone who cares about design and development rather than feature releases should get that info by subscribing to notifications about prs.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+z is a fresh type variable
+------------------------------------------------------------ [variant]
+Variant([ tag |-> "<TAG>", f_1 |-> e_1, ..., f_n |-> e_n ]):
+  Variant(
+    RowCons("<TAG>",
+      Rec(
+        RowCons("tag", Str,
+          RowCons(f_1, t_1,
+            ...
+              RowNil)...)
```
gabrielamafra, line 393: These dots are in the place of a 4th parameter for `RowCons` which don't exist.
```suggestion
              RowNil)
            )
```

gabrielamafra (approved):
Great explanation! The paper helped me a not because of the more familiar notation it uses. 

I have one issue with the notation in this document that I don't know if we should take into account considering its usage, but here it goes: I believe we are exchanging term types and row types in several places of this doc, and they are not the same thing. IIUC, `Rec` and `Variant` constructors take a row type and turn it into a term type. My suggestion would be to clarify this distinction, to use different symbols to represent them (i.e. t_1, t_2 for term types and r, z for row types) and fix some text that refers to the third argument of `RowCons` and `Variant` as a type variable (for example, in the `[variable]` rule).

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+feature. On the downside, we would have to:
+
+1. Increase the complexity of the type checker.
+
+1. Slightly update the rewriting rules in the model checker.
+
+1. Require the users to modify their specs to use the variant operators.
+
+We have tried to make type annotations as non-intrusive and compatible with
+TLC, as possible. After precisely specifying the requirements for the variant
```
rodrigo7491, line 37: The comma at the end of the first phrase is a bit odd.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+1. To construct a variant, whose only record is known at the time,
+   while the rest is captured with the type variable `z`.
+
+Going back to **Example 5.2.1**, the set constructor would produce a set of
+equalities:
+
+```
+  a = Set(b)
+  a = Set(d)
```
rodrigo7491, line 418: What is the reason for using `a` twice here?

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+**Example 5.2.1.** Now we have to figure out how to deal with TLA+ expressions
+like:
+
+```tla
+  { [ tag |-> "1a", bal |-> 3 ], [ tag |-> "2a", bal |-> 4, val |-> 0 ] }
+```
+
+Obviously, we cannot fit both of the records into a single plain record type,
+provided that we want to precisely track the fields that are present in a
+record.  So the type checker should report a type error, if we only implement
```
rodrigo7491, line 347:
```suggestion
record. So the type checker should report a type error, if we only implement
```

rodrigo7491 (commented):
For some reason the link to the paper is not working for me, maybe it is a temporary issue. I could the get general idea, but would probably need to do another pass and have a look at the paper to understand every detail. I believe that there is no expectation for an ADR to be self-contained, so this is fine. As a side note, the new ADR template is quite nice.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+```
+e_1: t_1, ..., e_n: t_n
+-------------------------------- [rec]
+[ f_1 |-> e_1, ..., f_n |-> e_n]:
+  Rec(
+    RowCons(f_1, t_1,
+      RowCons(f_2, t_2,
+        ...
+          RowNil)
+      )...)
```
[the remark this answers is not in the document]
gabrielamafra: Ooooh now I see it, ok, thanks! I'm not sure about the notation between `)...)` and `...))`. The first one seems to make it more clear that the dots are about parenthesis, although I didn't get it at first :stuck_out_tongue:

docs/src/adr/014adr-precise-records.md
gabrielamafra: It's the same discussion in https://github.com/informalsystems/apalache/pull/1170/files#r768904906, so let's centralize there.

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
rodrigo7491: For parenthesis opening and closing we have the following.
```
Rec(RowCons(f_1, t_1,RowCons(f_2, t_2,...,RowNil
    )      )                 ...        )      )
```
So the second to last closing is matching the `...`. I read it in terms of position, if you read `)...)` as an arbitrary number of `)` then I think it works as well, we just have to agree on the semantics :-).

Looking at this again, I think it would be good to have `RowCons(f_n, t_n,RowNil)` instead of simply `RowNil`.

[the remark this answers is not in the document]
gabrielamafra: Yes! That would be much better. So we'd be left with some like this:
```
e_1: t_1, ..., e_n: t_n
-------------------------------- [rec]
[ f_1 |-> e_1, ..., f_n |-> e_n]:
  Rec(
    RowCons(f_1, t_1,
      RowCons(f_2, t_2,
        ...
          RowCons(f_n, t_n, RowNil)
      ...)
    )
  )
```

shonfeder (commented):
Just submitting my first batch of notes. Still working through!

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+human reader, as it simply requires us to do the top-to-the-bottom type
+propagation, Snowcat constructs a set of equality constraints that are solved
+by unification. This approach requires that we capture the record access
+operator `.` as an equality over type variables and types.
+
+### 4.3. Support plain records and variants, including row typing
+
+This solution follows the approach outlined by [Leijen05][], except that we
+require field names to be unique, not scoped. We present our approach in the
+next section.
```
shonfeder, line 256: One point of concern, discussed briefly in our standup, but which I should flag here. In `Leijen05`, the author notes that 

> Row extension. Wand [35, 36] was the first to introduce row vari-
> ables, and described a type system that is very close to our proposal.
> The types of the basic operations are given as type rules, but these
> rules have basically the same effect as the type signatures in our
> system. However, this system also uses an ambiguous interpreta-
> tion of free extension as a mixture of update and extension. This
> gives rise to fundamental problems in the type system. Take for
> example the following expression:
>
> \r → if True then {x = 1 | r } else {x = 2 | {}}
>
> If extension overwrites previous fields, we can assign two different
> types to r , namely {x :: Int} or the empty record {}. This is
> why Wand’s original record system proved incomplete [35]. With
> scoped labels, the choice is unambiguous and r must have type {}.
> We show completeness (and soundness) of type inference formally
> in Section 7.

IIUC, the idea is that scoped labels are necessary to differentiate the extension and update operations, because otherwise any extension is necessarily an update in the event that the extended record already has a field with the same name. As a result, in the example given, it is impossible to give a unique type to `r`.  

For example, suppose `r = {x = 3}`. If extension entails update, then `{x = 1 | r} = {x = 1}`. This type checks. But now suppose `r = {}`, then, again `{x = 1 | r} = {x = 1}`. So we end up with ambiguous typing for `r`. This problem is avoided by scoped labels, because when `r = {x = 3}`, then `{x = 1 | r } = {(|x = 3::(|x=1::(||)|)|)}, but when `r = {}`, then we just get `{x = 1 | r} = {x = 1}`. This prevents us from ending up with an undecidable typing.

Given this warning in the paper, if my concern here seems sensible, then it is probably worth devoting a section to exploring how we can ensure that our decision to not adopt scoped labels won't leave us with problems. I would imagine there's more recent work that address this problem?

(Note that I tried to find the technical report to which the other refers readers to look for the proof of completeness, in the hopes that we might see how exactly the scoped labels are involved, but I can't find it anywhere online, except in the citation of paper.)

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+```tla
+Next ==
+  \E m \in msgs:
+    P
+```
+
+In the current encoding, `m` is a super-record that contains three fields:
+`tag`, `bal`, and `val`, even if some of these fields are not required by the
+actual type of `m`. In the alternative encoding, `m` is a tuple `tup`,
```
shonfeder, line 624: How is possible for `m` to include all three fields in all cases? E.g., suppose 

```
msgs = {[tag: "1a", bal: 3]}
```

Where do you get your value to associate with the `val` field?

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+```
+tup = IF m.tag = "1a"
+      THEN << { [ tag |-> "1a", bal |-> b ] }, {} >>
+      ELSE << {}, [ tag |-> "2a", bal |-> b, val |-> v ] }, {} >>
+```
+
+Since it is impossible to statically compute the actual type of `m`, the
+rewriting rules would have to replicate the structure of both elements of the
+tuple `tup`. This would lead to a blow-up in the number of constraints.
```
shonfeder, line 636: Could you show the equivalent here in your proposed encoding? I am having trouble understanding what that's meant to look like.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+### 5.4. Additional requirements
+
+Given the decisions in Section 5.3, we additionally require that all records in
+a variant type have compatible field types. In more detail, if a variant type
+contains two record types `[ tag: "A", value: a, ... ]` and `[ tag: "B", value:
+b, ...]`, then the types `a` and `b` must be unifiable.  In practice, this
+often implies that `a` and `b` are simply the same type.
```
shonfeder, line 644: This seems really strange to me: In this proposal, we will introduce fancy, row types, which are designed to support pretty remarkable flexibility and generality. Then we limit them to only being applicable to records, then we further limit them to only applying to records which have fields of the same types. It makes me feel like there must be a simpler way... If our options are so constrained, it would seem like we don't nee such generality, but if we do opt for such a general solution, then it would seem that we shouldn't need all this additional book keeping to deal with one particular case.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+will be distributed with Apalache. As is custom in the TLA+ community, the
+users should be also able to copy `Variants.tla` next to their specification.
+
+## 6. Consequences
+
+This will be a relatively big change in the types and the type checker.
+Additionally, it would render many existing specifications ill-typed, as the
+proposed solution imposes a stricter typing discipline. On the positive, the
+proposed solution is backwards-compatible with TLC, as we are proposing the
+default untyped implementation for the operators.
```
shonfeder, line 659: Note that the main point of this section is to provide a space for us to evaluate the consequences of the decision *after* we have actually seen the effects. :)

shonfeder:
I agree with Jure that it is great to see this written up and starting to take more concrete and full wrought form! :tada: 

My principle concerns after reading through are:

1. I would expect that the inclusion of row types would be motivated by the need for extensible, polymorphic records and/or variants. But I don't actually see either of those claimed or justified here. So I don't yet understand why we can't solve the stated problems with normal product and sum types (generalized to records and variants in the usual way). Note that I can certainly imagine specifications might have need for such generality: I just think we should clearly articulate and justify that need.
2.  I think the issue around scoped labels and completeness of the type system should be derisked. This might as simple as showing how we can give an unambiguous typing to `r` in `\r → if True then {x = 1 | r } else {x = 2 | {}}`, or perhaps reference to a paper that explains the limits and shows why our modified system won't get stuck.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,670 @@
...
+RowAccess(m) ==
+    m.a > 0         \* should flag a type error
+```
+
+**Blocker.** It is not clear to me, what type we would assign to the record
+access operator. Although the type of `m` in `FieldAccess` is obvious to a
+human reader, as it simply requires us to do the top-to-the-bottom type
+propagation, Snowcat constructs a set of equality constraints that are solved
+by unification. This approach requires that we capture the record access
+operator `.` as an equality over type variables and types.
```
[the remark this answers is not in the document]
konnov: Sure, Pierce is giving an inference rule for a known record type. This is a calculus, not an algorithm for type inference. What I mean is that I don't know how to construct equality constraints in the type checker without either using row types, or requiring the user to explicitly specify the record type.

docs/src/adr/014adr-precise-records.md
konnov: This will make things more complicated on the SMT side. In this ADR, variants are just a type wrapper around records, so we do not any additional data structures to carry variants. If we allow arbitrary value, we would have to introduce another data structure for variants, which would be very similar to records.

docs/src/adr/014adr-precise-records.md
konnov: We are constructing a system of equalities that are solved by unification. Are you talking about your branch on type inference with SMT?

docs/src/adr/014adr-precise-records.md
konnov: It's just the form of the constructed constraints. The left-hand side is alway a type variable. We could write it as one equality.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+Moreover, the following example would require a type annotation:
+
+```tla
+RowAccess(m) ==
+    m.a > 0         \* should flag a type error
+```
+
+In the above example, the type checker would not be able to infer the type of
+`m` and would require an explicit type annotation:
```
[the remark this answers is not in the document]
konnov: 1. We require type annotations for constants and variables. In some cases, the users have to write type annotations, when the type checker cannot decide between a function and a sequence. In most of the cases, we actually do not have to annotate user-defined operators. I see row types as a way to avoid unnecessary type annotations.

2. We do not have type declarations as in OCaml. The annotation `@typeAlias` is just an alias that is getting replaced with the type before the type checker starts doing anything. I am not even sure how OCaml does it.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+### 4.3. Support plain records and variants, including row typing
+
+This solution follows the approach outlined by [Leijen05][], except that we
+require field names to be unique, not scoped. We present our approach in the
+next section.
+
+## 5. Solution
+
+In the following, we present row types as type terms. We discuss the
```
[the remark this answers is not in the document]
konnov: I did not want to complicate the matters with kinds. We would need them, but that is more of an implementation issue.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+There are several solutions to the issue of precise record typing.
+
+### 4.1. Support only plain records
+
+In this case, we would only allow to mix records that have exactly the same
+shape. As a result, when we use variants, we would have to add extra
+fields to all of them. This solution is not very different from the current
+implementation of record type checking, though it would allow us to quickly
+detect spelling errors.
```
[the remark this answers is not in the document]
konnov: The moment you use `G(dinner)`, you should get a type error. If `dinner` appears nowhere, then `G` is just a polymorphic operator.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+In the following, we present row types as type terms. We discuss the
+user-facing syntax of the type system later in the text.
+
+### 5.1. Plain records
+
+By using [Row types][], we should be able to infer the type of `m` in
+`RowAccess`:
+
+```
+Rec(RowCons("a", Int, z))
```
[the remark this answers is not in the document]
konnov: This is the syntax of type terms that we are going to use. The syntax in the paper is ambiguous, as they wanted to make paper easier to read. Sure, we have to find a compact syntax for the users. The syntax that we report to the users may even deviate from the type grammar, e.g., we could even write `[ a: Int | ... ]`.

docs/src/adr/014adr-precise-records.md
konnov: But we do not have record extension in TLA+. I don't even think we would allow `EXCEPT` to change types of the fields. So yes, probably the type system of Wand [35, 36] is incomplete. But I do not think this applies to us. The problem with the recent research is that people are pushing more and more features into papers, because they cannot publish otherwise. We actually need a relatively simple type system and relevant exposition can be only found in the old papers.

docs/src/adr/014adr-precise-records.md
konnov: If `msgs` has the type `[tag: Str, bal: Int, val]`, which is possible in the current implementation, then the model checker would just introduce a default value for `val`. It's actually not so different from unions on C, where you would have to reserve memory for all fields, even though some of them will never be used.

docs/src/adr/014adr-precise-records.md
konnov: The arena encoding does not change but stays the same. A record will contain all possible fields. In this case, it will be three fields: `tag`, `bal`, and `val`. Since it is a single record, not a tuple of records, the rewriting rules would have to pick values only from that record instead of introducing copies of all possible records in the tuple.

docs/src/adr/014adr-precise-records.md
konnov: I don't see row type being fancy. It is just a natural approach when you have equality constraints instead of arbitrary subtyping constraints.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+```
+
+In `FieldAccess`, we use row types to construct a series of type equations
+(over free type variables `k`, ..., `q`):
+
+```
+// from LET m == [ a |-> 2, b |-> "B" ] IN
+m_type = Rec("a", Int, RowCons("b", Str, RowNil))
+// from m.a > 1
+m_type = Rec(RowCons("a", k, l))
```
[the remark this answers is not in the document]
konnov: Who is "we"? The type checker constructs equalities and then solves them. The rule `.` knows nothing about the actual type of the left-hand side, so it introduces a type variable, which will be matched with the actual type of `m` after some equalities have been solved.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+-------------------------------- [rec_acc]
+r.f: t_1
+```
+
+Note that the above rule can be rewritten into a series of equalities over
+types variables and type terms, which is how this would be implemented in the
+type checker.
+
+If we only had to deal with plain records, that would be a complete solution.
+Unfortunately, variants introduce additional complexity.
```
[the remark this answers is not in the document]
konnov: I mean that if we had only plain records, we would have to use rows only to talk about shapes of the records. Once we have variants, we have to use rows at two levels: for the shapes of the records and for the shapes of the variants.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+```tla
+  { [ tag |-> "1a", bal |-> 3 ], [ tag |-> "2a", bal |-> 4, val |-> 0 ] }
+```
+
+Obviously, we cannot fit both of the records into a single plain record type,
+provided that we want to precisely track the fields that are present in a
+record.  So the type checker should report a type error, if we only implement
+type inference for the case explained in Section 5.1. To support this important
+pattern, we introduce variants. They are similar to [unions in
```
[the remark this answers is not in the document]
konnov: What I meant by plain records were records that cannot carry different sets of fields, independently of whether they are row-typed or not. Yes, that holds for both cases.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+TypeScript](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types).
```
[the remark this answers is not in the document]
konnov: Right. Perhaps, the variants in the current form are further away from the TypeScripts unions.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+**Variant constructor.** We introduce a special TLA+ operator `Variant`
+that extracts the tag from a record and wraps the record into a variant.
+We need this operator to distinguish between plain records and records that
+belong to a variant. The operator `Variant` is defined in TLA+ as follows:
+
+```tla
+  Variant(r) ==
+    \* fallback untyped implementation
+    r
+```
```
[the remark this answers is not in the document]
konnov: Then we would have to introduce a yet another data structure. Records are already carrying all necessary data.

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
konnov: I guess, a record could be also polymorphic, true.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+                       RowCons("bal", Int,
+                         RowCons("val", Int, RowNil)))),
+                 t
+     ))))
+```
+
+Note that we still do not know the precise shape of the variant, as it
+closes with the type variable `t`. This is actually what we expect, as the set
+may be combined with records of other shapes. Normally, the final shape of a
+variant propagates via state variables of the TLA+ specification.
```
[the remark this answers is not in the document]
konnov: We would have to specify the type of a variant in `VARIABLES`, or in an operator annotation. If it's not specified anywhere, the variant would stay open, which might be ok for some operators.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+------------------------------------------------ [variant_match_only]
+MatchOnly(variant, ThenOper):
+  (Variant(RowCons("<TAG>", r, RowNil)),
+   r => z
+  ) => z
+```
+
+It looks like the solution with `MatchTag` and `MatchOnly` introduce a lot of
+boilerplate. However, this is probably the best solution that we can have,
+unless we can extend the grammar of TLA+.
```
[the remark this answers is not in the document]
konnov: That would be definitely great to have. However, the TLA+ syntax does not allow for polyadic user-defined operators and you cannot pass operators in a sequence, as operators are not values. Yep.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+
+### 5.3. Changes in the model checker
+
+Having precise types for variants, we have two options:
+
+ 1. Keep the current encoding, that is, a variant is encoded as a super-record
+ that contains all possible fields of the member records. The type checker
+ will guarantee that we do not access the fields of the super-record that are
+ not present in the actual record type.
```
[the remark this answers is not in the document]
konnov: Sure. The current encoding essentially implements C unions. This ADR does not suggest a new encoding in the model checker, it only makes record access safe at the type-level.

docs/src/adr/014adr-precise-records.md
Kukovec: No, I'm talking about cases where the current type checker infers a record with spurious fields, and then cannot catch illegal field access. In the post-ADR world, you could use variant information to avoid this.

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
shonfeder: Oh man. Then I am really confused. I thought this was exactly the thing that the row types were for. In the cited paper, they give this example:

> The selection operation (r .l ) selects the value of a label
> l from a record r . For example, we can define a function distance
> that calculates the distance of a point to the origin:
> distance p = sqrt ((p.x ∗ p.x ) + (p.y ∗ p.y))
> In contrast to many programming languages, the distance function
> works for any record that contains an x and y field of a suitable
> numeric type.

My focus here is on this point:

> the distance function works for any record that contains an x and y field of a suitable numeric type.

So I was expecting that `G` and `F` would the same, namely that `F` would work on any record that contains a `food` of a suitable `Str` type, and that `G`, since it's just doing a record extension, would work to add the field `fool` to any record that (simply because of the type annotation) has a `foo : Str`. But I think you've noted elsewhere that you don't plan to support record extension? Yeah, I am quite confused.

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
shonfeder: Yes, I confess I am pro making things easier to read :D

docs/src/adr/014adr-precise-records.md
shonfeder: Could you show an example of how you intend to represent a value which has a variant with two different alternatives?

E.g., what would be the equivalent of

```
ev =〈key = ’\t’〉::〈key :: Char | mouse :: Point 〉
```

I think seeing that written out here would help me understand what you envision. I expect adding such an example/comparison might useful for the clarity of the ADR too?

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
shonfeder: "we" is meaning to take the perspective of the type the type checker: I assume the type checker is keeping a context that tracks the types discovered or inferred so far for variables. Isn't `m` in the context by this line? Haven't we already inferred it's type from line 299? 299 Says `m_type = Rec("a", Int, RowCons("b", Str, RowNil))`, why can't the checker use that constraints here? Is this a different `m_type`? Is it shadowing the one on line 299?

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
shonfeder: Ah I see. It seems I am confused by the phrase `plain records` here and elsewhere. So perhaps change this to?

```suggestion
If we only had to deal with records, that would be a complete solution.
Unfortunately, variants introduce additional complexity.
```

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
shonfeder: Ah, yeah, I guess we could just use records of the shape `[tag: Str, value: a]` for any value `a`.

docs/src/adr/014adr-precise-records.md
shonfeder: Right, I'd think to use a record to represent the variants, but allow variants themselves to hold any type. But this is not essential.

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
shonfeder: Ah yeah, my bad for confusing operator LAMBDAs with functions.

docs/src/adr/014adr-precise-records.md
shonfeder: Ah ok. Right if we have no record extension, then that would address this concern.

docs/src/adr/014adr-precise-records.md
shonfeder: I guess I mean "fancier than non-polymorphic variants/records". 

The main thing I wish I understood around this ADR is why non-polymorphic variants and records wouldn't work for all of our needs. I apologize if I'm just being dense, but since you say records aren't extensible, and from our discussion [above](https://github.com/informalsystems/apalache/pull/1170#discussion_r771185191) it sounds like we won't support operators that are polymorphic over record fields, and now the variants supported are even more constrained than the variants supported by standard, *non-polymorphic* variant types, I can't figure out what aspect of the row typing is actually going to be used or needed.

docs/src/adr/014adr-precise-records.md
shonfeder: Quick note here that @gabrielamafra helped me understand this point bout TLA+ not supporting record extension. I had assumed that `EXCEPT` could be used to add new fields to records, but apparently, in both TLC and Apalache, using EXCEPT with a record field which is not already in the record leads to the update just be ignored? (Gabriela noted that this also issues a warning in TLC).

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
gabrielamafra: I talked with Shon on slack about `EXCEPT` behavior because it was not clear to us whether it could be used for extending records. Just to register here: it's not, at least not on TLC. Trying to extend a record gives a warning:
```
Warning: The EXCEPT was applied to non-existing fields of the value
```
and the operation is ignored.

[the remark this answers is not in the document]
Kukovec: `EXCEPT` won't change the function domain, see https://www.microsoft.com/en-us/research/uploads/prod/2018/05/book-02-08-08.pdf (p302, 16.7)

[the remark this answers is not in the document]
konnov: Right. In TLA+, records are functions, and functions are not allowed to extend their domains. So, yes, records are not allowed to extend their domains.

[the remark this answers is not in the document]
konnov: In my understanding, row types allow to infer types of records whose type is not fully known inside an operator (or a function in a functional language). So we would have to produce a polymorphic record, whose type can be instantiated with a record that has more fields. Whether the language itself supports record extension or not, is less important, though record extension probably makes things more complicated.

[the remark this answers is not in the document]
shonfeder: Ah yeah, I see. 

Having that narrower scope helps me understand the stakes and aim here better. I wonder if you that framing and narrow focus could be added to the ADR? I'm not sure if it's just me, but I think many of my confusions here would have been avoided if the setup noted these significant deviations and limits (compared against the paper) up front. I'll suggest that change below.

Thanks, all (especially @gabrielamafra), for helping me walk through this bit!

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+human reader, as it simply requires us to do the top-to-the-bottom type
+propagation, Snowcat constructs a set of equality constraints that are solved
+by unification. This approach requires that we capture the record access
+operator `.` as an equality over type variables and types.
+
+### 4.3. Support plain records and variants, including row typing
+
+This solution follows the approach outlined by [Leijen05][], except that we
+require field names to be unique, not scoped. We present our approach in the
+next section.
```
shonfeder, line 256: As per my comment above -- that clearly stating the many aspects of the paper are not relevant here would help with clarity -- what about extending this with something like the following:

```suggestion
This solution is inspired by the approach outlined by [Leijen05][], but is much more limited. The following features discussed in [Leijen05][] are neither needed needed nor supported:

- extension for records
- extension for variants
- record restriction
- scoped labels

Our use is limited to a few special cases that give support for subtyping and incremental inference of anonymous record types, in which we cannot know the full set of fields up front. 
```

docs/src/adr/014adr-precise-records.md
konnov: You would just introduce a variant over tagged records, instead of having arbitrary types.

docs/src/adr/014adr-precise-records.md
konnov: Yes, exactly. So the point is that the arena encoding is still using spurious fields, but by knowing types we can guarantee that spurious fields are never accessed.

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
konnov: fixed that

docs/src/adr/014adr-precise-records.md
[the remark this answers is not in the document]
konnov: I see. I think you are assuming that the constraint solver is solving equalities as soon as they are added. It does not have to, and actually it does not do that in our implementation.

docs/src/adr/014adr-precise-records.md
```diff
@@ -0,0 +1,668 @@
...
+`FieldAccess` is ill-typed.
+
+Hence, we formulate the type inference rule for record access in our type
+system as follows:
+
+```
+r: Rec(RowCons("f", t_1, t_2))
+-------------------------------- [rec_acc]
+r.f: t_1
+```
```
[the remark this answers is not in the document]
konnov: Yes, you could have a rule for `.` that assumes a record type without using row types. The point is that you would not be able to infer a type of unannotated `RowAccess`.

docs/src/adr/014adr-precise-records.md
gabrielamafra: I'll just add my understanding of why row types are necessary here and you guys can tell me if it makes sense.

PS: I'm mixing the paper notation with the ADR notation with the intention to clarify the relation between the two things
For what I understand about the ADR and Igor's comments, the scenario where we actually need row types is this one:

```
a = [ tag |-> "t1", x |-> 2 ] :: { tag :: String, x :: Int }
b = [ tag |-> "t2", x |-> 5, y |-> "hello" ]  :: { tag :: String, x :: Int, y :: String }

msgs = { a, b } :: Set(Variant(RowCons( "t1", Rec({ x :: Int })), RowCons("t2", Rec{ x :: Int, y :: String })))
```

And this is actually a scenario I believe we need subtyping. Suppose a general type rule for Sets

```
x_0 :: t, ... x_n :: t
------------------------------- [set]
{ x_0, ..., x_n } :: Set(t)
```

for `msgs` to be correctly typed, both `{ tag :: String, x :: Int }` and `{ tag :: String, x :: Int, y :: String }` should be subtypes of `Set(Variant(RowCons( "t1", Rec({ x :: Int })), RowCons("t2", Rec{ x :: Int, y :: String })))`

The subtyping relation could also be done with a sum type. The difference between a sum type and a variant is that a variant carries a row type variable and therefore can be extended. The extension I'm referring to here is not record extension as defined in the paper, but actually the type extension that has to happen when a record of a different row type is added to a set. Therefore, in my understading, we need row types because:
1. A set can't have values of different types, but through a variant, it can have values of different row types.
2. Unification rules for sum types would not be enough to accumulate information about the record type throughout extensions such as adding a new record to a set. We need more "dynamic" unification rules such as the ones provided in the row types paper.

shonfeder: Thanks, for your analysis here @gabrielamafra: I find it quite helpful and I think it helps me understand the motivating need much better.

shonfeder:
If everyone else is happy with the proposal as is, I'm fine with that, and I'll mark my approval now.

There are some parts of the proposed implementation that still don't totally make sense to me (mainly the way variants will be encoded). If it'd be useful, I'm happy to try to work through my understanding, but don't want to block on that.

shonfeder (approved):

docs/src/adr/014adr-precise-records.md
konnov: Once this ADR is implemented, we have to extend the memo on type inference with more details. It's hard to add more details without writing a full paper.

konnov:
I will merge this ADR now, to go ahead with small changes related to it. We can revisit this ADR, when we have a better understanding of the solution. A better solution would be to aggregate all of our knowledge into the memo on the type checker.

PR #1179, opened 2021-12-16
Remove snakeyaml dependency

shonfeder:
We don't need it anymore now that we're using PureConfig for the
configuration management.

-------


konnov (approved):
LGTM

PR #1181, opened 2021-12-16
Add IR operators for compound set update

### LINKED ISSUE #1171: [FEATURE] Streamline set update operations
The evaluation results of the current state of the arrays encoding are very poor, as shown by following reports.
- https://github.com/informalsystems/apalache-tests/pull/108/files
- https://github.com/informalsystems/apalache-tests/pull/107/files
- https://github.com/informalsystems/apalache-tests/pull/106/files
- https://github.com/informalsystems/apalache-tests/pull/105/files

One possible cause is the definition of intermediary arrays during array definitions/updates. These are quite numerous, with a new array being created for the update of every single element, and are probably slowing down the solver. To improve performance we should collapse sequences of SMT `store` operators into a single clause.

rodrigo7491:

This PR adds two new IR operators, `storeInSetOneStep` and `storeInSetLastStep`, which are variants of `storeInSet`. The existing store operator updates one entry of a SMT array. The new operators, which need to be used in conjunction, allow for the update of an arbitrary number of entries to be encoded into a single clause, thus avoiding the creation of unnecessary intermediary arrays. These new operators are intended to be used only for sequences of updates, thus the `storeInSet` operator remains useful for single updates. Closes #1171.

The arrays encoding will change from
```
(assert (= ite cond_1 (= (set_1 (store set_0 elem_1 true))) (= (set_1 set_0))))
(assert (= ite cond_2 (= (set_2 (store set_1 elem_2 true))) (= (set_2 set_1))))
(assert (= ite cond_3 (= (set_3 (store set_2 elem_3 true))) (= (set_3 set_2))))
```
to
```
(assert (= (set_1 (store (store (store set_0 elem_1 cond_1) elem_2 cond_2) elem_3 cond_3))))
```
\
To avoid having different compound set update operations depending on the encoding, the oopsla19 encoding will also use the new IR operators. Adding elements to a set will change from
```
(assert (= ite cond_1 (elem1_in_set) (not elem1_in_set)))
(assert (= ite cond_2 (elem2_in_set) (not elem2_in_set)))
(assert (= ite cond_3 (elem3_in_set) (not elem3_in_set)))
```
to
```
(assert (and (= cond_1 elem1_in_set) (= cond_2 elem2_in_set) (= cond_3 elem3_in_set)))
```
In `(= cond elem_in_set)`, if `cond` evaluates to `true` the expression is simplified to `elem_in_set`.

### DIFF
--- a/docs/src/adr/011adr-smt-arrays.md
+++ b/docs/src/adr/011adr-smt-arrays.md
@@ -157,8 +157,7 @@ The following changes will be made to implement the new encoding of sets:
   - The `storeInSetOneStep` and `storeInSetLastStep` IR operators, encoding a sequence of `store`
     operations into a single clause, are used to more efficiently encode array updates dealing
     with many elements. Their benefit stand from avoiding the declaration of many intermediary
-    arrays. They represent a shift in the mechanical way in which set updates are done in the
-    OOPSLA'19 encoding, and thus should only be used in the new encoding.
+    arrays.
   - Cases for `FinSetT` and `PowSetT` will be added to `getOrMkCellSort`, as these types are no
     longer represented by uninterpreted constants.
   - `cellCache` will be changed to contain a list of cells, in order to handle the effects of
--- a/docs/src/adr/011adr-smt-arrays.md
+++ b/docs/src/adr/011adr-smt-arrays.md
@@ -140,7 +140,7 @@ The following changes will be made to implement the new encoding of sets:
   `PreproSolverContext`, to represent the array operations.
   - The `selectInSet` IR operator represents the SMT `select`.
   - The `storeInSet` IR operator represents the SMT `store`.
-  - The `storeInSetOneStep` and `storeInSetLastStep` IR operators represent a compound SMT `store`.
+  - The `chain` and `assignChain` IR operators represent compound SMT operations.
   - The `unchangedSet` IR operator represents an equality between the current and new SSA array
     representations. This is required to constraint the array representation as it evolves. It is
     important to note that this operator assumes that all arrays are initially empty, so an element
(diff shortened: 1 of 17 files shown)


konnov (commented):
I see what you are trying to do here. I think we could introduce a more general operator instead of `StoreInSetLastStep` and `StoreInSetOneStep`, which we could use to chain arithmetic operators too! Essentially, you have a list of triples `(op, elem, cond)`, which you are right folding by applying `op` to the intermediate result and `elem` under the condition `cond`.

konnov (commented):
I have to look at it again. The remaining question is why we have to pass a chain of `apalacheChain` instead of passing a list to one instance of `apalacheChain`?

rodrigo7491:
@konnov, it should be possible to rewrite apalacheChain to be non-recursive, but it makes its implementation in `Z3SolverContext` rather clumsy. This is specially the case if we want to chain different operations, e.g. addition and subtraction. The way things are now we can rely on `Z3SolverContext.toExpr` to cleanly handle chains with heterogeneous operations. Is there an efficiency reason to want to use sequences in Scala?

konnov:
> @konnov, it should be possible to rewrite apalacheChain to be non-recursive, but it makes its implementation in `Z3SolverContext` rather clumsy. This is specially the case if we want to chain different operations, e.g. addition and subtraction. The way things are now we can rely on `Z3SolverContext.toExpr` to cleanly handle chains with heterogeneous operations. Is there an efficiency reason to want to use sequences in Scala?

No, there is no efficiency considerations. If you think that we should keep it the way it is, let's keep it.

konnov (approved):
As we have discussed, this is a correct solution to the problem. I have checked the updates to the rules and no bugs seem to be introduced. @rodrigo7491 if you are blocked on your work with this PR, please merge it and go ahead. We can refactor the code later, when we deal with arithmetic expressions. It is easy to see which parts of the code are affected by the change.

PR #1184, opened 2021-12-17
Add documentation of our project management usage

### LINKED ISSUE #1180: Draft guidelines on project coordination
This is just to track ongoing work by @gabrielamafra and myself on iterating on and evolving our project planning, coordination, and communication processes.

- [x] Create milestones to group ongoing work on features (see https://github.com/informalsystems/apalache/milestone/47)
- [x] Create new project board that uses milestones and views to provide more focused insights into our plans and WIP (see https://github.com/orgs/informalsystems/projects/30/views/1)
- [x] Draft guidelines on how to interpret and utilize the various github features for project management, and how to prioritize work

shonfeder:
Closes #1180

This documentation was drafted and conceived together with @gabrielamafra and
informed by our team discussions. 

It aims to be a living document recording our shared understanding around the
meaning and use of the various project management tools we've incorporated into
our process so far.


shonfeder:
Open Questions:

- [ ] Should onboarding new members be a "milestone"?
    - On the one hand, it's a closely related set of work.
    - On the other hand, members might not like to have that vsibible, and it's not a milestone on a particular product (perhaps using a private repo for this?)
    - On the one hand again, why shouldn't we treat this kind of work grouping like any other? If we are using issues to track our work but not putting onboarding stuff into the issues, it ends up as untracked work.
- [ ] How can we visualize past, current and future milestones?
    - We currently don't create milestones for things we have not started
    - GitHub doesn't provide a board visualization of milestones
    - It's probably better to think about prioritization in terms of milestones

COORDINATION.md
```diff
@@ -0,0 +1,92 @@
...
+| label         | a more or less vague category/topic/quality             | attributes of the work |
+| project board | concurrent human processes with a common dynamic        | how work is done       |
+| discussion    | free ranging and open discussion                        | whatever               |
+
+(For more on projects vs. milestones, see
+https://stackoverflow.com/a/47542346/1187277)
+
+### Project board
+
+#### Overall workflow reflection
```
shonfeder, line 33: Just a shoutout ,since we did the initial draft on Hedgedoc so the commit history wasn't tracked: The part of the document from this point on is all @gabrielamafra's work. 

I find it super helpful to have this kind of big picture, philosophical grounding when considering how we manage the project, and this section on the temporal structure of the project is the part of the document I find most helpful, personally.

COORDINATION.md
```diff
@@ -0,0 +1,103 @@
...
+[issues]: https://github.com/informalsystems/apalache/issues
+[prs]: https://github.com/informalsystems/apalache/pulls
+[milestones]: https://github.com/informalsystems/apalache/milestones
+[labels]: https://github.com/informalsystems/apalache/issues/labels
+[project]: https://github.com/orgs/informalsystems/projects/30
+[discussion]: https://github.com/informalsystems/apalache/discussions
+
+### Project board
+
+Our project board is the main resource to help with our day-to-day coordination. We use a Kanban-styled continuous workflow board with a high level view of whats being done and some queues for what is up next.
```
gabrielamafra, line 40: We should still discuss if this type of continuous workflow board suits the entire team. Jure expressed that he likes to see some reduction in the amount of work to do (like a burndown chart), but if we are continuously moving cards to done and continuously adding tasks, we lose any relation between our capacity and the amount of work to do. This type of visualization requires that we compromise with a certain amount of work inside a time-box where we won't change what we are compromised to do (i.e. add new tasks), like in a sprint in scrum-based approches.

Personally, I believe that the main problems in compromising with an amount of work are
1. We lose flexibility
2. We have to work with some sort of time-box (otherwise we can just change what we are compromised to at any given time and it becomes continuous again). Time-boxes can be unfounded sources of pressure and disappointment if they are not treated 100% maturely. It's usually very hard for people to distinguish a time-box like this, that has a specific purpose, from a imposed deadline.

shonfeder: I think we can get signal on some of this via existing features. For example, we can use milestones with due dates to get time-limited progression on a certain set of issues. If we want a view of our throughput, we can filter all issues/prs closed within a certain date range (I think!).

iiuc your point here, I agree that we shouldn't expect that the kanban boards to give us this kind of perspective, and that the costs trying to work within time boxes are pretty high.

shonfeder:
One additional thought: once we've worked out mutual agreement on these points and perhaps tested them out internally for a few weeks, I think we should consider "upstreaming it" into our handbook, as it would likely be valuable to come to agreement on these kinds of practices throughout the company, and we're likely to receive some valuable additional perspectives.

COORDINATION.md
```diff
@@ -0,0 +1,103 @@
...
+| [discussion][discussion] | free ranging and open discussion                        | whatever               |
+
+(For more on projects vs. milestones, see
+https://stackoverflow.com/a/47542346/1187277)
+
+[issues]: https://github.com/informalsystems/apalache/issues
+[prs]: https://github.com/informalsystems/apalache/pulls
+[milestones]: https://github.com/informalsystems/apalache/milestones
+[labels]: https://github.com/informalsystems/apalache/issues/labels
+[project]: https://github.com/orgs/informalsystems/projects/30
```
rodrigo7491, line 35: I assume this document is going to be public, and since the project board is only accessible with the right permission, it might be good to point this out somehow.

COORDINATION.md
rodrigo7491: I agree with what was said, but would like to point two things out.

1. iirc the point of having different buckets on the project board was to see how the team's time was spent in different types of tasks, i.e., we spent a lot of effort recently on refactoring, so let's put more focus on usability next week. This information is lost if we keep refiling the buckets as Gabriela said. I always saw the buckets as a selection of issues without time constraints i.e., once we get all this done we discuss what to do next, but maybe it was just me.
1. I don't think we should use milestones with two different meanings, or at least not without a very clear agreement on how to do so.

COORDINATION.md
```diff
@@ -0,0 +1,103 @@
...
+help us define which tasks from the backlog should go in this queue. The process
+of deciding which tasks should be on the "Next" column is usually called
+"replenishment".
+
+The Next column needs a size limit since "if everything is a priority, then
+nothing is a priority".
+
+###### How do we prioritize work?
+
+We are currently considering 3 different prioritization schemes:
```
rodrigo7491, line 89: From my understanding, scheme 1 and 3 are similar, but in 1 we vote indirectly, by agreeing on the impact and effort of different issues. Is this right, or is there some aspect I am missing?

rodrigo7491 (commented):
Glad we are working towards a clearer coordination strategy, thanks @gabrielamafra and @shonfeder! My main concern is different team members using the same GitHub feature for different goals, specially the project board. Eager to hear what the others think about it.

COORDINATION.md
gabrielamafra: About 1., maybe we can just use labels and look at the recent labels on the "Done" column? If we are reflecting on the past (i.e. last week), the "Done" column should be our best source of information. We could event have buckets for the Done side such as "Done Usability", "Done Refactoring" if we think that's better than using tags.

COORDINATION.md
gabrielamafra: I wouldn't say they are similar, but 3 can be used in combination with any prioritization strategy since every voting person can use their own strategy. 1 and 2 are techniques to prioritizing something and 3 is a suggestion about how to do prioritization in a group. We can opt to use any combination of the three. Maybe this should be clearer in the text.

COORDINATION.md
Kukovec: Here's the way I think the project board should be used:
To start, we have a [plan][https://github.com/informalsystems/strategy/blob/igor/yearly2021-2022/projects/apalache/yearly2021-2022/plan2022.md] for 2022.
The plan is not absolute, and obviously can't predict spontaneous issues or ideas that come up, but it does act as a partial order on categories of tasks.
In that context, at least to me, the project board acts as the refinement of tasks for a fixed period of time (e.g. Q4 or one month of Q4).
Therefore, until the end of the fixed time interval, the number of tasks should monotonically decrease. This does not mean that the fixed time window is a deadline, since a task can be kept on the board after the window (deciding that is what meetings are for).
The purpose of monthly or quarterly meetings is then to decide what tasks to put on, or remove from, the board.
The purpose of weekly meetings is to evaluate whether some buckets on the boards are being emptied much faster than others, with the ultimate goal being that "less interesting" (according to subjective perspective) tasks (e.g. documentation, testing, MBT) don't get neglected.

I don't think milestones/labels are particularly relevant in this scenario (arguably, milestones are subsets of labels and buckets contain subsets of milestones).

Adressing concerns:
>  We lose flexibility

Yes, in the sense that you lose the ability to arbitrarily pick your next issue from outside the board. Not really, in the sense that the board is not immutable, and if there is a sudden need to start working on something not planned, there is a meeting at least once every week, during which we can add things to the board. I'll also mention that critical bugfixes are understood to be outside the board and can (should?) be fixed asap.

> We have to work with some sort of time-box (otherwise we can just change what we are compromised to at any given time and it becomes continuous again). Time-boxes can be unfounded sources of pressure and disappointment if they are not treated 100% maturely. It's usually very hard for people to distinguish a time-box like this, that has a specific purpose, from a imposed deadline.

Yes, that's what quarterly planning has always been. Time-boxes are only pressure if there exists some entity imposing consequences on non-adherence. Currently, the only source of external feature constraints (that aren't just requests by other teams at Informal) is the funding agency that is paying for the new Apalache project (which is to say, this is mostly a non-issue IMO, we don't have traditional deadlines for concrete deliverables).

shonfeder: @Kukovec In this proposal, this feature is handled by milestones:

- > until the end of the fixed time interval, the number of tasks should monotonically decrease

Note that milestones are very well equipped for meeting this need, as you can specify a deadline, track the proportion of finished vs. unfinished tickets, and each milestone has the same logic to support this kind of thing. 

> arguably, milestones are subsets of labels and buckets contain subsets of milestones

You could of course collapse all the distinction between the different feature sets and view everything as a "mode of label", but what is gained? Milestones have distinct features not supported by labels (deadlines, progress meters), and their intended use is quite orthogonal to general labels. The columns in projects boards area really intended to track *status* of issues/prs. A single work items progresses through these statuses. It gives us a temporal view of the process by which work is done. The current board *does* use the status columns to group things in a way that is redundant with the new milestones, but don't think it adds anything, except making the board more chaotic, because the same x axis is now encoding two different kinds of information simultaneously: temporal status of work and the "bucket" within which the work falls.

Perhaps we could have a sync discussion where we compare the old and new boards, and look at what milestones actually offer, so we can see how to meet your needs here?

It is possible too that the kind of metrics you want are best served by doing some report generation on tickets within a given timeframe. That should be quite easy to rig up, whether by pulling in a SasS integration or just by posing some queries to github.

shonfeder: @rodrigo7491 

> iirc the point of having different buckets on the project board was to see how the team's time was spent in different types of tasks, i.e., we spent a lot of effort recently on refactoring, so let's put more focus on usability next week. This information is lost if we keep refiling the buckets as Gabriela said. I always saw the buckets as a selection of issues without time constraints i.e., once we get all this done we discuss what to do next, but maybe it was just me.

These buckets have moved into milestones and/or labels. This information is not lost, it's just not represented in the **status** columns (because those columns are designed to reflect status). All this info on different buckets is still present, and readily available through filters and views.

> I don't think we should use milestones with two different meanings, or at least not without a very clear agreement on how to do so.

Where do you see milestones being used with two different meanings?

rodrigo7491: @shonfeder 

> These buckets have moved into milestones and/or labels. This information is not lost, it's just not represented in the **status** columns (because those columns are designed to reflect status). All this info on different buckets is still present, and readily available through filters and views.

That makes sense, thanks!

> Where do you see milestones being used with two different meanings?

Not being used, but being suggested to be used with time constraints in https://github.com/informalsystems/apalache/pull/1184#discussion_r772595598. Maybe I misinterpreted what you meant there.

shonfeder: I see. So my view is that this is a single way of using milestones: to track many different items of work that all build towards a (relatively) clearly delimited goal, and in particular I think this goal should have a well-defined end state. 

But milestones support deadlines. So we can either add a deadline, indicating that we are tracking work towards a goal which we aim to complete within a specific timeframe, or we can leave it without a deadline, in which case completion of the goal isn't time constrained.

shonfeder: Maybe to address the ambiguity around milestones we should add a section just to detail the features and proposed use of milestones... will think on this tommorrow.

konnov (approved):
This document is great, though it somehow ends abruptly. Perhaps, we could extend it with examples in the future. The role of labels is not clear to me at all.

As for priorities, the action-priority matrix makes most sense to me. It will still be subjective, as we are going to decide on the impact and effort.

COORDINATION.md
konnov: > Therefore, until the end of the fixed time interval, the number of tasks should monotonically decrease.

This is an idealistic point of view that assumes that all tasks can be planned in advance. They cannot. In reality, we often start with an issue and divide it into smaller issues as we identify them. Of course, it does not mean that we should not plan. We should. But instead of decomposing a single task into subtasks in advance, we can guess the complexity of the task, which would intuitively translate into the number of subtasks (issues) for the initial task. It's hard to guess the actual subtasks, but it's relatively easy to weigh a task.

That said, I find the number of tasks to be a bad metric. If we attached complexity to each task, we could say that the complexity of the milestone (as a sum of the task complexities) should decrease.

shonfeder:
> This document is great, though it somehow ends abruptly. Perhaps, we could extend it with examples in the future. 

Yeah, this should definitely be  a living document.

> The role of labels is not clear to me at all.

I've added a section on labels.

> As for priorities, the action-priority matrix makes most sense to me. It will still be subjective, as we are going to decide on the impact and effort.

I've made a note on this in the relevant section.

gabrielamafra (approved):
Great additions! I see this as a pretty good base for us to start from, and then improve from experimenting and discussing from now on.

COORDINATION.md
```diff
@@ -0,0 +1,203 @@
...
+- Label issues that are "chores" so you can prevent overloading anyone with
+  unpleasant work.
+
+There are only two constraints limiting the content of a label:
+
+1. Don't use labels to indicate the status of ongoing work. That is represented
+   by the *status* defined in whatever projects the work item belongs to.
+2. Don't use a label for trying to group issues that build towards a common goal
+   (e.g., issues that implement a new feature). That is represented by a
+   [milestone](#milestones).
```
rodrigo7491, line 198: From my understanding, this implies that we should remove all `F*` labels, e.g., `Farrays`, and substitute them with either a `bug` or a (not yet existing) generic `feature` label.

Currently the bug vs. feature split also shows up in the issue template, i.e., `[BUG]` or `[FEATURE]` in the title. If we are going to have a generic `feature` label, this becomes redundant, although some might find this redundancy useful.

COORDINATION.md
```diff
@@ -0,0 +1,203 @@
...
+
+A "Next" column is a subset of our backlog that we can fully control, and the
+idea is for it to hold the tasks that we will work on once we finish what is in
+progress. Therefore, there should be some sort of prioritization criteria to
+help us define which tasks from the backlog should go in this queue. The process
+of deciding which tasks should be on the "Next" column is usually called
+"replenishment".
+
+The Next column needs a size limit since "if everything is a priority, then
+nothing is a priority".
```
rodrigo7491, line 95: Any suggestions to what the size limit should be? I'm happy to have a rough estimate, and I don't think we should be too strict, in the sense that one or two issues above the "limit" should not be a problem. The fact that issues are not equally complex does not make this easier, and I guess we should also consider the team size when estimating a limit.

My point is that if we don't elaborate on it, this comment becomes moot. In which case it should be rewritten in a more generic "The Next column contains prioritized issues" way, without mentioning a limit.

rodrigo7491 (commented):
Very happy with how the documentation is evolving, specially with the new section on labels, thanks! I suggested some minor text changes and pointed two aspects that I think are not clear.

shonfeder:
We agreed in slack that we'd merge this in, adopting the recommendations. And we'd revisit the open questions in the future.

PR #1190, opened 2021-12-22
closes #1188: an ADR on the trace format

### LINKED ISSUE #1188: [FEATURE] Write an ADR on the trace format
Following the request #530, I propose to develop a specialized format for counterexamples. This format is not essential for apalache itself, but it will be quite useful for testing tools that aim at integrating trace-based testing techniques.

konnov:
closes #1188

This ADR introduces a simplified format for counterexamples and traces of distributed systems in general.

### DIFF
--- a/docs/src/adr/015adr-trace.md
+++ b/docs/src/adr/015adr-trace.md
@@ -208,12 +208,12 @@ like:
   }
 ```
 
-The optional field `parameters` is an array of names that must be set in the
+The optional field `params` is an array of names that must be set in the
 initial state (if there are any parameters). The parameters play the same role
 as `CONSTANTS` in TLA+. For example, the field may look like:
 
 ```js
-  "parameters": [ "Missionaries", "Cannibals" ]
+  "params": [ "Missionaries", "Cannibals" ]
 ```
 
 The field `vars` is an array of names that must be set in every state.
--- a/docs/src/adr/015adr-trace.md
+++ b/docs/src/adr/015adr-trace.md
@@ -158,6 +158,7 @@ in JSON.
     - Pros:
 
       - easy to understand, if you know TLA+.
+      - it looks amazing in PDF.
 
     - Cons:
 
@@ -173,7 +174,22 @@ in JSON.
     - Cons:
 
       - almost impossible to read.
-      - requires the knowledge of Apalache IR.
+      - too detailed and too verbose.
+      - requires the knowledge of Apalache IR and of TLA+.
+
+ 1. Use the Informal Trace Format, which is proposed in this ADR:
+
+    - Pros:
+
+      - almost no introduction is required to read the traces.
+      - relatively compact.
+      - easy to parse automatically.
+      - uses the idioms that are understood by the engineers.
+      - not bound to TLA+.
+
+    - Cons:
+
+      - consistency of the format is in conflict with the ease of writing.
 
 ## Solution
 
--- a/docs/src/adr/015adr-trace.md
+++ b/docs/src/adr/015adr-trace.md
@@ -1,0 +1,403 @@
+---
+authors: Igor Konnov
+proposed by: Vitor Enes, Andrey Kupriyanov
+last revised: 2021-12-22
+---
+
+# ADR-015: Informal Trace Format in JSON
+
+**Table of Contents**
+
+- [Summary (Overview)](#summary)
+- [Context (Problem)](#context)
+- [Options (Alternatives)](#options)
+- [Solution (Decision)](#solution)
+- [Consequences (Retrospection)](#consequences)
+
+## Summary
+
+We propose a simple format for counterexamples (traces) in [JSON][]. Although
+Apalache already supports serialization to [JSON][] in [ADR005][], it is a
+general serialization format for all the constructs of TLA+ that are supported
+by Apalache. This makes tool integration harder. It also make it hard to
+communicate counterexamples to engineers who are not familiar with TLA+. This
+ADR-015 contains a very simple format that does not require any knowledge of
+TLA+ and can be easily integrated into other tools.
+
+## Context
+
+A TLA+ execution (called a behavior in TLA+) is a very powerful concept. It can
+represent virtually any execution of a state machine, including sequential
+programs, concurrent programs, and distributed systems. Counterexamples that
+are produced by TLC and Apalache are just executions of a TLA+ state machine.
+These counterexamples have two shapes:
+
+ 1. A *finite execution*, that is, a sequence of states.
+
+ 1. A *lasso execution*, that is, a finite sequence of states (prefix) followed
+ by an infinitely repeated sequence of states (loop). Any infinite execution of
+ a *finite-state system* can be represented with a lasso.  (Not all executions
+ of infinite-state systems cannot be represented with a lasso.)
+
+Although the concept of an execution in TLA+ is quite simple, it builds upon
+the vocabulary of TLA+. Moreover, TLA+ counterexamples are using the expression
+language of TLA+.
+
+To illustrate the problem, consider a very simple TLA+ specification of the
+MissionariesAndCannibals puzzle (specified by Leslie Lamport). We use a typed
+version of this specification, see [MissionariesAndCannibalsTyped][]. Consider
+the following instance of the specification:
+
+```tla
+------------------- MODULE MC_MissionariesAndCannibalsTyped -----------------    
+Missionaries == { "m1_OF_PERSON", "m2_OF_PERSON" }
+Cannibals == { "c1_OF_PERSON", "c2_OF_PERSON" }
+
+VARIABLES
+    \* @type: Str;
+    bank_of_boat,
+    \* @type: Str -> Set(PERSON);
+    who_is_on_bank
+
+INSTANCE MissionariesAndCannibalsTyped
+
+NoSolution ==
+    who_is_on_bank["E"] /= {}
+=============================================================================
+```
+
+By checking the invariant `NoSolution`, we obtain the following counterexample
+in TLA+:
+
+```tla
+---------------------------- MODULE counterexample ----------------------------
+
+EXTENDS MC_MissionariesAndCannibalsTyped
+
+(* Constant initialization state *)
+ConstInit == TRUE
+
+(* Initial state *)
+State0 ==
+  bank_of_boat = "E"
+    /\ who_is_on_bank
+      = "E"
+          :> { "c1_OF_PERSON", "c2_OF_PERSON", "m1_OF_PERSON", "m2_OF_PERSON" }
+        @@ "W" :> {}
+
+(* Transition 0 to State1 *)
+State1 ==
+  bank_of_boat = "W"
+    /\ who_is_on_bank
+      = "E" :> { "c1_OF_PERSON", "m1_OF_PERSON" }
+        @@ "W" :> { "c2_OF_PERSON", "m2_OF_PERSON" }
+
+(* Transition 0 to State2 *)
+State2 ==
+  bank_of_boat = "E"
+    /\ who_is_on_bank
+      = "E" :> { "c1_OF_PERSON", "m1_OF_PERSON", "m2_OF_PERSON" }
+        @@ "W" :> {"c2_OF_PERSON"}
+
+(* Transition 0 to State3 *)
+State3 ==
+  bank_of_boat = "W"
+    /\ who_is_on_bank
+      = "E" :> {"c1_OF_PERSON"}
+        @@ "W" :> { "c2_OF_PERSON", "m1_OF_PERSON", "m2_OF_PERSON" }
+
+(* Transition 0 to State4 *)
+State4 ==
+  bank_of_boat = "E"
+    /\ who_is_on_bank
+      = "E" :> { "c1_OF_PERSON", "c2_OF_PERSON" }
+        @@ "W" :> { "m1_OF_PERSON", "m2_OF_PERSON" }
+
+(* Transition 0 to State5 *)
+State5 ==
+  bank_of_boat = "W"
+    /\ who_is_on_bank
+      = "E" :> {}
+        @@ "W"
+          :> { "c1_OF_PERSON", "c2_OF_PERSON", "m1_OF_PERSON", "m2_OF_PERSON" }
+
+(* The following formula holds true in the last state and violates the invariant *)
+InvariantViolation == who_is_on_bank["E"] = {}
+
+================================================================================
+(* Created by Apalache on Wed Dec 22 09:18:50 CET 2021 *)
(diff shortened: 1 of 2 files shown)


docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+programs, concurrent programs, and distributed systems. Counterexamples that
+are produced by TLC and Apalache are just executions of a TLA+ state machine.
+These counterexamples have two shapes:
+
+ 1. A *finite execution*, that is, a sequence of states.
+
+ 1. A *lasso execution*, that is, a finite sequence of states (prefix) followed
+ by an infinitely repeated sequence of states (loop). Any infinite execution of
+ a *finite-state system* can be represented with a lasso.  (Not all executions
+ of infinite-state systems cannot be represented with a lasso.)
```
Kukovec, line 40:
```suggestion
 a *finite-state system* can be represented by a lasso.  (In general, executions
 of infinite-state systems cannot be represented by lassos.)
```

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+
+ 1. The expressions are using a very small subset of TLA operators:
+
+   1. Integer and string literals.
+
+   1. Set constructor, sequence/tuple constructor, record constructor.
+
+   1. TLC operators over functions: `:>` and `@@`.
+
+In the hindsight, the above expressions are not very far from the JSON format.
```
Kukovec, line 150:
```suggestion
In hindsight, the above expressions are not very far from the JSON format.
```

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+
+In this ADR, we propose a very simple format that represents executions of
+state machines that follows the concepts of TLA+ and yet avoids complexity of
+TLA+. It is so simple that we call it *"Informal Trace Format"* (ITF).
+(Obviously, it is formal enough to be machine-readable).  By convention, the
+files in this format should end with the extension `.itf.json`.
+
+### The ITF Format
+
+**Trace object**. A trace in TTF is a JSON object:
```
Kukovec, line 188:
```suggestion
**Trace object**. A trace in ITF is a JSON object:
```

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+
+```js
+  "loop": 1
+```
+
+**State object**. A state is a JSON object:
+
+```js
+  {
+    "#meta": <optional object",
```
Kukovec, line 245:
```suggestion
    "#meta": <optional object>,
```

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+    "<varN>": <expr>
+  }
+```
+
+As in the trace object, the field `#meta` may be an arbitrary object.
+Different tools may use this object to write their metadata into this object.
+
+The names `<var1>, ..., <varN>` are the names of the variables that are
+specified in the field `vars`. The syntax of `<expr>` is specified below.
+
```
Kukovec, line 257:
```suggestion
The names `<var1>, ..., <varN>` are the names of the variables that are
specified in the field `vars`. Each state must define a value for every specified variable. The syntax of `<expr>` is specified below. 

```

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+    JSON array. TLA+ sequences are written as lists in this format.
+
+ 1. A record of the form `{ "field1": <expr>, ..., "fieldN": <expr> }`. A
+    record is just a JSON object. Field names should not start with `#`
+    and hence should not pose any collision with other constructs.
+    TLA+ records are written as records in this format.
+
+ 1. A tuple of the form `{ "#tup": [ <expr>, ..., <expr> ] }`. There is no
+ strict rule about when to use sequences or tuples. Apalache differentiates
+ between tuples and sequences and it may produce expressions of the both forms.
```
Kukovec, line 283:
```suggestion
 1. A tuple of the form `{ "#tup": [ <expr>, ..., <expr> ] }`. There is no
 strict rule about when to use sequences or tuples. Apalache differentiates
 between tuples and sequences, and it may produce both forms of expressions.
```

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+      }
+    }
+  ]
+}
+```
+
+Compare the above trace format with the TLA+ counterexample. The TLA+ example
+looks more compact. The ITF example is heavier on the brackets and braces.
+However, the example in the ITF format is self-explanatory and it does not
+require any dive in TLA+.
```
Kukovec, line 380:
```suggestion
Compare the above trace format with the TLA+ counterexample. The TLA+ example
looks more compact. The ITF example is heavier on the brackets and braces, but it is also designed with machine-readability and tool automation in mind, whereas TLA+ counterexamples are not. 
However, the example in the ITF format is also self-explanatory and does not
require any understanding of TLA+.
```

Kukovec (approved):
LGTM, pretty straightforward format.

gabrielamafra (approved):

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+
+   1. Set constructor, sequence/tuple constructor, record constructor.
+
+   1. TLC operators over functions: `:>` and `@@`.
+
+In hindsight, the above expressions are not very far from the JSON format.
+As many engineers know [JSON][], it seems natural to write these counterexamples
+in JSON.
+
+## Options
```
shonfeder, line 154: Seems like the options here don't include the option you're proposing? I think that's valuable because it lets us clearly compare the pros and cons of the options not selected against the option which is.

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+    in this format.
+
+ 1. A JSON integer, e.g., 123. According to [RFC7159][], JSON integers
+    must be in the range: `[-(2**53)+1, (2**53)-1]`. Integers in this
+    range *may be* written as JSON integers.
+
+ 1. A big integer of the following form: `{ "#bigint": "[-][0-9]+" }`.
+    We are using this format, as many JSON parsers impose limits on
+    integer values, see [RFC7159][]. Big and small integers *may be*
+    written in this format.
```
shonfeder, line 271: I'm slightly unsure about this complication, since it introduces two different ways of representing what should be, iiuc, understood as just a single kind of value. If any program that consumes this data is going to have to support bigints  anyhow, should we perhaps only have `{"#int":  "[-][0-9]+" }` as the supported value for integer-expressions?

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+
+**Expressions.** As usual, expressions are inductively defined. An expression
+`<expr>` is one of the following:
+
+ 1. A JSON string literal, e.g., `"hello"`. TLA+ strings are written as strings
+    in this format.
+
+ 1. A JSON integer, e.g., 123. According to [RFC7159][], JSON integers
+    must be in the range: `[-(2**53)+1, (2**53)-1]`. Integers in this
+    range *may be* written as JSON integers.
```
shonfeder, line 266: As a potential use of this encoding, I don't like that some values just the literal JSON value, while others are inferred from a special encoding. I suggest instead just making every value of the form `{"#type": <type-name>, "#value": <value>}`.

Otherwise people writer deserializers have to look at each object and inspect it first to see if it's one of these special objects, by getting all keys, then checking for the presence of the special key, or if it's just a normal object (which is actually supposed to be interpreted as a TLA+ record). This ends up introducing two different dimensions of ambiguity: some json values are puns for certain TLA+ values, and some json values are actually special instructions that are meant to indicate an alternative interpretation.

I think you can make the format much more regular if instead every value is tagged, then the function to dispatch for deserialization can just be something like

```scala
def deser(o: Json): Tla = o.type match {
  case "#string" => Some(o.value)
  case "#int" => Option(o.value.toInt)
  case "#rec" => convertRecord(o.value)
  // ...
}
```

My main claim here is that requiring tags for some values but not others, and putting these tags each as their own dictionary key instead of assigning a single key to pull out tags, actually doesn't provide any simplification, it just makes things more irregular.

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+        "#map": [
+          [ "E", { "#set": [ ] } ],
+          [ "W", { "#set": [ "c1_OF_PERSON", "c2_OF_PERSON",
+                             "m1_OF_PERSON", "m2_OF_PERSON" ] } ]
+        ]  
+      }
+    }
+  ]
+}
+```
```
shonfeder, line 375: In contrast to the other format, it looks like this one does not include data about the invariant that was violated. It'd think that was important info. Should it be included?

docs/src/adr/015adr-trace.md
```diff
@@ -0,0 +1,403 @@
...
+Nothing yet. However, we expect this format to become an exchange format
+for the tools that consume traces at their input:
+
+ - End-to-end testing tools,
+
+ - Bug reproduction tools,
+
+ - Trace visualization tools,
+
+ - Trace fuzzers.
```
shonfeder, line 398: This is space is for retrospection. And we can't retrospect at the outset! I suggest including list useful list in the intro instead, since this is bout motivation, not retrospective evaluation.

```suggestion
```

shonfeder (changes requested):
I can definitely see the value in the simplification proposed here.

I have some concers about the specifics of the proposed format, and one other nit, however.

docs/src/adr/015adr-trace.md
Kukovec: A trace is a general concept, it doesn't have to come from a counterexample to an invariant.

shonfeder: That makes sense. But the main motivating force for this to report on counter examples, and those are used as the examples. So it seems like we should also account for how we’ll supply this data when needed, and probably make a clearer statement about this being a general form for traces which can be specialized to meet the needs of reporting counter examples. (Especially since this linked from our public docs).

docs/src/adr/015adr-trace.md
konnov: I had the same feeling when writing this document. Actually, I wanted to keep JSON values only, but that seemed to be too much loss of information. We could make the values more regular; it's easy. However, I think that the priorities are different here:

1. Some of these traces may be written by hand, when people are developing their own tools. In this case, people would probably limit themselves to arrays and records.
2. When I am looking at the testing frameworks that we do for audits, the structure of the counterexamples is fixed, so we already know the shape of the states.

docs/src/adr/015adr-trace.md
Kukovec: Generally speaking, at least for the audit case, the invariant isn't referenced by any of the tools that consume counterexample traces. But since this is JSON, you're always allowed to add non-specified fields as you wish, it'd be easy enough to add a toplevel `"(violated) invariant" : expr` entry.

docs/src/adr/015adr-trace.md
konnov: True. However, if we put focus on usability in this format, not on the easy of writing a parser, it makes sense to write small integers without typing too much.

docs/src/adr/015adr-trace.md
konnov: This can go into `#meta`. I am thinking of this format as of a trace format, not being really specific to TLA+.

shonfeder: Adding info on this one way or another would help, even if just to say: "we don't actually need to include the invariant".

docs/src/adr/015adr-trace.md
shonfeder: Imo, this is a shortsighted view of usability. It’s optimizing for the simple case instead of regularity and generality. It means users have to study documentation and memorize which values get encoded which way and why.

in general, and I know I keep saying this, optimizing data exchange formats for human readability is a bad idea. If you want a readable view, just implement a small script to print a markdown table or a dot graph or an avg or whatever. 

Every time I’ve had to work with a data format where authors compromised the regularity of the machine encoding so that is has nice features as ascii art I’ve found it to slow me down and make things more tediuous (YAML in general is a great example).

konnov: I see your point. However, we already have a perfect machine-readable format. I really hope that JSON fixes this issue of bounded integers in the future. For instance, TypeScript supports big integers, but they are not serialized into JSON very well.

shonfeder (approved):

docs/src/adr/015adr-trace.md
danwt: I agree with Shon here for what it's worth. The ADR looks good a whole

shonfeder:
For the record, I thought I'd drop the two other suggestions that came up during our standup today:

- Use [Json Schema](https://cswr.github.io/JsonSchema/#what-is-json-schema?) to separate out the json output from the typing information that could be used for deserialization. This would allow keeping the direct JSON output focused on the simplest possible, bare-bones representation, then for users who want the richer typing info, they could also the generated schema.
- We could support two alternative encodings, differentiated by a toggle in the metadata. One using  using plain json types and one with consistently specified types (the latter being strictly more expressive, the former being focused on the less typing use case).

docs/src/adr/015adr-trace.md
konnov: Fixed.

konnov:
> * Use [Json Schema](https://cswr.github.io/JsonSchema/#what-is-json-schema?) to separate out the json output from the typing information that could be used for deserialization. This would allow keeping the direct JSON output focused on the simplest possible, bare-bones representation, then for users who want the richer typing info, they could also the generated schema.

I also thought that JSON schema could be an option. But would it solve the issue of reinterpreting some of the expressions differently, e.g., arrays as sets?

Another question about JSON schema is who is going to write it. I have seen the same story with XML schemas. What seemed to be in principle the right approach turned out to be heavy-weight and obscure. I guess, there is a benefit in using schemas when you want to make sure that your APIs do not break. It is not obvious to me why would one need this safety measure in tests, which would break anyways, if there is a problem with the input data.

konnov:
> * We could support two alternative encodings, differentiated by a toggle in the metadata. One using  using plain json types and one with consistently specified types (the latter being strictly more expressive, the former being focused on the less typing use case).

Let's try it and get some feedback first :-)

docs/src/adr/015adr-trace.md
konnov: I have added a discussion on that, so we can come back to your point later :-)

Issue #451, opened 2021-01-19
[FEATURE] Cover ConstSimplifier and ConstSimplifierForSmt with tests


konnov:
These classes do a lot of simplifications. It is easy to do a stupid mistake in those simplifications.

konnov:
It's probably a good target for Scalacheck

konnov:
`at.forsyte.apalache.tla.lir.IrGenerators` may help to write PBT tests

shonfeder:
Was this closed by #1206, or does it need a followup for the `ConstSimplifierForSmt` part?

PR #1191, opened 2021-12-22
Tests for ConstSimplifier arithmetical operations

gabrielamafra:
Hello :octocat: 

This adds several property-based tests and some unit tests to the `ConstSimplifier` class, fixes some bugs on its base class `ConstSimplifierBase`, and includes some missing simplification scenarios.
1. A new jar is built in order to reuse tlair's `IrGenerator` in tla-pp
2. New errors are introduced for the following scenarios:
  a. Division by 0
  b. Mod (%) by 0
  c. Negative expoents
  d. Expoents bigger than an Integer (we used to try and make the operation used doubles, but that was not working) 
  e. Expoential operations that would overflow `BigInt`
3. TLA builder for `TlaInt` now accepts BigInts as well

The tests are structured like this:
- Generate any TLA expression, then combine with operations and parameters that result in a simplification:
  - any expression with <operation> with <operation's neutral element> should equal the expression itself. 
    - i.e. `x + 0 = x` 
  - any expression with  <operation> with <operation's absorbing element> should equal the same constant result regardless of the expression
    - i.e. `x * 0 = 0`, `x % 1 = 0`
    - These tests also include assertions like `x / x = 1` since the testing structure is the same
    - There are two groups for this: simplifications that result in 0 and simplifications that result in 1
- Generate pairs of numbers, combine them in an arithmetical expression and check that the result of the simplification equals the evaluation of the expression
- For exponential operations, use fixed values (in regular unit tests) since the result is highly dependent on the combination of the two values and that's a pain to generate.

Helps with https://github.com/informalsystems/apalache/issues/451 but is not enough for closing it. More tests to come!


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+      }
+
+    // x + 0 = x
+    case OperEx(TlaArithOper.plus, leftEx, ValEx(TlaInt(right))) =>
+      if (right == 0) {
+        leftEx
       } else {
-        ValEx(TlaInt(left + right))(intTag)
+        OperEx(TlaArithOper.plus, leftEx, ValEx(TlaInt(right))(intTag))(intTag)
       }
```
shonfeder, line 48: Don't mean to be code golfing here, and I know it's a bit orthogonol to these changes, but do you think a set of patterns like this might work here? 

```scala
    case OperEx(TlaArithOper.plus, ValEx(TlaInt(0)), ex) => ex
    case OperEx(TlaArithOper.plus, ex, ValEx(TlaInt(0))) => ex
    case ex @ OperEx(TlaArithOper.plus, _, _) => ex
```

If so, I think it's reads much more clearly. If would work, and you agree it's easier to read, you might consider swapping that out here and in the other similar cases.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+        ValEx(TlaInt(0))(intTag)
+      } else if (right == 1) {
+        leftEx
+      } else {
+        OperEx(TlaArithOper.mult, leftEx, ValEx(TlaInt(right))(intTag))(intTag)
+      }
+
+    case ex @ OperEx(TlaArithOper.div, ValEx(TlaInt(left)), ValEx(TlaInt(right))) =>
+      if (right == 0) {
+        throw new IllegalArgumentException(s"Division by zero at ${ex.toString}")
```
shonfeder, line 100: I think a custom exception might be more fitting here. iiuc, Java's [`IlegalArugemntException`](Thrown to indicate that a method has been passed an illegal or inappropriate argument.) is meant to be

> Thrown to indicate that a method has been passed an illegal or inappropriate argument.

But I don't think that's really what's happening here. It's more like an input error, iiuc. 

Perhaps this should raise a `TalInputError`? https://github.com/informalsystems/apalache/blob/406fcf0e3b46b4c6abde49c57b2981eea3a43c4a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/exceptions.scala#L29-L37

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+
+    // x / 1 = x
+    case ex @ OperEx(TlaArithOper.div, leftEx, ValEx(TlaInt(right))) =>
+      if (right == 0) {
+        throw new IllegalArgumentException(s"Division by zero at ${ex.toString}")
+      } else if (right == 1) {
+        leftEx
+      } else {
+        OperEx(TlaArithOper.div, leftEx, ValEx(TlaInt(right))(intTag))(intTag)
+      }
```
shonfeder, line 121: I was surprised to see a second check for division by zero here. Might we get away with a single check if one of the cases were

```scala
    case ex @ OperEx(TlaArithOper.div, leftEx, ValEx(TlaInt(0))) =>
        throw new IllegalArgumentException(s"Division by zero at ${ex.toString}")
```

Then allow everything after that to just assume the div by zero was already handled?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+
+    // x / x = 1
+    case ex @ OperEx(TlaArithOper.div, leftEx, rightEx) =>
+      if (leftEx == rightEx) {
+        ValEx(TlaInt(1))(intTag)
+      } else {
+        OperEx(TlaArithOper.div, leftEx, rightEx)(intTag)
```
shonfeder, line 128: Could this be:

```suggestion
        ex
```

if not for some reason, then I think the `ex` is unused in this case?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+        throw new IllegalArgumentException(s"Mod by zero at ${ex.toString}")
+      } else {
+        ValEx(TlaInt(left % right))(intTag)
+      }
+
+    // x % 1 = 0
+    case ex @ OperEx(TlaArithOper.mod, leftEx, ValEx(TlaInt(right))) =>
+      if (right == 0) {
+        throw new IllegalArgumentException(s"Mod by zero at ${ex.toString}")
       } else if (right == 1) {
```
shonfeder, line 142: My same question re: moving the by zero case into a single, general case should apply here too I think.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+        OperEx(TlaArithOper.mod, leftEx, ValEx(TlaInt(right))(intTag))(intTag)
+      }
+
+    // x % x = 0
+    case OperEx(TlaArithOper.mod, leftEx, rightEx) =>
+      if (leftEx == rightEx) {
+        ValEx(TlaInt(0))(intTag)
       } else {
-        ValEx(TlaInt(left * right))(intTag)
+        OperEx(TlaArithOper.mod, leftEx, rightEx)(intTag)
```
shonfeder, line 153: Perhaps another case where biding the patter to `ex @ ...` then returning the `ex` rather than having to reconstruct the expression would simplify things a bit?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+      if (right == 0) {
+        throw new IllegalArgumentException(s"Mod by zero at ${ex.toString}")
       } else if (right == 1) {
-        ValEx(TlaInt(left))(intTag)
+        ValEx(TlaInt(0))(intTag)
+      } else {
+        OperEx(TlaArithOper.mod, leftEx, ValEx(TlaInt(right))(intTag))(intTag)
+      }
+
+    // x % x = 0
```
shonfeder, line 148: These comments are super helpful. :+1:

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+        OperEx(TlaArithOper.mod, leftEx, rightEx)(intTag)
       }
 
-    case OperEx(TlaArithOper.div, ValEx(TlaInt(left)), ValEx(TlaInt(right))) =>
-      ValEx(TlaInt(left / right))(intTag)
+    case ex @ OperEx(TlaArithOper.exp, ValEx(TlaInt(base)), ValEx(TlaInt(power))) =>
+      if (power < 0) {
+        throw new IllegalArgumentException(s"Negative power at ${ex.toString}")
+      } else if (!power.isValidInt) {
+        throw new IllegalArgumentException(s"Power of ${power} is bigger than an integer at ${ex.toString}")
```
shonfeder, line 160: Perhaps the error message might be improved by (in addition to using a custom exception) adding info about the size limit?

```suggestion
        throw new IllegalArgumentException(s"Power of ${power} is bigger than the max allowed ${Int.MaxValue} at ${ex.toString}") 
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
-      ValEx(TlaInt(left % right))(intTag)
+    // x ^ 0 = 1
+    // x ^ 1 = x
+    case OperEx(TlaArithOper.exp, leftEx, ValEx(TlaInt(right))) =>
+      if (right == 0) {
+        ValEx(TlaInt(1))(intTag)
+      } else if (right == 1) {
+        leftEx
+      } else {
+        OperEx(TlaArithOper.exp, leftEx, ValEx(TlaInt(right))(intTag))(intTag)
```
shonfeder, line 180: Another use case for a `ex @ ....` in the head? And perhaps breaking this case analysis into three cases?

shonfeder (changes requested):
Really nice PR. The description of the work and rational in the description is awesome. Thanks!

The only change request that I think is blocking on my end is selecting or introducing a custom exception to raise for the error cases. Everything else is just stylistic or suggestions that I subjectively think are more readable, feel free to disregard (tho I'd be interested in your rationale if you think the suggestions aren't feasible or are worse for readability).

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
Kukovec, line 180:
```suggestion
    case ex @ OperEx(TlaArithOper.exp, leftEx, ValEx(TlaInt(right))) =>
      if (right == 0) {
        ValEx(TlaInt(1))(intTag)
      } else if (right == 1) {
        leftEx
      } else {
        ex
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+      }
 
-    case OperEx(TlaArithOper.exp, ValEx(TlaInt(base)), ValEx(TlaInt(power))) =>
-      if (power.isValidInt) {
-        ValEx(TlaInt(base.pow(power.toInt)))(intTag)
+    // 0 ^ x = 0 (except if x = 0 which will match the previous case)
```
Kukovec, line 183: Literal `0^0` will match the l156 case and default to the `BigInt` implementation of `0.pow(0)`. We can catch it earlier.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+    // 1 ^ x = 1
+    case OperEx(TlaArithOper.exp, ValEx(TlaInt(left)), rightEx) =>
```
Kukovec, line 185:
```suggestion
    case ex @ OperEx(TlaArithOper.exp, ValEx(TlaInt(left)), rightEx) =>
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+      if (left == 0) {
+        ValEx(TlaInt(0))(intTag)
+      } else if (left == 1) {
+        ValEx(TlaInt(1))(intTag)
       } else {
-        // the power does not fit into an integer. That is a lot. Use doubles.
-        val pow = Math.pow(base.toDouble, power.toDouble)
-        val powAsBigInt = BigDecimal(pow).setScale(0, BigDecimal.RoundingMode.DOWN).toBigInt()
-        ValEx(TlaInt(powAsBigInt))(intTag)
+        OperEx(TlaArithOper.exp, ValEx(TlaInt(left))(intTag), rightEx)(intTag)
```
Kukovec, line 191:
```suggestion
        ex
```

Kukovec (approved):
Good PR, but I feel like it opens up an unfortunate philosophical discussion:
There are many undefined expressions - `k/0`, `0/0`, `0 % 0`, etc., that can occur dynamically:
```
Init == x = -1
A == 0 / x
B == x / x
C == x ^ 0
...
Next == x' = x + 1 /\ ...
```
Considering this, does it then actually make sense to simplify non-constant expressions involving `/, *, ^, %`? (`+,-` are safe). I approve the changes of this PR, but we have to add comments acknowledging these can be runtime issues.

My second major comment has nothing to do with the contents of this PR (and is not directed at Gabriela), but concerns the simplifier itself: Why are the simplification methods, which are by all counts transformations, not using `TransformationTracker`? We're losing the ability to track source information. (this is the reason for all the `ex @ ...`/`return ex` suggestions, in a tracked system you'd want to avoid copy-creation)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
Kukovec: I'll +1 this sentiment, I feel like there's some pattern-match simplification that can happen here. Either flat, like Shon suggested or something of the sort:
```
case OperEx( arithOper, l, r ) =>
  (l,r) match {
   case (literal, literal) =>
   case (any, literal) => ...
   ...
  }
```

konnov:
> My second major comment has nothing to do with the contents of this PR (and is not directed at Gabriela), but concerns the simplifier itself: Why are the simplification methods, which are by all counts transformations, not using `TransformationTracker`? We're losing the ability to track source information. (this is the reason for all the `ex @ ...`/`return ex` suggestions, in a tracked system you'd want to avoid copy-creation)

These simplifications are happening in the rewriting rules. If we start tracking changes there, we will create a massive map, for no reason. It probably makes sense to decorate simplification in the optimization pass, but not in the rewriting rules.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
gabrielamafra: Yes! I moved this around way too much, that's indeed redundant.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
gabrielamafra: I've just comitted a version similiar to the one Shon proposed. I couldn't actually do `ValEx(TlaInt(0))` because 0 has to be a `BigInt`, and I can't do `ValEx(TlaInt(BigInt(0)))` because `BigInt` is not a case class. I could do something like this: https://codereview.stackexchange.com/questions/14561/matching-bigints-in-scala, but I found it more idiomatic to instead use actual if expressions inside the case statements like
```scala
case OperEx(TlaArithOper.plus, ValEx(TlaInt(left)), ex) if left == 0 => ex
```

This helped a lot with readability and I've also grouped this flat simplifications by arithmetical operation, which I personally believe is better to reason about than having a generalized form that avoids code duplication and increases complexity. But that's my point of view, let me know what you guys think :smile:

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
gabrielamafra: The limit is actually the `BigInt` limit, which is `2 ^ Int.MaxValue` according to the [docs](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html). I'll add that to the message.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+      } else {
+        try {
+          // This can take a long time for big base values i.e. 2147484647 ^ 1100000
+          // Maybe we should consider implementing a timeout
+          ValEx(TlaInt(base.pow(power.toInt)))(intTag)
+        } catch {
+          case _: ArithmeticException =>
+            throw new IllegalArgumentException(s"The result of ${ex.toString} exceedes the limit")
+        }
+      }
```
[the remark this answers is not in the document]
gabrielamafra: Didn't remember that, I just replicated the behavior from scala implementation. Thanks!

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,170 @@ abstract class ConstSimplifierBase {
...
+        ValEx(TlaInt(left / right))(intTag)
+      }
+
+    // 0 / x = 0
+    case OperEx(TlaArithOper.div, ValEx(TlaInt(left)), rightEx) =>
+      if (left == 0) {
+        ValEx(TlaInt(0))(intTag)
+      } else {
+        OperEx(TlaArithOper.div, ValEx(TlaInt(left))(intTag), rightEx)(intTag)
+      }
```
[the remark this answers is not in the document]
gabrielamafra: The best way to deal with undefined behavior is reporting it instead of trying to compute anything, but that's not always possible. In my understanding, the changes on this PR helps a bit with reporting in some cases and does not actually change the result in other cases. It just changes what is computed when behavior is undefined, but there's no better or worse in that, right? Here's some examples:
 
Expressions like `x = 0 /\ e = 0 / x`
Before: `e` can be anything
Now: `e = 0`
None of them throws an error which would be the best thing to do on undefined behavior. Not better not worse.

Expressions like `x = 1 /\ e = 0 / (x - x)``
Before: `e` can be anything
Now: Throws an error
Better than the current implementation.

PS: This assumes that the SMT solver is actually ignoring (evaluating as `true`) any expressions with undefined behavior, which seems to be the case for the examples I tested.

Regarding your example, I've tested on both this branch and `unstable` and the behavior is the same. Probably because the solver is just setting the hole `(0 = 0 / x)` expression to true when the operation is undefined.

I believe we should document better what we do with undefined behavior, which ones we can detect and maybe even explore adding constraints as you suggested earlier this week. I still have to open an issue for that as we agreed, and my suggestion would be to go further in this discussion when we address the issue. That is, if you agree that this PR is not making anything worse.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -29,49 +29,89 @@ abstract class ConstSimplifierBase {
...
+    // x + 0 = x
+    case OperEx(TlaArithOper.plus, leftEx, ValEx(TlaInt(right))) if right == 0 => leftEx
+
+    // Evaluate constant subtraction
+    case OperEx(TlaArithOper.minus, ValEx(TlaInt(left)), ValEx(TlaInt(right))) => ValEx(TlaInt(left - right))(intTag)
+    // 0 - x = -x
+    case OperEx(TlaArithOper.minus, ValEx(TlaInt(left)), rightEx) if left == 0 =>
+      OperEx(TlaArithOper.uminus, rightEx)(intTag)
+    // x - 0 = x
+    case OperEx(TlaArithOper.minus, leftEx, ValEx(TlaInt(right))) if right == 0 => leftEx
```
shonfeder, line 45: Any particular reason you've opted for guards here (the `if .... => `) rather than just matching against the `0` value?

E.g.,

```suggestion
    case OperEx(TlaArithOper.plus, ValEx(TlaInt(0)), rightEx) => rightEx
    // x + 0 = x
    case OperEx(TlaArithOper.plus, leftEx, ValEx(TlaInt(0))) => leftEx

    // Evaluate constant subtraction
    case OperEx(TlaArithOper.minus, ValEx(TlaInt(left)), ValEx(TlaInt(right))) => ValEx(TlaInt(left - right))(intTag)
    // 0 - x = -x
    case OperEx(TlaArithOper.minus, ValEx(TlaInt(0)), rightEx) => OperEx(TlaArithOper.uminus, rightEx)(intTag)
    // x - 0 = x
    case OperEx(TlaArithOper.minus, leftEx, ValEx(TlaInt(0))) => leftEx
```

shonfeder (approved):
Changes generally LGTM, tho I am curious about one choice.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
gabrielamafra: Yes! I actually explained it here: https://github.com/informalsystems/apalache/pull/1191#discussion_r774526282

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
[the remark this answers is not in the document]
Kukovec: Again, I support all of the simplifications introduced in the PR in the dynamically undefined cases, on the condition that we have a document (or at least comments) explicitly mentioning that the cases were considered and a design decision was made to e.g. treat `0^0` as 1 for consistency.
This PR is long enough as it is, but it might make sense to open an issue for adding these decisions to the manual.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
shonfeder: Sorry I missed this! Makes perfect sense to me. TIL :)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
shonfeder: I see! My bad. Makes perfect sense.

UNRELEASED.md
```diff
@@ -10,3 +10,14 @@
...
+### Features
+
+* New errors for the following constant simplification scenarios (see #1191):
+  1. Division by 0
+  1. Mod (%) by 0
+  1. Negative expoents
+  1. Expoents bigger than an Integer
+  1. Expoential operations that would overflow `BigInt`
+
+### Bug Fixes
```
shonfeder, line 23: We can leave this to be added if/when bug fixes get added during this release period.
```suggestion
```

UNRELEASED.md
```diff
@@ -10,3 +10,14 @@
          * Some bug fix, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
+
```
shonfeder, line 13:
```suggestion
```

shonfeder (approved):
Two little tweaks to the release notes, just due to the quirks of our ad hoc system :) 

Otherwise LGT!

PR #1202, opened 2022-01-09
Factor out the injectorFactory

shonfeder:
This tiny refactor is just some incidental cleanup on the CLI plumbing.

- It removes the `injectorFacotry`
- It reorders the arguments to `handleExceptions`

In the current state of the code, the injectorFactory looks like it's
just adding unhelpful indirection. Removing it allows us to save about
10 lines of code and remove an unreachable throw.

The arguments to `handleExceptions` are reordered to reflect the the
underlying logic better. I.e., we call the `runner` on the `injector`
and `cmd`, so the `runner` is now supplied first.

-------


konnov (approved):
Looks good!

Issue #480, opened 2021-01-23
[FEATURE] Algebraic transformations in the preprocessor


konnov:
There are a number of expressions that can be significantly simplified by using logical and algebraic equivalences.

**Example 1:**

```tla
x \in X \union Y
```

can be simplified to

```tla
x \in X /\ x \in Y
```

**Example 2:**

```tla
\E x \in a..b: P
```

can be simplified as

```tla
\E x \in Int: a <= x /\ x <= b /\ P
```

We are doing some of simplifications in the preprocessors already, but we should do that more systematically by introducing a rewriting framework. By doing so, we should eliminate a large class of issues, where our encoding is suboptimal.

konnov:
By a rewriting framework I meant something like [the worklist algorithm](https://en.wikipedia.org/wiki/Data-flow_analysis#The_work_list_approach). The point is to avoid unnecessary recursive calls when subexpressions were not optimized. However, I am not sure about how to do that in Scala.

gabrielamafra:
I'm not capturing the necessity of this type of algorithm here. My understanding is that we never need to re-simplify/optimize subexpressions of an expression we are rewriting, as long as we always rewrite using a bottom-up approach and never change the contents of a subexpression while rewriting. This is how `ConstSimplifier` is working: there is a single deep recursion and anything that changes halfway of the recursion is simplified before continuing to recurse and simplify its parent.

We could take a look at other rewrite implementations besides `ConstSimplifier` and check that they can work under this 2 assumptions and, if so, make sure that new nodes (created from a rewrite rule) are shallow simplified/optimized. This could be a first step towards unifying all rewrite implementations as suggested in https://github.com/informalsystems/apalache/issues/138

konnov:
Right. We don't need any special framework to implement this kind of preprocessing. It would be great, if we could call different optimizers when one of the optimizers has changed a subexpression. A naive fixpoint approach applied to a whole expression over and over again would be really slow. I think we would need some form of recursion that applies different optimizations to subexpressions and once nothing new has been produced, it produces the result upwards to the parent.

PR #1206, opened 2022-01-10
ConstSimplifier remaining tests

gabrielamafra:
Hello :octocat: 

This PR adds several tests to `ConstSimplifier` which now should be fully covered. The simplification rules in `ConstSimplifierBase` are refactored to follow the inline pattern introduced in #1191. Addresses https://github.com/informalsystems/apalache/issues/451.

The new tests are grouped by expected result (to improve readability) and necessary generators (to improve performance).

Some simplification rules were changed/corrected to make simplification more sound (and therefore easier to test):
1. `x /= y` is now always simplified to `!(x = y)` (not only when under negation as it was). Since all `TlaOper.ne` occurrences are eliminated, other simplification rules for this operator are removed.
2. `x = x` is now simplified to `TRUE` for any expression, not only constants
3. When simplifying `IF x THEN FALSE ELSE y` to `!x /\ y`, we apply additional simplification for `!x` and then for the new expression `!x /\ y` so recursion is consistent*
4. `TRUE -> x` is now correctly simplified to `x` (was `!x`)
5. `x /\ TRUE` and `x \/ FALSE` are now simplified to `x` (was `/\ x`). This broke `AndRule` which expected the simplification result to be either a conjunction or a value, so pattern matching for `AndRule` was also updated to expect any expression as a conjunction simplification result.

\* All tests check that the expected result is deep simplified


test/tla/Bug931.tla
```diff
@@ -3,6 +3,6 @@ EXTENDS  FiniteSets
 
 Init == TRUE
 Next == TRUE
-Inv == Cardinality({}) = 0
+Inv == Cardinality({}) /= 1
```
gabrielamafra, line 6: The expression optimizer transforms `Cardinality(S) = 0` into `S = {}` so, in this case, `Cardinality({}) = 0` into `{} = {}`. The new more generic simplification rule for equality transforms `{} = {}` into `TRUE`, and therefore suppresses the bug this test is trying to reproduce. I understand that suppressing the polymorphic type error here is a good thing, therefore I changed the test instead of changing the simplification behavior. Let me know if you guys disagree, please.

konnov:
That's great! Would it make sense to have a look at ExprOptimizer and #480. There is a lot of room for optimizations?

konnov (approved):
Great work! I just noticed that we've never implemented constant simplifications over sets, e.g., `{1, 2} \union {3}` could be simplified to `{1,2,3}`. These expressions occur quite often when we combine a few constants for model checking. Shall we introduce these simplifications as well (in a separate issue/PR)?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -113,130 +128,93 @@ abstract class ConstSimplifierBase {
...
-      ValEx(TlaBool(true))(boolTag)
-
+    // TRUE -> x = x
+    case OperEx(TlaBoolOper.implies, ValEx(TlaBool(true)), right) => right
+    // x -> FALSE = !x
     case OperEx(TlaBoolOper.implies, lhs, ValEx(TlaBool(false))) =>
       simplifyShallow(OperEx(TlaBoolOper.not, lhs)(boolTag))
 
+    // Evaluate equivalence between constants
     case OperEx(TlaBoolOper.equiv, ValEx(TlaBool(left)), ValEx(TlaBool(right))) =>
```
Kukovec, line 185: I was under the impression that one of the passes transforms `<=>` into `=`. Do any of these ever fire?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -113,130 +128,93 @@ abstract class ConstSimplifierBase {
...
-      if (thenEx != elseEx) {
-        ite
-      } else {
-        thenEx
-      }
+      simplifyShallow(OperEx(TlaBoolOper.not, pred)(boolTag))
+    // IF x THEN FALSE ELSE y = !x /\ y
+    case OperEx(TlaControlOper.ifThenElse, pred, ValEx(TlaBool(false)), elseEx) =>
+      val newPredicate = simplifyShallow(OperEx(TlaBoolOper.not, pred)(boolTag))
+      simplifyShallow(OperEx(TlaBoolOper.and, newPredicate, elseEx)(boolTag))
```
Kukovec, line 212: Just for symmetry, do we also want `IF x THEN TRUE ELSE y <=> x \/ y` ? What about the `IF x then y ELSE (T/F)` cases?

Kukovec (approved):
LGTM, some minor comments.

Side-note: Simplify your life and introduce `def trueEx : TlaEx = ValEx(TlaBool(true))(boolTag)` and `falseEx`

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
gabrielamafra: I checked and they don't, at least not from our integration test suite. I got rid of them, thanks!

PR #1208, opened 2022-01-12
Fix expansion of ~ in configured paths

shonfeder:
@Kukovec reported that `~` was not be expanded in configured paths correctly.
Indeed, this was an oversight on my part. I failed to correctly override
pureconfig's reader for `Path` and `File` types. This PR fixes that mistake.

Followup to #1160

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
@@ -11,7 +11,6 @@ import com.typesafe.config.ConfigObject
 private object Converters {
   import pureconfig.ConvertHelpers._
 
-
   private def expandedFilePath(s: String): Path = {
     Paths.get(if (s.startsWith("~")) s.replaceFirst("~", System.getProperty("user.home")) else s)
   }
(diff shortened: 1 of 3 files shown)


konnov:
It does not seem to work with the docker tests

shonfeder:
integration tests on docker are fixed.

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
```diff
@@ -4,39 +4,23 @@ import pureconfig._
...
+import com.typesafe.config.ConfigValueFactory
+import com.typesafe.config.ConfigObject
 
-object Converters {
+// Provides implicit conversions used when deserializing into configurable values.
+private object Converters {
+  import pureconfig.ConvertHelpers._
 
   private def expandedFilePath(s: String): Path = {
     Paths.get(if (s.startsWith("~")) s.replaceFirst("~", System.getProperty("user.home")) else s)
```
konnov, line 15: Shall we check, whether a string is starting with `"~" + File.separator`. Otherwise, we would misinterpret strings like `~user1`, which have a different meaning in shell.

konnov (approved):
Looks much simpler than before. Nice!

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
shonfeder: Sorry. Had this configured for auto merge once approved.

yeah, the expansion is not strictly correct. But that’s not actually introduced in this PR: it’s just a continuation of the initial code from the outputManager.

I’ll address this in a follow up tho.

PR #1221, opened 2022-01-13
Fix build config to allow skipping tests

### LINKED ISSUE #1219: [BUG] Build error when using `make apalache-jar`
## Description

Building Apalache using `make apalache-jar`, as described in the [documentation](https://apalache.informal.systems/docs/apalache/installation/source.html), leads to the following error:
```
[ERROR] Failed to execute goal on project tla-pp: Could not resolve dependencies for project at.forsyte.apalache:tla-pp:jar:0.19.2-SNAPSHOT: Could not find artifact at.forsyte.apalache:tlair:jar:tests:0.19.2-SNAPSHOT in oss-sonatype (https://oss.sonatype.org/content/repositories/snapshots/) -> [Help 1]
```
Building with `make` works without problems. The problem with `make apalache-jar` happens after running both `mvn clean && mvn package` and `make`.

## Expected behavior

No error.

## System information

- Apalache version [`apalache-mc version`]: v0.19.2
- OS [e.g. Ubuntu Linux or Mac OS, and the current version]: Linux Mint 20.2
- JDK version [e.g. OpenJDK 0.8.0]: openjdk 11.0.13

shonfeder:
Closes #1219 

51228a7 had the unintended consequence of triggering the maven/surefire
design bug reported in https://issues.apache.org/jira/browse/MJAR-138.

The effect of this was to break some of our build targets that skip
tests by setting `maven.test.skip=true`, because it would not build the
tlair jar when tests were skipped, but would then still require that the
jar be present to compile.

This fix implemented the workaround described at
https://issues.apache.org/jira/browse/MJAR-138?focusedCommentId=16160394&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16160394
-- namely, we add a maven profile that only includes the dependency when
`mavne.test.skip=false`.

-------

### DIFF
--- a/tla-pp/pom.xml
+++ b/tla-pp/pom.xml
@@ -14,6 +14,28 @@
     <name>tla-pp</name>
     <url>https://github.com/informalsystems/apalache</url>
 
+    <profiles>
+        <!-- Workaround to allow a test-only internal dependency without breaking maven.test.skip -->
+        <!-- See https://issues.apache.org/jira/browse/MJAR-138?focusedCommentId=16160394&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16160394 -->
+        <profile>
+            <activation>
+                <property>
+                    <name>maven.test.skip</name>
+                    <value>!true</value>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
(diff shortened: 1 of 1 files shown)


konnov (approved):
Yes, let's merge to unblock the builds

Issue #1235, opened 2022-01-18
[BUG] Simplificiation of neutral arithmetic expression can be inconsistent


shonfeder:
## Description

Initially reported by @Kukovec on #1228, the recent changes to the `ConstSimplifierBase` in #1231 seem to introduce some non-confluent arithmetical simplifications.

## Failing test

```scala
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running at.forsyte.apalache.tla.pp.TestDesugarer
Tests run: 28, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.729 sec
Running at.forsyte.apalache.tla.pp.TestExprOptimizer
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.065 sec
Running at.forsyte.apalache.tla.pp.TestNormalizer
Tests run: 24, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.049 sec
Running at.forsyte.apalache.tla.pp.TestParameterNormalizer
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.01 sec
Running at.forsyte.apalache.tla.pp.TestConstSimplifier
Tests run: 19, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 25.014 sec <<< FAILURE!
simplifies arithmetical operations with their neutral values(at.forsyte.apalache.tla.pp.TestConstSimplifier)  Time elapsed: 0.388 sec  <<< ERROR!
org.scalatest.exceptions.GeneratorDrivenPropertyCheckFailedException: TestFailedException was thrown during property evaluation.
  Message: 1930172387802112260 was not equal to 1930172387802112377 when simplifying (-1926341659 * -1001988603) ^ 1
  Location: (TestConstSimplifier.scala:59)
  Occurred when passed generated values (
    arg0 = -1926341659 * -1001988603
  )
```

## Expected behavior

In the property test `simplifies arithmetical operations with their neutral values` we expect that the simplification of an arbitrary expression is always the same as the simplification of that expression with a neutral value applied (e.g. if our base expression is `x`, then `x + 0 = x`). However, we are seeing divergence  between the simplified base expression and the simplification derived by applying `f x = x ^ 1`.

[...]

PR #1237, opened 2022-01-18
Fix exponentiation inconsistency

gabrielamafra:
Hello :octocat: 

This fixes https://github.com/informalsystems/apalache/issues/1235 by using `BigInt#pow` to obtain consistent results without floating point problems, but only after using `Math.pow` to estimate the result size and ensure its not too big and, therefore, too slow to compute.


konnov (approved):
Looks great! If Scalacheck detects a problem with that too, we should think about limiting the powers with a constant.

PR #1250, opened 2022-01-24
Fix docker continuous deployment

shonfeder:
This is a followup to #1234, which accidentally broke the continuous
deployment of docker containers to the github container registry.

This happened because we are now generating our docker files based 
on the sbt build configuration, but the current github workflow expects
a Dockerfile in the root of the project. The proposed fix here (to be
tested once the PR is merged into unstable) is to generate the docker
file using our make target, and then point the github action to the 
"context" (directory) where the build file and its related artifacts are stored
in our artifact dir.

This patch fixes that buy first building the docker image, before
invoking the github action to upload the image to the registry

-------

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -13,6 +13,7 @@ on:
   push:
     branches:
       - unstable
+      - shon/fix-docker-cd
 
 env:
   REGISTRY: ghcr.io
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -45,10 +45,16 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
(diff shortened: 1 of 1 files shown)


gabrielamafra (approved):
Just a suggestion: we could temporarily change this workflow's condition to run on this branch/all pull requests so we can validate the fix before merging. Totally optional, just in case you think it would be more practical.

shonfeder:
Good idea. I'll do that, since it's already passed CI and review, it's not really risky to push the container to our registry. Thanks for the idea!

shonfeder:
Nice. Now we can be confident the fix is actually a fix :D https://github.com/informalsystems/apalache/pkgs/container/apalache/13764782?tag=shon-fix-docker-cd

Very happy to err on the side of correctness here.

PR #1252, opened 2022-01-25
`LanguagePred` for reTLA

### LINKED ISSUE #1241: Implement `LanguagePred` for reTLA
Like Kera, reTLA has an associated supported language fragment and rejects specs if they use constructs not permitted by the fragment.

Kukovec:

closes #1241


Kukovec:
No idea what's up with the formatting check, I ran `make fmt-fix` again and there was no diff. @shonfeder any advice?

shonfeder:
@Kukovec the formatting check is run on the diff from `origin/unstable`. I'm guessing you didn't have the latest `origin/unstable` fetched. We fetch origin in the `make fmt-check` target, but not in the `fmt-fix` target. Maybe I should add that?

Probably the easiest for devs tho is just to ensure you have enabled format on save in your editor.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
```diff
@@ -0,0 +1,179 @@
...
+//        TlaArithOper.plus,
+//        TlaArithOper.minus,
+//        TlaArithOper.mult,
+//        TlaArithOper.div,
+//        TlaArithOper.mod,
+//        TlaArithOper.exp,
+//        TlaArithOper.lt,
+//        TlaArithOper.gt,
+//        TlaArithOper.le,
+//        TlaArithOper.ge,
```
shonfeder, line 162: What's up with these commented out parts?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
```diff
@@ -0,0 +1,179 @@
+package at.forsyte.apalache.tla.lir.transformations.standard
```
shonfeder, line 1: It looks to me on first evaluation like pretty much all of the logic here is shared with, e.g., the [KeraLanguagePred](https://github.com/informalsystems/apalache/tree/unstable/tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/KeraLanguagePred.scala).

Would it make sense to try to abstract the core logic and share it between the different language preds?

IIUC, the core logic is mostly just concerned with traversing the AST. It almost looks to me like you could make `TlaEx` and instance of the [Foldable](https://eed3si9n.com/learning-scalaz/Foldable.html) type class, and then just use [`all`](https://javadoc.io/doc/org.scalaz/scalaz-core_2.12/latest/scalaz/Foldable.html) along with a predicate `isFooOper : TlaOper -> Boolean`, where `Foo` is defined for the particular instance of `LanguagePred`.

If something like this is feasible, I think you might be able to introduce this change while consolidating the logic of this family of classes, and maybe even yielding a net reduction in LOC. What do you think?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
```diff
@@ -0,0 +1,179 @@
...
+        // go inside the let definitions (similar to FlatLanguagePred)
+        def eachDefRec(ctx: Set[String], ds: List[TlaOperDecl]): PredResult = {
+          ds match {
+            case Nil =>
+              PredResultOk()
+
+            case head :: tail =>
+              isOkInContext(ctx, head.body) // check the first operator definition
+                .and(eachDefRec(ctx + head.name, tail)) // check the other operator definitions
+          }
```
shonfeder, line 109: Since even these comments are copied are from `KeraLanguagePred` (or one of it's ancestors or descendants), it does seem like some code sharing would be in order. If you don't like the `Foldable` suggestion, you could perhaps subclass these from a shared parent class?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
```diff
@@ -0,0 +1,179 @@
...
+      case e =>
+        PredResultFail(List((e.ID, e.toString)))
+    }
+  }
+}
+
+object ReTLALanguagePred {
+  private val singleton = new ReTLALanguagePred
+
+  protected val unaryOps: HashSet[TlaOper] =
```
shonfeder, line 138: I know this structure wasn't introduced in your code here, but do you know what is the motive for having these distinct `HashSets` rather than one `HashSet` (or, equivalently, a predicate `TlaOper -> Boolean`)?  Is it a kind of micro optimization? Or are these subsets used in some other way I'm missing?

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestReTLALangaugePred.scala
```diff
@@ -0,0 +1,159 @@
...
+  test("reTLA: logic expressions") {
+    expectOk(pred.isExprOk(or(bool(false), name("b"))))
+    expectOk(pred.isExprOk(and(bool(false), name("b"))))
+    expectOk(pred.isExprOk(equiv(bool(false), name("b"))))
+    expectOk(pred.isExprOk(impl(bool(false), name("b"))))
+    expectOk(pred.isExprOk(neql(bool(false), name("b"))))
+  }
+
+  /**
+   * **************************** the tests from TestFlatLanguagePred ********************************************
```
shonfeder, line 117: I don't think I understand this comment. Why are there tests for a different predicate in this test suite? Does that mean we're running the same tests twice?

shonfeder (changes requested):
I've got a couple questions! Note the caveat that it's my first time looking at these language predicate classes, so I may be missing some context.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
Kukovec: There was a discussion on whether the 1st version of reTLA should include support for integer arithmetic or not. It's no problem in SMT, but not as great for IVy. Since we'll very likely be adding these operators in the future (and since my 1st draft had them), I just commented them out and made a note, so we can go to v2 just by uncommenting.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
Kukovec: It's a bit tricky, because that would _almost_ work. The problem with shared parentage is that 1) the difference is in one method's `case` classes, and that method is 95% of the class functionality (you could have a shared parent class with all the aux methods though) and 2) Unfortunately, reTLA isn't 100% a restriction on the allowed operators, because of functions. Since SANY expands `[f EXCEPT ![a] = b]` to `[f EXCEPT ![<<a>>] = b]` we have to support tuples, but only in function argument positions and not anywhere else. So you can't have `TlaFunOper.tuple -> false` or `TlaFunOper.tuple -> true`, because they're _sometimes_ allowed.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
Kukovec: Yes, I suppose the helper method sharing makes sense.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
Kukovec: You know, I honestly hadn't thought about it. Looking at it again it does seem unnecessary, fold should work in all cases, even for `bindingOps`.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestReTLALangaugePred.scala
Kukovec: They're unique tests for `pred.isExprOk`, just that this part tests for the presence of LET-operators _like_ the tests in `TestFlatLanguagePred`. I copied the comment with the rest and didn't bother changing it, but the tests are different.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
shonfeder: I see. Thanks for considering.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
shonfeder: Oh I see. I didn't understand what `IntArith not in v1` means without this context.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReTLALanguagePred.scala
shonfeder: The helper sharing does indeed help a bit. Still more duplicated code here than I'd prefer, but don't want to block on that.

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestReTLALangaugePred.scala
shonfeder: The fixed comment helps clear this up for me. Thanks!

shonfeder (approved):

PR #1257, opened 2022-01-26
Apply scalafmt changes to entire codebase

shonfeder:
Now that we've moved off maven, perhaps we'll have better luck doing a reformat
of the entire codebase.

Note that this also adds a few rules, mostly of them motivated by the [Scala
style guide](https://docs.scala-lang.org/style/). See
a46b733c38eb8eac7ef076af0c2eb866b84240cf for details.

Once this configuration is approved, I'll coordinate with the team to ensure any
ongoing work is merged before this changeset is applied.


.scalafmt.conf
```diff
@@ -1,4 +1,7 @@
 version = "2.7.5"
 preset = Scala.js
-maxColumn = 120
-trailingCommas = always
+maxColumn = 100
```
gabrielamafra, line 3: Why this reduction from 120 to 100? I like big columns :stuck_out_tongue_closed_eyes:

shonfeder: You are in good company with Igor then! I thought I could the Scala Style Guide as a pretext for inching things towards my preference for narrower code :stuck_out_tongue_winking_eye: 

If you and/or others on the team would prefer it stay at 120, I'll revert this bit :)

shonfeder:
I'll need to rebase here.

.scalafmt.conf
shonfeder: I've put this back to 120 :)

.scalafmt.conf
```diff
@@ -2,4 +2,7 @@ version = "3.4.2"
 runner.dialect = scala212
 preset = Scala.js
 maxColumn = 120
-trailingCommas = always
+verticalMultiline.atDefnSite = true
+verticalMultiline.arityThreshold = 4
+trailingCommas = multiple
```
shonfeder, line 7: This way we only add trailing commas when there's more than 1 element in the sequence (and they are across multiple lines).

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/TransitionPassImpl.scala
```diff
@@ -20,8 +20,13 @@ import java.nio.file.Path
...
  */
-class TransitionPassImpl @Inject() (options: PassOptions, sourceStore: SourceStore, tracker: TransformationTracker,
-    changeListener: ChangeListener, incrementalRenaming: IncrementalRenaming, writerFactory: TlaWriterFactory,
+class TransitionPassImpl @Inject() (
+    options: PassOptions,
+    sourceStore: SourceStore,
+    tracker: TransformationTracker,
+    changeListener: ChangeListener,
+    incrementalRenaming: IncrementalRenaming,
+    writerFactory: TlaWriterFactory,
```
shonfeder, line 29: This is so much more readable for me! :heart_eyes_cat:

shonfeder:
The format check is failing in CI, but succeeding on my machine.

gabrielamafra (approved):

shonfeder:
Things I've tried to get the formatting working consistently on my machine and CI:

- updating sbt (using coursaire)
- installed the scalafmt CLI tool (via coursier)
  - Running this directly produced a different set of failures and changes than those produced by running via sbt :(
  - However, after I had the CLI tool installed, I tried running via sbt again, and it finally caught the two outstanding files :shrug:

shonfeder:
yeesh, but now that I've rebased to pull in the latest changes from master, the format changes are not being applied any more :facepalm:

shonfeder:
Ok, I was able to get the changes to take effect by running `scalafmt` from the CLI (not via sbt). So it seems like this is a feasible backup in case running via sbt is flaky again.

For the record, I still have some misgivings about merging in this fix, but the team is all in agreement that the costs of dealing with the unrelated formatting changes in PRs is too high, and they all agreed it was worth the risk.

PR #1258, opened 2022-01-26
Improve clarity of test output

shonfeder:
Igor hit a case earlier where the build of a test was failing but there was so
much noise in the output that it wasn't clear what was causing the failure. 

This pr adds some setup/teardown to tests that write noisly to stdout, and some
config flags to make the test output more concise and easier to read.

There are still some tests in `tla-io` dumping stuff to stdout, but afaict those tests seem to be written in such a way that I can't redirect stdout without breaking the tests...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -53,7 +53,10 @@ lazy val tlaModuleTestSettings = Seq(
 )
 
 lazy val testSettings = Seq(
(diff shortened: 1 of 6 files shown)


konnov (approved):
Nice! Thanks for taking care of that!

shonfeder:
For some reason the code coverage isn’t being generated now tho :/ 

I’ll have to trouble shoot that.

PR #1261, opened 2022-01-27
Fix a typo in packaging `Apalache.tla` in jar

shonfeder:
I was including the file in the fat jar as `Apalacha.tla` :facepalm: 

This PR also removes the part of the runner script that appends the location of the Apalache.tla modules into the path, since we don't need that and it adds more complexity to the end user artifact.


konnov (approved):
LGTM! I will merge it right away.

PR #1265, opened 2022-01-28
remove hint finder and lazy short-circuiting

konnov:
The first fruits of having code coverage in the CI.

`HintFinder` has zero coverage and it is only used in lazy translation of `/\`. This feature has not been used for years. Removed.


shonfeder:
Wonderful! :tada:

PR #1275, opened 2022-02-01
Refactor the annotation preprocessor

### LINKED ISSUE #954: Confusing error message on missing semicolon in type
I forgot a semicolon on a type annotation and received this error message:

`Typing input error: Unexpected annotation of action: @type`

Here is the faulty code:

```
VARIABLES 
    \* mapping from chain id to its data
    \* @type: Str -> CHAIN;
    chains,

    \* last action performed
    \* @type: ACTION
    action,

    \* string with the outcome of the last operation
    \* @type: Str;
    actionOutcome
```

IMO this error message was unclear because it did not say anything about the semicolon problem. However, it was good that it mentioned `action`.

konnov:
Closes #954. This PR introduces `CommentPreprocessor` that extracts potential annotations from comments. This should improve the messages about incorrect annotations. As a nice side effect, the comments free of annotations are now propagated in the annotation store as a `#freeText` annotation. They are printed by `PrettyWriter`. `CommentPreprocessor` happened to be much harder to write. It  needs careful review.


tla-io/src/main/scala/at/forsyte/apalache/io/annotations/AnnotationParserError.scala
```diff
@@ -2,6 +2,6 @@ package at.forsyte.apalache.io.annotations
 
 class AnnotationParserError(message: String) extends Exception(message) {
   override def toString: String = {
-    "Error parsing annotation: " + message
+    "Annotation error: " + message
```
shonfeder, line 5: Would it still be useful to signal that this is an error at the syntax level, rather than an error with the meaning of the annotation?

```suggestion
    "Annotation syntax error: " + message
```

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/AnnotationLexer.scala
```diff
@@ -61,37 +57,21 @@ object AnnotationLexer extends RegexParsers {
...
     """"[a-zA-Z0-9_~!@#\\$%^&*\-+=|(){}\[\],:;`'<>.?/ \t\r\f\n]*"""".r ^^ { name =>
-      STRING(removeLeadingCommentsAndControlChars(name.substring(1, name.length - 1)))
+      STRING(name.substring(1, name.length - 1))
     }
   }
 
   private def inline_string: Parser[INLINE_STRING] = {
-    """:[a-zA-Z0-9_~!#\\$%^&*\-+=|(){}\[\],:`'<>.?/ \t\r\f\n]*;""".r ^^ { name =>
-      INLINE_STRING(removeLeadingCommentsAndControlChars(name.substring(1, name.length - 1)))
+    """:[a-zA-Z0-9_~!#\\$%^&*\-+=|(){}\[\],:`'<>.?/ \t\r\f\n]*;""".r ^^ { text =>
```
shonfeder, line 65: This looks like the same regex as on line 59. Can we instead share a value?

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/AnnotationLexer.scala
```diff
@@ -61,37 +57,21 @@ object AnnotationLexer extends RegexParsers {
...
-   * a group of carriage returns, form feeds, line feeds, and tabs.</p>
-   *
-   * @param text text to preprocess
-   * @return text without leading "\*" and ASCII control characters
-   */
-  private def removeLeadingCommentsAndControlChars(text: String): String = {
-    text.replaceAll("""\n[ \t\r\f]*\\\*""", "").replaceAll("[\n\t\r\f]*", "")
+  private def unexpected_char: Parser[Nothing] = {
+    failure("Unexpected character. Missing ')' or ';'?")
   }
```
shonfeder, line 75: Could we return what the unexpected character is? I think that would help identify the source of the problem.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/CommentPreprocessor.scala
```diff
@@ -0,0 +1,258 @@
...
+ * <li>Exclude candidates for annotations in double comments, e.g., (* (* ... *) *).</li>
+ * <li>Preserve the text outside the annotations and double comments.
+ * </ul>
+ *
+ * <p>The main design principle of the preprocessor is that it should not fail on any input.
+ * The preprocessor should respect the structure of the comments, but it may return ill-formed
+ * annotations, if they are ill-formed in the original text. This preprocessor prunes the comment
+ * tokens, even in arguably valid cases (e.g., inside a string). Although it is usually not a good idea
+ * to write an ad-hoc lexer, it is not obvious to me, how to use a lexer generator here.
+ * However, if someone knows how to write this preprocessor with a lexer generator, it would be great!
```
shonfeder, line 18: Did you look at some existing lexer generators for Scala? If so, could you share a pointer?

Even if there is not a good lexer generator utility for Scala, I would think that it would be natural write the lexer using the same parser combinator library we use for parsing. Any particular reason why you didn't go that way?

shonfeder (commented):
One main preliminary question around lexing strategy.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/CommentPreprocessor.scala
konnov: I think the problem is different here. Lexers are turning strings into a stream of tokens. If some strings are not recognizable, a lexer would throw an error. I tried to use the standard lexer approach in AnnotationLexer. It was just too sensitive to the input stream. `CommentPreprocessor` identifies strings that look like annotations and delegates the actual parsing to `AnnotationLexer` and `AnnotationParser`. This approach is less fragile in the presence of arbitrary text.

shonfeder: I'm not sure which part of my question that is answering.. :) Could you explain the connections a bit?

konnov: The short answer is that a lexer generator would solve the problem that is different from our problem. A lexer recognizes good strings and rejects bad strings. A good string is converted into a sequence of tokens. What we need is a preprocessor, that is, a program that receives an arbitrary string, extracts only relevant pieces out of it and passes them to an actual parser. Perhaps, some lexers can do that too, but I have not seen any lexer doing that.

shonfeder: Just to note: This seems at odds with your comment in the code. In the comment you say you don't know how to user a lexer generator for this purpose, but here you are saying the problem is that they are not suitable in this case. Is it worth updating the comment to include these insights?

Why we can't deal with bad strings in the lexer via a fallback rule or using a `bad-token` token type? Both approaches are described here: https://stackoverflow.com/a/48896227/1187277

konnov: True. The comment is a bit outdated. I feel like it should be possible to use a generator for this problem, because I am virtually defining a finite automaton by hand. But since it is not the common case, using a generator for it might happen to be quite hard. I am not sure `bad-token` is exactly solving the problem, but I only had a quick glance.

konnov: It's a compromise between a framework that one has to learn and 300 LOC that are hard to read. `CommentPreprocessor` is hard to write and read.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/AnnotationLexer.scala
konnov: They look very similar, but they are different. The latter one does not admit `@` and `;`. I will add a comment.

shonfeder: Ah. Please add a comment noting the difference? Otherwise it'll be up to a human to carefully parse the regex and they are liable to make the same mistake I did.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/AnnotationLexer.scala
konnov: I tried to figure out how to do it with Scala parser combinators, and I failed. I can produce `failure` or `err`, but neither of them accepts a token.

shonfeder: This seems to work:

```suggestion
  private def unexpected_char: Parser[Nothing] = {
    (".".r ^^ { char =>
      s"Unexpected character '${char}'. Missing ')' or ';'?"
    }).flatMap(failure)
  }
```

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/CommentPreprocessor.scala
shonfeder: So my other question is why it wouldn't work to use the same combinator library for the preprocessor, shouldn't that help remove a lot of manual bookkeeping here? Did you find that to be infeasible?

konnov: Yes, `AnnotationLexer` is using Scala combinators. The previous version tried to parse arbitrary string with it, and it did not work well. I don't know how to a preprocessor with the combinators.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/AnnotationLexer.scala
konnov: Oh, that's a good one. Thanks a lot!

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/CommentPreprocessor.scala
shonfeder: Ok. I think we could probably work out a `Parser[String]` for the preprocessing, but we can always do that as a following. The rest of the code here looks fine to me.

shonfeder (approved):

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/parser/CommentPreprocessor.scala
konnov: I don't think that the problem is in picking the library. If you like to put use a lexer generator (or Scala combinators), try to write a state machine first. The edges of this state machine will be regexes. For normal languages, the structure would be something like `(regex1|regex2|...|regexN)`. However, this preprocessor has a non-trivial internal state and it has lots of branching. This is why I do not see how a standard tool will immediately help here.

PR #1279, opened 2022-02-02
Clean up output when printing version

### LINKED ISSUE #1278: Enable clean version output from CLI
Towards  #1277

benchexec expects to be able to get the version info of a tool. Currently there's all kinds of noise surrounding the version. Let's clean this up, so that it'll be easier for tools to ask Apalache version it is :)

shonfeder:
Closes #1278

- Reduces output noise by not printing the tool location etc. by default. (Tho
  this can be enabled via environment variable).
- No longer print the header and stats config when default clist commands are
  run
- No longer print the EXITCODE output when default clist commands are run
- Fixes a regression introduced by #1234, causing build info and version to
  no longer be available for CLI output
  - This is achieved by introducing the sbt-buildinfo plugin to our build
    config, which generates a scala file holding this data for us.

See individual commits for more details.


bin/apalache-mc
```diff
@@ -56,6 +56,14 @@ sigterm() {
 
 trap sigterm SIGTERM SIGINT
 
+if [[ "$APALACHE_PKG_INFO" = true ]]
```
shonfeder, line 59: I'm not sure if it make sense to document this functionality, and if so, where? CONTRIBUTING? Usage?

Basically, it seems to like something that's only useful when trying to debug a broken installation.

konnov: Good point. Shall we report this, if the `--debug` is passed?

konnov (approved):
Cool! We could also de-noise the output by not printing timestamps when apalache does not output to tty. Many utilities have this optional progress tracking, when the output is not redirected. But I have no idea how to do that.

konnov:
btw, the docker tests are broken

bin/apalache-mc
shonfeder: Ah yeah, that's a better idea. I'll change this to check for that flag. Thanks!

PR #1294, opened 2022-02-04
Implement `--output` for `typecheck`

### LINKED ISSUE #1284: [FEATURE] Add the flag `--output` to typecheck
Related to #1281, there is no simple way to produce json after the `typecheck` pass. It's possible to use `--write-intermediate=1` and check the produced intermediate file. However, this is too hard to use.

Kukovec:

Also fixes decoding of recursive functions (untracked bug).

Closes #1284

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,7 +15,12 @@
 * `version` command only prints the version, see #1279
 * tool and jar location no longer output by default, see #1279
 
+### Features
+
+* Added support for `--output` with `typecheck`, see #1284
+
 ### Bug fixes
 
 * Fixed JSON decoder failing on inputs with `"Untyped"`, see #1281
+* Fixed JSON decoder failing on inputs with `"FUN_REC_REF"` or `"FUN_REC_CTOR"`
 * Correctly resolve higher-order operators in names instances, see #1289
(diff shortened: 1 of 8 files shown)


gabrielamafra (approved):
LGTM

konnov:
Maybe we should reuse the spec from `tla-types/src/test/resources/MegaSpec1.tla`. It contains pretty much all variations of operators.

shonfeder (approved):

shonfeder:
@konnov iiuc, this PR is only introducing the output file flag. So I wouldn't think it's the right place to check a megaspec. Seems that would be better fit for a PR that was concerned with the parser or serializer itself, rather than just the plumbing of printing to a file.

Kukovec:
> @konnov iiuc, this PR is only introducing the output file flag. So I wouldn't think it's the right place to check a megaspec. Seems that would be better fit for a PR that was concerned with the parser or serializer itself, rather than just the plumbing of printing to a file.

I understood this as Igor wanting to run the integration test on `MegaSpec1`, as opposed to `Annotations`. Is that right @konnov?

konnov:
yes, that's right. We can also do it in a unit test

shonfeder:
Yeah, I'm suggesting that we probably don't want to run the integration test on the megaspec, esp not if we're just testing that output files are generated.

If we are testing that the JSON or TLA constructed is correct, then we'd want the megaspec I guess? But I wonder if that's not a better suited for unit tests or even PBT?

shonfeder:
My point being, iiuc, this PR is only introducing dumping some stuff to a file. It is not introducing the logic of how to parse a spec, type check it, then reconstruct TLA or JSON from the AST. But the mega spec seems targeted at the latter.

PR #1295, opened 2022-02-04
Documentation for the IF-THEN-ELSE idiom

gabrielamafra:
Hello :octocat: 

While reviewing https://github.com/informalsystems/mbt/pull/87, some `IF-THEN-ELSE` expressions were flagged as misusages, and inspired by that we are updating our Idiomatic TLA+ documentation to explain why and when is better to use boolean operators instead of `IF-THEN-ELSE`.

This is pretty much my understanding of the problem. If any of you have a different opinion or a more objective approach to explain the issue, please let me know.


docs/src/idiomatic/007if-then-else.md
```diff
@@ -0,0 +1,94 @@
+# Idiom 7: Use Boolean operators in actions, not `IF-THEN-ELSE`
+
+## Description
+
+TLA+ provides an `IF-THEN-ELSE` operator, and it can be pretty tempting to use it for flow control as it's done in procedural programming. However, TLA+ is about transitions over a state machine, and a transition-defining action declared with `IF-THEN-ELSE` can be more complex than 2 actions declared without it. Considering that any expression in the form `IF b THEN x ELSE y` with `x` and `y` being booleans (and most TLA+ expressions are booleans) can be rewritten as `(b /\ x) \/ (!b /\ y)`, there's a pattern we can apply to get rid of some potentially troubled `IF-THEN-ELSE` definitions.
```
Kukovec, line 5:
```suggestion
TLA+ provides an `IF-THEN-ELSE` operator, and it can be pretty tempting to use it for flow control, as it's done in procedural programming. However, TLA+ is about transitions over a state machine, and a transition-defining action declared with `IF-THEN-ELSE` can be more complex than 2 actions declared without it. Considering that any expression of the form `IF b THEN x ELSE y`, where `x` and `y` are Booleans, can be rewritten as `(b /\ x) \/ (~b /\ y)`, there's a pattern we can apply to get rid of some potentially troublesome `IF-THEN-ELSE` definitions.
```

docs/src/idiomatic/007if-then-else.md
```diff
@@ -0,0 +1,94 @@
...
+
+The `IF-THEN-ELSE` operator can be used either to define a value or to branch some action as a sort of flow control. Defining values with `IF-THEN-ELSE` is common practice and is similar to the use of `IF` expressions on declarative programming languages. However, flow control in TLA+ can be done naturally by behavior definition through actions, making the use of `IF-THEN-ELSE` for flow control unnecessary. This idiom aims to clarify different usages of `IF-THEN-ELSE` expressions keeping in mind the TLA+ essence of declaring actions to define sets of transitions.
```
Kukovec, line 7:
```suggestion
The `IF-THEN-ELSE` operator can be used either to define a value, or to branch some action as a sort of flow control. Defining values with `IF-THEN-ELSE` is common practice and is similar to the use of `IF` expressions in declarative programming languages. However, flow control in TLA+ can be done naturally by behavior definition through actions, making the use of `IF-THEN-ELSE` for flow control unnecessary. This idiom aims to clarify different usages of `IF-THEN-ELSE` expressions, keeping in mind the TLA+ essence of declaring actions to define transitions. 
```

docs/src/idiomatic/007if-then-else.md
```diff
@@ -0,0 +1,94 @@
...
+
+The `IF-THEN-ELSE` operator can be used either to define a value or to branch some action as a sort of flow control. Defining values with `IF-THEN-ELSE` is common practice and is similar to the use of `IF` expressions on declarative programming languages. However, flow control in TLA+ can be done naturally by behavior definition through actions, making the use of `IF-THEN-ELSE` for flow control unnecessary. This idiom aims to clarify different usages of `IF-THEN-ELSE` expressions keeping in mind the TLA+ essence of declaring actions to define sets of transitions. 
+
+## When to use `IF-THEN-ELSE`
+
+### When the result is not Boolean
+
+When the `IF-THEN-ELSE` expression doesn't evaluate to a boolean value, it cannot be rewritten using boolean operators, so this idiom doesn't apply.
```
Kukovec, line 13:
```suggestion
When the `IF-THEN-ELSE` expression doesn't evaluate to a boolean value, it cannot be rewritten using boolean operators, so this idiom doesn't apply. For example:
```

docs/src/idiomatic/007if-then-else.md
```diff
@@ -0,0 +1,94 @@
...
+
+## When (and how) *not* to use `IF-THEN-ELSE`
+
+Mixing `IF-THEN-ELSE` expressions with action operators introduces unnecessary branching to definitions that could be self-contained and look more like a transition definition. 
+
+```tla
+Withdraw(amount) == IF balance >= amount 
+                     THEN /\ balance' = balance - amount
+                          /\ response' = "SUCCESS"
+                     ELSE /\ UNCHANGED<balance>
```
Kukovec, line 44:
```suggestion
                     ELSE /\ UNCHANGED balance
```

docs/src/idiomatic/007if-then-else.md
```diff
@@ -0,0 +1,94 @@
...
+We could separate the two branches into their own actions with clarifying names and explicit conditions, and use a disjunction between the two actions instead of the `IF-THEN-ELSE` block:
+
+```tla
+WithdrawSuccess(amount) == /\ balance >= amount 
+                           /\ balance' = balance - amount
+                           /\ response' = "SUCCESS"
+                            
+WithdrawFailure(amount) == /\ balance < amount 
+                           /\ response' = "FAILURE"
+                           /\ UNCHANGED<balance>
```
Kukovec, line 57:
```suggestion
                           /\ UNCHANGED balance 
```

docs/src/idiomatic/007if-then-else.md
```diff
@@ -0,0 +1,94 @@
...
+```tla
+Next == \/ /\ C1()
+           /\ A1()
+        \/ /\ C2()
+           /\ A2()
+        \/ A3()
+           
+```
+
+This second definition can allow more behaviors than the first one (depending on whether `C1()` and `C2()` overlap), and therefore is safer in most cases.
```
Kukovec, line 90: This is a bit disingenuous, IMO. If you _wanted_ to define the behavior written in ITE, your disjunction would be _less_ readable, not more, and more prone to errors.
>This second definition can allow more behaviors ... and therefore is safer in most cases.

I disagree, the assumption that allowing more behaviors is safer is questionable.

Kukovec (approved):
I approve this doc, but I have to wonder, to what extent is this actually an issue?
In the presence of action-ITE, the transition finder, followed by the simplifier would prune the actual sliced transitions anyway:
`IF p THEN A ELSE B` slices into `IF p THEN FALSE ELSE B` and `IF p THEN A ELSE FALSE`, which simplify into `p /\ A` and `~p /\ B` already.

Also, if we wanted, we could fix this automatically for them (w/o introducing new operators) in preprocessing, that is, rewrite action-ITEs as disjunctions in exactly this way:
```tla
IF p 
THEN A
ELSE B
```
to 
```tla
\/ /\ p
   /\ A
\/ /\ ~p
   /\ B
```
since this transformation is purely syntactic (assuming level analysis is done beforehand).

docs/src/idiomatic/007if-then-else.md
gabrielamafra: I meant it in the strict sense, buy your example is actually a good one to be included in the "When to use" section

docs/src/idiomatic/007if-then-else.md
```diff
@@ -0,0 +1,94 @@
...
+                           /\ response' = "FAILURE"
+                           /\ UNCHANGED<balance> 
+                            
+Withdraw(amount) == WithdrawSuccess(amount) \/ WithdrawFailure(amount)
+```
+
+## Advantages
+
+- Each action declares fewer transitions, so it's easier to reason about it
+- A disjunction of actions is closer to a union of transition sets than an `IF-THEN-ELSE` expression
```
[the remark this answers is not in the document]
gabrielamafra: I mean the set of transitions derived from an action

shonfeder (approved):
LGTM, and I learned things :)

Look forward to learning more as Jure's comments are addressed!

docs/src/idiomatic/007if-then-else.md
gabrielamafra: > If you wanted to define the behavior written in ITE
This is a idiom, the aim is to help people decide what they want when they are not sure. I wrote this assuming a lot of people use ITE just because they are way more used to programming languages than logic. My point in this example is: if you have a bunch of independent "when the state is this, do that" , its better to use a disjunction, but most people will use ITE because that's how you do it in programming languages. 

As for allowing more behaviors being safer: yeah, that's kinda pretentious, I'll get rid of it.

konnov: I think Gabriela's point here is that the author of such a spec would not have thought about the case of `A1` and `A2` being simultaneously enabled. A disjunction makes it clear. I agree that it is often the case that people wants to specify a concurrent system via `IF-THEN-ELSE`, which is a purely sequential operator.

gabrielamafra:
> to what extent is this actually an issue? In the presence of action-ITE, the transition finder, followed by the simplifier would prune the actual sliced transitions anyway

@Kukovec I'm not even considering optimization for the solver here (I couldn't if I wanted), only spec readability and good practices to avoid human mistakes.

docs/src/idiomatic/007if-then-else.md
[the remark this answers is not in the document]
Kukovec: I know what you meant, but as a general rule, I'd recommend caution with terminology, because, especially with TLA+, it's easy for someone to misunderstand this and assume there is a TLA+ expression/construct that is a literal set.
We have a lot of documentation (and papers) on the treatment of transition systems, full of terminology (e.g. https://apalache.informal.systems/docs/tutorials/symbmc.html), and I'd like to advocate against diverging on terminology in the docs. 
This might come across as nitpicking, but Igor and I have spent many hours discussing terminology in docs, and you become sensitive to this kind of stuff.

konnov (approved):
Nice! I think the main message is like: Prefer disjunctions to IF-THEN-ELSE, if you don't know what you are doing :)

docs/src/idiomatic/007if-then-else.md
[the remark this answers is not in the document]
gabrielamafra: Not at all, I'll totally agree with consistent terminology, I just didn't know/recall the standard term here, sorry! I've quickly scrolled through the doc but I'm still not sure about the best term here. Would you write "set of transitions"? Or should I change the whole phrase to address transitions as a relation?

PR #1300, opened 2022-02-04
Fix polymorphic types when the type checker is called on its output

### LINKED ISSUE #1255: [BUG] Type variables clash with fresh variables when the type checker is applied twice
An interesting bug has popped up in #1253. The type checker sometimes fails when it is applied to its own results. This is caused by polymorphic signatures. Concretely, `ToEtcExpr` is using a pool of fresh variables that collides with the type variables that were introduced by `ToEtcExpr` itself in the first run. The reason for this bug is that `ToEtcExpr` was first written under the assumption that user-defined operators are not polymorphic (back in 2020).

This test was found by a unit test in `TestTypeCheckerTool`. In #1253, the failing test is isolated in the test: `"the tool consumes its output on TlcSpec1"`

Since the new rewiring approach in #1253 is using polymorphic operators, this bug appears. I could not reproduce it on individual runs of `typecheck`, so the type checker is relatively safe to use in the CLI.

konnov:
closes #1255.


Kukovec (approved):
LGTM

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
```diff
@@ -105,7 +105,7 @@ class TestTypeCheckerTool extends FunSuite with BeforeAndAfterEach with EasyMock
   }
 
   // fixing this test is scheduled in: https://github.com/informalsystems/apalache/issues/1255
```
konnov, line 107:
```suggestion
```

Issue #1286, opened 2022-02-03
[FEATURE] Configure passes order through a list


gabrielamafra:
We use Google Guice to define our passes order by injecting a pointer to the next pass to each of our passes. This makes it more complicated to read and mainly to update, since it requires a bunch of pointer reassignments even for a small change.

We should have a more declarative way to configure this order by just having a `Seq` of passes in the order they're supposed to be run.

PR #1309, opened 2022-02-07
Refactor how passes are chained

gabrielamafra:
Hello :octocat: 

This is a refactor for our passes definitions. Injecting each pass as a `nextPass` for another pass is being a pain to maintain since a simple order change requires multiple injection updates. With the changes here proposed, the passes are defined in a simple sequence of classes, which is manipulated entirely by the `PassChainExecutor`. The interface for `Pass` is always change to work more purely, without manipulating any weird state and instead receiving a `TlaModule` and (if successful) returning another. This way, we get rid of `TlaModuleMixin` completely and dependency verification and manipulation is done entirely at `PassChainExecutor`.

In order to allow the pass list to be defined differently for each execution module, I've added an extension from Google Guice's `AbstractModule`, allowing both dependency injection and passes configuration to be done by the same class (which is a `ToolModule`). The pass list cannot be injected directly with Guice because it mixes actual instances with classes that depend on injection to be instantiated. This approach keeps dependency injection as is, since all passes that use injection will keep using it; but stops using injection to define pass order.

Closes https://github.com/informalsystems/apalache/issues/1286


mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/PassChainExecutor.scala
```diff
@@ -14,37 +14,37 @@ import at.forsyte.apalache.tla.lir.MissingTransformationError
...
+      transformedModule match {
+        case None => return None
+        case Some(m) =>
+          val newModule = new TlaModule(m.name, m.declarations) with TlaModuleProperties
+          newModule.properties = module.properties ++ pass.transformations
+
+          module = newModule
+      }
+    }
+    Some(module)
```
Kukovec, line 48:
```suggestion
    passes.zipWithIndex.foldLeft(Option[TlaModule with TlaModuleProperties](new TlaModule("empty", Seq()) with TlaModuleProperties)){ case (moduleOpt, (pass, index)) =>
      // Raise error if the pass dependencies aren't satisfied
      if (!pass.dependencies.subsetOf(module.properties)) {
        val missing = pass.dependencies -- module.properties
        throw new MissingTransformationError(
            s"${pass.name} cannot run for a module without the properties: ${missing.mkString(", ")}", module)
      }
      for {
        module <- moduleOpt
        transformedModule <- exec(index, pass, module)
      } yield {
        val newModule = new TlaModule(transformedModule .name, transformedModule .declarations) with TlaModuleProperties
        newModule.properties = module.properties ++ pass.transformations
        newModule
      }
    }
```

Kukovec: You might need to tweak this a bit before committing, because I wrote it in the github editor, but the basic idea is that 
```
for {
  x <- optionX // optionX: Option[A]
  y <- f(x) // f: A -> Option[A]
} yield y // None if optionX or f(x) are None, Some(y) otherwise
```
is a neat way to handle fold + options.

Kukovec (approved):
LGTM

shonfeder (approved):
Really nice cleanup!

Issue #1285, opened 2022-02-03
[BUG] `apalache check` throws `rewriter error: OrRule is not applicable` with `--view`


rnbguy:
## Input specification

```
------------------------------ MODULE Indices ------------------------------
\* This is the model of Substrate Frame Indices from 
\* https://github.com/paritytech/substrate/tree/master/frame/indices 
\* at this commit: https://git.io/Ju5Jv
\*
\* An index is a short form of an address.
\* This module handles allocation of indices for a newly created accounts.
\*
\* This TLA+ model is created for demostrative purposes of MBT capabilities.
\* s2021 Andrey Kuprianov, Informal Systems

EXTENDS Integers, FiniteSets

CONSTANTS 
  \* @typeAlias: ADDR = Int;
  \* @typeAlias: INDEX = Int;
  \* @typeAlias: BALANCE = Int;
  \* @typeAlias: INDEX_ENTRY = [ addr: ADDR, deposit: Int, perm: Bool ];
  \* @typeAlias: INDICES = INDEX -> INDEX_ENTRY;
  \* @typeAlias: RESERVES = ADDR -> BALANCE;
  \* @typeAlias: ACTION = [type: Str, who: ADDR, new: ADDR, index: INDEX, freeze: Bool];
  \* How many users and indices are there? 
  \* @type: Int;
  NUM_USERS,
  \* @type: Int;
  NUM_INDICES

VARIABLES
 \* @type: INDICES;
  indices,
  \* @type: RESERVES;
  reserved,
  \* @type: ACTION;
  action,
  \* @type: Str;
  actionOutcome

Indices == 1..NUM_INDICES
Addresses == 0..NUM_USERS

\* A non-address
None == -1
\* A root address
Root == 0


\* Possible outcomes

\* Operation successful
OK == "OK"
\* The index was not already assigned.
NotAssigned == "NotAssigned"
\* The index is assigned to another account.
NotOwner == "NotOwner"
\* Only root can do this.
NotRoot == "NotRoot"
\* The index was not available.
InUse == "InUse"
\* The source and destination accounts are identical.
NotTransfer == "NotTransfer"
\* The index is permanent and may not be freed/changed.
Permanent == "Permanent"


ClaimOutcome(who, index) ==
  IF indices[index].who /= None
    THEN InUse
  ELSE OK

\* @type: (ADDR, INDEX) => Bool;
Claim(who, index) ==
  /\ indices' = [indices EXCEPT ![index] = 
[...]

PR #1311, opened 2022-02-08
Bugfix for #1285

gabrielamafra:
Hello :octocat: 

The newly implemented simplifications for `Or` expressions allow a generic expression to arise from simplification
since expressions like `\/ x` and `x \/ true` are now simplified to `x`. However, `OrRule` calls a simplification over an `Or` expression and expects the result to be either a value (`TRUE` or `FALSE`) or another `Or` expression. This expectation breaks with the newly introduced simplification.

 Fixes https://github.com/informalsystems/apalache/issues/1285

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,6 @@
          * Some bug fix, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
+### Bug fixes
+
+* Fix a bug where some simplified `Or` expressions were not expected by the rewriting rules, see #1285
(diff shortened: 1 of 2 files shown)


Kukovec (approved):
LGTM

shonfeder (approved):

gabrielamafra:
> Cool! As an afterthought, can the same bug pop up in AndRule or other rules?

@konnov It did for `AndRule`, it was caught by the integration tests and I fixed back in the simplification PR. Yesterday I looked for this pattern in other rules and couldn't find it anywhere else, so we should be safe.

PR #1317, opened 2022-02-10
isolate calls to stats collection

konnov:
This PR isolates the settings for TLC/Toolbox/Apalache statistics collection, so the tests do not call home.


shonfeder (approved):
Nice solution!

konnov:
The docker tests are failing. How shall we deal with the docker setup?

shonfeder:
Ah! I think we can safely bind mount the user's `~/.tlaplus` from their host system into the container. I'll give that a shot here.

shonfeder:
lol, looks like i also found a workaround to approve PR's that you've contributed to :laugh

PR #1319, opened 2022-02-10
Implement SetAsFun

### LINKED ISSUE #1226: [FEATURE] Introduce an operator that converts a set of pairs to a function, use it in counterexamples
As we saw from [this discussion](https://github.com/informalsystems/apalache/discussions/1195#discussioncomment-1955079), it is often tempting to use the TLC operators `:>` and `@@` to iteratively construct a function. These operators applied in sequence produce a large set of constraints. In some sense, we encourage the users to use these operators, as we output functions with `:>` and `@@` in the counterexamples.

Alternatively, we could introduce an operator `KeyValuesAsFun` that converts a set of pairs to a function. Such an operator could be treated more efficiently. We would apply this operator in the counterexamples and, by doing so, show the users the more efficient way of constructing functions by enumeration.

@Kukovec what do you think about this?

konnov:
Closes #1226. This PR implements translation for the operator `SetAsFun` as a more efficient and more readable replacement for TLC's `:>` and `@@`. See the integration test for the examples. This operator surfaced a bug in `DomainRule`, which was never triggered before. It's also fixed here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,26 +6,14 @@ target/
 .bsp
 .bloop
 
+# Generated if running integration tests locally
+_apalache-out/
+
 # Auto generated file for metals
-
 metals.sbt
-log*.smt
-test/tla/testSpec\.smt2
-test/tla/counterexample*
-
-tla-assignments/src/test/resources/assignmentTest1.smt2
-tla-bmcmt/counterexample*
-tla-bmcmt/profile-rules.txt
-
-*.err
-*.out
-*.log
 
 # Local envars for direnv
 /.local-envrc
-
-# Ignore directories where test artifacts are stored
-x/
 
 # Ignore artifacts created when previewing the gh-pages site
 /Gemfile.lock
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -242,6 +242,14 @@ https://scalameta.org/metals/docs/editors/emacs.html
 
 ## Testing
 
+### Build and run Apalache from source
+
+Execute apalache from the unpackaged source, ensuring any updates are built, run
+
+```sh
+make run <arguments>
+```
+
 ### Unit tests
 
 Run the units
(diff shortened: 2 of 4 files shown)


src/tla/Apalache.tla
```diff
@@ -1,103 +1,109 @@
...
+ * two pairs <<x, y>> and <<u, v>> such that x = u and y /= v,
+ * then F is not uniquely defined. We use CHOOSE to resolve this ambiguity.
+ * Apalache implements a more efficient encoding of this operator
+ * than the default one.
+ *
+ * @type: Set(<<a, b>>) => (a -> b);
+ *)
+SetAsFun(S) ==
+    LET Dom == { x: <<x, y>> \in S }
+        Rng == { y: <<x, y>> \in S }
```
Kukovec, line 52: Not a criticism of your code, but wow, this syntax sugar is horrible. The fact that you're allowed to pattern-match on pairs, when `S` could be any set is outright strange.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetAsFunRule.scala
```diff
@@ -0,0 +1,80 @@
...
+          case tt =>
+            throw new RewriterException("Unexpected type in SetAsFunRule: " + tt, state.ex)
+        }
+
+      case _ =>
+        throw new RewriterException("%s is not applicable".format(getClass.getSimpleName), state.ex)
+    }
+  }
+
+  private def translateRelation(pairsSet: ArenaCell, state: SymbState): SymbState = {
```
Kukovec, line 47: Please add a comment on the purpose of this method and the expected output (maybe a small example).

Kukovec (approved):
Looks OK, but I would like to see a test (expecting an exception) for the case where `setAsFun` is called on a set of non-pairs.

src/tla/Apalache.tla
konnov: Yes, but it works in TLA+

Kukovec: I know, that's the weird part.

PR #1322, opened 2022-02-10
Cleaned up `.gitignore`

Kukovec:

I took the liberty of reviewing the `.gitignore` file, and pruned file formats that shouldn't be appearing in the directories at all anymore. Removing them from `.gitignore` also lets us see if these files still get produced, which would be a bug at this point (e.g. the missing `rm` in the tests file).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,26 +6,14 @@ target/
 .bsp
 .bloop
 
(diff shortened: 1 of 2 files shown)


shonfeder (approved):
Awesome. Thanks!

PR #1327, opened 2022-02-10
Use SetAsFun instead of :> and @@

### LINKED ISSUE #1305: [BUG] Support for `--view` is broken by TLC rewiring
We have migrated `:>` and `@@` to a custom TLC module in #1253. Hence, `:>` and `@@` should not appear in the code produced by Apalache. However, they do appear in counterexamples. As a result, generation of multiple counterexamples with `--view` got broken, as there is no SMT support for `:>` and `@@` anymore. Thus, we should expedite #1226 to fix this and support the MBT team.

konnov:
This PR replaces all uses of `:>` and `@@` with `SetAsFun`. This approach simplifies our code and the tests. It also improves readability of the counterexamples.

Closes #1305. 

### DIFF
--- a/docs/src/lang/booleans.md
+++ b/docs/src/lang/booleans.md
@@ -350,7 +350,8 @@ FALSE <=> 1      \* runtime error in TLC, type error in Apalache
 
 **Example in Python:**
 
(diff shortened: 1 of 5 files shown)


Kukovec (approved):
LGTM

PR #1331, opened 2022-02-11
Add a short Twitter widget for narrow screens

thpani:
The current Twitter widget has unbounded height. On narrow screens
(e.g., phones) one needs to scroll past the Twitter feed to get to the
main page contents.

This adds a second (shorter) Twitter widget below the main content, that
is only displayed on narrow screens. The other widget is then hidden.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -38,13 +38,19 @@
         </ul>
         {% endif %}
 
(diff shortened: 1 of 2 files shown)


shonfeder (approved):
Awesome. Thanks for this fix!

PR #1344, opened 2022-02-15
Add property-based tests to TestKeramelizer

### LINKED ISSUE #1272: [FEATURE] Add property-based tests to TestKeramelizer
[Keramelizer](https://github.com/informalsystems/apalache/blob/unstable/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Keramelizer.scala) transforms expressions into a simplified fragment KerA+. [KeraLanguagePred](https://github.com/informalsystems/apalache/blob/unstable/tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/KeraLanguagePred.scala) checks, whether an expression fits into this fragment. Looks like a natural fit for property-based testing. We already have some tests in [TestKeramelizer](https://github.com/informalsystems/apalache/blob/unstable/tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestKeramelizer.scala). Extending it with PBT, would be great.

thpani:
Closes #1272

This adds property-based tests for Keramelizer.

The new `KeramelizerInputLanguagePred` recognizes TLA+ expressions in the language fragment accepted as input by Keramelizer. The predicate is then used to generate TLA+ expressions via `IrGenerators`, passes them through Keramelizer and applies `KeraLanguagePred` to check that the result is in KerA+.

Also, some adjustments for operator arity (`\X` and record sets) are introduced to ensure construction of legal expressions.


konnov (approved):
Nice!

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestKeramelizer.scala
```diff
@@ -291,4 +296,26 @@ class TestKeramelizer extends FunSuite with BeforeAndAfterEach {
...
+
+    val prop = forAll(gens.genTlaEx(ops)(gens.emptyContext)) { ex =>
+      inputPred.isExprOk(ex) == PredResultOk() ==>
+        (try {
+          val keramelized = keramelizer(ex)
+          val inKera = outputPred.isExprOk(keramelized)
+          inKera shouldBe PredResultOk() withClue s"when keramelizing $ex to $keramelized"
+          true
+        } catch {
+          case _: MalformedTlaError => true
```
shonfeder, line 316: It's worth noting that this I think this has the potential to make the test very ineffective, since its possible the rng could roll out 1000 malformed Tla expression, meaning this test would always pass but never once exercise the intended functionality!

shonfeder: Oh! Perhaps I'm mistaken here. I see we have the conditional checking whether the `isExprOk` is true as an antecedent, and ScalaCheck will warn us if too few conditions satisfy the antecedent.

But why do we need the `catch` if we are first ensuring that `ex` is OK?

shonfeder (approved):

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestKeramelizer.scala
shonfeder: On idea to move this into an antecedent check that will warn on too many discarded results: You might use https://www.scala-lang.org/api/2.12.4/scala/util/control/Exception$.html to reify the exception into an `Option`. You could then use either a `suchThat` generator to only take `Some` or just a boolean `isSome` check in the antecedent.

thpani: Addressed in 41d4476 by adding checks to the language predicate that prevent throwing of the exception. I believe this is the best place for them.

@shonfeder PTAL?

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestKeramelizer.scala
```diff
@@ -291,4 +296,26 @@ class TestKeramelizer extends FunSuite with BeforeAndAfterEach {
     assert(expected == output)
   }
 
+  test("simplifies TLA+ expressions to KerA+") {
+    // Generator for PBT
+    val gens = new IrGenerators { override val maxArgs: Int = 3 }
+    // Generated operators
+    val ops = gens.simpleOperators ++ gens.logicOperators ++ gens.arithOperators ++ gens.setOperators
```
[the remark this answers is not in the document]
thpani: Do we have a generator that also applies function operators?
Otherwise this should probably go into a separate PR ;)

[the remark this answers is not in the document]
konnov: nope. It probably needs a separate one

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/KeramelizerInputLanguagePred.scala
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Test whether expressions fit into the input fragment of [[Keramelizer]], i.e., whether an expression
+ *   - is flattened (see [[FlatLanguagePred]]),
+ *   - operators `exists`, `forall`, `chooseBounded`, `filter`, `apply` take a [[NameEx]] as first argument, and
+ *   - `Seq(_)` is prohibited.
+ *
+ * To get a better idea of the accepted fragment, check [[TestKeramelizerInputLanguagePred]].
+ */
+class KeramelizerInputLanguagePred extends ContextualLanguagePred {
```
shonfeder, line 18: Something I don't quite understand: if these preconditions are all necessary for valid input for Keramelizer, then why would this check but put into the `apply` method of the `Keramelizer` itself?

shonfeder (commented):
One lingering question on my end!

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/KeramelizerInputLanguagePred.scala
thpani: I assume that nothing violating these constraints reaches `Keramelizer` in our toolchain. @Kukovec @konnov can you confirm?
As per my offline discussion with @konnov, I assumed `KeraLanguagePred` as ground truth, i.e., that if `Keramelizer(x)` isn't recognized by `KeraLanguagePred`, then `x` should not be legal input to Keramelizer.

shonfeder: So I'm still slightly unclear on the reasoning for why we'd need this extra predicate to guard inputs for the tests. Seems to me that all the logic for validation should either already be in `KreaLanguagePred`, or else we'd have some automated way of reducing expressions to valid valid Keramelizer input, and the generated expressions could be run through that?

But this may just be to lack of context on my part. So don't want to block on this, esp. since Igor has already approved :)  Please merge whenever you're satisfied on this point.

shonfeder (approved):
I'm still unclear on the question I asked, but the changes around the PBT implication look good to me!

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/KeramelizerInputLanguagePred.scala
thpani: My understanding is that these expressions wouldn't go through SANY, but that's a heavy dependency for a unit test.
I will merge this as-is to avoid lagging too far behind `unstable`.

PR #1345, opened 2022-02-15
Added optimization case for quantifiers with constant bodies

### LINKED ISSUE #1343: [BUG] Optimization pass removes assignments, causes exception
## Description

Optimization transforms `FALSE /\ x' := v` into `FALSE`, but doesn't propagate this to make the entire transition `FALSE`. Consequently, the transition causes an exception when `x'` is accessed:

```
Step 0: picking a transition out of 1 transition(s)               I@13:30:54.415
This error may show up when CONSTANTS are not initialized.        E@13:30:54.440
Check the manual: https://apalache.informal.systems/docs/apalache/parameters.html E@13:30:54.440
Input error (see the manual): SubstRule: Variable x' is not assigned a value E@13:30:54.442
```

## Input specification

```tla
-------------------- MODULE test ---------------------------

EXTENDS Integers, Apalache

VARIABLES   
  \* @type: Int;
  x

Init ==  x = 0

Next == 
  /\ \E t \in {0}:
    /\ FALSE
    /\ x' = 1
  /\ x' < 3

```

## Expected behavior

No exception.

### LINKED ISSUE #1347: [FEATURE] Introduce more simplifications for expressions with quantifiers
Implement the following simplifications:
```tla
\E x \in {}: P <=> FALSE
\A x \in {}: P <=> TRUE
\E x \in S: TRUE <=> S /= {}
\A x \in S: FALSE <=> S = {}
```

Kukovec:

Closes #1343 
Closes #1347


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -204,6 +205,11 @@ abstract class ConstSimplifierBase {
     // IF x THEN y ELSE y = y
     case OperEx(TlaControlOper.ifThenElse, _, thenEx, elseEx) if (thenEx == elseEx) => thenEx
 
+    // \E x \in S: FALSE = FALSE
+    case OperEx(TlaBoolOper.exists, _, _, ValEx(TlaBool(false))) => falseEx
+    // \A x \in S: TRUE = TRUE
```
gabrielamafra, line 210: Isn't `\A x \in S: FALSE = FALSE` also a simplification required to propagate this and avoid the bug?

Kukovec:
Seems like this is hitting #[928](https://github.com/informalsystems/apalache/issues/928) again, on hold while we we figure out if #928 was properly resolved.

konnov:
> Seems like this is hitting #[928](https://github.com/informalsystems/apalache/issues/928) again, on hold while we we figure out if #928 was properly resolved.

Oh, the optimizer is simplifying the expression in a way that makes it impossible to figure out the type. I think the only reasonable approach here is to optimize quantification over the empty set too.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
konnov: Perhaps, one more expression to simplify then? I always have to recall the definition of conjunction and disjunction over the empty sets :)

gabrielamafra (approved):

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
Kukovec: Yeah, I'll add some more equivalencies for the quantifiers.

Kukovec: See #1347

gabrielamafra (approved):

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestConstSimplifier.scala
```diff
@@ -405,21 +408,74 @@ class TestConstSimplifier extends FunSuite with BeforeAndAfterEach with Checkers
...
+      val existsFalse = tla.exists(n_x, n_S, tla.bool(false)) as BoolT1()
+      val existsFalseResult = simplifier.simplify(existsFalse)
+      // \A x \in S: TRUE
+      val forallTrue = tla.forall(n_x, n_S, tla.bool(true)) as BoolT1()
+      val forallTrueResult = simplifier.simplify(forallTrue)
+
+      existsEmptyResult shouldBe (tla.bool(false) as BoolT1()) withClue s"when simplifying ${existsEmpty.toString}"
+      forallEmptyResult shouldBe (tla.bool(true) as BoolT1()) withClue s"when simplifying ${forallEmpty.toString}"
+      forallTrueResult shouldBe (tla.bool(true) as BoolT1()) withClue s"when simplifying ${forallTrue.toString}"
+      existsFalseResult shouldBe (tla.bool(false) as BoolT1()) withClue s"when simplifying ${existsFalse.toString}"
```
thpani, line 435: Just a small one to put the checks in the order of the definitions above.

```suggestion
      existsFalseResult shouldBe (tla.bool(false) as BoolT1()) withClue s"when simplifying ${existsFalse.toString}"
      forallTrueResult shouldBe (tla.bool(true) as BoolT1()) withClue s"when simplifying ${forallTrue.toString}"
```

thpani (commented):
LGTM

Issue #1248, opened 2022-01-24
Migrate packagine to sbt-native-package and release pipeline to sbt-release


shonfeder:
#1234 brings us much closer to the prevailing Scala project setup. However, the way we are currently building and packaging our program is still pretty idiosyncratic (afaict). This prevents straightforward usage of nice plugins like `sbt-native-package` and `sbt-release`.

This issue tracks proposed work to reconfigure our packaging to conform with standard structures, so that we can make use of this community work and drop our various bespoke scripts for packaging and release.

PR #1357, opened 2022-02-17
Use sbt-native-package to prepare distribution

shonfeder:
Contributes to #1248

While working on https://github.com/informalsystems/apalache/milestone/50 I hit
some snags due to our idiosyncratic packaging. The sbt-native-plugin has helped
streamline the process of producing a more standard package for distribution.

With these changes in place, it should be a relatively simple followup to
produce `.dmg` bundles for MacOS and `.deb` for debian systems. It also solves
the problem blocking me on the benchmarking work.

The work to change the packaging is quite minor here. Most of the changes result
from my attempt to ensure that these changes won't impact devs and user's
current expectations about how to invoke the program.

## Key changes

- Add the `sbt-native-plugin` package and simple configurations to ensure the
  resulting package includes our license and fat jar.
- Simplify the runner script at `bin/apalache-mc` and move it to
  `src/universal/bin`, as per guidelines of the `sbt-native-plugin` docs.
- Define the `apalacheCurrentPackage` task that runs the packager, and then
  unpacks and symlinks to it for the convenience of people who are used to being
  able to run the packaged build using `bin/apalache-mc`
- Put a symlink at `bin/apalache-mc` to the location in the `target`
  where the runner script ends up during packaging.
- Add `dist` make target, to create all the archives we need for releases
- Take advantage of the simpler packaging to simplify our docker file generation
  and the release script.

-------


konnov:
Just curious, would symbolic links work on Windows? Probably, not an issue, since our tooling is not tuned towards windows anyways.

shonfeder:
@konnov Right, I don't think the symbolic links would work on windows per se, but of course neither would the bash script that we are replacing!  But I *would* expect both to work on WSL :)

shonfeder:
The symlinking is only to preserve existing behavior for current users who are used to running out of the build dir.

build.sbt
```diff
@@ -211,11 +217,26 @@ lazy val root = (project in file("."))
...
+        s"tar zxvf ${pkg} --overwrite -C ${target_dir}" ! log
+        log.info(s"Symlinking ${current_pkg} -> ${unzipped}")
+        s"ln -sfn ${unzipped} ${current_pkg}" ! log
+        file(unzipped)
+      },
   )
 
-// Specify and build the docker file
-enablePlugins(DockerPlugin)
-
```
shonfeder, line 218: I just moved this to consolidate the plugin enabling.

bin/apalache-mc
```diff
@@ -1,71 +0,0 @@
...
-
-# a trap for SIGINT and SIGTERM
-sigterm() {
-    echo "Premature termination requested. Killing apalache-mc (pid $child)" 1>&2
-    trap - SIGTERM SIGINT # unregister to avoid recursion below
-    # send termination to the child process
-    kill SIGTERM $child 2>/dev/null
-}
-
-trap sigterm SIGTERM SIGINT
```
shonfeder, line 57: @konnov note that, in my suggested changes I've removed this trap, and we're just running the jar int he foreground now. In my local testing, control-c worked to kill Apalache without any trouble, which seems to be the only reason this trap was used before?

Please let me know if I'm overlooking something!

docs/src/apalache/installation/source.md
```diff
@@ -12,10 +12,20 @@
...
+   move this wherever you'd like, and ensure that the `<dist-package-location>/bin` directory
+   is added to your `PATH`.
+
+## Running from within the Apalache repo
+
+If you prefer to keep the built package inside of the Apalache source
+repository, you have three options after running `make`:
+
+- Add the `./bin` directory in the source repository to your `PATH`, which will
+  make `apalche-mc` available.
```
gabrielamafra, line 25:
```suggestion
  make `apalache-mc` available.
```

docs/src/apalache/installation/source.md
```diff
@@ -12,10 +12,20 @@
...
+- Install [direnv][] and run `direnv allow`, which will put `apalche-mc` in your
```
gabrielamafra, line 26:
```suggestion
- Install [direnv][] and run `direnv allow`, which will put `apalache-mc` in your
```

gabrielamafra (approved):

shonfeder:
Thanks for the fixes, @gabrielamafra , and sorry for the sloppy typos!

src/universal/bin/apalache-mc
```diff
@@ -0,0 +1,57 @@
...
+
+APALACHE_JAR=${APALACHE_JAR:-"$DIR/../lib/apalache.jar"}
+JVM_ARGS=${JVM_ARGS:-""}
+TLA_PATH=${TLA_PATH:-""}
+
+if ! test -f "$APALACHE_JAR"
+then
+    echo "ERROR: No file found at ${APALACHE_JAR}"
+    echo "       Ensure you have run 'make package' and are running the script from the"
+    echo "       distribution pacakge, or else set APALACHE_JAR to point to your custom"
```
konnov, line 31:
```suggestion
    echo "       distribution package, or else set APALACHE_JAR to point to your custom"
```

src/universal/bin/apalache-mc
```diff
@@ -0,0 +1,57 @@
...
+done
+DIR=$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )
+
+APALACHE_JAR=${APALACHE_JAR:-"$DIR/../lib/apalache.jar"}
+JVM_ARGS=${JVM_ARGS:-""}
+TLA_PATH=${TLA_PATH:-""}
+
+if ! test -f "$APALACHE_JAR"
+then
+    echo "ERROR: No file found at ${APALACHE_JAR}"
```
konnov, line 29: I think we had a test for the case when the directory contained jars for two different version.

konnov:
> @konnov note that, in my suggested changes I've removed this trap, and we're just running the jar int he foreground now. In my local testing, control-c worked to kill Apalache without any trouble, which seems to be the only reason this trap was used before?

Would it work when `apalache-mc` is run from another script? I think the trap was intended for this case.

src/universal/bin/apalache-mc
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+#
+# Run the APALACHE model checker
+#
+# NOTE: The primary intended use for this script is to be copied into the
+# packaged produced.
```
konnov, line 6:
```suggestion
# produced package.
```

konnov (approved):
Looks good!

src/universal/bin/apalache-mc
shonfeder: That can’t happen anymore ! :)

konnov: this is awesome 👍

src/universal/bin/apalache-mc
shonfeder: That is clearer. Thanks!

shonfeder:
> Would it work when `apalache-mc` is run from another script? I think the trap was intended for this case.

Thanks for the context! I've researched up what looks like a slightly cleaner solution in [d268aa2](https://github.com/informalsystems/apalache/pull/1357/commits/d268aa26a895c1d6db5bcbd19facaeccc299c4f2)

I've tested wrapping the call in another script, and it exited as expected.

PR #1380, opened 2022-02-18
Catch `MalformedTlaError` in property-based tests of Keramelizer

### LINKED ISSUE #1364: [BUG] "simplifies TLA+ expressions to KerA+ fails" on some inputs
We are getting a failure on certain inputs to the `keramelizer` PBT test added in #1344: 
 
```
[info] - simplifies TLA+ expressions to KerA+ *** FAILED *** (1 minute, 8 seconds)
[info]   MalformedTlaError was thrown during property evaluation.
[info]     Message: Expected a set, found: Typed(Bool)
[info]     Occurred when passed generated values (
[info]       arg0 = SUBSET (LET ulSu ≜ -BOOLEAN IN Nat // LET iS(mYYcngxD) ≜ LET x0(zz, jHfcC) ≜ ( ≠ -1) ∖ Int IN x0 IN iS)
[info]     )
```

See for the failure https://github.com/informalsystems/apalache/runs/5240089131?check_suite_focus=true#step:5:183

This is possibly a sign that the KerA+ simplifier PBT test are paying off! :)

thpani:
Closes #1364

This is a workaround for when `IrGenerators` constructs an ill-typed expressions that fails in Keramelizer, as in #1364.


shonfeder (approved):
lgtm.

Issue #1368, opened 2022-02-18
[BUG] make package fails


konnov:
```
make package
...
[info] Unpacking package /Users/igor/devl/apalache/target/universal/apalache-0.20.4-SNAPSHOT.tgz
[error] tar: Option --overwrite is not supported
[error] Usage:
[error]   List:    tar -tf <archive-filename>
[error]   Extract: tar -xf <archive-filename>
[error]   Create:  tar -cf <archive-filename> [filenames...]
[error]   Help:    tar --help
[info] Symlinking /Users/igor/devl/apalache/target/universal/current-pkg -> /Users/igor/devl/apalache/target/universal/apalache-0.20.4-SNAPSHOT
[success] Total time: 59 s, completed Feb 18, 2022 9:06:30 AM
```

Platform: MacOS 12.1

The worst of it is that the command does not fail.

konnov:
Related to that, I guess we have seriously restricted our building setup to Linux/MacOS. We could probably build the project on Windows with maven (never checked that).

konnov:
If I remove the option `--overwrite`, `make package` produces another error:

```
[info] Unpacking package /Users/igor/devl/apalache/target/universal/apalache-0.20.4-SNAPSHOT.tgz
[error] x apalache-0.20.4-SNAPSHOT/
[error] x apalache-0.20.4-SNAPSHOT/LICENSE
[error] x apalache-0.20.4-SNAPSHOT/bin/
[error] x apalache-0.20.4-SNAPSHOT/lib/
[error] x apalache-0.20.4-SNAPSHOT/lib/apalache.jar
[error] x apalache-0.20.4-SNAPSHOT/bin/apalache-mc
[info] Symlinking /Users/igor/devl/apalache/target/universal/current-pkg -> /Users/igor/devl/apalache/target/universal/apalache-0.20.4-SNAPSHOT
[success] Total time: 27 s, completed Feb 18, 2022 9:14:38 AM
```

thpani:
For now, you can probably just undo #1357 / 22517bdc27be22f60f9c5fc220a032945364a0a3 ?

shonfeder:
@konnov FYI, sbt will prefix output to stderr with `error` log, but that doesn't mean the command has failed. See https://stackoverflow.com/questions/14504572/sbt-suppressing-logging-prefix-in-stdout 

Looks like I might be able to fix this to avoid future confusion.

shonfeder:
[...]

PR #1382, opened 2022-02-20
Don't send tar output to logger

shonfeder:
The logger prefixes outputs on stderr with `error`, but on macos, tar
outputs logging messages to stderr, which has produced confusion (see
https://github.com/informalsystems/apalache/issues/1368#issuecomment-1044121136).

This change just lets the output go to std{out,err} as appropriate,
rather than thru the logger, so no misleading prefixes are attached.

-------


shonfeder:
MacOS builds don't have the misleading `error` prefix on the tar output now:

image

konnov (approved):
That's great! Thanks for fixing this one.

PR #1386, opened 2022-02-21
Extend `IrGenerators` with function operators

### LINKED ISSUE #1355: [FEATURE] Add function operators to IrGenerators
Extend `IrGenerators` to also apply function operators. This would be useful in property-based tests such as the ones added in #1344. Original request: https://github.com/informalsystems/apalache/pull/1344#discussion_r807066075

- [x] Extend `IrGenerators` to also apply function operators
- [x] Use them for PBT in `TestKeramelizer`
- [x] Identify other places these can be used

thpani:
Closes #1355

* Extends `IrGenerators` with function operators.
* Adds them to PBT specs where applicable.

### DIFF
--- a/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
+++ b/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
@@ -81,6 +81,12 @@ trait IrGenerators extends TlaType1Gen {
       TlaSetOper.setminus, TlaSetOper.subseteq, TlaSetOper.times, TlaSetOper.union)
 
   /**
+   * Function operators (`TlaFunOper._`)
+   */
+  val functionOperators = List(TlaFunOper.enum, TlaFunOper.tuple, TlaFunOper.app, TlaFunOper.domain, TlaFunOper.funDef,
+      TlaFunOper.recFunDef, TlaFunOper.recFunRef, TlaFunOper.except)
+
+  /**
    * The list of action operators.
    */
   val actionOperators = List(TlaActionOper.prime, TlaActionOper.enabled, TlaActionOper.stutter, TlaActionOper.nostutter,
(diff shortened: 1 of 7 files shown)


tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
```diff
@@ -80,6 +80,12 @@ trait IrGenerators extends TlaType1Gen {
       TlaSetOper.in, TlaSetOper.notin, TlaSetOper.map, TlaSetOper.powerset, TlaSetOper.recSet, TlaSetOper.seqSet,
       TlaSetOper.setminus, TlaSetOper.subseteq, TlaSetOper.times, TlaSetOper.union)
 
+  /**
+   * Function operators (`TlaFunOper._`)
+   */
+  val functionOperators = List(TlaFunOper.enum, TlaFunOper.tuple, TlaFunOper.app, TlaFunOper.domain, TlaFunOper.funDef,
+      TlaFunOper.recFunDef, TlaFunOper.recFunRef, TlaFunOper.except)
```
Kukovec, line 87: I think there's very limited value in supporting `recFunDef` and `recFunRef`. Not only are they the next to go, they will also generate countless malformed expressions, due to the restrictions on their proper use.

Kukovec (approved):
LGTM

tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
konnov: I am not certain yet about letting the recursive functions to go (in contrast to recursive operators). I have seen some value in them recently. Though they can always be replaced with `CHOOSE` over a function set.

konnov (approved):
Looks good!

tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
thpani: Tested this on the Keramelizer PBT with `minSuccessful = 10,000` and it didn't produce a malformed expression, so I think it's fine in this regard. If `reckFun{Def,Ref}` become obsolete, we can clean them out then.

PR #1388, opened 2022-02-21
Apply additional operators in PBT

thpani:
Should we apply additional operators in these PBTs?

* `TestUJsonToTla`: function, action, temporal
* `TestConstSimplifier`: logic, function


gabrielamafra (approved):
Seems good to me. We can reevaluate in case some test breaks. Good catch :smile:

PR #1399, opened 2022-02-23
add an interesting test

konnov:
This is an interesting example that appeared in the [TLA+ user group](https://groups.google.com/g/tlaplus/c/cD1bhMcHxqA).  Adding it as an integration test.

### DIFF
--- a/test/tla/TestHash2.tla
+++ b/test/tla/TestHash2.tla
@@ -1,0 +1,32 @@
+--------------- MODULE TestHash2 --------------
+\* Run Apalache to check the invariant for all combinations of hashes
+\*
+\* apalache-mc check --cinit=ConstInit --inv=Inv TestHash2.tla
+EXTENDS Integers
+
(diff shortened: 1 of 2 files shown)


thpani (approved):
Cool, I saw this in the tlaplus group and then immediately after your PR 😄 

LGTM

Kukovec (approved):

shonfeder:
Very nice! Should the TLA spec or the integration test perhaps have a link to the google group discussion so we can trace the context?

test/tla/TestHash2.tla
```diff
@@ -0,0 +1,32 @@
+--------------- MODULE TestHash2 --------------
+\* Run Apalache to check the invariant for all combinations of hashes
+\*
```
konnov, line 3: `
\* This is an interesting example that appeared in the TLA+ user group:
\*(https://groups.google.com/g/tlaplus/c/cD1bhMcHxqA).
`

PR #1400, opened 2022-02-23
Exclude operators causing malformed expressions from PBT

thpani:
`IncrementalRenaming` checks argument arity of `map`, `funDef`, `recFunDef`.
Including these causes lots of `MalformedTlaError`s. I'm not sure about the cause, but this seems to slow down PBT significantly.
For now, exclude these operators in `TestIncrementalRenamingScalacheck`.

### DIFF
--- a/tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestIncrementalRenamingScalacheck.scala
+++ b/tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestIncrementalRenamingScalacheck.scala
@@ -22,21 +22,22 @@ class TestIncrementalRenamingScalacheck extends AnyFunSuite with BeforeAndAfter 
     }
 
     val prop = {
-      val ops = gens.simpleOperators ++ gens.arithOperators ++ gens.setOperators ++ gens.functionOperators
+      val ops =
+        (gens.simpleOperators ++ gens.arithOperators ++ gens.setOperators ++ gens.functionOperators).filterNot(op =>
+          // map, funDef, recFunDef cause lots of malformed expressions, exclude them
+          op match {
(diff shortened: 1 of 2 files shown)


thpani:
If you have an idea why this slows down the PBT so much, LMK!

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestIncrementalRenamingScalacheck.scala
```diff
@@ -22,21 +22,22 @@ class TestIncrementalRenamingScalacheck extends AnyFunSuite with BeforeAndAfter
...
-          val multiset = output.operDeclarations.foldLeft(emptyCounters) { case (set, d) =>
-            set ++ collectDefined(d.body)
-          }
-          multiset.find { case (_, i) => i > 1 } match {
-            case None            => passed
-            case Some((name, n)) => falsified :| s"Name $name is defined $n times"
-          }
-        } catch {
-          // the generator has produced a malformed expression
-          case _: MalformedTlaError => passed
```
shonfeder, line 39: Ah nice! So removing these operators makes it impossible for malformed expressions to get through here?

shonfeder (approved):
Thanks for the fix!

tlair/src/test/scala/at/forsyte/apalache/tla/lir/transformations/standard/TestIncrementalRenamingScalacheck.scala
thpani: At least not from `IncrementalRenaming`, and I haven't seen any others in testing.
It's a bit hard to tell, since Scala doesn't require thrown exceptions to be declared.

Issue #1064, opened 2021-10-27
Fail CI on compiler errors


shonfeder:
Compiler warnings create noise during the build, they can undermine the confidence of contributors or users, and sometimes they signal serious problems that can lead to runtime errors (e.g., for unmatched patterns). So, to improve the hygiene and reliability of our development, it would be a good idea to start failing the build for any unsilenced warnings.

If we decide that there are some cases where we don't want to, e.g., handle all possible match cases, we should explicitly silence the warnings for those cases. Otherwise, we should clean up all that cruft. 

- [x] #1526
  - [x] #1471
  - [x] #1472
- [x] #1473
- [x] Add linting CI task that runs with `set scalacOptions ++= Seq("-Ywarn-unused", "-Xlint:unused", "-Xfatal-warnings")` to fail when new warnings are introduced.

thpani:
@shonfeder Looks like all current compiler warnings are squashed :tada:

I'm not quite sure what's left to do for the last point:

> Add linting CI task that runs with set scalacOptions ++= Seq("-Ywarn-unused", "-Xlint:unused", "-Xfatal-warnings") to fail when new warnings are introduced

Should we just enable `-Xfatal-warnings` on all builds? Feel free to take this issue from here on if you want.

shonfeder:
I'll take this up! 

I think you're right that we should enable fatal warnings on all builds. And in fact iiuc the linting cannot enable fatal-warnings, or else scalafix won't be able to work it's magic.

PR #1404, opened 2022-02-24
Add scalafix and remove unused imports

### LINKED ISSUE #1389: [FEATURE] Automate imports management
It would be nice to use some tools to help us manage our imports, including:
- Removing unused imports
- Keeping a consistent order between imports

@thpani suggested [scalafix](https://scalacenter.github.io/scalafix/docs/rules/RemoveUnused.html)

shonfeder:
Closes #1389

Please individual commits for details on the changes. The only "substantive"
changes are:

- b0dc5d91b03790f2a5de99409cb19e301184323a just adds the scalafix plugin
- 240d414c46323b9573c2665c0568e63b6889a907 configures the plugin for our builds
- 7ab898880c51669eee19f6867f89c74899aa497e adds a scalafmt rule to sort imports

The other changes are just automated application of these configurations (or
documentation).

With this changes in place, our IDEs should now warn us on unused variables and
imports, and we'll get warnings during compilation too.

Running `sbt "scalafix RemoveUnused"` should remove the unused terms.

(Note that on my first run I had to make 3 manual fixes, in places with nested
`@` bindings resulted in syntactically incorrect rewriting by scalafix. So, be
on guard.)

to @gabrielamafra and @thpani for motivating and directing this
improvement.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Markdown files used for integration tests
 TEST_MD_FILES := $(wildcard test/tla/*.md)
 
-.PHONY: all apalache apalache-jar compile build-quick test integration clean deps promote docker dist
+.PHONY: all apalache apalache-jar compile build-quick test integration clean deps promote docker dist fmt-fix-unused
 
 all: apalache
 
@@ -56,6 +56,9 @@ fmt-check:
 
 fmt-fix:
 	sbt scalafmtAll scalafmtSbt
+
+fmt-fix-unused:
+	sbt "scalafix RemoveUnused"
 
 clean:
 	sbt clean
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,7 @@ fmt-check:
 		  exit 1 )
 
 fmt-fix:
-	sbt scalafmtAll scalafmtSbt
-
-fmt-fix-unused:
-	sbt "scalafix RemoveUnused"
+	sbt "scalafix RemoveUnused" scalafmtAll scalafmtSbt
 
 clean:
 	sbt clean
--- a/build.sbt
+++ b/build.sbt
@@ -43,10 +43,21 @@ ThisBuild / libraryDependencies ++= Seq(
     TestDeps.scalatestplusScalacheck,
 )
 
+////////////////////////////
+// Linting and formatting //
+////////////////////////////
+
+// scalafmt
 // TODO: Remove if we decide we are happy with allways reformatting all
 // Only check/fix against (tracked) files that have changed relative to the trunk
 // ThisBuild / scalafmtFilter := "diff-ref=origin/unstable"
 ThisBuild / scalafmtPrintDiff := true
+
+// scalafix
+// https://scalacenter.github.io/scalafix/docs/rules/RemoveUnused.html
+ThisBuild / scalacOptions ++= Seq("-Ywarn-unused")
+ThisBuild / semanticdbEnabled := true
+ThisBuild / semanticdbVersion := scalafixSemanticdb.revision
 
 ///////////////////////////////
 // Test configuration //
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExprGradeAnalysis.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExprGradeAnalysis.scala
@@ -76,8 +76,7 @@ class ExprGradeAnalysis @Inject() (val store: ExprGradeStoreImpl) {
    */
   def refineOrInExpr(expr: TlaEx): TlaEx = {
     expr match {
-      case OperEx(TlaBoolOper.or, args @ _*) =>
-        args.map(refineOrInExpr)
+      case OperEx(TlaBoolOper.or, _*) =>
         store.get(expr.ID) match {
           case Some(ExprGrade.Constant) | Some(ExprGrade.StateFree) | Some(ExprGrade.StateBound) =>
             expr // keep it
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExprGradeAnalysis.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExprGradeAnalysis.scala
@@ -1,7 +1,7 @@
 package at.forsyte.apalache.tla.bmcmt.analyses
 
 import at.forsyte.apalache.tla.lir._
-import at.forsyte.apalache.tla.lir.oper.{ApalacheOper, TlaActionOper, TlaBoolOper, TlaTempOper}
+import at.forsyte.apalache.tla.lir.oper.{TlaActionOper, TlaBoolOper, TlaTempOper}
 import at.forsyte.apalache.tla.lir.UntypedPredefs._
 import com.google.inject.Inject
 
@@ -77,7 +77,7 @@ class ExprGradeAnalysis @Inject() (val store: ExprGradeStoreImpl) {
   def refineOrInExpr(expr: TlaEx): TlaEx = {
     expr match {
       case OperEx(TlaBoolOper.or, args @ _*) =>
-        val newArgs = args.map(refineOrInExpr)
+        args.map(refineOrInExpr)
         store.get(expr.ID) match {
           case Some(ExprGrade.Constant) | Some(ExprGrade.StateFree) | Some(ExprGrade.StateBound) =>
             expr // keep it
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/AssignmentRule.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/AssignmentRule.scala
@@ -16,7 +16,7 @@ import at.forsyte.apalache.tla.lir.{NameEx, OperEx}
  *   Igor Konnov
  */
 class AssignmentRule(rewriter: SymbStateRewriter) extends RewritingRule {
-  private val pickRule = new CherryPick(rewriter)
+  new CherryPick(rewriter)
 
   override def isApplicable(state: SymbState): Boolean = {
     def isUnbound(name: String) =
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/AssignmentRule.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/AssignmentRule.scala
@@ -1,7 +1,6 @@
 package at.forsyte.apalache.tla.bmcmt.rules
 
 import at.forsyte.apalache.tla.bmcmt._
-import at.forsyte.apalache.tla.bmcmt.rules.aux.CherryPick
 import at.forsyte.apalache.tla.lir.oper.{ApalacheOper, TlaActionOper}
 import at.forsyte.apalache.tla.lir.{NameEx, OperEx}
 
@@ -16,7 +15,6 @@ import at.forsyte.apalache.tla.lir.{NameEx, OperEx}
  *   Igor Konnov
  */
 class AssignmentRule(rewriter: SymbStateRewriter) extends RewritingRule {
-  new CherryPick(rewriter)
 
   override def isApplicable(state: SymbState): Boolean = {
     def isUnbound(name: String) =
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
@@ -5,7 +5,7 @@ import at.forsyte.apalache.tla.bmcmt.rules.aux.{CherryPick, DefaultValueFactory,
 import at.forsyte.apalache.tla.lir.convenience.tla
 import at.forsyte.apalache.tla.lir.TypedPredefs._
 import at.forsyte.apalache.tla.lir.oper.TlaOper
-import at.forsyte.apalache.tla.lir.{BoolT1, OperEx, SetT1, TlaEx, TypingException}
+import at.forsyte.apalache.tla.lir.OperEx
 
 /**
  * <p>Rewriting rule for CHOOSE. Similar to TLC, we implement a non-deterministic choice. It is not hard to add the
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
@@ -2,11 +2,10 @@ package at.forsyte.apalache.tla.bmcmt.rules
 
 import at.forsyte.apalache.tla.bmcmt._
 import at.forsyte.apalache.tla.bmcmt.rules.aux.{CherryPick, DefaultValueFactory, OracleHelper}
-import at.forsyte.apalache.tla.bmcmt.types.FinSetT
 import at.forsyte.apalache.tla.lir.convenience.tla
 import at.forsyte.apalache.tla.lir.TypedPredefs._
 import at.forsyte.apalache.tla.lir.oper.TlaOper
-import at.forsyte.apalache.tla.lir.{BoolT1, OperEx, SetT1, TlaEx, TypingException, UID}
+import at.forsyte.apalache.tla.lir.{BoolT1, OperEx, SetT1, TlaEx, TypingException}
 
 /**
  * <p>Rewriting rule for CHOOSE. Similar to TLC, we implement a non-deterministic choice. It is not hard to add the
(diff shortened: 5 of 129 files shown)


tla-types/src/main/scala/at/forsyte/apalache/tla/types/TypeReduction.scala
```diff
@@ -34,15 +34,15 @@ class TypeReduction(private val gen: SmtVarGenerator) {
...
+      ts.length
       val recIdx = gen.getFreshInt
       val constraints = ts.flatMap { case KvPair(k, v) =>
         val ReductionResult(t, phi) = apply(v, m)
         hasField(recIdx, k, t) +: phi
       }
       ReductionResult(rec(recIdx), constraints)
     case SparseTupT(ts @ _*) =>
-      val l = ts.length
+      ts.length
```
shonfeder, line 45: I guess these should be removed, unless there's a logic error by us not using these values.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -457,15 +457,15 @@ class ToEtcExpr(annotationStore: AnnotationStore, aliasSubstitution: ConstSubsti
...
-        etcExpr.typeErrorExplanation = (expectedTypes: List[TlaType1], actualTypes: List[TlaType1]) =>
+        etcExpr.typeErrorExplanation = (_: List[TlaType1], _: List[TlaType1]) =>
           Some(s"Cannot apply $fun to the argument $arg in $ex.")
         etcExpr
 
       case OperEx(TlaFunOper.domain, fun) =>
         // DOMAIN f
         val a = varPool.fresh
         val b = varPool.fresh
-        val c = varPool.fresh
```
shonfeder, line 468: Can be removed.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/AnnotationParser.scala
```diff
@@ -91,7 +91,7 @@ class AnnotationParser extends Parsers {
 }
 
 object AnnotationParser {
-  private val parser: AnnotationParser = new AnnotationParser
+  new AnnotationParser
```
shonfeder, line 94: rm

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -374,8 +374,8 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext {
    *   the name of the new function (declared in SMT)
    */
   def declareCellFun(cellName: String, argType: CellT, resultType: CellT): Unit = {
-    val domSig = argType.signature
-    val resSig = resultType.signature
+    argType.signature
+    resultType.signature
```
shonfeder, line 378: These can be removed, but that seems a bit suspicious too. I guess we should carefully consider this  code to ensure there's not a logic error?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetCupRule.scala
```diff
@@ -40,8 +40,7 @@ class SetCupRule(rewriter: SymbStateRewriter) extends RewritingRule {
         // introduce a new set
         val newType = types.unify(leftSetCell.cellType, rightSetCell.cellType)
         if (newType.isEmpty) {
-          val msg =
-            s"Failed to unify types ${leftSetCell.cellType} and ${rightSetCell.cellType} when rewriting ${state.ex}"
+          s"Failed to unify types ${leftSetCell.cellType} and ${rightSetCell.cellType} when rewriting ${state.ex}"
           throw new TypingException(s"Failed to unify types ${leftSetCell.cellType}"
                 + " and ${rightSetCell.cellType} when rewriting ${state.ex}", state.ex.ID)
         }
```
shonfeder, line 46: rm

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/NegRule.scala
```diff
@@ -14,7 +14,7 @@ import at.forsyte.apalache.tla.lir.{OperEx, TlaEx}
  *   Igor Konnov
  */
 class NegRule(rewriter: SymbStateRewriter) extends RewritingRule {
-  private val substRule = new SubstRule(rewriter)
+  new SubstRule(rewriter)
```
shonfeder, line 17: rm (unless there's some state updates happening in the rewriter?)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/IntArithRule.scala
```diff
@@ -17,7 +17,7 @@ import at.forsyte.apalache.tla.lir.UntypedPredefs._
  */
 class IntArithRule(rewriter: SymbStateRewriter) extends RewritingRule {
   private val intConstRule: IntConstRule = new IntConstRule(rewriter)
-  private val simplifier = new ConstSimplifierForSmt()
```
shonfeder, line 20: rm

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/IntCmpRule.scala
```diff
@@ -15,7 +15,7 @@ import at.forsyte.apalache.tla.lir.UntypedPredefs._
  *   Igor Konnov
  */
 class IntCmpRule(rewriter: SymbStateRewriter) extends RewritingRule {
-  private val simplifier = new ConstSimplifierForSmt()
```
shonfeder, line 18: rm

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/FunExceptRule.scala
```diff
@@ -103,7 +103,7 @@ class FunExceptRule(rewriter: SymbStateRewriter) extends RewritingRule {
     }
 
     // compute all updated cells in case we are dealing with a function over non-basic indices
-    val updatedCells = relationCells.map(eachRelationPair)
+    relationCells.map(eachRelationPair)
```
shonfeder, line 106: check whether there are side effects we need here or if this can be removed.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
```diff
@@ -75,74 +73,5 @@ class ChooseRule(rewriter: SymbStateRewriter) extends RewritingRule {
...
-    val nextState = rewriter.rewriteUntilDone(state.setRex(set))
-    val setCell = nextState.asCell
-    val isFinSet = PartialFunction.cond(setCell.cellType) { case FinSetT(_) => true }
-    if (isFinSet) {
-      happyChooseFromFinite(nextState, setCell, varName, pred)
-    } else {
-      happyChooseFromOther(nextState, setCell, varName, pred)
-    }
-  }
-
```
shonfeder, line 93: Is this something we want to keep around, @konnov? I think we could remove the `private` qualifier if so.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
```diff
@@ -49,7 +47,7 @@ class ChooseRule(rewriter: SymbStateRewriter) extends RewritingRule {
           defaultValueFactory.makeUpValue(nextState, setCell)
         } else {
           val elems = nextState.arena.getHas(setCell)
-          val nelems = elems.size
+          elems.size
```
shonfeder, line 50: rm

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
```diff
@@ -36,7 +35,6 @@ class ChooseRule(rewriter: SymbStateRewriter) extends RewritingRule {
       case OperEx(TlaOper.chooseBounded, varName, set, pred) =>
         // This is a general encoding, handling both happy and unhappy scenarios,
         // that is, when CHOOSE is defined on its arguments and not, respectively.
-        def solverAssert = rewriter.solverContext.assertGroundExpr _
```
shonfeder, line 39: This is curious. Will need to check what the assert does.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/AssignmentRule.scala
```diff
@@ -16,7 +16,7 @@ import at.forsyte.apalache.tla.lir.{NameEx, OperEx}
  *   Igor Konnov
  */
 class AssignmentRule(rewriter: SymbStateRewriter) extends RewritingRule {
-  private val pickRule = new CherryPick(rewriter)
+  new CherryPick(rewriter)
```
shonfeder, line 19: rm?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExprGradeAnalysis.scala
```diff
@@ -77,7 +77,7 @@ class ExprGradeAnalysis @Inject() (val store: ExprGradeStoreImpl) {
   def refineOrInExpr(expr: TlaEx): TlaEx = {
     expr match {
       case OperEx(TlaBoolOper.or, args @ _*) =>
-        val newArgs = args.map(refineOrInExpr)
+        args.map(refineOrInExpr)
```
shonfeder, line 80: rm

shonfeder (commented):
Some quite interesting unused values and private methods here! I'll have to make a second pass to remove stuff that was being stored into tmp variables that were removed. I've marked those in some comments.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
shonfeder: Oh, I see. This was just giving a short naming to the lambda here, and at some point we stopped calling the function.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
```diff
@@ -466,9 +466,11 @@ class ToEtcExpr(annotationStore: AnnotationStore, aliasSubstitution: ConstSubsti
         val a = varPool.fresh
         val b = varPool.fresh
         val c = varPool.fresh
+        // The possible types to which which DOMAIN can be be applied,
+        // and the corresponding type of the domain when so applied:
         val funType = OperT1(Seq(FunT1(a, b)), SetT1(a)) // (a -> b) => Set(a)
         val seqType =
-          OperT1(Seq(SeqT1(a)), SetT1(IntT1())) // Seq(c) => Set(Int)
+          OperT1(Seq(SeqT1(c)), SetT1(IntT1())) // Seq(c) => Set(Int)
         val recType = OperT1(Seq(RecT1()), SetT1(StrT1())) // [] => Set(Str)
```
shonfeder, line 474: This is actually a small logic fix, that was caught by the unused removal trying to eliminate the `c`. I don't think it could have caused problems in our type checking, since only one of *either* the `funType` or the `seqType` could be relevant for any application of the domain operator, and `b` is always fresh in any case, but nice to have it cleaned up just in case.

shonfeder: There’s a failing unit test due to this change. I’ll fix it tomorrow.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
konnov: Yes, we should remove this unused definition.

konnov: and the one below too.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/FunExceptRule.scala
konnov: Yes, `eachRelationPair` has side effects -- it is submitting queries to the solver. It would be better to have `forEach` here.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala
konnov: Yes, it was supposed to be a fix, but the actual variable was not substituted.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/ChooseRule.scala
konnov: Yes, it was just a shorthand

konnov (approved):
Looks good! Most of the changes are syntactic, except for the ones you have already flagged.

Makefile
```diff
@@ -57,6 +57,9 @@ fmt-check:
 fmt-fix:
 	sbt scalafmtAll scalafmtSbt
 
+fmt-fix-unused:
+	sbt "scalafix RemoveUnused"
```
gabrielamafra, line 61: Would it be safe to run this under `fmt-fix` as well? I mean fmt-fix: `sbt scalafmtAll scalafmtSbt && sbt "scalafix RemoveUnused"`. I think there's a better chance we stick to using this if its part of the process we are already used to do.

gabrielamafra (approved):
This is awesome Shon, thank you for making this happen :heart:

Makefile
shonfeder: That's a good point. I did try putting it in the `fmt-fix` target at first, but it seemed to me resulted in a significant slowdown (by a factor of 3 or 4), due to the need to compile all the code to identify unused identifiers. I was worried that we might find that too costly.

Maybe I can take temperature on the tradeoff in today's meeting.

shonfeder: In the meeting we all agreed that we'd like to check for unused values and imports in CI. The feelings on whether this should always be included in the format fix were a little less clear, but I think people were generally in favor of it.

shonfeder: So, afaict, there's no straightforward way to fail CI when unused values/imports are found other then by enabling `-Yfatal-warnings`. I think we *should* do this, but as part of #1064, the prerequisite for which is fixing all of the compiler warnings. 

I have run a couple micro benchmarks, and for a no-op build, the cost of adding unused removal is in the order of 5 to 20 extra seconds on top of `fmt-fix`. That's pretty slow, but I think our current workflows are that this isn't run more than once or twice per PR anyhow. Hopefully we all have "format on save" and linting in our IDEs set up to catch most things before they make it to CI?

So I've added remove unused to the make target for now. We can always remove in followup if it proves too onerous.

shonfeder:
I reported the syntactically incorrect rewrites we hit with nested pattern binders here:  [#1555](https://github.com/scalacenter/scalafix/issues/1555)

Makefile
```diff
@@ -55,7 +55,7 @@ fmt-check:
 		  exit 1 )
 
 fmt-fix:
-	sbt scalafmtAll scalafmtSbt
+	sbt "scalafix RemoveUnused" scalafmtAll scalafmtSbt
```
bjaglin, line 58: did you consider running scalafix on test sources? `scalafixAll` can help with that (just like `scalafmtAll`)

build.sbt
```diff
@@ -43,11 +43,22 @@ ThisBuild / libraryDependencies ++= Seq(
...
+// scalafmt
 // TODO: Remove if we decide we are happy with allways reformatting all
 // Only check/fix against (tracked) files that have changed relative to the trunk
 // ThisBuild / scalafmtFilter := "diff-ref=origin/unstable"
 ThisBuild / scalafmtPrintDiff := true
 
+// scalafix
+// https://scalacenter.github.io/scalafix/docs/rules/RemoveUnused.html
+ThisBuild / scalacOptions ++= Seq("-Ywarn-unused")
+ThisBuild / semanticdbEnabled := true
```
bjaglin, line 59: I see scalafix is wired through a specific make target, so you might want to use [`scalafixEnable`](https://scalacenter.github.io/scalafix/docs/users/installation.html#enable-semanticdb-for-current-shell-session) as a first step of that command to avoid the overhead of semanticdb for everyone upon compilation.

Makefile
shonfeder: We should do that! I'll make this change. Thanks very much for taking a look here, @bjaglin! You're tips are very helpful :pray:

PR #1406, opened 2022-02-24
Use type information to simplify membership tests

### LINKED ISSUE #723: [FEATURE] Use types to simplify `TypeOK`
Blocked by [#401 Precise type checking for records](https://github.com/informalsystems/apalache/issues/401)

We should use type information to simplify membership tests such as `x \in BOOLEAN`, `x \in Int`, `x \in STRING`, `x \in Seq(...)`. By knowing the types, we could write a preprocessing pass that rewrites the above tests to `TRUE`, if the types are matching the tests.

- [x] `n \in Nat` (#1406)
- [x] `i \in Int, b \in BOOLEAN, \r \in Real` (#1406)
- [x] `seq \in Seq(_)` (#1406)
- [x] `fun \in [Int -> Int]` (#1443)
- [x] `<<{{1}}>> \in Seq(SUBSET Int)` or `{<<1>>} \in SUBSET (Seq(Int))` (#1443)
- [x] `tup \in S1 \X ... \X SN` (#1445)
- [x] records (#1453, blocked by #401 – maybe alternative solution in #1629)

thpani:
Addresses #723

Uses type information to simplify membership tests such as `x \in BOOLEAN`, `x \in Int`, `x \in STRING`, `x \in Seq(...)` when we know that `x` has the appropriate type.


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -0,0 +1,51 @@
...
+      trueVal
+    case OperEx(TlaSetOper.in, name, OperEx(TlaSetOper.seqSet, ValEx(TlaStrSet)))
+        if name.typeTag == Typed(SeqT1(StrT1())) =>
+      trueVal
+    case OperEx(TlaSetOper.in, name, OperEx(TlaSetOper.seqSet, ValEx(TlaIntSet)))
+        if name.typeTag == Typed(SeqT1(IntT1())) =>
+      trueVal
+    case OperEx(TlaSetOper.in, name, OperEx(TlaSetOper.seqSet, ValEx(TlaRealSet)))
+        if name.typeTag == Typed(SeqT1(RealT1())) =>
+      trueVal
```
Kukovec, line 44: I feel like there's some unnecessary repetition here. You could write something along the lines of just:
```scala
    case OperEx(TlaSetOper.in, name, set) if isMatch(name.typeTag.asTlaType1(), set) => trueEx
    case ex => ex
```
where
```scala
def typeOfElem(ps: TlaPredefSet): TlaType1 = ...
def isMatch(nameType, set) = set match {
  case ValEx(ps: TlaPredefSet) => nameType == typeOfElem(ps)
  case OperEx(TlaSetOper.seqSet, ValEx(ps: PredefSet)) => nameType == SeqT1(typeOfElem(ps))
  case _ => false 
}
```

Kukovec (changes requested):

tlair/src/main/scala/at/forsyte/apalache/tla/lir/Builder.scala
```diff
@@ -148,6 +148,14 @@ class Builder {
    */
   def booleanSet(): BuilderVal = BuilderVal(TlaBoolSet)
 
+  /**
+   * The set STRING of all strings.
+   *
+   * @return
+   *   the value expression that corresponds to STRING.
+   */
+  def stringSet(): BuilderVal = BuilderVal(TlaStrSet)
```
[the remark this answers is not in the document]
thpani: Yeah, I was unsure about this... I can change it if that's preferred.
Are we fine with some heterogeneity  in the tests (builder/non-builder expressions) then?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
thpani: I prefer my solution; it's more verbose, but also more readable.

Kukovec (approved):
I'll sign off, since this was the only blocking issue I had, but I'd like to hear @konnov and @shonfeder 's thoughts on https://github.com/informalsystems/apalache/pull/1406#discussion_r813921011.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
shonfeder: I'm in favor of @Kukovec's suggestion. I'm not worried about LOC here if it impacts readability, but having to manually repeat the same logic `n` times introduces `n` chances for errors to creep in. @Kukovec's suggestion is also much better for maintainability, as it makes it possible to add new operators under `in`, or change the names of operators or types, without needing to update the case analysis here. 

A compromise might be something like:

```suggestion
    case OperEx(TlaSetOper.in, name, ValEx(valEx)) if name.typeTag.asTlaType1() == typeOfElem(valEx) => trueVal
    case OperEx(TlaSetOper.in, name, OperEx(TlaSetOper.seqSet, ValEx(valEx))) if name.typeTag.asTlaType1 == SeqT1(typeOfElem(valEx) =>
```

But either way, I think manually having to type out the correlation of terms with their types is illdvised when an automatic solution is readily available.

thpani: @thpani @Kukovec You both assume that code is more often written/edited than read; in my experience this is almost never true.

But Jure's comment inspired me to do [aac5a6f](https://github.com/informalsystems/apalache/pull/1406/commits/aac5a6fd3d84b2bf2f56833045f3f09151a47e8c), which is similar to Shon's suggestion.

shonfeder (approved):
LGTM!

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
shonfeder: I don't think our motives here are primarily about privileging writing over reading. I also don't agree, personally, that having 15 lines of the same logic, with small variations, is more readable than having two lines that clearly communicate the abstract pattern. But readability is a pretty subjective, and what is readable varies from person to person, and in the same person depending on their familiarity and experiences.

However, the difference between doing stuff manually and having clean, automated abstractions seems pretty objective I think. And, IOM, even if it were proven that extensional definitions of functions were more readable by some objective standard, I still think we'd have good reason to favor intensional definitions for any functions with more than 4 or 5 elements in the domain. That is, even if we could measure readability objectively, I think I'd still be in favor of clean abstractions over dozens of lines of repetition :)

shonfeder: Thanks for making this change! I find it much easier to parse out at a glance.

thpani: Look, it's about good abstractions, and I like the one we arrived at :)

All I was saying is that abstractions in computer programming aren't pure mathematical objects and come at a (in this case) cognitive cost. For me, there's definitely a trade-off between copy-pasting a line (that gets code review) versus spatially locating and loading/unloading an abstractions into my brain.

thpani:
Well here we go... there was a bug in the abstraction (membership test with `Nat`), and we missed it in code review. It was obvious in the original exhaustive enumeration. PTAL.

@Kukovec @shonfeder Do you think this is easier to read and maintain than my original solution?

Kukovec:
I don't know how we missed `Nat`, but rather than omitting it, why not just translate it to `x >= 0`?

thpani:
> I don't know how we missed `Nat`, but rather than omitting it, why not just translate it to `x >= 0`?

At least [the OOPSLA encoding already does this](https://github.com/informalsystems/apalache/blob/09b66d3b49d01670606df6db8e08fa4cf609ddaf/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRule.scala#L138), is there any benefit to doing it early?

konnov:
> > I don't know how we missed `Nat`, but rather than omitting it, why not just translate it to `x >= 0`?
> 
> At least [the OOPSLA encoding already does this](https://github.com/informalsystems/apalache/blob/09b66d3b49d01670606df6db8e08fa4cf609ddaf/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRule.scala#L138), is there any benefit to doing it early?

I don't remember the details, but we probably do it only for state variables. You can write something like this in TLA+

```tla
\E i \in Int:
  i \in Nat => ...
```

Kukovec (approved):
LGTM, though I think you could get away with not doing any checks on the type tags. It's not wrong to check, in any case, but if you ever wanted to write
```tla
VARIABLE
  \* @type: Int;
  x
...
TypeOK == x \in BOOLEAN
```
Snowcat would reject that outright, since `\in` has type constraints.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -25,19 +25,41 @@ class SetMembershipSimplifier(tracker: TransformationTracker) extends AbstractTr
...
+   * Returns the type of a TLA+ predefined set, if this transformation is applicable. In particular, it is *not*
+   * applicable to `Nat`, since `i \in Nat` does not hold for all `IntT1`-typed `i`.
+   */
+  private def typeOfSupportedPredefSet: PartialFunction[TlaPredefSet, TlaType1] = {
     case TlaBoolSet => BoolT1()
     case TlaIntSet  => IntT1()
     case TlaRealSet => RealT1()
     case TlaStrSet  => StrT1()
+    // intentionally omits TlaNatSet, see above.
   }
```
shonfeder, line 39: I'm afraid I was overly hasty and did a bad job of getting up speed on this PR and the proposed changes after being tagged in. I had missed that this function was necessary as an auxiliary, and was replying under the misconception that we could avoid having to manually enumerate this mapping. It was the prospect of avoiding this manual enumeration that mainly motivated my reasoning in my replies. If we still have to enumerate this mapping, but we just move it into a less direct location, I agree we don't get really get a beneficial abstraction or improved readability. My bad. 

I'm sorry for the misunderstanding, and I'll be more careful in my reviews/opinions in the future.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/Builder.scala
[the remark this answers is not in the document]
konnov: +1. Perhaps we do not need `STRING` in the builder.

konnov (approved):
Great! This PR transforms basic types. How about `seq \in Seq(Int)` and `fun \in [Int -> Int]`. Perhaps, something could be done about records and tuples too. Having the type information, we can do a lot of interesting simplifications.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
thpani: @shonfeder No worries, I also missed it at first! Luckily my brain kept processing overnight 😅

thpani:
> LGTM, though I think you could get away with not doing any checks on the type tags.

Right, I was wondering about it, but was unsure if me might use this in a different context where Snowcat isn't run first.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/Builder.scala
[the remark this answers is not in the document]
thpani: How do we decide what goes into `Builder` and what doesn't? It's definitely useful for unit tests, where Apalache should still support constructs that aren't necessarily Apalache-style.

thpani:
> Great! This PR transforms basic types. How about `seq \in Seq(Int)` and `fun \in [Int -> Int]`. Perhaps, something could be done about records and tuples too.

@konnov This already does `seq \in Seq(_)` as well! I've recorded the other ideas in #723, we can keep expanding the list there.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/Builder.scala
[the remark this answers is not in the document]
konnov: There is no decision process in place yet. How about keeping this method in the builder and adding a warning in the comments?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -0,0 +1,76 @@
...
+   */
+  private def transformMembership: PartialFunction[TlaEx, TlaEx] = {
+    // n \in Nat  ->  x >= 0
+    case OperEx(TlaSetOper.in, name, ValEx(TlaNatSet)) if name.typeTag == Typed(IntT1()) =>
+      OperEx(TlaArithOper.ge, name, ValEx(TlaInt(0))(intTag))(boolTag)
+    // b \in BOOLEAN, i \in Int, r \in Real  ->  TRUE
+    case OperEx(TlaSetOper.in, name, ValEx(ps: TlaPredefSet)) if isApplicable(name, ps) => trueVal
+    // seq \in Seq(_)  ->  TRUE
+    case OperEx(TlaSetOper.in, name, OperEx(TlaSetOper.seqSet, ValEx(ps: TlaPredefSet))) if isApplicableSeq(name, ps) =>
+      trueVal
```
konnov, line 68: I feel like we need some form of recursion here. For instance, we can write `<<{{1}}>> \in Seq(SUBSET Int)` or `{<<1>>} \in SUBSET (Seq(Int))`. This transformation would miss that, though it could be recursively matched. Perhaps, `SUBSET` is another form of expressions that we have to handle separately though.

konnov (commented):
Added one more comment.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/Builder.scala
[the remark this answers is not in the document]
thpani: Done in ec3c4de0

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
thpani: I'm adding it to #723

PR #1409, opened 2022-02-24
Remove unused `SymbStateRewriter` arg to IntOracle.create

shonfeder:
Towards #1064

Identified by compiler warnings, this parameter is not used in the
method, or required by an interface.

-------


konnov (approved):
Yes, looks like an unused argument!

PR #1410, opened 2022-02-24
Fix compiler warnings for unused mutable variables

shonfeder:
Towards #1064

The main thrust of this changeset is to fix compiler warnings of the form

``` sh
local var foo in method bar is never updated: consider using immutable val
```

This PR addresses the warnings by replacing the identified cases of unneeded
`var` with `val`.

I also made two unrelated fixes in

- 67ec68849cdb08a5b0cf4cdba44bc41ebffa52ab
- 6ff163b3fda56a9c06d69877f93b9bc77f977a98

These are special cases of unused values, too complicated for scalafix to
address automatically. See the individual commits for details.

### DIFF
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
@@ -11,8 +11,11 @@ import at.forsyte.apalache.tla.lir.values._
 abstract class Printer {
   def apply(p_ex: TlaEx): String
 
-  def apply(p_decl: TlaDecl): String = ""
-
+  def apply(p_decl: TlaDecl): String = {
+    // Telling the compiler we know we're not using p_decl here
+    locally(p_decl);
+    ""
+  }
 }
 
 object UTFPrinter extends Printer {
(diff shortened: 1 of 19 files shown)


tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
```diff
@@ -11,8 +11,11 @@ import at.forsyte.apalache.tla.lir.values._
 abstract class Printer {
   def apply(p_ex: TlaEx): String
 
-  def apply(p_decl: TlaDecl): String = ""
-
+  def apply(p_decl: TlaDecl): String = {
+    // Telling the compiler we know we're not using p_decl here
+    locally(p_decl);
```
thpani, line 16: This is orthogonal to this PR, but it might be cleaner to keep this abstract and override in implementations?

thpani (approved):
LGTM, thanks @shonfeder!
I was wondering what we're doing about compiler warnings, and it's great to see it's already being worked on 😀

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
shonfeder: It's a good point. I didn't actually dig in to see if there's every a reason to expect this to default to an empty string. Perhaps it should always be overriden?

thpani: I'd say so, @Kukovec you have thoughts on this?

PR #1417, opened 2022-02-28
Fix infallible ToEtcExpr tests

shonfeder:
These two asserts were comparing a value of type `Optional[String]` with
a value of type `String` for inequality, so they were guaranteed to
always succeed.

-------


konnov (approved):
LGTM

Issue #1169, opened 2021-12-13
[FEATURE] Add support for TLA+ functions to the SMT encoding with arrays


rodrigo7491:
See title. The changes to be implemented are described in ADR11 (see #1154).

PR #1419, opened 2022-02-28
Make records as they are work in the arrays encoding

### LINKED ISSUE #1288: [FEATURE] Add support for TLA+ records in the arrays encoding
See title. The initial idea is to enable records as per the oopsla19 encoding to work with sets and functions in the arrays encoding.

rodrigo7491:

This PR fixes three errors that were blocking the use of records in the arrays encoding, they are detailed below. Closes #1288.

1) `DomainRuleWithArrays` was computing `funEx` twice due to call to `super`. This was solved by moving the remaining cases of `apply` from `DomainRule` to `DomainRuleWithArrays`.
2) As pointed out by @konnov, when equating sets or functions we need to generate equalities for their elements using `LazyEquality`. This was solved by calling `cacheEqConstraints` with the sets'/functions' elements. For this fix `EqRuleWithArrays` was deleted and its logic moved to `LazyEquality`, making that class less lazy than before :-).
3) Picking a record via `CherryPick.pickRecord` was generating unsatisfiable formulas in some cases. The problem was that when constructing the domain of the picked record we could lose the chain of SSA updates, due to a `storeNotInSet` operation being in the rhs of an implication. This was solved by changing the operation to `not(selectInSet)`, which ensures absence from the set without updating the array, given that sets are initially empty.


rodrigo7491:
The integration tests caught a bug that surfaced due to the changes made for in point 2 (see PR description). When handling  function application we were comparing cells directly, which is problematic in the presence of quantifiers, e.g., `\A x \in S . f(x)`. The fix I made involved using part of the logic in `FunAppRule.applyFun`, using an oracle to avoid the direct cell comparison. In contrast to `FunAppRule` however, the constraints generated are for the domain and not the range. See the latest commit for these changes.

The integration tests are now passing, but the fix introduced a major efficiency loss. In #1169, the integration tests check for the arrays encoding took 4min, now it takes 1h. This should be addressed before merging the PR.

rodrigo7491:
The slowdown seen before was due to having to use an oracle to ascertain the argument of the function application. I made a fix in which, if the argument can be found in Scala by comparing cells, we generate a single SMT constraint querying the array that encodes the function. If we can't, e.g., in the presence of quantifiers, we use an oracle and need to have O(N) SMT constraints, with N being the cardinality of the domain of the function.

This change made the integration tests for the arrays encoding go back to their previous runtime, observed in #1169. Shall we merge the PR @konnov?

konnov (approved):
Looks good! Perhaps, we can simplify things when we add SMT-native support for records, tuples, etc.

PR #1420, opened 2022-02-28
Remove {[BUG],[FEATURE]} from issue template title

shonfeder:
Given that we have a labeling system, I think that these bracketed words
in the title are redundant and a bit noisy. So I'm suggesting removing
them.

-------


gabrielamafra (approved):
I'm in favor of this. The difference between bugs and features is not always clear, and having that on the title increases our issue management overhead.

konnov (approved):
Great idea!

Kukovec (approved):
+1

Issue #1412, opened 2022-02-26
[FEATURE] Support PlusCal specs


lemmy:
The PlusCal translator removes all comments when translating an algorithm to TLA+.

```tla
---- MODULE Foo ----

(*--algorithm OneBitClock {
  variable
    \* @type: Int;
    clock \in {0, 1};
  {
    while (TRUE) {
      if (clock = 0)
        clock := 1
      else 
        clock := 0    
    }
  }
}*)
\* BEGIN TRANSLATION (chksum(pcal) = "d384ffcd" /\ chksum(tla) = "8dd09982")
VARIABLE clock

vars == << clock >>

Init == (* Global variables *)
        /\ clock \in {0, 1}

Next == IF clock = 0
           THEN /\ clock' = 1
           ELSE /\ clock' = 0

Spec == Init /\ [][Next]_vars

\* END TRANSLATION 
====
```

shonfeder:
Related to #1436

PR #1427, opened 2022-03-01
add the Bakery pluscal spec to test/tla

konnov:
This PR adds specs in `test/tla/bakery-pluscal`. This is an example of how one can check Pluscal specs with Apalache. This is the example for the issue #1412, which will be addressed in a separate PR.

The fix in [QuantRule.scala](https://github.com/informalsystems/apalache/compare/ik/pluscal1412?expand=1#diff-5a1e4fbfceb53d647606bc9106d8bb6880857e5737ef39804bafa5c253c0e49a) was already merged. However, I accidentally branched it off this branch.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,6 @@
          * Some bug fix, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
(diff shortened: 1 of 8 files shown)


Kukovec (approved):
👍

Issue #1407, opened 2022-02-24
[FEATURE] Rewrite `\E x \in a..b: p` to `\E x \in Int: a <= x /\ x <= b /\ p`


Kukovec:
`ExprOptimizer` already translates `x \in a..b` to `a <= x <= b`, and while there's no good way to handle the set `a..b` for non-constant `a,b` explicitly, quantification should be one of the cases where we can introduce optimizations.

We should also consider translating `\A x \in a..b: p` to `\A x \in Int: (a <= x /\ x <= b) => p` (though at the moment, `\A x \in Int: ...` might be a bad idea).

Related: #446

Kukovec:
@konnov thoughts?

Kukovec:
Also, from what I can tell, this would not throw in cases like #425

konnov:
Also mentioned in #481

konnov:
It is a very good idea and we should do it. It is not just an optimization. It may enable support of `a..b` for the case when `a` and `b` are not constants. As we discussed during the stand up, I think we should do that in two steps:

 1. Translate expressions over unbounded sets of integers into the unbound forms:
  - Translate `\E x \in Int: P` and `\A x \in Int: P` into `\E x: P` and `\A x: P`.
  - Translate `\E x \in Nat: P` and `\A x \in Nat: P` into `\E x: x >= 0 /\ P` and `\A x: x >= 0 => P`
  - Translate `\E x \in a..b: P` and `\A x \in a..b: P` into `\E x: a <= x /\ x <= b /\ P` and `\A x: (a <= x /\ x <= b) => P`

 2. Add a special rule for the unbound forms `\E x: P` and `\A x: P`. By having the types, we can figure the type of `x`. The only complication here is that `Z3SolverContext` immediately pushes assertions on `assertGroundExpression`. We have to introduce a stack of quantified contexts, similar to database transactions. Whenever `\E x: P` or `\A x: P` is met, it would call `solver.introQuantifier(...)`. All calls to `assertGroundExpressions` should be buffered, if the quantifier stack is non-empty. Once the rewriting rule is done with `P`, it calls `solver.commitQuantifier(...)` and this pushes the buffered expressions either to the solver context, or to the higher level context (if quantifiers are nested).

konnov:
Also related is #844

konnov:
This one is also related #480. All these issues call for a general approach.

konnov:
[...]

PR #1431, opened 2022-03-02
Added transformations for quantification over special integer sets

Kukovec:

First part of #1407, without the plumbing.


tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestQuantificationOptimizer.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  def inBounds(lowerBound: TlaEx, upperBound: TlaEx): TlaEx = {
+    val lConstraint = tla.le(lowerBound, varname).as(BoolT1())
+    val uConstraint = tla.le(varname, upperBound).as(BoolT1())
+    tla.and(lConstraint, uConstraint).as(BoolT1())
+  }
+
+  val ge0 = tla.ge(varname, tla.int(0).as(IntT1())).as(BoolT1())
+
+  def postPredE(setEx: TlaEx, ex: TlaEx): TlaEx = setEx match {
```
gabrielamafra, line 73: I suggest `s/postPredE/postPredExists` and `s/postPredA/postPredForall`. Took me a while to understand what this was referring to (my first guess was E = expected and A = actual).

gabrielamafra (approved):
LGTM, I just left some opinion/questions for the testing part.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestQuantificationOptimizer.scala
```diff
@@ -0,0 +1,104 @@
...
+    b <- gens.genTlaEx(ops)(gens.emptyContext)
+  } yield tla.dotdot(a.as(IntT1()), b.as(IntT1())).as(SetT1(IntT1()))
+
+  val boolTag = Typed(BoolT1())
+  val varname = tla.name("x").as(IntT1())
+
+  def forSetWithPredicates(
+      setGen: Gen[TlaEx],
+      postPredE: (TlaEx, TlaEx) => TlaEx,
+      postPredA: (TlaEx, TlaEx) => TlaEx): Unit = {
```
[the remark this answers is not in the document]
Kukovec: Ah, this seems to be an artifact of refactoring. They used to be separate functions, but then I changed the tests enough to where all patterns fit in 1 match. Good catch.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
```diff
@@ -0,0 +1,88 @@
...
+    // Assume typechecking succeeded -> name has type Int
+    case OperEx(oper @ (TlaBoolOper.forall | TlaBoolOper.exists), nameEx, ValEx(TlaIntSet), pred) =>
+      OperEx(mkUnbounded(oper), nameEx, pred)(boolTag)
+    case OperEx(oper @ (TlaBoolOper.forall | TlaBoolOper.exists), nameEx @ NameEx(name), ValEx(TlaNatSet), pred) =>
+      modifyPred(oper, nameEx, ge0(name), pred)
+    case OperEx(oper @ (TlaBoolOper.forall | TlaBoolOper.exists), nameEx @ NameEx(name),
+            OperEx(TlaArithOper.dotdot, lowerBound, upperBound), pred) =>
+      modifyPred(oper, nameEx, inRange(name, lowerBound, upperBound), pred)
+    case ex => ex
+  }
```
shonfeder, line 87: NOTE: This comment is just a suggestion, and more to discuss the general way we break down this kind of logic than to insist on any particular change. If you don't think the suggestion helps clarify the logic, feel free to disregard.

This may come down to a matter of taste, but the way the logic of this function is currently factored is a bit confusing to me. I think the following things throw me off while reading:

- We end up pattern matching on the quantifiers in 3 different places
- The helper function `mkUnbounded` gets applied in two logically different places (once on line 74, and again on line 80) even tho the underlying logic of the transformations in each case is really quite similar.
- Instead of applying a few different transformations to construct the new data we need, and then constructing the transformation using that new data, we end up sort of (imo) "hiding" the sequence of transformations in nested function calls.

An alternative that aims to consolidate the related changes, to give what is (imo) a clearer and more concise diagram of the overall transformation being made, might look something like this:

```suggestion
  def transform: TlaExTransformation = tracker.trackEx {
    // Assume typechecking succeeded -> name has type Int
    case OperEx(oper @ (TlaBoolOper.forall | TlaBoolOper.exists), name, bound, pred) if isAllowedSet(bound) => {
      val (unboundedOper, outerPred) = oper match {
        case TlaBoolOper.forall => (TlaBoolOper.forallUnbounded, TlaBoolOper.implies)
        case TlaBoolOper.exists => (TlaBoolOper.existsUnbounded, TlaBoolOper.and)
      }
      val newPred = bound match {
        case ValEx(TlaIntSet) => pred
        case ValEx(TlaNatSet) => OperEx(outerPred, ge0(name), pred)
        case OperEx(TlaArithOper.dotdot, lowerbound, upperbound) =>
          OperEx(outerPred, inRange(name, lowerbound, upperbound), pred)
      }
      OperEx(unboundedOper, name, newPred)
    }
  }
```

And the `mkUnbounded` would also not be needed.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * This is the only place we support unbounded quantification, as we can leverage the type annotation of `x` (which
+ * should be IntT1()). We can then introduce a special rewriting rule for this quantification form.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+
+@Singleton
+class IntegerQuantificationOptimizer(tracker: TransformationTracker) extends TlaExTransformation {
```
shonfeder, line 22: What motivates the use the `@Singleton` annotation here instead of defining this as an `object`?

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestQuantificationOptimizer.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  test("Translation of Int") {
+    val intSet = tla.intSet().as(SetT1(IntT1()))
+
+    forSetWithPredicates(Gen.const(intSet), postPredE, postPredA)
+  }
+
+  test("Translation of a..b") {
+    forSetWithPredicates(rangeGen, postPredE, postPredA)
+  }
```
shonfeder, line 103: What are the properties being tested for here? It's not clear to me from the names of the test. I can see that you are testing certain transformations, but it would help with readability of the tests IMO to include in the name what properties of the transformations are being checked for.

shonfeder: In my experience, the point of property based tests it to end up with something that looks like: "forall x: T. P(x)` where the `T` is determined by generators of expressions of the appropriate type, and `P` is the test of what should hold of every instance of `x`. I am not really able to figure out whether that is happening in these tests in some way. Maybe after the fix discussed with Gabriela that will be clearer for me. I'll check back after those changes are int!

shonfeder (commented):
A few small comments and suggestions, one one more substantive (but totally optional) suggestion about the organization of the logic in the transformation.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
```diff
@@ -0,0 +1,88 @@
...
+  // name >= 0
+  private def ge0(name: String): TlaEx =
+    OperEx(TlaArithOper.ge, NameEx(name)(intTag), ValEx(TlaInt(0))(intTag))(Typed(BoolT1()))
+
+  // a <= name <= b
+  private def inRange(name: String, lowerBound: TlaEx, upperBound: TlaEx): TlaEx = {
+    val lConstraint = OperEx(TlaArithOper.le, lowerBound, NameEx(name)(intTag))(boolTag)
+    val uConstraint = OperEx(TlaArithOper.le, NameEx(name)(intTag), upperBound)(boolTag)
+    OperEx(TlaBoolOper.and, lConstraint, uConstraint)(boolTag)
+  }
```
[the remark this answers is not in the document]
Kukovec: The purpose of the string parameter is to explicitly avoid passing `nameEx`, because you always have to consider UID uniqueness. This way, each constructed `NameEx` has a fresh UID (except the `NameEx` directly below the quantifier, which _is_ an exact replica, but it still maintains uniqueness if it is used in exactly one place)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
Kukovec: Yeah, this is celeaner.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
Kukovec: I think I copied `TypeSubstitutor.scala` as a template. It's probably not needed. It definitely can't be an `object` though, as it's parameterized by `tracker`.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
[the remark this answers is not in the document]
shonfeder: Ah yes, the uuid. I forgot about those!

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestQuantificationOptimizer.scala
Kukovec: If you want the above formulation, `P(ex) := transform((\E | \A) x \in S: P)  = (\E | \A) x: Q`,

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
shonfeder: h right. Makes sense. We might want to figure out what that singleton annotation is used for at some point!

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestQuantificationOptimizer.scala
Kukovec: Basically, does the transformation have the expected shape for `Q`

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestQuantificationOptimizer.scala
```diff
@@ -0,0 +1,104 @@
...
+      val forallEx = tla.forall(varname, setEx, ex).as(BoolT1())
+
+      val existsCond = postPredE(setEx, ex)
+      val forallCond = postPredA(setEx, ex)
+
+      val existsExpected: TlaEx = OperEx(TlaBoolOper.existsUnbounded, varname, existsCond)(boolTag)
+      val forallExpected: TlaEx = OperEx(TlaBoolOper.forallUnbounded, varname, forallCond)(boolTag)
+
+      val existsActual = optimizer(existsEx)
+      val forallActual = optimizer(forallEx)
```
thpani, line 57: I think this should be refactored to separate building expected and actual values (now they're interleaved), and actually separate them via a comment. It took me quite some time to track what's happening here.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestQuantificationOptimizer.scala
```diff
@@ -0,0 +1,104 @@
...
+      exRaw <- gens.genTlaEx(ops)(gens.emptyContext)
+    } yield (setEx, exRaw)
+    val prop = forAll(pairGen) { case (setEx, exRaw) =>
+      // gens produces garbage w.r.t. type tags, we just force the toplevel ex to be boolean.
+      val ex = exRaw.withTag(boolTag)
+      val existsEx = tla.exists(varname, setEx, ex).as(BoolT1())
+      val forallEx = tla.forall(varname, setEx, ex).as(BoolT1())
+
+      val existsCond = postPredE(setEx, ex)
+      val forallCond = postPredA(setEx, ex)
```
thpani, line 51: Neither `{exists,forall}Cond` nor `postPred{A,E}` tell me that these are the _expected_ conditions. I think they should be renamed to something more descriptive.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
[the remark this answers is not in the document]
konnov: Not really. IMO a simplifier translates an expression into another expression that is equivalent under certain conditions and structurally simpler. An optimizer may produce a more complex expression, which nevertheless improves performance.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  // (\A | \E) x \in S: P transforms into (\A | \E) x: Q
+  // For \E x \in S: P,
+  // Q := R /\ P, for some R
+  // and for \A x \in S: P,
+  // Q := R => P, for some R
```
konnov, line 61: I would prefer separate explanations for the two cases (`\E` and `\A`). The comments look like a puzzle.

konnov (approved):
Just a few comments, as there are good reviews already. I agree with Thomas that these transformations are not about optimizations. Maybe they are simplifications.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/IntegerQuantificationOptimizer.scala
```diff
@@ -0,0 +1,88 @@
...
+import at.forsyte.apalache.tla.lir.values.{TlaInt, TlaIntSet, TlaNatSet}
+
+import com.google.inject.Singleton
+
+/**
+ * This transformation rewrites expressions of the form `(\A | \E) x \in S: P`, where S is one of `Int`, `Nat`, `a..b`,
+ * into `(\A | \E) x: Q` (unbounded quantification).
+ *
+ * This is the only place we support unbounded quantification, as we can leverage the type annotation of `x` (which
+ * should be IntT1()). We can then introduce a special rewriting rule for this quantification form.
```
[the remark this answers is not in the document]
Kukovec: No, that's the next PR.

shonfeder (approved):
LGTM!

Thank you for taking the time to add this feature in incrementally! :)

PR #1432, opened 2022-03-03
Add references to the source files and diffs

konnov:
This PR adds references to the specs in the tutorial, as well as diffs (for a future PR).


Kukovec (approved):
👍

PR #1401, opened 2022-02-23
Exclude operators causing malformed expressions from PBT

thpani:
`TlaFunOper.{funDef,recFunDef}` introduced in #1386 seem to cause lots of malformed expressions in PBT that slow down unit tests. Jure suggested this may happen in [#1386 (comment)](https://github.com/informalsystems/apalache/pull/1386#discussion_r811016316).
Therefore, disable them for now.

PR #1434, opened 2022-03-03
Prevent infinite chain of generator calls and re-enable function constructors in PBT

### LINKED ISSUE #1402: Support `TlaFunOper.{funDef,recFunDef}` in PBT
#1401 disabled `TlaFunOper.{funDef,recFunDef}` in all property-based tests for performance regression on CI.
It should be re-introduced wherever possible. Use this issue to track progress.

@konnov suggested in https://github.com/informalsystems/apalache/issues/1401#issuecomment-1048894262 that this may only affect recursive functions.

thpani:
This fixes an infinite (non-`size`-decreasing) chain of generator calls via `genTlaEx` -> `genLetInEx` -> `genTlaOperDecl` -> `genTlaEx`, where in `genTlaEx` a non-resized reference of `genTlaEx` is passed via the `exGen` parameters of `genLetInEx` and `genTlaOperDecl`, where it is (recursively) called.

It seems this infinite chain was only eventually broken by the ScalaCheck RNG.
We fix this by resizing `genTlaEx` in `genTlaOperDecl`.

I checked the call graph for other such cycles, but haven't found any.

This PR also re-enables function constructors in PBT, which where disabled for performance regression in #1401.
Hypothetically, this regression was not introduced directly the additional operators, but by the infinite chain of generator calls fixed in this PR.

Closes #1402

### DIFF
--- a/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
+++ b/tlair/src/test/scala/at/forsyte/apalache/tla/lir/IrGenerators.scala
@@ -260,10 +260,10 @@ trait IrGenerators extends TlaType1Gen {
    * @return
    *   a generator of operator declarations
    */
-  def genTlaOperDecl(exGen: UserContext => Gen[TlaEx])(ctx: UserContext): Gen[TlaOperDecl] = {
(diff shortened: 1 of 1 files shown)


konnov (approved):
Great find!

PR #1442, opened 2022-03-07
Support for FunAsSeq

### LINKED ISSUE #569: [FEATURE] Translate FunAsSeq in the model checker
Following #223, we have to properly translate `Apalache!FunAsSeq` to SMT.

konnov:
Closes #569. Instead of translating `FunAsSeq` to SMT, we are simply using its TLA+ definition via `MkSeq`.


tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/ApalacheOper.scala
```diff
@@ -102,10 +102,15 @@ object ApalacheOper {
...
-   * Attempt to dynamically cast an Int -> T function to a Seq(T). The first argument should be the function expression
-   * and the second argument should be an integer, denoting the maximal length of the sequence.
+   * <p>Attempt to dynamically cast an Int -> T function to a Seq(T). The first argument should be the function
+   * expression and the second argument should be an integer, denoting the maximal length of the sequence.</p>
+   *
+   * <p>We keep this operator in the IR. However, we are using the definition of this operator from Apalache.tla. Hence,
+   * if you construct an expression that contains `funAsSeq`, its constructor will throw.</p>
    */
   object funAsSeq extends ApalacheOper {
+    require(false, "This operator is defined in Apalache.tla. Do not construct it.")
```
Kukovec, line 112: Why not just delete the object, if you're going to add `require(false)`?

Kukovec (approved):
Looks fine, but I wonder if we couldn't change the interface of `FunAsSeq(fn, maxSeqLen)` to `FunAsSeq(fn)` instead, and just compute `maxSeqLen == FoldSet(MaxOf2, 0, DOMAIN fn)` (at the cost of performance)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/ApalacheOper.scala
konnov: I also wanted to delete the object. But it is better to keep it as documentation.

konnov:
> Looks fine, but I wonder if we couldn't change the interface of `FunAsSeq(fn, maxSeqLen)` to `FunAsSeq(fn)` instead, and just compute `maxSeqLen == FoldSet(MaxOf2, 0, DOMAIN fn)` (at the cost of performance)

That would not help us, as `MkSeq` requires a constant length.

konnov:
However, we should think further. Maybe there is a way to implement `FunAsSeq` without passing constant length directly.

PR #1443, opened 2022-03-07
Simplify `SUBSET` and sets of functions `[S1 -> S2]` in `TypeOK`

thpani:
Going forward with #723, this PR adds code to simplify the following in `TypeOK`:
- power sets (`SUBSET`),
- function sets (`[S1 -> s2]`), and
- nested expressions of these and previously supported expressions

:bulb: Probably easiest to review commit-by-commit. The first 2 commits contain refactoring that will become useful for the later commits.

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
@@ -7,9 +7,15 @@ import at.forsyte.apalache.tla.lir.transformations.{LanguageWatchdog, Transforma
 import at.forsyte.apalache.tla.lir.values._
 
 /**
- * A simplifier that rewrites vacuously true membership tests based on type information.
+ * A simplifier that rewrites expressions commonly found in `TypeOK`. Assumes expressions to be well-typed.
  *
- * For example, `x \in BOOLEAN` is rewritten to `TRUE` if x is typed BoolT1.
+ * After Apalache's type-checking, we can rewrite some expressions to simpler forms. For example, the (after
+ * type-checking) vacuously true `x \in BOOLEAN` is rewritten to `TRUE` (as `x` must be a `BoolT1`).
+ *
+ * We currently perform the following simplifications:
+ *   - `n \in Nat` ~> `x >= 0`
+ *   - `b \in BOOLEAN`, `i \in Int`, `r \in Real` ~> `TRUE`
+ *   - `seq \in Seq(_)` ~> `TRUE`
  *
  * @author
  *   Thomas Pani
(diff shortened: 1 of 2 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -7,9 +7,18 @@ import at.forsyte.apalache.tla.lir.transformations.{LanguageWatchdog, Transforma
...
+ * After Apalache's type-checking, we can rewrite some expressions to simpler forms. For example, the (after
+ * type-checking) vacuously true `x \in BOOLEAN` is rewritten to `TRUE` (as `x` must be a `BoolT1`).
+ *
+ * We currently perform the following simplifications (for applicable sets AS, see [[isApplicable]]):
+ *   - `n \in Nat` ~> `x >= 0`
+ *   - `b \in BOOLEAN`, `i \in Int`, `r \in Real` ~> `TRUE`
+ *   - `seq \in Seq(AS)` ~> `TRUE`
+ *   - `set \in SUBSET AS` ~> `TRUE`
+ *   - `fun \in [AS1 -> AS2]` ~> `TRUE`
+ *   - `fun \in [Dom -> AS]` ~> `DOMAIN fun = Dom`
```
Kukovec, line 21: Can you add `tup \in S1 \X ... \X SN` ~> `TRUE`? 
Maybe even `tup \in SimplifiableSet1 \X ... \X SimplifiableSetN` ~> ` cond(SimplifiableSet1) /\ ... /\ cond(SimplifiableSetN)`, where `SimplifiableSet` is such that `transformMembership` has a case for it?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -27,46 +36,55 @@ class SetMembershipSimplifier(tracker: TransformationTracker) extends AbstractTr
...
 
   /**
-   * Rewrites vacuously true membership tests based on type information, and rewrites `i \in Nat` to `i \ge 0`.
+   * Simplifies expressions commonly found in `TypeOK`, assuming they are well-typed.
    *
-   * For example, `x \in BOOLEAN` is rewritten to `TRUE` if `x` is typed `BoolT1`.
+   * @see
+   *   [[SetMembershipSimplifier]] for a full list of supported rewritings.
    */
   private def transformMembership: PartialFunction[TlaEx, TlaEx] = {
```
Kukovec, line 73: Since all the cases have to be `OperEx( TlaSetOper.in, name, ...)` you could do
```scala
private def transformMembership: TlaEx => TlaEx = {
  case OperEx(TlaSetOper.in, name, setEx) => 
    setEx match {
      // x \in AS  ~>  TRUE
      case set if isApplicable(set) => trueVal
      // n \in Nat  ~>  x >= 0
      case ValEx(TlaNatSet) => OperEx(TlaArithOper.ge, name, ValEx(TlaInt(0))(intTag))(boolTag)
      // fun \in [Dom -> AS]  ~>  DOMAIN fun = Dom    (fun \in [AS1 -> AS2] is handled above)
      case OperEx(TlaSetOper.funSet, domain, set2) if isApplicable(set2) =>
        OperEx(TlaOper.eq, OperEx(TlaFunOper.domain, name)(domain.typeTag), domain)(boolTag)
      case _ => ex
    }
  case _ => ex
}
```

I'm also not too keen on the use of partial functions; they make it harder to track down bugs, because the code will just skip cases where the partial function isn't applicable, instead of complaining that all cases aren't covered by the transformation (which may or may not be an implementation error).

Kukovec (changes requested):

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
thpani: Tuples, records etc are a TODO in https://github.com/informalsystems/apalache/issues/723

Kukovec: Fair enough, ignore this comment then.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
thpani: Good idea with the refactoring!

Regarding the partial function, I'm overriding `AbstractTransformer.partialTransformers` above which is a `Seq[PartialFunction[TlaEx, TlaEx]]` and `Function1` isn't a subtype of `PartialFunction`. I suppose there's nothing to be done here, unless you have an idea how to work around this?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -7,9 +7,18 @@ import at.forsyte.apalache.tla.lir.transformations.{LanguageWatchdog, Transforma
...
 
 /**
- * A simplifier that rewrites vacuously true membership tests based on type information.
+ * A simplifier that rewrites expressions commonly found in `TypeOK`. Assumes expressions to be well-typed.
  *
- * For example, `x \in BOOLEAN` is rewritten to `TRUE` if x is typed BoolT1.
+ * After Apalache's type-checking, we can rewrite some expressions to simpler forms. For example, the (after
+ * type-checking) vacuously true `x \in BOOLEAN` is rewritten to `TRUE` (as `x` must be a `BoolT1`).
+ *
+ * We currently perform the following simplifications (for applicable sets AS, see [[isApplicable]]):
```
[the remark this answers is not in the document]
thpani: Awesome, I was looking for a better name but couldn't find one 👍

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
Kukovec: Then keep it as Partial I suppose, lets not derail this PR.
But @konnov, I propose to make `AbstractTransformer` use actual functions. Using `PartialFunctions` fells like it's supporting laziness, not efficiency.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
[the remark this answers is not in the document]
thpani: Addressed in [372742d](https://github.com/informalsystems/apalache/pull/1443/commits/372742d200afe0b2e4b406f82b93e26a597e6d21)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
thpani: Refactoring in [9387f09](https://github.com/informalsystems/apalache/pull/1443/commits/9387f0937111e7c2e63b83531c3aea090ed1dec2)

Kukovec (approved):
👍

PR #1449, opened 2022-03-08
Split TestSetMembershipSimplifier

thpani:
Simple refactoring that splits `TestSetMembershipSimplifier` into multiple, smaller unit tests.

This is in prep for supporting records in #723


Kukovec (approved):
👍

PR #1450, opened 2022-03-08
adding length to FunAsSeq

konnov:
After discussing `FunAsSeq` in #1442, I have added another parameter to this operator:

```tla
(**
 * As TLA+ is untyped, one can use function- and sequence-specific operators
 * interchangeably. However, to maintain correctness w.r.t. our type-system,
 * an explicit cast is needed when using functions as sequences.
 *
 * The parameters have the following meaning:
 *
 *  - fn is the function from 1..len that should be interpreted as a sequence.
 *  - len is the length of the sequence, len = Cardinality(DOMAIN fn),
 *    len may be a variable, a computable expression, etc.
 *  - capacity is a static upper bound on the length, that is, len <= capacity.
 *
 * @type: ((Int -> a), Int, Int) => Seq(a);
 *)
FunAsSeq(fn, len, capacity) ==
    LET __FunAsSeq_elem_ctor(i) == fn[i] IN
    SubSeq(MkSeq(capacity, __FunAsSeq_elem_ctor), 1, len)
```

This new form lets the user to specify the actual length of the sequence, which may be unknown in advance, while providing the tool with a statically known bound.


konnov:
@Kukovec, what do you think?

Kukovec (approved):
LGTM, assuming we're happy with the current error message for "constant" LET-INs

Issue #401, opened 2021-01-07
More precise type inference for records


Kukovec:
We should rethink type inference for sets of records of different types.
One possible direction would be to introduce something similar to variant records (https://en.wikibooks.org/wiki/Ada_Programming/Types/record#Variant_record). We could then treat expressions of the form `r.type = X` as casts, possibly accompanied by a change in the internal representation.

konnov:
Here is a piece of text from https://apalache.informal.systems/docs/adr/002adr-types.html for the context:

> Note: For the above example of a set of records, we are considering to introduce union types. So the type of the set
> 
> 
> {[type |-> "1a", bal |-> 1],
>  [type |-> "2a", bal |-> 2, val |-> 3]}
> would be something like:
> 
> 
> Set([type |-> "1a"], bal |-> 1] + [type |-> "2a", bal |-> 2, val |-> 3])
> The value of the field type would serve as a type tag. However, we would have to fix a set of patterns that turn a union type into a precise record type. One such pattern is a set comprehension, e.g., { r \in S: r.type = "1a" }. If you have suggestions on this line of thought, please let us know.

konnov:
It looks like at least three users ran into hard-to-debug issues that would not appear if we had a precise record type checker.

konnov:
This seems to be a usability issue after all

konnov:
Let's figure patterns first and write an ADR in addition to https://apalache.informal.systems/docs/adr/002adr-types.html

shonfeder:
Might we also consider using https://en.wikipedia.org/wiki/Row_polymorphism ? This would just enable the (imo) most natural sub-typing relationship on records, where a record `r1 <: r2` iff `r2` has all the `fieldName : type`s in `r1` (but possibly others as well). I think that would cover  the case you gave, without having to introduce a special meaning for a particular key.

konnov:
I am not sure about row polymorphism, but I guess it is different. What we need are records that have different combinations of fields and are distinguished by a tag field. That's the pattern introduced by Lamport, which would be very hard to replace with something else.

shonfeder:
What would be the difficulty in replacing this with a more principled approach?

konnov:
One thing is to add a few additional tests where needed, another one is to completely rewrite a spec. Here is a canonical example of how records are mixed in a set: https://github.com/tlaplus/Examples/blob/master/specifications/Paxos/Paxos.tla

[...]

PR #1453, opened 2022-03-08
Simplify records in TypeOK

### LINKED ISSUE #723: [FEATURE] Use types to simplify `TypeOK`

thpani:
Blocked by [#401 Precise type checking for records](https://github.com/informalsystems/apalache/issues/401)
Closes #723

Add record sets to the type-defining sets and simplify membership tests against them.

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
@@ -64,7 +64,7 @@ class SetMembershipSimplifier(tracker: TransformationTracker) extends AbstractTr
     // 3. [s1 -> s2] for type-defining sets `s1` and `s2`
     case OperEx(TlaSetOper.funSet, set1, set2) => isTypeDefining(set1) && isTypeDefining(set2)
     // 4. s1 \X ... \X sn for type-defining sets `s1` to `sn`
-    case OperEx(TlaSetOper.times, args @ _*) => args.forall(set => isTypeDefining(set))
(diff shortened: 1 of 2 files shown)


Kukovec:
Records are one place where I'd be careful, probably to the point of not implementing this simplification at all. The problem with current type inference w.r.t. records, is that it overapproximates record types. 
A record, the value of which is `r = [a |-> 5]` can be typed `[a: Int, b: Int]`, and "pass" the check `r \in [a: Int, b: Int]` with this rule.
I think this simplification should be postponed until after variant-type support is implemented.

thpani:
> Records are one place where I'd be careful, probably to the point of not implementing this simplification at all. [...]
> I think this simplification should be postponed until after variant-type support is implemented.

@Kukovec Aaaah, I read about records, but given how we currently type them, I assumed that `[a |-> 5] \in [a: Int, b: Int]` holds. My bad. I will close this and we can revisit it after #401.

PR #1458, opened 2022-03-09
Fix some classes of compiler warnings

thpani:
Towards #1064. Fixes all compiler warnings related to

- empty files (00f4926)
- catching all `Throwables` (ece65b5)
- Using `val` instead of `var` for immutable values (542d6e7)
- Type arguments eliminated by erasure (cb0742498ec5543423e814738b24fdc44fefbb06)

Further info in the commit comments.


konnov (approved):
LGTM

PR #1462, opened 2022-03-09
Fix compiler warnings on non-exhaustive matches

thpani:
Towards #1064 

Fix compiler warnings on non-exhaustive matches.
Rationale for each fix in the commit message.


konnov (approved):
Nice. Thanks for taking care of this!

PR #1463, opened 2022-03-09
Fix for improper recursion in embedding

Kukovec:


konnov (approved):
Looks good. But are we sure that this is the ultimate solution?

Kukovec:
> Looks good. But are we sure that this is the ultimate solution?

Without answering your question, the "ultimate solution" would probably be merging `Inliner`, `Embedder` and `LET-expander` into one utility, replacing `bodyMap`-based computation with an operator dependency-graph traversal and explicitly encodiding scope.

PR #1464, opened 2022-03-10
Fix compiler warnings for unused pattern vars

thpani:
Towards #1064. Fixes/silences compiler warnings of the shape

    pattern var v in value V is never used: use a wildcard `_` or suppress this warning with `v@_`

Similar modifications are grouped into the same commit.

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/DomainRule.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/DomainRule.scala
@@ -63,7 +63,7 @@ class DomainRule(rewriter: SymbStateRewriter, intRangeCache: IntRangeCache) exte
   }
 
   protected def mkSeqDomain(state: SymbState, seqCell: ArenaCell): SymbState = {
-    val (protoSeq, len, capacity) = proto.unpackSeq(state.arena, seqCell)
+    val (protoSeq @ _, len, capacity) = proto.unpackSeq(state.arena, seqCell)
     // We do not know the domain precisely, as it depends on the length of the sequence.
     // Hence, we create the set `1..capacity` and include only those elements that are not greater than `len`.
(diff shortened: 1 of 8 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/DomainRule.scala
```diff
@@ -63,7 +63,7 @@ class DomainRule(rewriter: SymbStateRewriter, intRangeCache: IntRangeCache) exte
   }
 
   protected def mkSeqDomain(state: SymbState, seqCell: ArenaCell): SymbState = {
-    val (protoSeq, len, capacity) = proto.unpackSeq(state.arena, seqCell)
+    val (protoSeq @ _, len, capacity) = proto.unpackSeq(state.arena, seqCell)
```
Kukovec, line 66: `a @ _` is just `a`

Kukovec (approved):
👍

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/DomainRule.scala
thpani: This silences the warning that `protoSeq` is unused; I figured it makes sense here to see `protoSeq` instead of just binding it to `_`.

Kukovec: But why wouldn't you want `_` if `protoSeq` is actually unused?

thpani: For me, the statement is more readable (i.e., more obvious what `len` and `capacity` refer to and what `unpackSeq` does) if `protoSeq` is named. YMMV.

PR #1472, opened 2022-03-11
Fix unused parameter compiler warnings

### LINKED ISSUE #1526: Fix unused compiler warnings
Blocks https://github.com/informalsystems/apalache/issues/1064. Fix compiler warnings caused by `-Ywarn-unused`.
- [x] #1471
- [ ] #1472

thpani:
Towards https://github.com/informalsystems/apalache/issues/1064. Closes #1526

Second chunk of fixes for compiler warnings of the shape

    parameter value arg in {class, method} MySymbol is never used

Rationale for the fix in the commit message.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/CardinalityConstRule.scala
```diff
@@ -76,7 +76,7 @@ class CardinalityConstRule(rewriter: SymbStateRewriter) extends RewritingRule {
     }
 
     // create the inequality predicate
-    val witnesses = 1.to(threshold).map(pick).toList
+    val witnesses = 1.to(threshold).map(_ => pick()).toList
```
Kukovec, line 79:
```suggestion
    val witnesses = List.fill(threshold)(pick())
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetUnionRule.scala
```diff
@@ -61,15 +61,15 @@ class SetUnionRule(rewriter: SymbStateRewriter) extends RewritingRule {
...
+            def isPointedBySet(setElems: Set[ArenaCell]): Boolean = setElems.contains(elemCell)
             def inPointingSet(set: ArenaCell) = {
               // this is sound, because we have generated element equalities
               // and thus can use congruence of in(...) for free
               tla.and(tla.apalacheSelectInSet(set.toNameEx, topSetCell.toNameEx),
                   tla.apalacheSelectInSet(elemCell.toNameEx, set.toNameEx))
             }
 
-            val pointingSets = (sets.zip(elemsOfSets).filter((isPointedBySet _).tupled)).map(_._1)
+            val pointingSets = sets.zip(elemsOfSets).filter { case (_, elems) => isPointedBySet(elems) }.map(_._1)
```
Kukovec, line 72:
```suggestion
            val pointingSets = sets.zip(elemsOfSets).collect { case (e, elems) if isPointedBySet(elems) => e }
```

Kukovec (commented):
Regarding `@unused`, why not just change the API?
> Kept for argument compatibility with the other rewrite* methods at the callsite

I don't get this comment.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetUnionRule.scala
thpani: Good point, applied in 04e4ae05

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/CardinalityConstRule.scala
thpani: I'd rather not do this. To understand its correctness, you'd have to both know that the parameter is call-by-name and that it's evaluated for each element. Neither is obvious from the snippet.

thpani:
> Regarding `@unused`, why not just change the API?
> 
> > Kept for argument compatibility with the other rewrite* methods at the callsite
> 
> I don't get this comment.

What I meant is that calls to other methods at the same location expect these arguments; changing the API of a single one makes that entire code misaligned and hard to read.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/CardinalityConstRule.scala
Kukovec: If you want to emphasize that pick is called separately each time you can still do
```suggestion
    val witnesses = List.fill(threshold)(_ => pick())
```
instead of `to` + `map`

Kukovec:
> What I meant is that calls to other methods at the same location expect these arguments; changing the API of a single one makes that entire code misaligned and hard to read.

I would personally still just delete them, except in cases where methods are overridden from some base class (not the case here, afaict). The notion of "alignment" is a bit questionable to me, I don't see why different methods that do similar, but not identical, things would need to have the same parameter set.

thpani:
> I would personally still just delete them, except in cases where methods are overridden from some base class (not the case here, afaict).

All of the `*Rule` classes extend `RewritingRule`. I suppose the only reason this doesn't take a `SymbStateRewriter` argument is because it's implemented as Trait. I found no double inheritance there, we can open a separate PR to replace it with an abstract class.

konnov (approved):
Looks great!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/IntDotDotRule.scala
```diff
@@ -8,14 +8,15 @@ import at.forsyte.apalache.tla.lir.io.UTFPrinter
...
+import scalaz.unused
 
 /**
  * Rewrites an integer range a..b.
  *
  * @author
  *   Igor Konnov
  */
-class IntDotDotRule(rewriter: SymbStateRewriter, intRangeCache: IntRangeCache) extends RewritingRule {
+class IntDotDotRule(@unused rewriter: SymbStateRewriter, intRangeCache: IntRangeCache) extends RewritingRule {
```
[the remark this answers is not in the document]
thpani: I explained this in https://github.com/informalsystems/apalache/pull/1472#issuecomment-1067681067. WDYT?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/FunExceptRule.scala
```diff
@@ -163,7 +164,7 @@ class FunExceptRule(rewriter: SymbStateRewriter) extends RewritingRule {
   def rewriteTuple(
       state: SymbState,
       oldTuple: ArenaCell,
-      tupleT: TupT1,
+      @unused tupleT: TupT1,
```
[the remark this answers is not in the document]
thpani: Yes.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
```diff
@@ -252,7 +253,7 @@ class CherryPick(rewriter: SymbStateRewriter) {
    *   a new symbolic state with the expression holding a fresh cell that stores the picked element.
    */
   def pickRecord(
-      cellTypeToIgnore: CellT,
+      @unused cellTypeToIgnore: CellT,
```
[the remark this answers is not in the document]
thpani: Fixed in 2d8ffd8

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/IntDotDotRule.scala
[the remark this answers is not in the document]
konnov: OK. I guess, most of these rules were created by copy & paste. In any case, the annotation clearly marks that the parameter is not used.

Issue #1473, opened 2022-03-11
Fix compiler deprecation and feature warnings


thpani:
Blocks #1064

`scalac` currently reports 7 deprecation and 14 feature warnings across our codebase. These become fatal when passing `-Xfatal-warnings` (https://github.com/informalsystems/apalache/issues/1473#issuecomment-1065161610) in #1064.

We should

- [x] enable `-feature` and `-deprecation`
- [x] #1481 
- [x] #1482

shonfeder:
Presumably related to #1064, which is why I'm assigned/

thpani:
@shonfeder Sorry, I meant to assign myself 😅

I don't know if the current warnings á la

    [warn] four deprecations (since 2.11.0); re-run with -deprecation for details
    
would fail with `-Xfatal-warnings`, so either this is blocked by #1064 or blocking it.

shonfeder:
no worries! I'm happy to take the issue if you decide to change gears :)

PR #1474, opened 2022-03-11
Enable feature and deprecation warnings

thpani:
Towards #1473, enables feature and deprecation warnings


build.sbt
```diff
@@ -44,6 +44,15 @@ ThisBuild / libraryDependencies ++= Seq(
...
 
+//////////////////////
+// Compiler options //
+//////////////////////
+
+ThisBuild / scalacOptions ++= Seq(
+    "-feature", // enable feature warnings
+    "-deprecation", // enable compiler warnings
+)
+
```
shonfeder, line 55: We are also setting compiler flags on line 67. I think it would be best to keep them all together. What do you think?

thpani: I had them all together, but then the options below are requirements of `scalafix` so I thought they should go under the "Linting and formatting" section. Usually I'd prefer to have them all in the same place, but this being a Scala `Seq` makes it really tempting to split them up... I guess we have to decide if we group thematically or by sbt variable.
I'm undecided, lmk what you prefer.

shonfeder: If we were conditionally setting the scalac options for scalafix, I could see splitting them up, but since these options will always be enabled for all the builds, and since this section of the config is called `Compiler Options`, I'm inclined to keep them together. Already the addition of `Ypatmat-exhaust-depth` is not relevant to `scalafix` so I guess that comment and placement is just outdated?

build.sbt
```diff
@@ -44,6 +44,22 @@ ThisBuild / libraryDependencies ++= Seq(
...
+ThisBuild / scalacOptions ++= Seq(
+    // Enable deprecation and feature warnings
+    "-deprecation",
+    "-feature",
+    // Enable `unused` compiler warnings; required by scalafix
+    // https://scalacenter.github.io/scalafix/docs/rules/RemoveUnused.html
+    "-Ywarn-unused",
+    // Fixes warning: "Exhaustivity analysis reached max recursion depth, not all missing cases are reported."
+    "-Ypatmat-exhaust-depth",
+    "22",
```
thpani, line 60: I dislike that this is broken up, but it's an artifact of `scalafmtSbt`

build.sbt
thpani: Good points, PTAL?

shonfeder (approved):
Looks great!

PR #1481, opened 2022-03-14
Fix compiler deprecation warnings

thpani:
Towards #1473, #1064

Fixes all current compiler deprecation warnings.
This addresses two kinds of warnings; rationale for each fix in the commit message.


Kukovec (approved):
👍

PR #1482, opened 2022-03-14
Fix compiler feature warnings

### LINKED ISSUE #1473: Fix compiler deprecation and feature warnings

thpani:
Towards https://github.com/informalsystems/apalache/issues/1064. Closes #1473.

Fixes all current compiler feature warnings.
Rationale for each fix in the commit message.

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
@@ -1,22 +1,21 @@
 package at.forsyte.apalache.tla.bmcmt.smt
 
 import at.forsyte.apalache.io.OutputManager
-
-import java.io.PrintWriter
-import java.util.concurrent.atomic.AtomicLong
 import at.forsyte.apalache.tla.bmcmt._
 import at.forsyte.apalache.tla.bmcmt.profiler.{IdleSmtListener, SmtListener}
 import at.forsyte.apalache.tla.bmcmt.rewriter.ConstSimplifierForSmt
-import at.forsyte.apalache.tla.bmcmt.types.{BoolT, CellT, FinSetT, FunT, IntT, PowSetT}
+import at.forsyte.apalache.tla.bmcmt.types._
+import at.forsyte.apalache.tla.lir.UntypedPredefs._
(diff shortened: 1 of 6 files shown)


Kukovec (approved):
LGTM, but I'd prefer not having explanations inside git commit messages.

shonfeder:
@Kukovec why the preference against informative git messages? I tend to think these are great. But do you see them causing some sort of problem?

shonfeder (approved):
Looks awesome to me.

Was super easy to review this by commit, thanks to the careful organization and clear messages. Thanks!

Kukovec:
> @Kukovec why the preference against informative git messages? I tend to think these are great. But do you see them causing some sort of problem?

That's a loaded question. My preference isn't against informative messages, it's against information dispersion. As it stands, we have explanations a) in the issue b) in the PR c) in issue/PR comments d) in the source code comments, and now also in e) commit messages. (a)-(d) are searchable, and (a)-(c) show up in emails. (d) is persistently visible. To view PR comments, one has to either manually invoke git commands or check the GithubUI (where PR messages are greyed out and behind an "expand"/"..." button).

shonfeder:
I’ll echo Thomas in noting that each of these locations serves a different purpose.

- (a) is to direct and inform future work
- (b) is to communicate and give high-level context for merge requests (which conclude WIP)
- (c) is for dynamic dialogue and discussion around planning and review
- (d) is for elucidating/explaining complex code for a reader who is looking at the current state of the program (i.e. looking at the source code at one particular point in time)
- (e) is for leaving a persistent record of the reason for a *change* **or** (in the event of a squashed pr) to give granular insight into changes to inform review (i.e. looking at the source code across two points in time)

afaict, there is no really useful place to record the kind of granular information about changes that Thomas has provided here outside of commit messages. Perhaps the only other contender would be as comments on the pr? That can help with review but isn't actually recorded in way that remains useful after the branch is merged.

The kind of disciplined commits Thomas is making here will prove very useful if we ever need to, e.g., run git bisect to track down an issue or pull up got blame to see why certain lines were changed (which I recommend doing in your editor, btw!).

Issue #1320, opened 2022-02-10
[FEATURE] Use output management for `dependencies.dot` produced by `DeclarationSorter`


Kukovec:
These files just get dumped in the current working directory, instead of into `outdir`, with the rest of the outputs. Since we have centralized IO management, we should use it for these files as well.

shonfeder:
This one will require some refactoring of the current package dependencies:

- generation of `dependencies.dot` is in the `tlair` package
- `OutputManager` is in the `tla-io` package
- `tla-io` depends on `infra` and `infra` depends on `tlair`

So if we pulled `OutputManager` into `tlair` we'll end up with a cyclical
dependency: `tla-io` -> `infra` -> `tlair` -> `tla-io`.

Therefore, the first order of business is to break out the general purpose IO
handling into it's own package. This makes sense to me: I can't think of a good
reason why our manager for configurations and I/O should depend on other
infrastructure.

konnov:
On the other hand, many transformations in `at.forsyte.apalache.tla.lir.transformations.standard` would be better placed in `tla-pp`, as they are about preprocessing. I would rather move these transformations to `tla-pp`, keeping only the interfaces and maybe `ModuleByExTransformer`. Jure @Kukovec, what do you think?

PR #1489, opened 2022-03-16
Clean graphviz output after cyclic declaration tests

thpani:
`DeclarationSorter` writes a witness file `dependencies.dot` when it detects cyclic dependencies.
In unit tests with cyclic input: assert that this file is written, then remove it.

Ensures a clean working directory after running `make test`.


Kukovec (approved):
LGTM. I'm just going to ping #1320 here for visibility. We should definitely revisit it in Q2.

Issue #1352, opened 2022-02-16
[FEATURE] Register Apalache with scala-steward


shonfeder:
One unforeseen cost of switching from maven to sbt is that we lost use of dependabot (see https://github.com/dependabot/dependabot-core/issues/352). However, there is a special purpose bot for updating Scala dependencies: https://github.com/scala-steward-org/scala-steward#scala-steward

To use Scala Steward, it seems we just register our rep in an md file in that repo, then the bot will open PRs for us.

shonfeder:
Pending PR: https://github.com/scala-steward-org/repos/pull/970

shonfeder:
We're merged :)

PR #1494, opened 2022-03-16
Fix references to `pom.xml`

shonfeder:
There were 4 left over references to `pom.xml` from migrating to sbt (#1234).
This PR updates these locations to reflect the current build system.

    $ ag pom.xml
    CONTRIBUTING.md
    372:Assuming the current version recorded in the project's `pom.xml` files is
    
    tla-io/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporterStandardModules.scala
    500:    // We have to set TLA-Library, in order to look up Apalache.tla. This is done automatically in pom.xml.
    
    script/release-notes.sh
    5:# version of the software specified in the pom.xml.


shonfeder:
Whoops. I didn't mean to "convert" the issue into a PR, only to link them!

thpani (approved):
LGTM, thanks for clearing these up!

Out of curiosity, do we have something in place to replace dependabot? I know it doesn't support sbt.

shonfeder:
> Out of curiosity, do we have something in place to replace dependabot? I know it doesn't support sbt.

we've an issue with the proposed replacement, but it's not implemented yet: #1352

PR #1500, opened 2022-03-16
Implement optimization in function application

### LINKED ISSUE #1493: Add an optimization for function application
Similar to what @rodrigo7491 did in the arrays encoding, we can statically check, whether the function argument as a cell belongs to the function domain. If so, we can immediately return the result.

konnov:
Closes #1493. Implement the optimization in function application in the oopsla19 encoding. This optimization was introduced by @rodrigo7491 in the arrays encoding, see #1492 .


shonfeder:
The integration test failure is happening only on the array encoding, and it appears to be because the particular tests is exceeding our timeout. I've been able to reproduce it on my machine with the following invocation of the integration test:

```
SMT_ENCODING=arrays test/mdx-test.py "check TestFunctions.tla reports no error" --debug
```

Or more directly with 

```
apalache-mc check --smt-encoding=arrays --length=0 --inv=AllTests test/tla/TestFunctions.tla
```

Perhaps the running time is just at the threshold of our timeouts, which allowed the breakage to slip in?

I think our options are either: 

1. Disable this test for the arrays encoding
2. Increase the timeout (for this test, or perhaps just for when we run the arrays encoding tests?)
3. Improve the performance of the arrays encoding :)

rodrigo7491 (approved):
LGTM!

PR #1525, opened 2022-03-18
Inliner rework

Kukovec:

This is a re-implementation of inlining, which merges `InlinerOfUserOper`, `LetInExpander` and `CallByNameOperatorEmbedder`. The key difference is that the module declarations are sorted by dependency-order beforehand, then traversed iteratively (instead of pre-building a `BodyMap` "in parallel"). This should help avoid many bugs arising from tracking "stale" operator definitions.


Kukovec:
Note that this PR does not rewire `InlinePass` to use this class. However, I've run the integration tests locally for https://github.com/informalsystems/apalache/pull/1525/commits/d786b234cf25c5ac30b7f94eac55d6efd6d27796 with this rewiring implemented, and they all pass.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,238 @@
...
+  // main access method, performs the transformations described above
+  def transform(scope: Scope): TlaExTransformation = tracker.trackEx {
+    // Standard application
+    case OperEx(TlaOper.apply, NameEx(name), args @ _*) if scope.contains(name) =>
+      instantiateWithArgs(scope)(name, args)
+
+    // Roundabout application, happens if you pass a lambda to a HO operator
+    case OperEx(TlaOper.apply, LetInEx(NameEx(name), decl @ TlaOperDecl(declName, params, _)), args @ _*)
+        if name == declName && params.size == args.size =>
+      // not 100% optimal, but much easier to read/maintain: just push the lambda into scope and apply the std case
```
thpani, line 160: Why not optimal, what could be improved?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,238 @@
...
+}
+
+object Inliner {
+
+  type Scope = BodyMap
+  def emptyScope: Scope = Map.empty
+
+  type DeclFilter = TlaOperDecl => Boolean
+
+  def negateFilter(df: DeclFilter): DeclFilter = d => !df(d)
```
thpani, line 225: Is this only in the companion object because it's static in principle?
I'd prefer to see it in the class, next to `nonNullaryFilter`.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,238 @@
...
+
+  object FilterFun {
+    def ALL: DeclFilter = _ => true
+    def NAMED(set: Set[String]): DeclFilter = d => set.contains(d.name)
+  }
+
+  // Checks for the specific LET-IN form used by LAMBDA and call-by-name
+  def isCallByName(letInEx: LetInEx): Boolean = letInEx match {
```
thpani, line 233: As above, is this here because it's static?
```suggestion
  private def isCallByName(letInEx: LetInEx): Boolean = letInEx match {
```

thpani (dismissed):
I've only made it through the Inliner, will look at tests tomorrow.
Nice code, this was a joy to read! I have some questions for clarification and some minor stylistic remarks.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Given a module m, with global operators F1,...,Fn, Inliner performs the following transformation:
+ *   - For each i, replaces all application instances Fi(a1i, ..., aki) in m, where Fi(p1i,...,pki) == ei, with
+ *     ei[a1/p1i, ..., aki/pki]
+ *   - Replaces each instance of LET G1(q11,...,qk1) == g1 ... Gm(q1m, ..., qkm) == gm IN f, with:
+ *     - f, inlined by the previous rule, as if G1, ..., Gm were global, if `keepNullary` is `false`, or,
+ *     - LET Gj1 == gj1 ... Gjl == gjl IN ff, where Gj1, ..., Gjl are all the nullary operators among G1,...,Gm and ff
+ *       is the expression obtained by inlining f as though all G1,...,Gm, that are not among Gj1, ..., Gjl, were global
+ *   - For each instance of call-by-name A (where a definition A(p1,...,pk) == e exists in scope), replaces
+ *     - A with LET A_LOCAL(p1,...,pk) = e IN A_LOCAL
```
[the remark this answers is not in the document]
Kukovec: This was (is?) a consideration for a future CMD flag. 
Basically, if `keepNullary` is set, we keep `LET A == ... IN e`, because it turns out that this sort of manual caching is actually easier on the solver with the current encoding. However, depeneding on what sort of encoding we use in the future, it might be more beneficial to fully get rid of LET-IN expressions, in which case you'd set `keepNullary` to false.
TLDR, having this as a flag lets us tune performance, where it is currently the case that `true` is the more optimal value.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,238 @@
...
+            s"Inliner: Unable to unify generic signature $genericType of ${decl.name} with the concrete type $targetType",
+            decl.ID)
+
+      case Some(pair) => pair
+    }
+  }
+
+  // Assume name is in scope.
+  // Creates a fresh copy of the operator body and replaces formal parameter instances with the argument instances.
+  private def instantiateWithArgs(scope: Scope)(name: String, args: Seq[TlaEx]): TlaEx = {
```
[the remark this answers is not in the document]
Kukovec: We could, but `embedCallByName` needs `NameEx` for type information, whereas `instantiateWithArgs` doesn't, and so the 1st line would just be `val name = nameEx.name`

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,238 @@
...
+
+    // To make a local definition, we use a fresh name
+    val newName = nameGenerator.newName()
+    val newLocalDecl = TlaOperDecl(newName, decl.formalParams, newBody)(tpTag)
+
+    LetInEx(NameEx(newName)(tpTag), newLocalDecl)(tpTag)
+  }
+
+  // main access method, performs the transformations described above
+  def transform(scope: Scope): TlaExTransformation = tracker.trackEx {
```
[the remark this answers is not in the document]
Kukovec: That'd break all the tests. Either
```suggestion
  private[pp] def transform(scope: Scope): TlaExTransformation = tracker.trackEx {
```
or 
```suggestion
  def transformEx: TlaExTransformation = transform(emptyScope)
  private def transform(scope: Scope): TlaExTransformation = tracker.trackEx {
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
Kukovec: Technically you could just do the part that's under
```scala
case opDecl: TlaOperDecl => ...
```
and ignore the `List[TlaDecl]` accumulation, but that's a reeeeeeally minor performance benefit.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
Kukovec: Yeah, I moved this and `isCallByName` out because they were static.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
Kukovec: Why? Also, Not sure if this will even work with `import Inliner._` in the class above.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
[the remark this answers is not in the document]
Kukovec: But also, there's no real downside to having it public as is.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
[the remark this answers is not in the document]
thpani: Thanks, I got what `keepNullary` does, but not how I would use it. I guess we can mark this as resolved.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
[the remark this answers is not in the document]
thpani: Right, but it would express a contract on the caller's side that `name` is not an arbitrary string.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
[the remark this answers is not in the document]
thpani: I find it confusing that this has two entry points given that it's a `TlaModule => TlaModule`. Can we do `transformEx` and add a Scaladoc that clarifies that module-level transformation is provided through `apply()`?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
thpani: I'd remove the `not 100% optimal...` then (or explain it).

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
thpani: It'd work, companion object/class have access to each other's private fields.
But I just saw that this is used in the unit tests, never mind.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
thpani: So are `nonNullaryFilter` and `negateFilter` public API? Looks like they're only used to set up the default parameters of `pushDeclsIntoScope`.
```suggestion
  private def negateFilter(df: DeclFilter): DeclFilter = d => !df(d)
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,237 @@
...
+ *     - A with LET A_LOCAL(p1,...,pk) = e IN A_LOCAL
+ *
+ * @author
+ *   Jure Kukovec
+ */
+class Inliner(
+    tracker: TransformationTracker,
+    nameGenerator: UniqueNameGenerator,
+    keepNullary: Boolean = true,
+    toplevelFilter: Inliner.DeclFilter = FilterFun.ALL)
```
thpani, line 32: Since we expose `topLevelFilter`, can we add tests for `FilterFun.NAMED`?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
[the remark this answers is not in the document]
Kukovec: It's not that unusual for transformations to have two versions: a `TlaModule => TlaModule` and ad `TlaEx => TlaEx` . I suppose we should get rid of `extends TlaModuleTransformation` in this case and define a `transformEx` and a `transformModule` instead of `apply`.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
Kukovec: I guess this is a matter of preference, since both are side-effect free. My default philosophy is that 
1. `object` methods are public and all `class` methods are private except for a handful of entry points. 
2. Static methods belong in the `object`

In this case, since `nonNullaryFilter` relies on a class parameter, you end up with one `nonNullaryFilter` and `negateFilter` having different privacy levels as a consequence.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
Kukovec: These will come in once `Inliner` is plugged into `InlinePassImpl` in a follow-up PR, since `InlinePassImpl` uses `keepRelevant` to filter out toplevel declarations.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,237 @@
...
+  type Scope = BodyMap
+  def emptyScope: Scope = Map.empty
+
+  type DeclFilter = TlaOperDecl => Boolean
+
+  def negateFilter(df: DeclFilter): DeclFilter = d => !df(d)
+
+  object FilterFun {
+    def ALL: DeclFilter = _ => true
+    def NAMED(set: Set[String]): DeclFilter = d => set.contains(d.name)
```
konnov, line 228:
```suggestion
    def RESTRICT_TO(names: Set[String]): DeclFilter = d => names.contains(d.name)
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  // Checks for the specific LET-IN form used by LAMBDA and call-by-name
+  def isCallByName(letInEx: LetInEx): Boolean = letInEx match {
```
konnov, line 232: This form is only used to encode lambdas. Why don't we call it `isLambda`? Technically, `F(x)` is always call by name, not by value.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestInliner.scala
```diff
@@ -0,0 +1,349 @@
...
+import at.forsyte.apalache.tla.lir.oper.ApalacheOper
+import at.forsyte.apalache.tla.lir.transformations.impl.IdleTracker
+import at.forsyte.apalache.tla.pp.Inliner.emptyScope
+import org.junit.runner.RunWith
+import org.scalatest.BeforeAndAfterEach
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatestplus.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class TestInliner extends AnyFunSuite with BeforeAndAfterEach {
```
konnov, line 15: I think it's helpful to have all these unit tests. But we should also introduce comprehensive integration tests for the inliner, similar to `TestSets.tla` and `TestSequences.tla`.

konnov (approved):
This code looks much better than the previous implementation! I still find that it's hard to grasp the exact description of the input and the output. We have to write an ADR that specifies the behavior of the inliner.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestInliner.scala
Kukovec: Since inlining is a fundamental transformation the entire collection of integration tests also acts as tests for inlining. What would a `TestInlining.tla` even look like?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+  // Default scope filter, we add nullary operators only if keepNullary is disabled
+  private def nonNullaryFilter(d: TlaOperDecl): Boolean = !keepNullary || d.formalParams.nonEmpty
+
+  // Given a declaration (possibly holding a polymorphic type) and a monotyped target, computes
+  // a substitution of the two. A substitution is assumed to exist, otherwise TypingException is thrown.
+  private def getSubstitution(targetType: TlaType1, decl: TlaOperDecl): (Substitution, TlaType1) = {
+    val genericType = decl.typeTag.asTlaType1()
+    new TypeUnifier().unify(Substitution.empty, genericType, targetType) match {
```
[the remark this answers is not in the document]
Kukovec: I mean, it might also be a monotype, in which case you'd get the empty substitution.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestInliner.scala
konnov: It may duplicate the unit tests and richer combinations of inlining. Tests in TLA+ are way easier to read than the tests in IR.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,237 @@
...
+      case Some(pair) => pair
+    }
+  }
+
+  // Assume an operator declaration named name is in scope.
+  // Creates a fresh copy of the operator body and replaces formal parameter instances with the argument instances.
+  private def instantiateWithArgs(scope: Scope)(name: String, args: Seq[TlaEx]): TlaEx = {
+    val decl = scope(name)
+
+    // Note: it can happen that the body and the decl have desynced types,
```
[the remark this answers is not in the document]
Kukovec: Oh, I guess this comment is a bit misplaced. The "body" here should refer to `replacedBody`. I moved some comments around and this must have slipped.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -0,0 +1,237 @@
...
+    // due to the way type-checking is currently implemented for polymorphic operators.
+    // We fix that below with type unification.
+    val freshBody = deepCopy(decl.body)
+
+    // Each formal parameter gets instantiated independently.
+    val replacedBody = decl.formalParams.zip(args).foldLeft(freshBody) { case (partialBody, (fParam, arg)) =>
+      ReplaceFixed(tracker)(NameEx(fParam.name)(arg.typeTag), arg)(partialBody)
+    }
```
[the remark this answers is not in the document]
Kukovec: Yep

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
Kukovec: Hmm, I guess a better name would have been `isPassByName`, since this refers to when operator names are used as arguments to HO operators.

Kukovec:
Ok, the only thing missing now are integration tests, but since this PR doesn't actually rewire `InlinePass` yet, those would not be able to test the new inliner anyway, so I'll be merging this and adding integration tests with #1566 .

PR #1527, opened 2022-03-18
Bags In Apalache

Kukovec:

Custom implementation of the `Bags` module, with type annotations.


Kukovec:
Just some context for people who were not present during the Zoom call where this was discussed:
The reason `SubBag` is unsupported is that, no matter how you try to implement it, you need to instantiate some set of the form `1..n`, where `n` is a dynamic upper bound (and you need the set as a value itself, you can't quantify over it and make use of the constraint rewriting).

konnov (approved):
Looks awesome! Let's merge

PR #1542, opened 2022-03-22
Make warnings fatal in CI

### LINKED ISSUE #1064: Fail CI on compiler errors

shonfeder:
Closes #1064

This setting is controlled via the environment variable
`APALACHE_FATAL_WARNINGS`, making it possible for developers to opt in
or out of having compiler warnings fail their builds locally.

Instructions on configuration have been added to CONTRIBUTING.md

-------

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -139,6 +139,19 @@ The necessary shell environment is specified in [.envrc](./.envrc). You can:
 
 [direnv]: https://direnv.net/
 
+#### Compiler warnings
+
+We fail CI builds on compiler warnings.
+
+To have compiler warnings also fail on local builds, set
+
+```sh
+export APALACHE_FATAL_WARNINGS=true
+```
+
+You may add this to your `.local-envrc` file at the root of this repo to have
+`denvenv` load this into your development environment.
+
 ## Development Environment
 
 If you use a different development environment or editor set up, please document
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,7 +150,7 @@ export APALACHE_FATAL_WARNINGS=true
 ```
 
 You may add this to your `.local-envrc` file at the root of this repo to have
-`denvenv` load this into your development environment.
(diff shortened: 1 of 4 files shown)


shonfeder:
Here's proof that the new settings will cause CI to fail on compiler warnings:

image

CONTRIBUTING.md
```diff
@@ -139,6 +139,19 @@ The necessary shell environment is specified in [.envrc](./.envrc). You can:
...
+We fail CI builds on compiler warnings.
+
+To have compiler warnings also fail on local builds, set
+
+```sh
+export APALACHE_FATAL_WARNINGS=true
+```
+
+You may add this to your `.local-envrc` file at the root of this repo to have
+`denvenv` load this into your development environment.
```
thpani, line 153: I think this is a typo?

```suggestion
You may add this to your `.local-envrc` file at the root of this repo to have
`direnv` load this into your development environment.
```

thpani (approved):
LGTM, one small change inline.

Thank you for adding the option to also fail local builds!
I'm pretty sure I'd overlook warnings otherwise 😅

CONTRIBUTING.md
shonfeder: Yep. thanks!

PR #1539, opened 2022-03-21
Support for SequencesExt

konnov:
Closes #1486. This PR contains alternative definitions for `SequencesExt` of the community modules. This module required a lot of work. All comments are derived from the original spec: https://github.com/tlaplus/CommunityModules/blob/master/modules/SequencesExt.tla

This PR also extends `MkSeq`, as was required by more complex use cases in `SequencesExt.tla`. These extensions are intended for the internal use.

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [ ] Documentation added for any new functionality
- [x] Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality

PR #1543, opened 2022-03-23
fix a bug in default values

konnov:
This is a one-line fix for the bug discovered in #1539.


thpani (approved):
LGTM, should we have a test for this?

konnov:
I also thought about this. We don't have tests for caches :(

konnov:
I have opened the issue #1544 on that

Issue #1541, opened 2022-03-22
Pack arithmetic expressions


thpani:

thpani:
Whoops, I didn't mean to convert this into an issue... closing.

PR #1545, opened 2022-03-23
Pack integer arithmetic comparisons in IntCmpRule

### LINKED ISSUE #1342: [FEATURE] Pack arithmetic operations in a single SMT expression

thpani:
Closes #1342 

Pack arithmetic expressions in `IntCmpRule` and `EqRule`. Translate TLA+ arithmetic comparisons into a single SMT constraint, without introducing intermediate cells.

Uses the same logic for packing arithmetic expressions as #1541, which is refactored into a trait in 835e717.

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/IntArithPacker.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/IntArithPacker.scala
@@ -19,14 +19,14 @@ trait IntArithPacker {
    * Rewrite [[state]]'s expression into an expression that is purely arithmetic, referring to arena cells for
    * non-arithmetic subexpressions.
    *
-   * @param state
-   *   current rewriter state, `state.ex` is the arithmetic expression to rewrite
    * @param rewriter
    *   rewriter to use for non-arithmetic subexpressions
+   * @param state
+   *   current rewriter state, `state.ex` is the arithmetic expression to rewrite
    * @return
    *   the rewritten state
    */
-  protected def packArithExpr(state: SymbState, rewriter: SymbStateRewriter): SymbState = state.ex match {
+  protected def packArithExpr(rewriter: SymbStateRewriter, state: SymbState): SymbState = state.ex match {
     /* *** base case: integer literals + names *** */
 
     // keep integer literals as-is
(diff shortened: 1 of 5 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/IntArithPacker.scala
```diff
@@ -0,0 +1,56 @@
...
+   * non-arithmetic subexpressions.
+   *
+   * @param state
+   *   current rewriter state, `state.ex` is the arithmetic expression to rewrite
+   * @param rewriter
+   *   rewriter to use for non-arithmetic subexpressions
+   * @return
+   *   the rewritten state
+   */
+  protected def packArithExpr(state: SymbState, rewriter: SymbStateRewriter): SymbState = state.ex match {
```
konnov, line 29:
```suggestion
  protected def packArithExpr(rewriter: SymbStateRewriter, state: SymbState): SymbState = state.ex match {
```

I usually collect the longer-living objects on the left. This is of course a matter of taste.

konnov (approved):
Looks good!

PR #1491, opened 2022-03-16
Lookup modules located next to the file supplied on CLI

thpani:
Look up modules in the directory containing the main TLA+ file supplied on CLI. This was requested multiple times #426, #1477 and is what TLC does for module lookup.
Drop support for lookup in `TLA-Library`, unless the main file has no parent directory.
Drop support for `TLA_PATH`, since it becomes virtually useless.

This mirrors the behavior of TLC since https://github.com/tlaplus/tlaplus/commit/b905a86e31714bf5002290141b51b41ffe280c8e.

If we need `TLA_PATH`-like functionality down the road, we should parse the env variable in Apalache, then pass it into the filename resolver `SanyNameToStream` as additional library path.

The only remaining use of `TLA-Library` is to wire in the Apalache standard library in `src/tla/` for unit tests.

Fixes #426. Fixes #1477.

PR #1553, opened 2022-03-23
Find Apalache standard library outside of fat JAR

### LINKED ISSUE #1550: Find Apalache standard library outside of fat jar
#1491 removed support for `TLA_PATH` and `-DTLA-Library`.

Since then, Apalache fails to lookup modules from its standard library when run from

- `make run` (viz `tool / run <args>`)
- an IDE.

For unit tests,
https://github.com/informalsystems/apalache/blob/e42c54f1c4ad362cbd905f1db01bb3b566870e43/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala#L99 currently handles this by passing `useRelativePathLookup=false` to `loadFromFile`, which initializes a `SanyNameToStream` that honors `-DTLA-Library`.

thpani:
Fixes #1550

#1491 dropped support for `-DTLA-Library` except for unit tests. This fails (some) other runs that don't utilize the fat JAR, e.g., `make run` (viz `sbt tool / run <args>`) or running from an IDE, when the tool must look up TLA+ modules from Apalache's standard library.

I propose to fix only what is broken, which is looking up the standard library when not running from the fat JAR.
To this extent, in this PR we:

- Introduce the environment variable `APALACHE_HOME` (we already use this name throughout the codebase to refer to the root of the Apalache source tree)
- Only if _not_ running from the fat JAR: Read `APALACHE_HOME` and add it to the library lookup path in the same place where the Apalache standard lib would be looked up in the fat JAR.
- Update `.envrc` and docs on running from IDEs accordingly.

Note that this uncovered two bugs in unit tests with shadowed module names that are fixed in 998f95cd7. I think this demonstrates why we should be extra careful about shadowing when it comes to library lookup.

Curious about your comments 🤩


CONTRIBUTING.md
```diff
@@ -212,6 +212,10 @@ IDEA](https://www.jetbrains.com/idea/) and set up a new project. If you already
 have IntelliJ installed, please ensure you using version 2021.3.1 or later.
 We've had reports of build failures with earlier versions.
 
+For some tests to succeed, you will have to [set the environment
+variable](https://www.jetbrains.com/help/objc/add-environment-variables-and-program-arguments.html#add-environment-variables)
+`APALACHE_HOME` to the root of the Apalache source tree.
```
shonfeder, line 217: This is not accurate, afaik. No tests should require this. This is just about debugging practices used in development, not about testing.

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -16,37 +17,56 @@ import scala.util.{Failure, Success, Try}
...
+        // Running from outside the fat JAR, we don't have a standard library bundled.
+        // Add $APALACHE_HOME/src/tla/ to lookup paths.
+        System.getenv("APALACHE_HOME") match {
+          // Warn if environment variable APALACHE_HOME is not set
+          case null =>
+            logger.warn("Not running from fat JAR and APALACHE_HOME is not set;")
+            logger.warn("will not be able to find the Apalache standard library.")
+            Nil
+          case apalacheHome: String => Seq(s"${apalacheHome}/src/tla")
+        }
```
shonfeder, line 63: I really don't like the idea that we have two alternative module lookup strategies coded into the application logic in order to accommodate a particular development practice. I think it'd be much more preferable to expose the general purpose functionality we've discussed.

This looks to me like we're hardcoding a hack to inspect our own source code, and I don't understand why this is preferable to supporting general logic for looking up tla modules.

shonfeder (commented):
Given my current understanding of the tradeoffs and costs, I'm not in favor of this approach.


Could you explain the connection with these changes and the bugs you noted regarding module lookup? It's not clear to me how those unit tests lookups are related to issues around path loading.

tla-io/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporterStandardModules.scala
```diff
@@ -1,22 +1,21 @@
...
 
 /**
- * <p>A collection of tests that check how SanyImporter loads the standard modules.</p>
+ * A collection of tests that check how SanyImporter loads the standard modules.
  *
- * <p>If you run this test in an IDE, and the test fails, add the following line to the VM parameters (don't forget to
- * replace <APALACHE_HOME> with the directory where you checked out the project):
- * -DTLA-Library=<APALACHE_HOME>/src/tla </p>
+ * If you run this test in an IDE and the test fails, set environment variable `APALACHE_HOME` to the root of the
+ * Apalache source tree.
```
shonfeder, line 18: I would argue that if these tests require the file layout we produce after assembling the package, then they should moved into integration tests: IMO, we should not hack the environment to make things that are testing integration look like unit tests.

shonfeder (approved):

CONTRIBUTING.md
thpani: Some unit tests have always required setting `-DTLA-Library` when run from IDEA, see [TestSanyImporterStandardModules.scala](https://github.com/informalsystems/apalache/pull/1553/files/998f95cd75bdcc33bcd0093224b5114cfb2b26fc#diff-3a0cbae0ae23361e326f995194bf7a9d207190b62f666d3eeb5ef75943bc4ccd)

tla-io/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporterStandardModules.scala
thpani: We have done so before by passing `-DTLA-Library`, no? What has changed?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
thpani: We will always have two alternative module lookup strategies, no matter what. Fat JAR looks up from the baked-in files, anything else doesn't. Why would we introduce a way that potentially affects both, possibly shadowing the fat JAR lookup, if we can just fix the one that needs to do an extra lookup?

shonfeder:
afaict, the disagreement on the best approach here seems to trace back to some basic difference in view of the tradeoffs and risks of the different approaches. I think it might help if you'd articulate what your concerns are around supplying something like a `TLA_PATH`, and why you think that approach should be avoided.

I'll also try to summarize my concerns with the approach proposed here:

I see a risk to complicating our application logic with special treatment for development practices, and I see a benefit to making the application more configurable.

In my experience, putting special allowance for development practices into core application logic introduces the risk that later changes or features end up depending on this exceptional behavior. It also complicates the business logic in way that is purely accidental, in so far as it is only required to enable that development practice, and doesn't actually correspond to any real need of the problem domain. 

On the other hand, I think we end up with more useful and robust software by ensuring that the core application logic is focused on addressing the needed functionality, and providing flexible configuration that allows testing and developing as needed, but sing the same application logic that we deliver to users.  In my view, this allows us to keep the application logic cleaner (i.e., avoiding complications like hard-coded support for special file locations in the source code) while also pushing us to provide a more flexible and configurable tool for users.

tla-io/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporterStandardModules.scala
shonfeder: I think this is a sign that these always should have been integration tests :)

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
shonfeder: I'm not opposed to having multiple module lookup strategies. I'm for it! What I'm opposed to is having one special module lookup strategy that only applies when the code is run in a pre-assembly, development state.

CONTRIBUTING.md
shonfeder: Yes, I see that now, thank you. I think this concern is subsumed by my view that those tests should be moved into integration tests.

thpani:
> Could you explain the connection with these changes and the bugs you noted regarding module lookup? It's not clear to me how those unit tests lookups are related to issues around path loading.

Consider one of the unit tests fixed in [998f95c](https://github.com/informalsystems/apalache/pull/1553/commits/998f95cd75bdcc33bcd0093224b5114cfb2b26fc) shortened and dumped into a file:

    $ reals.tla
    ---- MODULE reals ----
    EXTENDS Reals
    RealSet == Real
    ==================

 TLC says
 
    $ java -jar ~/Downloads/tla2tools.jar reals.tla
    TLC2 Version 2.17 of 02 February 2022 (rev: 3c7caa5)
    Fatal errors while parsing TLA+ spec in file reals
    [...]
    Unknown location
    File name 'Reals' does not match the name 'reals' of the top level module it contains.

Apalache says

    $ apalache-mc parse reals.tla
    [...] 
    Parse reals.tla                                                   I@19:27:00.228
    PASS #0: SanyParser                                               I@19:27:00.234
    Parsing file /Users/thomas/src/apalache/unstable/reals.tla
    Parsing file /Users/thomas/src/apalache/unstable/reals.tla
    Error by TLA+ parser: *** Abort messages: 1
    
    Unknown location
    
    File name 'Reals' does not match the name 'reals' of the top level module it contains.

Yet our unit test happily passed for the past 5 years.

The argument in https://github.com/informalsystems/apalache/pull/1553#discussion_r833567274 was that these tests should probably be integration tests. Maybe; but how do we know which tests those are? Shouldn't we still be able to catch these errors when run as unit tests?

> I think it might help if you'd articulate what your concerns are around supplying something like a `TLA_PATH`, and why you think that approach should be avoided.

This is a good idea, I will follow up tomorrow morning!

tla-io/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporterStandardModules.scala
thpani: If I need to know what code paths my test hits internally in order to decide whether it's a unit test or an integration test, something's wrong.

shonfeder: To my mind, this is not about looking at what code paths it hits internally, it's about considering whether or not we can test an isolated unit, or whether we are testing the integration of our system with (relatively) external systems. In this case, we are testing the integration with the file system and the external dependency on SANY's importing logic. That seems like a clear case of integration testing to me. What do you think? Maybe we are disagreeing on terms here?

I guess, thinking strictly, if we wanted a unit test for this same logic, we'd stub or mock out SANY's import logic.

thpani:
I think we disagree on a few basic assumptions. Mine lead me to believe that this issue should not be addressed by `TLA_PATH`, so I will lay them out here:

1. If I run the packaged JAR from the source tree, it should behave _exactly_ as it does when a user runs it from a release archive.
2. If I run anything other than the packaged JAR (i.e., unit tests, debugger runs, from IntelliJ...), it should behave as close to the JAR as possible.
3. If we were to use `TLA_PATH` to patch the Apalache stdlib into the runs described by (2), we would need to ensure that `TLA_PATH` is only altered to include the stdlib on runs of (2), and not those of (1).
4. I cannot think of any way of achieving this in a uniform, non-flaky way. In particular, this approach would need to either (a) patch the stdlib into `TLA_PATH` on all runs of (2) which can only be achieved by doing it in various places, or (b) filter the stdlib out of `TLA_PATH` on runs of (1) which would mean we prescribe a specific way of running the JAR.
5. In fact, the only reliable way of determining whether the JAR is being run, is from within the Apalache code, which is what this PR proposes. 

- Risks of using `TLA_PATH`: It looks to me that it will lead to a non-uniform, flaky way of patching in the Apalache standard library. It will cause weird issues with module lookup down the road, when we forget to mangle `TLA_PATH` in the right way, or when someone accidentally exports `TLA_PATH` with the standard library included when it shouldn't be. These issues will be super hard to debug. That said, I'm happy to look at a PR that addresses point (4) from above without its downsides.

- Benefits of using `TLA_PATH`:
  - It would fix the issue without putting separate logic into the Apalache source code. My concern is that this logic will instead be scattered across `.envrc` files, run scripts, the `build.sbt`, etc.
  - It gives a more general way of extending the module lookup path. My concerns from above lead me to believe that the present issue should be addressed in a narrower, more fail-safe and future-proof way.

### On going with this PR

#### Risks

I think there's also a few misconceptions about what this PR proposes:

> putting special allowance for development practices into core application logic

> It also complicates the business logic in way that is purely accidental, in so far as it is only required to enable that development practice, and doesn't actually correspond to any real need of the problem domain.

This PR does not make special allowance for a single development practice. It fixes Apalache stdlib lookup for _everything_ except when running the JAR (which is not broken in the first place). Again, it fixes all of running Apalache code from IDE, sbt, command line debugging, ... and whatever future use case we may have.

> later changes or features end up depending on this exceptional behavior

This PR does not introduce exceptional behavior. It makes sure the stdlib is looked up when and where it should, when it currently is not.

> [On using `TLA_PATH`:] I think we end up with more useful and robust software by ensuring that the core application logic is focused on addressing the needed functionality, and providing flexible configuration that allows testing and developing as needed, but sing the same application logic that we deliver to users.  In my view, this allows us to keep the application logic cleaner (i.e., avoiding complications like hard-coded support for special file locations in the source code) while also pushing us to provide a more flexible and configurable tool for users.

In my mind, the core application logic is currently not addressing the needed functionality, which is looking up the stdlib on all runs.

I'm not against supporting `TLA_PATH`. I'm against using it to address this specific issue, see my concerns above.

#### Benefits

Main benefit of going with this PR: It is future-proof. Whatever new development practice / way of running Apalache we introduce, it is already handled in an automated fashion.

tla-io/src/test/scala/at/forsyte/apalache/tla/imp/TestSanyImporterStandardModules.scala
thpani: I agree that this is, conceptually, an integration test. I also believe that SANY is such an important dependency that we should be able to run Scalatest fixtures against it, the same we way currently run unit tests.

konnov:
Another option is to enable the additional lookup from `APALACHE_HOME` only if the `--debug` flag is supplied. Then we can use the release mode by default and the debug mode, only when we need it.

shonfeder:
Thanks for the careful explanation.  I think it has really helped me understand the points of disagreement.

> 1.  If I run the packaged JAR from the source tree, it should behave exactly as it does when a user runs it from a release archive.

I mostly agree on this point in principle, but I think my POV differs in that I think of "behaving exactly the same" 
in this case to mean, "loading modules in the same ways and from the same locations".  But the 
But I can understand your perspective, that "behaving the  same" means being able to load all the 
same TLA modules that we end up injecting into the tla2tools  jar.

I think the key difference here is I'm thinking of the `Apalache.tla` and friends as external resources, but you are thinking of them as part of the source code. Your view is probably closer to how Igor and Jure think of them.

If we consider `Apalache.tla` and friends as part of the source, then I think the most "correct" way to address this might be to vendor the tla2tools jar, and add in our additions to the jar's resources when they are compiled as a dependency, rather then injecting during packaging. Of course, that increases our maintenance burden tho.

If we consider `Apalache.tla` and friends as external dependencies, then the "correct" way might be only provide them via library extensions via something like `TLA_PATH`, and not slip them into our dependencies during packaging.

Instead, afaiu, this PR takes a middle course, and I think it's this ambiguity that is disconcerting to me.

I think Igor is right in pointing out that this change would effectively introduces two different compilation modes, inferred implicitly by the presence or absence of the `Apalache.tla` file in the resource directory.

One issue I that could arise from this change is that devs might add new modules to `src/tla`, and their dev and unit tests will pass, but the black box integration tests will fail, because they have not included the module in the `build.sbt`. This is inverse of our current problem. But we should catch that in the integration tests.

This isn't the course I'd prefer, but I won't block it if the rest of the time thinks it's the right way to address the issue!

build.sbt
```diff
@@ -86,9 +86,9 @@ ThisBuild / semanticdbVersion := scalafixSemanticdb.revision
 
 // NOTE: Include these settings in any projects that require Apalache's TLA+ modules
 lazy val tlaModuleTestSettings = Seq(
-    // we have to tell SANY the location of Apalache modules for the tests
-    Test / fork := true, // Forking is required for the system options to take effect in the tests
-    Test / javaOptions += s"""-DTLA-Library=${(ThisBuild / baseDirectory).value / "src" / "tla"}""",
+    // Forking is required for the system options to take effect in the tests
+    // It also ensures that tests run from their respective sub-project directories
+    Test / fork := true
 )
```
shonfeder, line 92: Do we still need the forking if we don't need set this option any more? What is the importance of running from the sub-project directories?

Regardless, I think the comment about the "system options" can be removed, since those are not being set anymore?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -16,37 +17,56 @@ import scala.util.{Failure, Success, Try}
...
-      file.getAbsoluteFile().getParent() match {
-        case null          => SanyNameToStream()
-        case parentDirPath => SanyNameToStream(parentDirPath)
+    // Construct library lookup path:
+    //   1. Include the file's parent directory
+    val parentDirPath = file.getAbsoluteFile().getParent() match {
+      case null                  => Nil
+      case parentDirPath: String => Seq(parentDirPath)
+    }
+    //   2. Patch in Apache standard library.
```
shonfeder, line 43:
```suggestion
    //   2. Patch in Apalache standard library.
```

shonfeder (approved):

thpani:
Excellent, I think we understand each other's concerns now 😄 
I agree that this PR very much addresses the status quo, and maybe we want a different way going forward.

IMO the way we are currently packaging the Apalache TLA+ stdlib is not optimal; it obviously leads to issues. If this can be avoided, I'm all for it!

> vendor the tla2tools jar, and add in our additions to the jar's resources when they are compiled as a dependency, rather then injecting during packaging

I don't understand this part. Could you sketch how it would be implemented and maybe estimate how much time this would take?

> If we consider `Apalache.tla` and friends as external dependencies, then the "correct" way might be only provide them via library extensions via something like `TLA_PATH`, and not slip them into our dependencies during packaging.

I can totally see this in the future, but I think it would need a lot of infrastructure work to do it right. This essentially requires a package management infrastructure like pip or CTAN. The current approach allows us to version and distribute the Apalache stdlib in lock-step with the source code. I'm also worried that this would over-complicate things for first-time users, who just want to play around with Apalache and take it for a test-drive.

shonfeder:
That’s a very good point about the status quo and our room for future improvement.

That perspective helps convince me this is a sensible solutions for the current state of things. Thanks for helping hash this out!  I’ll follow up with proper relies on Monday!

build.sbt
thpani: There's two issues here, paths and failing tests due to parallelism; I've opened https://github.com/informalsystems/apalache/issues/1577 to follow up on this.
The comment's fixed in e56c96e9

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
thpani: I think we've hashed this out in the comments below.

.envrc
```diff
@@ -18,6 +18,9 @@ exports () {
 
     # Add executables to path
     export PATH=$DIR/bin:$PATH
+
+    # Base path for looking up the Apalache standard library
+    export APALACHE_HOME=$DIR
```
thpani, line 23: Should we choose a name that more explicitly says that this is a dev-only variable?
Something like `APALACHE_INTERNAL_HOME` or `APALACHE_DEV_HOME`?

FWIW, even if a user sets this, nothing bad will happen, as we only use it when not running from the packaged JAR.

Kukovec (approved):
LGTM.

shonfeder (approved):
LGTM too! Thanks again for working through the options here :)

PR #1568, opened 2022-03-28
ADR on inlining

### LINKED ISSUE #1567: Inliner ADR
Since #1525 changes how the inliner works, and since the current approach wasn't documented all that well (in the manual), we need an ADR that describes what `InlinerPass` is supposed to do.

Kukovec:

closes #1567


docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+## Context
+
+<!-- Communicates the forces at play (technical, political, social, project).
+     This is the story explaining the problem we are looking to resolve.
+-->
+The term "inlining" typically refers to the process of replacing instances of operator application `A(x1,...,xn)`, with `e[x1/p1,...,xn/pn]` - the expression obtained by replacing each instance of `pi` with `xi` -  where the operator `A` is defined as `A(p1,...,pn) == e` and is _non-recursive_.
+We elect to use the term in a broader sense of "replacing an operator with its definition", in the two ways listed below:
+  1. Standard inlining: the instantiation described above
+      1. Non-nullary inlining: the instantiation described above, except the inlining skips nullary LET-defined operators
+  2. Pass-by-name inlining: replacing an operator name `A` with a local LET-definition: `LET A_LOCAL(p1,...,pn) == A(p1,...,pn) IN A_LOCAL`
```
konnov, line 31: Before giving these definitions, please introduce "nullary LET-defined operators" and explain the need for local LET-definitions.

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+
+This ADR outlines the transformations that happen in the Apalache inliner pass.
+Since we have recently reworked the inliner in #1569, we saw it fit to document exactly how inlining is supposed to work and why the inlining pass is the way it is. 
+
+## Context
+
+<!-- Communicates the forces at play (technical, political, social, project).
+     This is the story explaining the problem we are looking to resolve.
+-->
+The term "inlining" typically refers to the process of replacing instances of operator application `A(x1,...,xn)`, with `e[x1/p1,...,xn/pn]` - the expression obtained by replacing each instance of `pi` with `xi` -  where the operator `A` is defined as `A(p1,...,pn) == e` and is _non-recursive_.
```
konnov, line 27: Since it is a foundational document for inlining, we have to explain that there a top-level operators and nested operators that are defined with LET-IN.

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+We elect to use the term in a broader sense of "replacing an operator with its definition", in the two ways listed below:
+  1. Standard inlining: the instantiation described above
+      1. Non-nullary inlining: the instantiation described above, except the inlining skips nullary LET-defined operators
+  2. Pass-by-name inlining: replacing an operator name `A` with a local LET-definition: `LET A_LOCAL(p1,...,pn) == A(p1,...,pn) IN A_LOCAL`
+
+The reason for doing (1) is that, in order to encode `A(x1,...,xn)`, we need to know `e[x1/p1,...,xn/pn]`, not just `e`, since the latter may contain free variables (e.g. `p1` is free in `e`).
```
konnov, line 33: I do not understand this explanation.

docs/src/adr/018adr-inlining.md
konnov: I believe that the current write up requires a deep understanding of the problem. We should define the problem first.

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Options
+
+<!-- Communicate the options considered.
+     This records evidence of our circumspection and documents the various alternatives
+     considered but not adopted.
+-->
+1. Perform no inlining in preprocessing and inline only as needed in the rewriting rules.
+    - Pros: Spec intermediate output remains small, since inlining increases the size of the specificaiton
+    - Cons: Fewer optimizations can be applied, as some are only applicable to the syntactic forms obtained after inlining
```
konnov, line 46:
```suggestion
    - Cons: Fewer optimizations can be applied, as some are only applicable to the syntactic forms obtained after inlining
    - Cons: Rewriting rules for different encodings have to deal with operators in their generality.
```

docs/src/adr/018adr-inlining.md
konnov, line 31: Also, the inliner does not replace all occurrences of `A`, but only those, where `A` is passed as an argument, right?

konnov (commented):
This document requires a very deep understanding of the problem. The goal of this ADR is to document this deep understanding and motivation behind it. I think we should ask @thpani and @gabrielamafra to read this ADR and tell us, whether it makes the problem and the solution clear.

shonfeder (changes requested):
Queuing up a review as per Igor’s request.

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+We elect to use the term in a broader sense of "replacing an operator with its definition", in the two ways listed below:
+  1. Standard inlining: the instantiation described above
+      1. Non-nullary inlining: the instantiation described above, except the inlining skips nullary LET-defined operators
+  2. Pass-by-name inlining: replacing an operator name `A` with a local LET-definition: `LET A_LOCAL(p1,...,pn) == A(p1,...,pn) IN A_LOCAL`
+
+The reason for doing (1) is that, in order to encode `A(x1,...,xn)`, we need to know `e[x1/p1,...,xn/pn]`, not just `e`, since the latter may contain free variables (e.g. `p1` is free in `e`).
+
+The reason for doing (2) is more pragmatic; In order to rewrite expressions which feature any of the higher-order (HO) built-in operators, e.g. `ApaFoldSet(A, v, S)`, we need to know, at the time of rewriting, how to evaluate an application of `A` (e.g. `A(partial, current)` for folding). 
+Performing (2) allows us to make the rewriting rule local, since the definition becomes available where the operator is used, and frees us from having to track scope in the rewriting rules.
```
thpani, line 36: Maybe we can include an example for each?

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+- [Context (Problem)](#context)
+- [Options (Alternatives)](#options)
+- [Solution (Decision)](#solution)
+- [Consequences (Retrospection)](#consequences)
+
+## Summary
+
+<!-- Statement to summarize, following the following formula: -->
+
+This ADR outlines the transformations that happen in the Apalache inliner pass.
```
thpani, line 19: Is this true? It's more like it discusses different options for inlining and their tradeoffs.

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+The reason for doing (2) is more pragmatic; In order to rewrite expressions which feature any of the higher-order (HO) built-in operators, e.g. `ApaFoldSet(A, v, S)`, we need to know, at the time of rewriting, how to evaluate an application of `A` (e.g. `A(partial, current)` for folding). 
+Performing (2) allows us to make the rewriting rule local, since the definition becomes available where the operator is used, and frees us from having to track scope in the rewriting rules.
+
+## Options
+
+<!-- Communicate the options considered.
+     This records evidence of our circumspection and documents the various alternatives
+     considered but not adopted.
+-->
+1. Perform no inlining in preprocessing and inline only as needed in the rewriting rules.
```
thpani, line 44: Based on the document so far, I was unprepared that this will actually discuss different options.

docs/src/adr/018adr-inlining.md
thpani, line 46: What syntactic forms? I believe its important to document this.

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+1. Perform only standard inlining (1)
+    - Pros: Allows for additional optimizations after inlining
+    - Cons: Rewriting rules still need scope, to resolve higher-order operator arguments in certain built-in operators (e.g. folds)
+
+1. Perform non-nullary inlining and pass-by-name inlining
+    - Pros: 
+        - Enables further optimizations 
+        - Using non-nullary inlining has all of the benefits of standard inlining, while additionally being able to avoid repetition (e.g. not inlining `A` in `A + A`)
+        - Pass-by-name inlining allows us to keep rewriting rules local
+    - Cons: Implementation is more complex
```
thpani, line 57: Somehow this is not an exhaustive decision matrix. Aren't there three dimensions that can be discussed separately?

1. have a separate inlining pass? -> yes/no
2. if (1) is affirmative, perform full inlining or non-nullary inlining -> full/non-nullary
3. if (1) is affirmative, perform pass-by-name? -> yes/no

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+     This records evidence of our circumspection and documents the various alternatives
+     considered but not adopted.
+-->
+1. Perform no inlining in preprocessing and inline only as needed in the rewriting rules.
+    - Pros: Spec intermediate output remains small, since inlining increases the size of the specificaiton
+    - Cons: Fewer optimizations can be applied, as some are only applicable to the syntactic forms obtained after inlining
+
+1. Perform only standard inlining (1)
+    - Pros: Allows for additional optimizations after inlining
+    - Cons: Rewriting rules still need scope, to resolve higher-order operator arguments in certain built-in operators (e.g. folds)
```
thpani, line 50: Based on the below:

```suggestion
    - Cons: Rewriting rules still need scope, to resolve higher-order operator arguments in certain built-in operators (e.g. folds)
    - Cons: Nullary inlining prone to repetition
```

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+
+1. Perform non-nullary inlining and pass-by-name inlining
+    - Pros: 
+        - Enables further optimizations
```
thpani, line 54: Which ones?

thpani (commented):
I agree with @konnov that this would benefit from giving a bit more context.

It's also a bit vague in places, e.g., if an option enables/prohibits us from performing certain optimizations, I believe it's important for an ADR to characterize what kind of optimizations.

Finally, it seems to me that there are really 3 design dimensions proposed by this ADR, and I suspect that it would be much clearer if they were discussed separately, see https://github.com/informalsystems/apalache/pull/1568#discussion_r853127609

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+We elect to use the term in a broader sense of "replacing an operator with its definition", in the two ways listed below:
+  1. Standard inlining: the instantiation described above
+      1. Non-nullary inlining: the instantiation described above, except the inlining skips nullary LET-defined operators
+  2. Pass-by-name inlining: replacing an operator name `A` with a local LET-definition: `LET A_LOCAL(p1,...,pn) == A(p1,...,pn) IN A_LOCAL`
+
+The reason for doing (1) is that, in order to encode `A(x1,...,xn)`, we need to know `e[x1/p1,...,xn/pn]`, not just `e`, since the latter may contain free variables (e.g. `p1` is free in `e`).
+
+The reason for doing (2) is more pragmatic; In order to rewrite expressions which feature any of the higher-order (HO) built-in operators, e.g. `ApaFoldSet(A, v, S)`, we need to know, at the time of rewriting, how to evaluate an application of `A` (e.g. `A(partial, current)` for folding).
```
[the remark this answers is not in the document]
shonfeder: I don't understand why the "Pass-by-name inlining" tells us "how to evaluate an application of `A`". Should this be obvious if I had the right background knowledge, or should this connection be explained?

docs/src/adr/018adr-inlining.md
shonfeder: Why doesn't this  just move the problem to the definition of the expanded `A`?

docs/src/adr/018adr-inlining.md
shonfeder: I agree. I think it's partially because the summary and context don't present the problem to be solved, which would invite thinking about possible solutions, but instead frame the document as an explanation of a technique we are already using.

docs/src/adr/018adr-inlining.md
shonfeder, line 46: So is a key constraint here that we don't let later rewriting steps depend on earlier ones? Otherwise, I'd think you could still do inlining as part of rewriting, but just ensure it is the first rewriting step applied?

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,73 @@
...
+
+<!-- Communicate the options considered.
+     This records evidence of our circumspection and documents the various alternatives
+     considered but not adopted.
+-->
+1. Perform no inlining in preprocessing and inline only as needed in the rewriting rules.
+    - Pros: Spec intermediate output remains small, since inlining increases the size of the specificaiton
+    - Cons: Fewer optimizations can be applied, as some are only applicable to the syntactic forms obtained after inlining
+
+1. Perform only standard inlining (1)
```
shonfeder, line 48: What is the (1) here?

shonfeder: This seems orthogonal to the first option: aren't  the question of what inlining we do and when we do them independent?

shonfeder (commented):
Some feedback! In general I agree with Thomas' questions and think addressing his questions would help improve the clarity of the exposition.

docs/src/SUMMARY.md
```diff
@@ -103,3 +103,4 @@
 - [ADR-015: ITF: informal trace format](./adr/015adr-trace.md)
 - [ADR-016: ReTLA: Relational TLA](./adr/016adr-retla.md)
 - [PDR-017: Checking temporal properties](./adr/017pdr-temporal.md)
+- [ADR-018: Inlining](./adr/018adr-inlining.md)
```
shonfeder, line 106: It seems like this is more specific than just the problem of inlining in general?

docs/src/adr/018adr-inlining.md
Kukovec: > Since it is a foundational document for inlining, we have to explain that there a top-level operators and nested operators that are defined with LET-IN.

Are you serious? That's just TLA+ 101

docs/src/adr/018adr-inlining.md
Kukovec: > Why is 2. considered inlining at all? Do I understand correctly that this is essentially just eta expansion (i.e., converting `𝜆x.Mx` to `M`) only via a let expression?

It's inlinining in the sense that 
> We elect to use the term in a broader sense of "replacing an operator with its definition"

You can also view it as localization. Suppose you have an expression `e`, in which `A` appears as an argument to a HO operator (e.g. `FoldSet(A, v, S)`). The treatment of this expression depends on the scope surrounding `e`, that is, the definition of `A` is external to `e`. In the `Fold` example, `A` is supposed to be "applied" to the arguments, but we don't know what the arguments are until execution. (2) makes it so the definition is present where the name is passed as an argument, meaning `Fold` can now be analyzed locally, w/o needing to track "scope".

Kukovec: Yes, I suppose this should be clarified

docs/src/adr/018adr-inlining.md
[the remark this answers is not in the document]
Kukovec: > What is `A(partial, current)`?

Typically, in folds, the first operator argument is an accumulator/partial result and the 2nd is the collection element, so a common definition would be 
```tla
LET A(partial, current) == ... IN FoldSet(A, v, S)
```

> I don't understand why the "Pass-by-name inlining" tells us "how to evaluate an application of `A`". Should this be obvious if I had the right background knowledge, or should this connection be explained?

Because inlining gives us a definition of `A` _within_ the sub-scope of `Fold(A, ...,...)`, so it can be used when encoding the fold. Otherwise, the rewriting rules would have to track scope.

docs/src/adr/018adr-inlining.md
Kukovec: > Why doesn't this just move the problem to the definition of the expanded `A`?

Because in 
```tla
LET A(...) == ... IN FoldSet(A, ..., ...)
```
the definition is _outside_ the `Fold`-scope, while in
```tla
FoldSet( LET A(...) == ... IN A, ... ,... )
```
it is inside. The rewriting rule for `Fold` only sees the `Fold` sub-scope.

docs/src/adr/018adr-inlining.md
Kukovec: The thing is though, inlining isn't really a solution to some theoretical problem, it's a case of "this makes rewriting rules much more convenient".

docs/src/adr/018adr-inlining.md
Kukovec: Not really, because the point of inlining is that you can then run e.g. `ConstSimplifier`. If you do inlining _after_ other transformations you lose a lot of its value (basically the only value of doing late inlining is saving 1 rewrtiting rule)

docs/src/adr/018adr-inlining.md
Kukovec: RE your (1.), there's little functional difference between having a separate pass, or just having it be the 1st transformation applied. 
RE (2.) non-nullary is strictly better, so I'll merge the sections.

docs/src/adr/018adr-inlining.md
thpani: I meant that this is an inaccurate summary of the document.

docs/src/adr/018adr-inlining.md
thpani: I suppose most real-world ADRs address practical problems, not theoretical ones.

> "this makes rewriting rules much more convenient"

Right. Any why this is a problem we're trying to solve, and what forces an effective solution must take into account, should be discussed as context for the reader.

docs/src/adr/018adr-inlining.md
thpani: See my comment below. This is meant as context to frame the problem for the reader, irrespective of their experience with TLA+.

docs/src/adr/018adr-inlining.md
shonfeder: So I guess the real kernel of my question is that

> We elect to use the term in a broader sense of "replacing an operator with its definition"

does not seems accurate to the spirit of the phrase "inlining" if we are replacing it with newly synthesized definitions. This seems like a distinct kind of rewriting to me (again, it looks to me like eta expansion). There might not be any stakes to this difference tho, so I'm ok with leaving it as is, even if I personally find the terminology a bit misleading.

docs/src/adr/018adr-inlining.md
```diff
@@ -24,17 +24,41 @@ Since we have recently reworked the inliner in [#1569](https://github.com/inform
...
+
+Then, the result of performing (1) for `A(1, 2)` would be `1 + 2 * 2`. 
+The constant simplification could take the inlined expression and simplify it to `5`, whereas it could not do this across the application boundary of `A(1,2)`.
+
+The result of performing (2) for `ApaFoldSet(A, 0, {1,2,3})` would be
+```tla
+ApaFoldSet(LET A_LOCAL(p,q) == p + 2 * q, 0, {1,2,3})
+```
+
+While this resulting expression isn't subject to any further simplification, notice that it _does_ contain all the required information to fully translate to SMT, unlike `ApaFoldSet(A, 0, {1,2,3})`, which requires external information about `A`.
```
shonfeder, line 112: I've a general, unrelated question here: why *not* simplify folds like this to a single scalar value?

shonfeder (approved):
Clarifications really help for me! Thanks for the great writeup!

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,156 @@
...
+  2. Pass-by-name inlining: replacing an operator name `A` used as an argument to a higher-order operator with a local LET-definition: `LET A_LOCAL(p1,...,pn) == A(p1,...,pn) IN A_LOCAL`
+
+The reason for doing (1) is that, at some point, a rewriting rule would have to generate constraints from `A(e1,...,en)`. 
+To do this, we couldn't just separately encode `body` and `e1,...,en`, because the richness of the data structures allowed in TLA+ makes it difficult to combine independently generated constraints, in cases where the operator parameters are complex expressions (e.g. `e1` is some record with varied nested constructs).
+This is mostly due to the fact that there is no 1-to-1 correspondence between TLA+ objects and SMT datatypes, so encoding object equality is more complicated (which would be needed to express that `ei` instantiates `pi`).
+Therefore we must, no later than at the point of the rewriting rule, know `body[e1/p1,...,en/pn]`. 
+
+While inlining non-nullary operators is  strictly necessary, inlining nullary operators is not, because nullary operators, by definition, do not have formal parameters.
```
thpani, line 66:
```suggestion
While inlining non-nullary operators is strictly necessary, inlining nullary operators is not, because nullary operators, by definition, do not have formal parameters. 
```

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,156 @@
...
+The non-nullary variant of (1) is therefore strictly better for performance, because it allows for a sort of caching, which avoids repetition. Consider for example:
+```tla
+A1(p) == 
+  LET pCached == p
+  IN F(pCached,pCached)
+
+A2(p) == F(p,p)
+```
+
+If we apply the substitution `{p -> e}`, for some complex value `e`, to the bodies of both operators, the resutls are
```
thpani, line 81:
```suggestion
If we apply the substitution `{p -> e}`, for some complex value `e`, to the bodies of both operators, the results are 
```

docs/src/adr/018adr-inlining.md
```diff
@@ -0,0 +1,156 @@
...
+```tla
+LET pCached == e
+IN F(pCached,pCached)
+```
+and
+```
+F(e,e)
+```
+
+In the first case, we can translate `pCached` to a cell and reuse the cell expression twice, whereas in the second, `e` is rewritten twice, independently.
```
thpani, line 92: Maybe link

> translate `pCached` to a cell

to the OOPSLA paper (or another source)? Cells weren't introduced in the scope of this ADR.

thpani (approved):
Looks great! :tada:

I think this is an excellent explanation now, will be a much easier read for people without prior knowledge. I really appreciate how crisp the Options + Solution sections are now because the necessary context is given before.

Just 2 typos and a reference to add in the comments.

docs/src/adr/018adr-inlining.md
Kukovec: Because you can't in general. E.g.:
```tla
VARIABLE 
  \* @type: Set(Int);
  S

Op == FoldSet(+, 0, S) \* == ??
```
With inlining, the args are always given (even if they're symbolic), but with folds, if the args are symbolic you're SOL, since you need the contents.

PR #1579, opened 2022-03-30
Rename `--tuning-options` to `--tune-here`

thpani:
We currently have 2 tuning flags, `--tuning=<file>` and `--tuning-options=key1=val1:key2=val2`. However, `--tuning-options` is called `--tune-here` throughout the documentation (the mismatch was already in the original PR).

As [discussed below](https://github.com/informalsystems/apalache/pull/1579#discussion_r838815069):
* Rename `--tuning` to `--tuning-options-file`.
* Fix references to `--tune-here` in the docs (to `--tuning-options`).

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,6 @@
          * Some bug fix, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
+### Breaking changes
+
+ * Rename `--tuning` to `--tuning-options-file`, see #1579
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,3 +13,7 @@
 ### Breaking changes
 
  * Rename `--tuning` to `--tuning-options-file`, see #1579
+
+### Bug fixes
+
+ * Fix references to `--tune-here` (actually `--tuning-options`), see #1579
--- a/docs/src/apalache/running.md
+++ b/docs/src/apalache/running.md
@@ -12,7 +12,7 @@ The model checker can be run as follows:
 $ apalache-mc check [--config=filename] [--init=Init] [--cinit=ConstInit] \
     [--next=Next] [--inv=Inv] [--length=10] [--algo=(incremental|offline)] \
     [--discard-disabled] [--no-deadlock] \
-    [--tuning-options-file=filename] [--tune-here=options] \
+    [--tuning-options-file=filename] [--tuning-options=key1=val1:...:keyn=valn] \
     [--smt-encoding=(oopsla19|arrays)] \
     [--out-dir=./path/to/dir] \
     [--write-intermediate=(true|false)] \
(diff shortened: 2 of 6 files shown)


Kukovec (approved):

UNRELEASED.md
```diff
@@ -10,3 +10,6 @@
          * Some bug fix, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
+### Breaking changes
+
+ * Rename `--tuning-options` to `--tune-here`, see #1579
```
shonfeder, line 15: I'm actually in favor of correction the docs, just because these do indeed seem to be options used for tuning, whereas I'm not at clear on what "here" means in this context.

shonfeder (approved):
LGTM

Approving cause I don't want to block, but I also feel that `tuning-options` is a nicer name. I'd suggest being sure others agree on this name change before merging it in.

UNRELEASED.md
thpani: We currently have `--tuning=<file>` and `--tuning-options=key1=val1:key2=val2`.
I think `--tune-here` makes it clearer that tuning options for this flag go _here_, on the command line.

Or maybe these should be `--tuning-file` and `--tuning-options`?
In any case, maybe we need to be more careful in naming these things in the first place.

shonfeder: What about calling the one `tuning-options-file`?

konnov: I think we can always distinguish between the syntax `key=val` and a filename. I would prefer to introduce just one option called `--tune=<value>`, and depending on `<value>` either treat it as in-place assignments or a filename.

thpani: > I think we can always distinguish between the syntax `key=val` and a filename. I would prefer to introduce just one option called `--tune=<value>`, and depending on `<value>` either treat it as in-place assignments or a filename.

I think the idea was that `--tune=key=val` is able to override `--tune=file`. We'd lose that ability?

thpani: > What about calling the one `tuning-options-file`?

+1 from my side.

konnov: > I think the idea was that `--tune=key=val` is able to override `--tune=file`. We'd lose that ability?

That is true. But the main use is to quickly define options instead of introducing a file. I doubt that we need such a complicated overriding mechanism anyways.

thpani: I'm torn. `key=val` is a valid unix filename. How would we parse it to distinguish files from key-value pairs? Look for a valid `key`? Likely this would complicate our parsing logic for the benefit of dropping one very straight-forward flag.

konnov: Well, `key=val` is a valid filename. I don't know. I would prefer simplicity to such a weird corner case.

gabrielamafra: I like having a different flag for file and inline options. It's the more common alternative in my experience and precedence is easy to document.

konnov: Yeah, let's converge on this one and merge this one :)

shonfeder: It sounds like we agreed to rename `tuning` to `tuning-options-file`?

rodrigo7491 (approved):
LGTM

UNRELEASED.md
thpani: Okay, converting this to draft until I have that renaming ready.

shonfeder (approved):
LGTM!

PR #1583, opened 2022-03-31
ADR017 on checking temporal properties

konnov:
Closes #1303. This PR contributes ADR017, which explains a plan for implementing temporal properties. I am not going into technical details of the SAT/SMT encodings of LTL. One has to read the cited papers, in order to understand those encodings.

PR #1585, opened 2022-04-01
refactor ADR017 into PDR017

konnov:
Given the comments by @thpani and the team, this PR downgrades ADR017 into PDR017, a preliminary design record. It also includes the comments by @thpani that were made after the merge in https://github.com/informalsystems/apalache/pull/1583.


thpani (approved):
👍
Thanks for exploring options on this! 🙌

PR #1597, opened 2022-04-05
Extend TlaType1 with the row types

### LINKED ISSUE #1592: Extend TlaType1 with row types, new records, and variants
As planned in #401

konnov:
Closes #1592. This is the first step in the series of baby steps towards #401. It extends `TlaType1` with `RowT1`, `RecRowT1`, and `VariantT1`. It also overrides `toString`. This is the second iteration. The first one, which was closer to the paper on scoped labels, was harder to handle, due to explicit manipulation over sequences, which we did not need.

The big plan is explained in [ADR014](https://github.com/informalsystems/apalache/blob/unstable/docs/src/adr/014adr-precise-records.md) and #401.


tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
```diff
@@ -310,3 +313,112 @@ case class OperT1(args: Seq[TlaType1], res: TlaType1) extends TlaType1 {
...
+      case RowT1(fieldTypes, other) =>
+        // the special user-friendly form, e.g., { foo: Int, g: Bool } or { foo: Int, g: Bool, a }
+        val fields = fieldTypes.map(p => "%s: %s".format(p._1, p._2)).mkString(", ")
+        other match {
+          case None    => s"{ $fields }"
+          case Some(v) => s"{ $fields, $v }"
+        }
+
+      case _ =>
+        // the general case, which is probably never reachable
```
gabrielamafra, line 373: I feel like making a distinction of kinds will help us here. `RecRowT1` and `VariantT1` should be constructed with parameters of kind `row`, and then we can make sure that `RowT1` (of kind row) is matched at scala's type system level.

As I see it, `TlaType1` is the only kind we have right now. We could introduce a `TlaRow = RowT1 | VariantT1` and remove this two types from `TlaType1`.

I see there is an "Implement kind inference?" at the meta-issue, and I think we should do it to avoid mixing things up. We can do it in another iteration, of course.

gabrielamafra (approved):
LGTM, just a note on how representation for kinds could help us

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
konnov: Right. I do not have a clear opinion about kinds yet. `VariantT1` and `RecRowT1` are not rows, but they require rows as the arguments. The main reason for having `row: TlaType1` and not `row: RowT1` was the need for accommodating `VarT1`. However, we also have `VarT1` as an optional parameter of `RowT1` now. So maybe we can just require `row` to have the type `RowT1`?

gabrielamafra: Yes, you're right! I guess that's good enough for now, I'm just concerned that we might have problems if `RowT1` is a `TlaType1` as it is right now. We shouldn't unify term types with rows - but we can figure out how to distinguish them when implementing unification for this.

I'll leave this image here for future reference because I keep going back to the paper to check it:
image

PR #1598, opened 2022-04-06
Fix build from source docs

thpani:
Fix the instructions for building from source:

* Correct the Scala version `12.0` -> `2.12.0`
* Re-introduce instructions to source the shell env (otherwise unit tests will fail)

### DIFF
--- a/docs/src/apalache/installation/source.md
+++ b/docs/src/apalache/installation/source.md
@@ -10,6 +10,7 @@
    - On macOS / Homebrew: `brew install sbt`
 4. Clone the git repository: `git clone https://github.com/informalsystems/apalache.git`.
 5. Change into the project directory: `cd apalache`.
+6. Install [direnv][] and run `direnv allow`, or use [another way to set up the shell environment][shell environment].
 7. Run `make`.
    - To skip running the tests, you can run `make package`
 8. The distribution package will be built to `./target/universal/apalache-<VERSION>`, and you can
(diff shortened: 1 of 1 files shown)


docs/src/apalache/installation/source.md
```diff
@@ -2,16 +2,17 @@
 
 1. Install `git`.
 2. Install [OpenJDK8][] or [Zulu JDK8][].
```
konnov, line 4: Shall we update the JDK version numbers? I am currently working on Zulu 15, and it's all fine. We should probably check which versions are ok.

konnov (approved):
Looks good. I have noticed that the JDK version is terribly outdated

docs/src/apalache/installation/source.md
thpani: It looks like JDK 8 is recommended as long as we're on Scala 2.12.0? https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#version-compatibility-table

konnov: We are currently using 2.12.15 in `build.sbt`, so this is also outdated :)

thpani: Oh 😅 I already merged, we will need to follow up on this. Filed #1600

thpani: We should probably use an LTS release like JDK17 though?

konnov: true

PR #1599, opened 2022-04-06
Switch symlink creation to use os-independent Files.createSymbolicLink

p-offtermatt:
Change build.sbt to use [Files.createSymbolicLink](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createSymbolicLink(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.attribute.FileAttribute...))
for compatibility with more OS's

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -292,7 +292,7 @@ lazy val root = (project in file("."))
         // See https://github.com/informalsystems/apalache/pull/1382
         (s"tar zxvf ${pkg} -C ${target_dir}" !)
         log.info(s"Symlinking ${current_pkg} -> ${unzipped}")
-        val target_path: java.nio.file.Path = java.nio.file.Paths.get("", unzipped)
+        val target_path: java.nio.file.Path = java.nio.file.Paths.get(unzipped)
         if (!current_pkg.exists) {
           java.nio.file.Files.createSymbolicLink(current_pkg.toPath(), target_path)
         } else {
--- a/build.sbt
+++ b/build.sbt
@@ -293,9 +293,8 @@ lazy val root = (project in file("."))
         (s"tar zxvf ${pkg} -C ${target_dir}" !)
         log.info(s"Symlinking ${current_pkg} -> ${unzipped}")
         val target_path: java.nio.file.Path = java.nio.file.Paths.get("", unzipped)
-        if (!current_pkg.exists) 
-        { 
-          java.nio.file.Files.createSymbolicLink(current_pkg.toPath(), target_path) 
+        if (!current_pkg.exists) {
+          java.nio.file.Files.createSymbolicLink(current_pkg.toPath(), target_path)
         } else {
           log.info(s"${current_pkg} already exists")
         }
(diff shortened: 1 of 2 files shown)


thpani:
Can you please run `make fmt-fix` and push to make the CI checks pass?

konnov (commented):
I guess, the new behavior is not fully equivalent to `ln -sfn`. The flag `-f` should force the new link to be created.

build.sbt
```diff
@@ -292,7 +292,12 @@ lazy val root = (project in file("."))
         // See https://github.com/informalsystems/apalache/pull/1382
         (s"tar zxvf ${pkg} -C ${target_dir}" !)
         log.info(s"Symlinking ${current_pkg} -> ${unzipped}")
-        s"ln -sfn ${unzipped} ${current_pkg}" ! log
+        val target_path: java.nio.file.Path = java.nio.file.Paths.get("", unzipped)
```
thpani, line 295: Do we need the `""` first argument?

build.sbt
```diff
@@ -292,7 +292,12 @@ lazy val root = (project in file("."))
...
+        if (!current_pkg.exists) {
+          java.nio.file.Files.createSymbolicLink(current_pkg.toPath(), target_path)
+        } else {
+          log.info(s"${current_pkg} already exists")
+        }
```
thpani, line 300: We should mirror the behavior of `ln -f` in this case.

thpani (commented):
Couple of comments from my side.

Did you figure out why this didn't work on Windows before? Maybe you can update the issue with your findings, so we have it for future reference.

build.sbt
```diff
@@ -292,7 +292,12 @@ lazy val root = (project in file("."))
         // See https://github.com/informalsystems/apalache/pull/1382
         (s"tar zxvf ${pkg} -C ${target_dir}" !)
         log.info(s"Symlinking ${current_pkg} -> ${unzipped}")
-        s"ln -sfn ${unzipped} ${current_pkg}" ! log
+        val target_path: java.nio.file.Path = java.nio.file.Paths.get(unzipped)
+        if (current_pkg.exists) {
+          log.info(s"${current_pkg} already exists, overriding")
+          current_pkg.delete
+        }
+        java.nio.file.Files.createSymbolicLink(current_pkg.toPath, target_path)
```
thpani, line 300: I wonder if we could use the [File API exposed by sbt](https://www.scala-sbt.org/1.x/docs/Paths.html#Constructing+a+File) to `file(unzipped).toPath`?
Looks a bit more idiomatic sbt and symmetrical to the treatment of `current_pkg` below.

thpani (commented):
Sorry for the incremental comments, I'm learning more about sbt as we go in this PR 😅

p-offtermatt:
There is also one notable difference between the current behaviour and the behaviour of `ln -snf`, which is when there is not a file at the location of the symlink, but a directory.
`ln -snf` will place the symlink inside the directory, my implementation fails currently. This appears not to happen in the context of the current build setup, but could occur in the future if the build procedure changes.

konnov (approved):
Looks good! I am always suspicious of deletion. Hopefully, we cannot nuke the whole filesystem by that.

thpani (approved):
LGTM 👍
Thanks for working on this!

PR #1619, opened 2022-04-11
Document review locking and code curation

shonfeder:
As per our discussion in today's meeting, this PR documents our convention for
locking PRs until we a chance to review them (when desired) and for our use of
the CODEOWNERS file.

It also includes an unrelated update to the documentation of our branching model.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,8 @@ of landing changes:
 
 ## Making a pull request
 
-We develop on the `unstable` branch and release stable code on `master`.
+We develop on the `unstable` branch and practice [trunk-based
+development](https://trunkbaseddevelopment.com/).
 
 Nontrivial changes should start with a [draft pull request][] against
 `unstable`. The draft signals that work is underway. When the work is ready for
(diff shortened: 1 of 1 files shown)


konnov (approved):
Looks good!

gabrielamafra (approved):
I like this. I'm still skeptical about how much successful we will be at keeping CODEOWNERS updated so it is useful, but I don't think it will be a burden otherwise, so I'm on board with giving it a shot. I really like the locking review protocol :smile:

shonfeder:
> IIUC, the main problem we had in the past was that PRs where merged faster than people got to look at them.

iiuc, the situation discussed in the meeting was where someone was in the middle of doing a review when a PR got merged.

PR authors can and should still do the things you describe. If they think reviews for some particular contributors are critical, they shouldn’t merge their prs until they get the needed reviews.

Sanctioning the proposed mechanism is just a way of allowing reviewers to quickly communicate and enforce their conviction in the other direction, for cases when they feel their own review is critical. It doesn’t replace the tasks for PR authors.

If there is persistent and problematic disagreement on (b) I agree we should communicate to clarify.

I think Jure has already made use of this mechanism to lock a or pending review l, so it seems it can work in practice.

thpani:
> iiuc, the situation discussed in the meeting was where someone was in the middle of doing a review when a PR got merged.

To prevent this I would need to invoke the locking protocol every time I start a review?

PR #1629, opened 2022-04-13
Optimize membership for record sets

### LINKED ISSUE #1627: An optimization for membership in a record set
This is an MWE extracted from a larger specification:

```tla
------------------------- MODULE RecMem ------------------------------           
EXTENDS Integers

VARIABLES
  \* @type: [ pos: Int, q: Int, color: Str ];
  token       \* token structure

N == 3
Node == 0 .. N-1
Color == {"white", "black"}
Token == [pos : 1..3, q : Int, color : Color]

TypeOK ==
  token \in Token

Init ==
  token = [pos |-> 0, q |-> 0, color |-> "black"]

Next ==
  UNCHANGED token

===================================================================== 
```

When checking `TypeOK`, Apalache fails with an error:

```
apalache-mc check --inv=TypeOK RecMem.tla
...
Input error (see the manual): Found a set map over an infinite set of Int. Not supported.
```

The reason for that is the following invariant negation:

```tla
VCNotInv_si_0 ==
  ~(token
    \in {
      [pos |-> t_7$1, q |-> t_8$1, color |-> t_9$1]:
        t_7$1 \in 1 .. 3,
        t_8$1 \in Int,
        t_9$1 \in { "white", "black" }
    })
```

We could rewrite an expression `rec \in [ f_1: S_1, ..., f_k: S_k ]` as follows:

```tla
DOMAIN rec = { "f_1", ..., "f_k" } /\ rec.f_1 \in S_1 /\ ... /\ rec.f_k \in S_k
```

While it would not solve all the problems with infinite sets, it would help us in checking this membership pattern, which occurs often in `TypeOK`.

konnov:
Closes #1627. This PR introduces an optimization for record membership, which is quite common in `TypeOK`:

```tla
r \in { [a |-> x, b |-> y]: x \in S, y \in T }
\* becomes
DOMAIN r = { "a", "b" } /\ r.a \in S /\ r.b \in T
```

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
@@ -69,9 +69,35 @@ class ExprOptimizer(nameGen: UniqueNameGenerator, tracker: TransformationTracker
       // Transform x \in { y \in S: P } into x \in S /\ P[y/x]
 
       def memCopy = DeepCopy(tracker).deepCopyEx(mem)
+
       val predSubstituted = ReplaceFixed(tracker)(nameEx, memCopy)(pred)
       val b = BoolT1()
       tla.and(tla.in(mem, set).as(b), predSubstituted).as(b)
+
+    case memEx @ OperEx(TlaSetOper.in, rec,
+            OperEx(TlaSetOper.map, OperEx(TlaFunOper.`enum`, fieldsAndValues @ _*), varsAndSets @ _*))
+        if fieldsAndValues.length == varsAndSets.length =>
+      // Transform r \in { [f_1 |-> x_1, ..., f_k |-> x_k]: x_1 \in S_1, ..., x_k \in S_k }
+      // into
+      // DOMAIN r = { "f_1", ..., "f_k" } /\ r.f_1 \in S_1 /\ ... /\ r.f_k \in S_k
+      val (fields, values) = TlaOper.deinterleave(fieldsAndValues)
+      val (vars, sets) = TlaOper.deinterleave(varsAndSets)
+      assert(fields.length == vars.length)
+      if (values.zip(vars).exists(p => p._1 != p._2)) {
+        // The set has a more general form: { [f_1 |-> e_1, ..., f_k |-> e_k]: x_1 \in S_1, ..., x_k \in S_k }, where
+        //   e_1, ..., e_k are expressions over x_1, ..., x_k.
+        // We do not know how to optimize it.
+        memEx
+      } else {
+        // The set is of the nice form: { [f_1 |-> x_1, ..., f_k |-> x_k]: x_1 \in S_1, ..., x_k \in S_k }
+        val strSetT = SetT1(StrT1())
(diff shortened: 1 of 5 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
```diff
@@ -69,9 +69,35 @@ class ExprOptimizer(nameGen: UniqueNameGenerator, tracker: TransformationTracker
...
+
+    case memEx @ OperEx(TlaSetOper.in, rec,
+            OperEx(TlaSetOper.map, OperEx(TlaFunOper.`enum`, fieldsAndValues @ _*), varsAndSets @ _*))
+        if fieldsAndValues.length == varsAndSets.length =>
+      // Transform r \in { [f_1 |-> x_1, ..., f_k |-> x_k]: x_1 \in S_1, ..., x_k \in S_k }
+      // into
+      // DOMAIN r = { "f_1", ..., "f_k" } /\ r.f_1 \in S_1 /\ ... /\ r.f_k \in S_k
+      val (fields, values) = TlaOper.deinterleave(fieldsAndValues)
+      val (vars, sets) = TlaOper.deinterleave(varsAndSets)
+      assert(fields.length == vars.length)
```
Kukovec, line 85: This is dangerous with an `assert`. Theoretically, it's perfectly fine to have e.g. more vars than fields:
```tla
[a |-> 1] \in {[a |-> x]: x \in {1}, y \in T}
```
Dumb, but correct.
Or fewer:
```tla
{[a |-> x, b |-> x]: x \in S}
```
You should just default to the "return no optimization" case, not throw here.

Kukovec (changes requested):

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ExprOptimizer.scala
konnov: You are right, but there is a test above:

```
if fieldsAndValues.length == varsAndSets.length =>
```

Kukovec: Ah, must have missed that.

Kukovec: But then what's the point of this assert anyway?

konnov: Just in case. It is an assert about deinterleaved lists.

Kukovec: This is implied by ` if fieldsAndValues.length == varsAndSets.length`,`TlaSetOper.map.arity`, and `TlaFunOper.enum.arity`

thpani: I think this is a reasonable precondition to state before the `.zip` below.

thpani (commented):
LGTM

PR #1630, opened 2022-04-14
Support builtin operators as arguments to other operators

### LINKED ISSUE #1626: Support for the built-in operators in folds
We have got this question several times. It is very tempting to write expressions like these ones:

```tla
FoldSet(+, 0, 1..3)
FoldSet(\union, {}, SUBSET 1..3)
```

Those expressions are currently not supported by Apalache, as `+` and `\union` are treated as built-in operators in Apalache, whereas `FoldSet` requires a user-defined operator or lambda. TLA+ by itself does not pose any restriction like this. The standard workaround is:

```tla
LET Plus(x, y) == x + y IN
FoldSet(Plus, 0, 1..3)

LET Unite(S, T) == S \union T IN
FoldSet(Unite, {}, SUBSET 1..3)
```

We probably could automatically replace a built-in operator with a lambda or a let-definition when importing a spec from SANY.

@Kukovec, @shonfeder, @gabrielamafra, @thpani, @rodrigo7491, @p-offtermatt, what do you think about this feature? Do we want to support it?

konnov:
Closes #1626. This PR addresses the case of passing a built-in operator as an argument to another operator, e.g.:

```tla
FoldSet(+, 0, {1, 2, 3})
```

Since we do not want to handle the built-in operators and library operators as a special case, we simply wrap them with an auxiliary definition, similar to `LAMBDA`:

```
FoldSet(
  LET __PLUS_22(__p1, __p2) == __p1 + __p2 IN
  __PLUS_22,
  0,
  { 1, 2, 3}
)
```

This approach is compatible with the new version of operator inlining. So it is working out of the box 🎸 

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/ExprOrOpArgNodeTranslator.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/ExprOrOpArgNodeTranslator.scala
@@ -2,7 +2,7 @@ package at.forsyte.apalache.tla.imp
 
 import at.forsyte.apalache.tla.imp.src.{SourceLocation, SourceStore}
 import at.forsyte.apalache.tla.lir._
-import at.forsyte.apalache.tla.lir.oper.{TlaFunOper, TlaOper}
+import at.forsyte.apalache.tla.lir.oper.{FixedArity, TlaFunOper, TlaOper}
 import at.forsyte.apalache.tla.lir.values.{TlaDecimal, TlaInt, TlaStr}
 import at.forsyte.apalache.io.annotations.store._
 import at.forsyte.apalache.tla.lir.UntypedPredefs._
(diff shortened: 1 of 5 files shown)


tla-io/src/main/scala/at/forsyte/apalache/tla/imp/ExprOrOpArgNodeTranslator.scala
```diff
@@ -138,19 +138,41 @@ class ExprOrOpArgNodeTranslator(
...
+            ValEx(tlaValue)
+
+          case NoneUnit() =>
+            // No definition found. This may be a standard operator such as `\\union`.
+            OpApplTranslator.simpleOpcodeToTlaOper.get(name) match {
+              case Some(oper) =>
+                wrapBuiltinWithLetIn(oper)
+
+              case None =>
+                throw new SanyImporterException(s"Unknown built-in operator $name applied as an argument")
```
Kukovec, line 173: Can it be both unknown and buillt-in?

Kukovec (approved):
👍

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/ExprOrOpArgNodeTranslator.scala
konnov: Good question :D

Issue #1588, opened 2022-04-04
Nitpicker: Leverage Scalacheck to compare different encodings


konnov:
@rodrigo7491 has been trying to debug an evasive bug in the arrays encoding for some time. I think we should approach this problem automatically.

We have `IrGenerators` for property-based testing. It should be possible to generate an expression `generated` of type `Set(genType)` and then check the following spec:

```tla
------ MODULE Oracle --------
VARIABLES
  \* @type: ${genType};
  result,
  \* @type: Bool;
  found

Init ==
  /\ result \in generated
  /\ found \in BOOLEAN

Next ==
  UNCHANGED <<result, found>>

NotFound ==
  ~found
======================
```

By checking the invariant `NotFound` of the spec `Oracle`, we will obtain a value `witness` of `result` that belongs to the set `generated`, if such a value exists. We can feed this value into the `Verifier` spec:

```tla
------ MODULE Verifier --------
VARIABLES
  \* @type: ${genType};
  result,
  \* @type: Bool;
  found

Init ==
  /\ result \in generated
  /\ result = ${witness}
  /\ found \in BOOLEAN

Next ==
  UNCHANGED <<result, found>>

NotFound ==
  ~found
======================
```

When checking the invariant `NotFound` of the spec `Verifier`, we can have two possible outcomes:

 1. There is a counterexample. We compare the value of `result` to `witness`. They should match. If they don't, there is a bug in one of the encodings.
 1. There is no counterexample. There is a bug in one of the encodings.

Obviously, there is no need to go through the whole Apalache pipeline. We can simply generate expressions with `IrGenerators` and call the model checker, as it is done in [BoundedCheckerPassImpl.scala](https://github.com/informalsystems/apalache/blob/a535522b39fe1e3d5815e7a4422a400fc48d5b51/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala).

Since `IrGenerators` may generate ill-typed expressions, we have to resolve #1379 first.

We can use the same approach to compare Apalache against TLC, though it will require a bit more boilerplate.

- [x] #1632
- [x] #1638
- [x] #1639
- [x] #1666
- [ ] #1667

rodrigo7491:
For context on the elusive bug mentioned see #1589.

@konnov, it is not stated anywhere, but `Oracle` and `Verifier` would be checked on alternating encodings, right?

konnov:
[...]

PR #1631, opened 2022-04-14
Refactor SeqModelChecker to expose counterexamples via listeners

thpani:
Towards #1632, refactoring required for #1588:

Expose found counter-examples in `SeqModelChecker` via observer pattern:

* Introduce an observer trait `SeqModelCheckerListener`.
* Add a first observer `DumpCounterexamplesSeqModelCheckerListener`, to take over the file dumping previously done directly in `SeqModelChecker`.

We will introduce a second `SeqModelCheckerListener` in subsequent PRs to collect counterexamples for #1588.

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelCheckerListener.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelCheckerListener.scala
@@ -1,0 +1,31 @@
+package at.forsyte.apalache.tla.bmcmt
+
+import at.forsyte.apalache.tla.bmcmt.trex.DecodedExecution
+import at.forsyte.apalache.tla.lir.{TlaEx, TlaModule}
+
+/**
+ * Observe [[SeqModelChecker]]. State changes in model checker state are reported via callbacks.
+ */
+trait SeqModelCheckerListener {
+
+  /**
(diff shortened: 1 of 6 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelCheckerListener.scala
```diff
@@ -0,0 +1,31 @@
+package at.forsyte.apalache.tla.bmcmt
+
+import at.forsyte.apalache.tla.bmcmt.trex.DecodedExecution
+import at.forsyte.apalache.tla.lir.{TlaEx, TlaModule}
+
+/**
+ * Observe [[SeqModelChecker]]. State changes in model checker state are reported via callbacks.
+ */
+trait SeqModelCheckerListener {
```
konnov, line 9:
```suggestion
trait ModelCheckerListener {
```

It's not really related to Sequential vs. Parallel. (The parallel version is living in a branch, but it very unlikely that we will merge it.)

konnov (approved):
This is definitely a good improvement. Just one comment

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelCheckerListener.scala
thpani: Good point!

PR #1634, opened 2022-04-14
Typed Builder implementation Part 1

Kukovec:


Kukovec:
As per our discussion at Friday's meeting, I've added scope-validation to the builder. 
There's 2 levels to building: 
  1. "raw" building, which takes `n` (presumably well typed) arguments and constructs an IR node, with a given operator, and performs toplevel type computation for the resulting expression. For example, given arguments `a1: Set(Int), a2: Set(Int)`, the raw `_union` should produce a `OperEx(TlaSetOper.union, a1, a2)(Typed(Set(IntT1())))`, because `\union` is an operator with a known signature `(Set(t), Set(t)) => Set(t)`, which can be unified with the arguments `(Set(Int), Set(Int))`. Note that raw building does not guarantee total type correctness, e.g. if 
  ```a1 = OperEx(TlaSetOper.enum, NameEx("a")(Untyped()))(Typed(SetT1(IntT1())))``` 
nor scope correctness, e.g. in `forall(x: Int, S: Set(Int), x: Bool)`, as the signature of `forall` is `(t, Set(t), Bool) => Bool` and agnostic to actual names.

2. Scoped building, which operates over `BuilderWrapper` expressions, which are `State` values carrying the scope as internal state, producing `TlaEx` on demand. Scoped building guarantees that, as long as all IR components are built with the `ScopedBuilder`, the final expression will be both type-correct and scope-consistent.


At his point in time, you can ignore: 
  - Applicative
  - BoolOp
  - FunOp
  - TypeCalculatingBuilder

Comments on the draft would be appreciated at this time @konnov @thpani @shonfeder @gabrielamafra

Kukovec:
This is still missing tests for `TlaSetOper` methods, but that shouldn't be a blocker for reviewing.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  /** [ k1: v1, ... , kN: vN ], must have at least 1 key-value pair */
+  def recSet(k1W: ValWrapper, v1W: BuilderWrapper, kvsW: BuilderWrapper*): BuilderWrapper = for {
+    k1 <- k1W
+    v1 <- v1W
+    kvs <- buildSeq(kvsW)
+  } yield _recSet(k1, v1, kvs: _*)
+
+  /** Alternate call method, where Scala strings are passed in place of ValWrappers */
+  def recSet(kv1W: (String, BuilderWrapper), kvsW: (String, BuilderWrapper)*): BuilderWrapper = for {
```
thpani, line 88: You don't actually know what this is an alternative to when you're reading the Scaladoc.
I suggest copying the above doc comment and linking the other method with `@see`.

thpani (commented):
Nicely organized, clean code, I really like this! ✨
I have a few comments inline.

LGTM from my side; I'm only not approving this so others have time to take a look.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/BuilderUtil.scala
```diff
@@ -0,0 +1,46 @@
...
+  // Constructs an OperEx expression for oper applied to args. Uses typeCmp to validate args and determine the result type
+  def buildInstruction(oper: TlaOper, typeCmp: typeComputation, args: TlaEx*): TlaEx = typeCmp(args) match {
+    case Left(err) => throw err
+    case Right(typeRes) =>
+      OperEx(oper, args: _*)(Typed(typeRes))
+  }
+
+  // Removes elem from the scope, as the scope only contains types of free variables
+  def markAsBound(elem: NameEx): InternalState[Unit] = State[MetaInfo, Unit] { mi: MetaInfo =>
+    (mi.copy(mi.nameScope - elem.name), ())
```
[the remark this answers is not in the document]
Kukovec: Future-proofing. For a concrete example, I'm considering adding automatic coverage analysis (i.e. which "branches" assign to which variables) to the builder (very far off in the future), so that when you build an expression it is automatically correct w.r.t. assignments. In that case, I'd extend `MetaInfo` with e.g. another map, but `markAsBound` wouldn't have to care about other components of `MetaInfo`.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/SignatureGenerator.scala
```diff
@@ -0,0 +1,72 @@
+package at.forsyte.apalache.tla.typecmp
+
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.oper.{FixedArity, TlaOper}
+import at.forsyte.apalache.tla.typecheck.etc.{Substitution, TypeUnifier, TypeVarPool}
+import at.forsyte.apalache.tla.typecmp.BuilderUtil.throwMsg
+import at.forsyte.apalache.tla.typecmp.signatures.{ArithOperSignatures, BoolOperSignatures, SetOperSignatures}
+
+/**
+ * Some TNT operators have signatures. SignatureGenerator serves them, from the operator identifier.
```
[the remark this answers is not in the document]
Kukovec: I think I noticed this one myself after I had already pushed, this is leftover from when the generator did more things that are now delegated to other classes.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/SignatureGenerator.scala
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+  }
+
+  // Performs unificaiton on 2 types with a fresh unifier
+  private def singleUnification(
+      lhs: TlaType1,
+      rhs: TlaType1,
+      subst: Substitution): Option[(Substitution, TlaType1)] = {
+    (new TypeUnifier).unify(subst, lhs, rhs)
```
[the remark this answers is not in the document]
Kukovec: This method used to be part of the package object, so it had to be stateless. I guess we can have a persistent `Unifier` now, but the performance diff is probably marginal either way.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/package.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  // Builder methods return TlaEx (they throw exceptions)
+  type builderReturn = TlaEx
+
+  // A type computation returns the result type of an operator application, given the concrete arguments
+  // In general, argument types are not enough, because we need concrete fields/indices for records/sequences
+  type typeComputation = Seq[TlaEx] => typeComputationReturn
+  // For some (most) operators, we don't need the exact values (just types) of the arguments to determine the result type
+  // Pure computations are typeComputations that are associated with such operators
+  type pureTypeComputation = Seq[TlaType1] => typeComputationReturn
```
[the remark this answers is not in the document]
Kukovec: I don't remember, I think I just went with the conventions from the 1st draft. But it is strange, now that you point it out. I'll uppercase them.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Builder for TlaSetOper expressions.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait SetBuilder extends RawSetBuilder {
+
+  /** {x1, ..., xn}, must have >0 args. */
+  def enumSet(argW: BuilderWrapper, argsW: BuilderWrapper*): BuilderWrapper = for {
```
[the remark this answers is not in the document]
Kukovec: To me, `require` is a necessary evil in methods which both a) take a variable number of arguments, and b) impose constraints only on _some_ of the arguments (e.g. every other arg. has to be a `NameEx` or a string `ValEx`).
In your case, you can do 
```scala
val h +: tail = args
enumSet(h, tail:_*)
```
instead of `enumSet(args:_*)`, so it shouldn't be too bad.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/BuilderUtil.scala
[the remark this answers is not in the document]
thpani: 👍

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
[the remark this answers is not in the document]
thpani: I won't block on this, but I would like to hear others' thoughts.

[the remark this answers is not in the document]
thpani: Pinging @shonfeder @konnov @gabrielamafra

thpani (approved):
This has been sitting around for some time; if it doesn't get more looks I might as well LGTM it.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/package.scala
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * `nameScope` tracks the types of the variables currently considered as free
+   */
+  final case class MetaInfo(nameScope: Map[String, TlaType1])
+
+  /** An IntenalState is a computation (possibly) mutating some MetaInfo */
+  type InternalState[T] = State[MetaInfo, T]
+
+  /** Builders operate over BuilderWrappers, which hold MetaInfo as state */
+  type BuilderWrapper = InternalState[BuilderReturn]
```
konnov, line 50:
```suggestion
  type TBuilderPiece = InternalState[BuilderReturn]
```

Do the users of your API care about the fact that this is a wrapper? It's important that this is a piece of work to be done by the builder. So I would called it a "piece", a "part", etc.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/package.scala
```diff
@@ -0,0 +1,92 @@
...
+   * assignment analysis.
+   *
+   * `nameScope` tracks the types of the variables currently considered as free
+   */
+  final case class MetaInfo(nameScope: Map[String, TlaType1])
+
+  /** An IntenalState is a computation (possibly) mutating some MetaInfo */
+  type InternalState[T] = State[MetaInfo, T]
+
+  /** Builders operate over BuilderWrappers, which hold MetaInfo as state */
```
konnov, line 49: Since the users will see this type in their IDE quite often, it needs a good description, not a comment about its internal mechanics.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/package.scala
```diff
@@ -0,0 +1,92 @@
...
+  type BuilderWrapper = InternalState[BuilderReturn]
+
+  // Special wrapper cases for subclasses of TlaEx
+  /** Variant of BuilderWrapper, when the built expression must be NameEx */
+  type NameWrapper = InternalState[NameEx]
```
konnov, line 54:
```suggestion
  type TBuilderNamePiece = InternalState[NameEx]
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/package.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  /** Variant of BuilderWrapper, when the built expression must be ValEx */
+  type ValWrapper = InternalState[ValEx]
```
konnov, line 57:
```suggestion
  type TBuilderValPiece = InternalState[ValEx]
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/RawLeafBuilder.scala
```diff
@@ -0,0 +1,37 @@
...
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.values._
+
+/**
+ * Raw builder for names and literals (IR tree leaves)
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait RawLeafBuilder {
```
konnov, line 12:
```suggestion
trait RawLiteralAndNameBuilder {
```

I would prefer a descriptive name, especially if it's used only internally.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/ProtoBuilder.scala
```diff
@@ -0,0 +1,26 @@
...
+ */
+trait ProtoBuilder {
+  protected val varPool: TypeVarPool
+  protected val sigGen: SignatureHandler = new SignatureHandler(varPool)
+
+  /**
+   * Specialized `buildInstruction`, applicable when the TNT operator has a signature, that is, it is not overloaded. In
+   * that case, we fetch the computation from the [[SignatureHandler]].
+   */
+  protected def simpleInstruction(oper: TlaOper, args: TlaEx*): TlaEx =
```
konnov, line 24:
```suggestion
  protected def buildBySignatureLookup(oper: TlaOper, args: TlaEx*): TlaEx =
```

I think `simpleInstruction` is only meaningful in the eyes of the author. This method may construct operators that are not simple to me. Moreover, I (as a method user) don't really care whether it was simple to implement or not. It is probably important that such an expression is essentially constructed by a table lookup.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/RawArithmeticBuilder.scala
```diff
@@ -0,0 +1,49 @@
...
+import at.forsyte.apalache.tla.lir.TlaEx
+import at.forsyte.apalache.tla.lir.oper.TlaArithOper
+
+/**
+ * Raw builder for TlaArithOper expressions.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait RawArithmeticBuilder extends ProtoBuilder {
```
konnov, line 12:
```suggestion
trait UnsafeArithmeticBuilder extends ProtoBuilder {
```

Again, "raw" does not tell much to the reader. I would call it "unsafe", because it may construct ill-typed expressions. Hence, the API user should not even look at this class.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/RawBoolBuilder.scala
```diff
@@ -0,0 +1,40 @@
...
+import at.forsyte.apalache.tla.lir.{NameEx, TlaEx}
+import at.forsyte.apalache.tla.lir.oper.TlaBoolOper
+
+/**
+ * Raw builder for TlaBoolOper expressions.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait RawBoolBuilder extends ProtoBuilder {
```
konnov, line 12:
```suggestion
trait UnsafeBoolBuilder extends ProtoBuilder {
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/RawSetBuilder.scala
```diff
@@ -0,0 +1,139 @@
...
+
+import scala.collection.immutable.SortedMap
+
+/**
+ * Raw builder for TlaSetOper expressions.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait RawSetBuilder extends ProtoBuilder {
```
konnov, line 17:
```suggestion
trait UnsafeSetBuilder extends ProtoBuilder {
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/RawSetBuilder.scala
```diff
@@ -0,0 +1,139 @@
...
+  /** { x \in set: p } */
+  protected def _filter(x: NameEx, set: TlaEx, p: TlaEx): TlaEx =
+    simpleInstruction(TlaSetOper.filter, x, set, p)
+
+  /** { e: x1 \in set1 , ..., xN \in setN }, must have at least 1 var-set pair */
+  protected def _map(
+      e: TlaEx,
+      x1: NameEx,
+      set1: TlaEx,
+      pairs: TlaEx*): TlaEx = {
```
konnov, line 49: I found it annoying that the builder differentiates between the first parameter and the other parameters. So I would prefer just `TlaEx*` or maybe `(NameEx, TlaEx)*`, which you can nicely write as `n1 -> e1, n2 -> e2, ...`. Perhaps, shapeless can help us in doing static checks to make sure that the list is non-empty. @shonfeder ?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/RawSetBuilder.scala
```diff
@@ -0,0 +1,139 @@
...
+    require(pairs.size % 2 == 0)
+    require(TlaOper.deinterleave(pairs)._1.forall { _.isInstanceOf[NameEx] })
+    simpleInstruction(TlaSetOper.map, e +: x1 +: set1 +: pairs: _*)
+  }
+
+  /** Function set constructor [fromSet -> toSet] */
+  protected def _funSet(fromSet: TlaEx, toSet: TlaEx): TlaEx = simpleInstruction(TlaSetOper.funSet, fromSet, toSet)
+
+  /** Record set constructor [ k1: v1, ... , kN: vN ]; must have at least 1 key-value pair */
+  protected def _recSet(k1: ValEx, v1: TlaEx, kvs: TlaEx*): TlaEx = {
```
konnov, line 60: Same here. I would prefer not to distinguish the first field from the other fields.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/raw/RawSetBuilder.scala
```diff
@@ -0,0 +1,139 @@
...
+
+    BuilderUtil.buildInstruction(TlaSetOper.recSet, typeCmp, k1 +: v1 +: kvs: _*)
+  }
+
+  /**
+   * Alternate call method, where Scala strings are passed in place of ValEx(TlaStr(_))
+   * @see
+   *   _recSet[[_recSet(k1: ValEx, v1: TlaEx, kvs: TlaEx*)]]
+   */
+  protected def _recSet(kv1: (String, TlaEx), kvs: (String, TlaEx)*): TlaEx = {
```
konnov, line 113: Same here. There should be no difference between the first pair and the other pairs.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/LeafBuilder.scala
```diff
@@ -0,0 +1,65 @@
...
+import scalaz.Scalaz._
+import scalaz._
+
+/**
+ * Builder for names and literals (IR tree leaves)
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait LeafBuilder extends RawLeafBuilder {
```
konnov, line 15:
```suggestion
trait LiteralAndNameBuilder extends RawLeafBuilder {
```

Names should reflect the use, not the implementation or another unnecessary detail.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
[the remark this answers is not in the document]
konnov: Yeah, I also wrote above that I did not like that about the old builder. If we want to enforce some properties of a sequence at compile-time, we should not put this burden on the user. And @shonfeder mentioned shapeless.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
```diff
@@ -0,0 +1,117 @@
...
+    p <- pW
+    _ <- markAsBound(x)
+  } yield _filter(x, set, p)
+
+  /** { mapExpr: x1 \in set1 , ..., xN \in setN }, must have at least 1 var-set pair */
+  def map(
+      eW: BuilderWrapper,
+      x1W: NameWrapper,
+      set1W: BuilderWrapper,
+      pairsW: BuilderWrapper*): BuilderWrapper = for {
```
konnov, line 60: Same here. I would prefer a uniform interface. Perhaps, a sequence of pairs? Scala's `->` can be very nicely applied here.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
```diff
@@ -0,0 +1,117 @@
...
+      cmp.flatMap({ _ => markAsBound(xi) })
+    }
+  } yield _map(mapExpr, x1, set1, pairs: _*)
+
+  /** Function set constructor [fromSet -> toSet] */
+  def funSet(fromSetW: BuilderWrapper, toSetW: BuilderWrapper): BuilderWrapper =
+    binaryFromRaw(fromSetW, toSetW)(_funSet)
+
+  /** Record set constructor [ k1: v1, ... , kN: vN ], must have at least 1 key-value pair */
+  def recSet(k1W: ValWrapper, v1W: BuilderWrapper, kvsW: BuilderWrapper*): BuilderWrapper = for {
```
konnov, line 81: Wanted: uniform treatment of all arguments, e.g., `(ValWrapper, BuilderWrapper)*`.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
```diff
@@ -0,0 +1,117 @@
...
+    v1 <- v1W
+    kvs <- buildSeq(kvsW)
+  } yield _recSet(k1, v1, kvs: _*)
+
+  /**
+   * Alternate call method, where Scala strings are passed in place of ValWrappers
+   * @see
+   *   recSet[[recSet(k1W: ValWrapper, v1W: BuilderWrapper, kvsW: BuilderWrapper*)]]
+   */
+  def recSet(kv1W: (String, BuilderWrapper), kvsW: (String, BuilderWrapper)*): BuilderWrapper = for {
```
konnov, line 92: Wanted: uniform treatment of all arguments, e.g., (String, BuilderWrapper)*.

konnov (commented):
Another batch of comments. I like your approach! My main concern is that the implementation details sneak into the naming and the methods too much. Hence, most of the changes are purely syntactic.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecmp/TestArithmeticBuilder.scala
```diff
@@ -0,0 +1,95 @@
...
+import at.forsyte.apalache.tla.lir.oper.TlaArithOper
+
+class TestArithmeticBuilder extends BuilderTest {
+  def testBinaryOperAndBuilderMethod(
+      oper: TlaArithOper,
+      method: (BuilderWrapper, BuilderWrapper) => BuilderWrapper,
+      retT: TlaType1): Unit = {
+    val cmp = sigGen.computationFromSignature(oper)
+
+    val args = Seq.fill(2)(builder.int(1))
```
konnov, line 13:
```suggestion
    val args = Seq(builder.int(10), builder.int(20))
```

I don't like riddles in tests :)

tla-types/src/test/scala/at/forsyte/apalache/tla/typecmp/TestBoolBuilder.scala
```diff
@@ -0,0 +1,156 @@
...
+    val builtArgs = args.map(build)
+    val res = cmp(builtArgs)
+
+    assert(res.contains(BoolT1()))
+
+    val resEx = eval(args)
+
+    assert(resEx.eqTyped(OperEx(oper, builtArgs: _*)(Typed(BoolT1()))))
+
+    val badY: BuilderWrapper = builder.str("a")
```
konnov, line 25:
```suggestion
    val illTypedY: BuilderWrapper = builder.str("a")
```

konnov (commented):
Just a few comments about the tests. End of my comments :)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/BuilderUtil.scala
```diff
@@ -0,0 +1,51 @@
+package at.forsyte.apalache.tla.typecmp
```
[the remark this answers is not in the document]
Kukovec: "computation"

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/SignatureHandler.scala
```diff
@@ -0,0 +1,67 @@
...
+import at.forsyte.apalache.tla.typecmp.signatures.{ArithOperSignatures, BoolOperSignatures, SetOperSignatures}
+
+/**
+ * Some TNT operators have signatures (see [[signatures]]). SignatureHandler collects [[SignatureGenerator]]s for such
+ * operators and constructs signature-matching [[PureTypeComputation]]s on demand.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+class SignatureHandler(varPool: TypeVarPool) {
```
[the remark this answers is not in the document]
Kukovec: `-Factory` maybe, but not `Signature-`, rather `TypeComputation-`, since it imports signatures and makes computations.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/package.scala
Kukovec: `-Instruction`? Since an `InternalState[TlaEx]` defines how to build `TlaEx` on demand?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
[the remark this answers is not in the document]
Kukovec: Ok, so the general sentiment seems to be to enforce minimal arity via runtime checks, so I'll go ahead and change those methods. 
Just one question, regarding maps (or methods with asymmetric args in general): do you prefer
`builder.map(e: T, xsAndSs: T*)` + `require(P(xsAndSs))` or `builder.map(eAndXsAndSs: T*)` + `require(P(eAndXsAndSs))`. Here, `e` and `xsAndSs` play different roles, and it's very unlikely that `eAndXsAndSs` would be programmatically constructed, since the components have different roles and shapes (unlike `xsAndSs`, which is uniform in shape).
I'm more in favor of the former, but I'd like to hear your opinions.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
Kukovec: I can make a `BuilderWrapper*` variant and a `(A,B)*` variant, like I do with record sets.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecmp/TestArithmeticBuilder.scala
Kukovec: I'll hold off on these changes, because I'm working on the generalzied testing framework where this should be parameterized by arity anyway.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecmp/TestArithmeticBuilder.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    val res = cmp(args.map(build))
+
+    assert(res.contains(retT))
+
+    val Seq(x, y) = args
+    val resEx = method(x, y)
```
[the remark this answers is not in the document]
Kukovec: See above

Kukovec (commented):
I'll incorporate @konnov 's naming suggestions, but not with GH's "commit suggestions", since they need refactoring to work.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/subbuilder/SetBuilder.scala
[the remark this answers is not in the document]
konnov: I would definitely prefer `builder.map(e: T, xsAndSs: T*)`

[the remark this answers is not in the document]
thpani: +1

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/package.scala
```diff
@@ -0,0 +1,92 @@
...
+  type PureTypeComputation = Seq[TlaType1] => TypeComputationReturn
+
+  /**
+   * MetaInfo holds all of the information about the internal state of the builder. For now, this just includes the name
+   * scope, but it can be extended in the future, to have the builder perform additional static analysis, e.g.
+   * assignment analysis.
+   *
+   * `nameScope` tracks the types of the variables currently considered as free
+   */
+  final case class MetaInfo(nameScope: Map[String, TlaType1])
```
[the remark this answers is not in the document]
Kukovec: Hmm, this is a transitional issue at best, since there's not going to be another builder after the overhaul. But I'll change it for now.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecmp/BuilderUtil.scala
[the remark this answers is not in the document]
shonfeder: This would be clearer as

```suggestion
package at.forsyte.apalache.tla.typecomp
```

konnov (approved):
Nothing catches my eye! Looks great!

thpani:
I can't do a delta review here, because the package renaming breaks the diff. Also this PR has become so large that I don't want to do another full review.
IMO we should've merged this at some point and followed up with another PR to add the tests.

PR #1649, opened 2022-04-21
Fix Markdown end-of-codeblock

thpani:
Move inline <code>```</code> to a new line.
This caused the bibliography to print as part of the code block, and links not being rendered.


shonfeder (approved):
LGTM!

PR #1652, opened 2022-04-22
Integrate features flag into ConfigManager

shonfeder:
As per our discussion in the meeting today, this integrates the `--feature` CLI
flag into the configuration manager, allowing users to enable features via their
global or local config files.

Along the way, it reworks some of the base PR in order to improve code reuse,
and also fixes uncaught exceptions resulting from invalid config files.

Note that  this PR is into Igor's feature branch.

It should be easiest to review by commit, but the changes are hopefully pretty clear and well motivated anyhow.

### DIFF
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/General.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/General.scala
@@ -36,7 +36,7 @@ trait General extends Command with CliConfig {
       useEnv = true)
   var features = opt[Seq[Feature]](default = Seq(),
       description = {
-        val featureDescriptions = Feature.all.map(f => s" - ${f.name}: ${f.description}")
+        val featureDescriptions = Feature.all.map(f => s"  ${f.name}: ${f.description}")
         ("a comma-separated list of experimental features:" :: featureDescriptions).mkString("\n")
       })
 
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/General.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/General.scala
@@ -1,7 +1,7 @@
 package at.forsyte.apalache.tla.tooling.opt
 
 import at.forsyte.apalache.io.CliConfig
-import at.forsyte.apalache.tla.lir.{Feature, RowsFeature}
+import at.forsyte.apalache.tla.lir.Feature
 
 import java.io.File
 import org.backuity.clist._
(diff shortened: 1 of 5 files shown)


mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/General.scala
```diff
@@ -35,23 +35,26 @@ trait General extends Command with CliConfig {
         "write intermediate output files to `out-dir`, default: false (overrides envvar WRITE_INTERMEDIATE)",
       useEnv = true)
   var features = opt[Seq[Feature]](default = Seq(),
-      description = """a comma-separated list of experimental features:
-        | - rows: enable row typing as explained in ADR-014
-        |""".stripMargin)
+      description = {
+        val featureDescriptions = Feature.all.map(f => s" - ${f.name}: ${f.description}")
+        ("a comma-separated list of experimental features:" :: featureDescriptions).mkString("\n")
```
thpani, line 40: Wow, this even idents by the right amount! 🤩

The `-` could maybe be confusing?
 ```suggestion
        val featureDescriptions = Feature.all.map(f => s"  ${f.name}: ${f.description}")
        ("a comma-separated list of experimental features:" :: featureDescriptions).mkString("\n")
```

File: tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala [hunk 2]
```diff
@@ -40,7 +44,8 @@ case class ApalacheConfig(
     runDir: Option[File] = None,
     configFile: Option[File] = None,
     writeIntermediate: Boolean = false,
-    profiling: Boolean = false)
+    profiling: Boolean = false,
+    features: Seq[Feature] = Seq())
```
thpani, line 48:
```suggestion
    enabledFeatures: Seq[Feature] = Seq())
```

thpani (approved):
👍 LGTM!
Really enjoyed the individual commits + informative messages ✨

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/General.scala
shonfeder: Sounds good. Less noise is better :) thanks!

shonfeder:
Thanks for the review! I’ll leave it for eager to also check since it’s into his code.

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
shonfeder: im ok with this change, but then I think we should change the cli flag to match, as we should keep those names in sync. What do you think?

thpani: Good point, but I find `--enabled-features` a bit verbose. I'd say it's probably sufficiently clear by convention that `--features=a,b` enables features `a` and `b`; otherwise `--enable=a,b` might also be a good name.

thpani: I was more worried that in this code snippet, the type signature alone does not tell us what parameter `features` is supposed to do.

shonfeder: Good point! I added https://github.com/informalsystems/apalache/pull/1652/commits/aea03e8cc381009851553da656002bf883388921 hoping to address this without duplicating documentation :)

konnov (approved):
Very nice!

PR #1664, opened 2022-04-25
Replace assignments in invariants

### LINKED ISSUE #1663: Error when checking UNCHANGED in an invariant
Consider the following specification:

```tla
---------------------------- MODULE UnchangedAsInv ----------------------------  
EXTENDS Integers

VARIABLE
    \* @type: Int;
    x

Init == x = 0

Next == x' = x + 1

Inv == UNCHANGED x
===============================================================================
```

When I run apalache as follows:

```sh
apalache-mc check --inv=Inv --write-intermediate=1 UnchangedAsInv.tla
```

It throws an exception:

```
...
State 0: Checking 1 action invariants                             I@17:28:18.693
UnchangedAsInv.tla:15:5-15:15: rewriter error: No rewriting rule applies to expression: Apalache!:=(x', x) E@17:28:18.700
at.forsyte.apalache.tla.bmcmt.RewriterException: No rewriting rule applies to expression: Apalache!:=(x', x)
	at at.forsyte.apalache.tla.bmcmt.SymbStateRewriterImpl.doRecursive$1(SymbStateRewriterImpl.scala:397)
	at at.forsyte.apalache.tla.bmcmt.SymbStateRewriterImpl.rewriteUntilDone(SymbStateRewriterImpl.scala:416)
	at at.forsyte.apalache.tla.bmcmt.rules.NegRule.apply(NegRule.scala:28)
...
```

This is happening because `UNCHANGED x` was replaced with `x' := x` by Desugarer.

This kind of invariants appear when proving refinement. So we have to fix it.

konnov:
Closes #1663. As discussed in #1663, the model checker fails when an invariant contains assignments, e.g., `UNCHANGED x`. The bugfix is very simple: All assignments `x' := e` are replaced back to `x' = e` in the invariant candidates.

Instead of implementing a yet another class similar to `Prime` and `Deprime`, I have refactored `ReplaceFixed` to accept a partial function, which produces a replacement, if the function is defined on its argument. To disambiguate three versions of `apply`, I have renamed them to `whenEqualsTo`, `whenMatches`, and `withFun`. This refactoring touched a lot of files, but the changes are syntactic.


tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/package.scala
```diff
@@ -84,4 +87,32 @@ package object transformations {
...
+   */
+  def decorateWithPrime(vars: Set[String], tracker: TransformationTracker): TlaExTransformation = {
+    ReplaceFixed(tracker).withFun {
+      case ex @ NameEx(name) if vars.contains(name) =>
+        // add prime to a variable from vars
+        OperEx(TlaActionOper.prime, ex)(ex.typeTag)
+
+      case OperEx(TlaActionOper.prime, primedOnce @ OperEx(TlaActionOper.prime, _)) =>
+        // in case the variable was already primed, remove the second prime
+        primedOnce
```
Kukovec, line 115: Rather than defining de-priming of doubly primed expressions, which relies on the implementation detail that `ReplaceFixed` implements post-transform too 
https://github.com/informalsystems/apalache/blob/919e65948904156615a05980d98b335fd5a8fc2a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReplaceFixed.scala#L96-L99
I'd prefer defining this in a way that never attempts to add unnecessary primes in the first place:
```suggestion
      case ex@OperEx(TlaActionOper.prime, _) =>
        // in case the expression was already primed, ignore
        ex
        
      case ex @ NameEx(name) if vars.contains(name) =>
        // add prime to a variable from vars
        OperEx(TlaActionOper.prime, ex)(ex.typeTag)
```

Kukovec (approved):
👍

tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/package.scala
konnov: Are you sure that this code would not introduce two primes, i.e., by going to 113-115 first and then to 109-111 one level up in the recursion?

Kukovec: When it goes one level up it will already be primed, so 109 will skip.
```
x -(113)-> x' -(109)-> x'
x' -(109)-> x' -(109)-> x'
```

konnov: Ah, that makes sense.

Kukovec: I looked into it, seems like your implementation of `withFun` doesn't allow short-circuiting.
Unlike a dedicated transformation, like `Prime`, it must always first look into the generic `OperEx` case
https://github.com/informalsystems/apalache/blob/919e65948904156615a05980d98b335fd5a8fc2a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/ReplaceFixed.scala#L96-L99
This is potentially problematic in practice; a lot of transformations rely on the fact that certain synactic forms halt exploration, because transforming may ruin subexpressions, but these generic `PartialFunction` traversals do not, they always full explore the expression tree.

konnov: True. `Prime` had a very special behavior. We can think of generalizing `ReplaceFixed` even further.

Kukovec: It's not just Prime, a typical tx has cases like 
```scala
case InterestingCase => interestingCaseTx
case OperEx(args) => mapOverArgs
case LetIn => mapOverDecls
case default => defaultTx
```

where `interestingCaseTx` would typically not recurse any more. I think this can be fixed by changing
```scala
    ex match {
      case LetInEx(body, defs @ _*) =>
        // Transform bodies of all op.defs
        val newDefs = defs.map(tracker.trackOperDecl { d => d.copy(body = self(d.body)) })
        val newBody = self(body)
        val retEx = if (defs == newDefs && body == newBody) ex else LetInEx(newBody, newDefs: _*)(ex.typeTag)
        tr(retEx)


      case OperEx(op, args @ _*) =>
        val newArgs = args.map(self)
        val retEx = if (args == newArgs) ex else OperEx(op, newArgs: _*)(ex.typeTag)
        tr(retEx)


      case _ => tr(ex)
```
in `withFun` to
```scala
    ex match {
      case special if partialFun.isDefinedAt(special) =>
        partialFun(special)

      case LetInEx(body, defs @ _*) =>
        // Transform bodies of all op.defs
        val newDefs = defs.map(tracker.trackOperDecl { d => d.copy(body = self(d.body)) })
        val newBody = self(body)
        val retEx = if (defs == newDefs && body == newBody) ex else LetInEx(newBody, newDefs: _*)(ex.typeTag)
        retEx

      case OperEx(op, args @ _*) =>
        val newArgs = args.map(self)
        val retEx = if (args == newArgs) ex else OperEx(op, newArgs: _*)(ex.typeTag)
        retEx

      case _ => ex
```

konnov: It got automerged. Shall we open another PR for your suggestions?

PR #1510, opened 2022-03-17
Upgrade to Scala 2.13

shonfeder:
Closes #1509

I thought I'd see how difficult it would be upgrade, and turns out it was
relatively easy to fix the compiler errors.

The remaining compiler issues are warnings.

Opening this PR so we can see how it fairs on the tests.

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] Documentation added for any new functionality
- [x] Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality

PR #1677, opened 2022-04-27
Sync Scala version in docs

thpani:
After the version bump in #1510, sync Scala version in docs.


shonfeder (approved):
Sorry I missed this! Thanks for the cleanup :)

PR #1681, opened 2022-04-27
Small improvements to the encodings comparison

thpani:
Small improvements to the encodings comparison #1588:

- print RNG seeds
- label inputs
- use expected comparison `?=`


konnov (approved):
Nice!

PR #1684, opened 2022-04-27
Move version number printing and include in detailed log

p-offtermatt:
~- [ ] Tests added for any new code~
~- [ ] Documentation added for any new functionality~


konnov (approved):
Nice! There is probably a rare case, if a log misconfiguration happens and the users do not know the version number. Most likely this is not important though.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
```diff
@@ -443,11 +445,6 @@ object Tool extends LazyLogging {
   }
 
   private def printHeaderAndStatsConfig(): Unit = {
```
thpani, line 447:
```suggestion
  private def printStatsConfig(): Unit = {
```

thpani (approved):
LGTM! Thanks for implementing this!

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
p-offtermatt: Thanks, good catch!

PR #1686, opened 2022-04-27
Add silent logback config for unit tests

### LINKED ISSUE #1666: Divert logger output on unit tests
The transition executor and model checker classes produce a lot of logger output, especially when run in a loop as with the encodings comparison #1588.
While useful in the CLI, we need to find a way to limit this output on unit tests.

thpani:
Add silent logback configs for unit tests in in `mod-infra` and `tla-bmcmt`.

Since these are per (sub-)project, I'm starting with a minimal set for the unit tests in `TestPassChainExecutor` and `CrossTestEncodings`.

Closes #1666.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -301,6 +301,14 @@ Run the units
 make test
 ```
 
+### Logging in unit tests
+
+We disable unit test log output for subprojects where necessary, to avoid output
+flooding the console. This affects unit tests only, and is configured in a
+per-subproject logback configuration file `test/resources/logback-test.xml`.
+This file also contains a commented-out console appender that can be enabled if
+needed for debugging purposes.
+
 ### Integration tests
 
 #### Installing Dependencies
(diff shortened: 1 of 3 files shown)


konnov (commented):
Can we conditionally enable output to the console, which may be sometimes useful when debugging. We could enable this with an environment variable. There is a page on conditional configuration: https://logback.qos.ch/manual/configuration.html#conditional

thpani:
> Can we conditionally enable output to the console, which may be sometimes useful when debugging. We could enable this with an environment variable. There is a page on conditional configuration: https://logback.qos.ch/manual/configuration.html#conditional

Isn't this a bit overkill? The conditional config requires an external library.
Also, if you set an environment variable, you might as well just temporarily `rm` the logback config?

konnov:
> Isn't this a bit overkill? The conditional config requires an external library.

Ugh. I did not know that.

> Also, if you set an environment variable, you might as well just temporarily rm the logback config?

I am just thinking about a debugging session where I would need the log output. How about:

 - adding a commented-out appended in the xml config, that tells the developers to uncomment it when debugging.
 - adding a note about it here: https://github.com/informalsystems/apalache/blob/unstable/CONTRIBUTING.md#unit-tests

konnov (approved):
Perfect!

PR #1690, opened 2022-04-28
Improve Apalache manual intro sections

### LINKED ISSUE #1672: Clean up `system-reqs.md`
The info in [`docs/src/apalache/system-reqs.md`](https://github.com/informalsystems/apalache/blob/unstable/docs/src/apalache/system-reqs.md) is outdated, perhaps a bit confusing, and can probably be merged with `installation/index.md`.

thpani:
Currently, the first few subsections of the Apalache manual are rather short and partially outdated.

This PR attempts to:
- update outdated info
- merge short sections into more meaningful chunks
- add context where necessary

Probably best reviewed by commit. Closes #1672.


docs/src/apalache/index.md
```diff
@@ -1,20 +1,27 @@
...
+by [Microsoft Z3][]. That is, Apalache operates on formulas (i.e.,
+_symbolically_), not by enumerating states one by one (i.e., _state
+enumeration_).
+
+### Shall I use Apalache or TLC?
+
+We recommend starting with TLC. It is mature, well-documented, and
+well-integrated into TLA+ Toolbox. Once you have debugged your TLA+
+specification, and TLC is still producing too many reachable states, switch to
+Apalache. We are using this approach at [Informal Systems][].
```
konnov, line 21: Shall we remove this paragraph too? As we have found, people would write more efficient specs for Apalache, if they start with Apalache. So the old workflow is less obvious to me now.

konnov (approved):
Looks good. Thanks for taking care of that!

docs/src/apalache/index.md
thpani: Maybe we should include something like the snippet we wrote for the TLA+ website?

Something like
> Depending on the specification you wrote, either TLC or Apalache may be more efficient in checking it. While TLC is a mature tool, Apalache is still experimental, so be prepared to use the command-line and to help us discover bugs.

konnov: Yes, this looks good.

PR #1696, opened 2022-04-29
Refactor CellT

### LINKED ISSUE #1693: Refactor CellT
To accept `TlaType1` instead of introducing case classes that mirror those of `TlaType1`. This is needed for #401.

konnov:
Closes #1693. To enable implementation of new records, we have to refactor `CellT`. Instead of introducing new case classes that mirror the new types of `TlaType1`, we introduce a new case class `CellTFrom(tt)` that decorates `TlaType1`. This allowed me to:

 - Remove the other cases classes of `CellT` that mirrored some of `TlaType1`.
 - Remove the old unification code.
 - Clean up old code that is not needed anymore.

Unfortunately, this change affects plenty of files in the model checker. Most of these changes are mechanical though.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/types/package.scala
```diff
@@ -165,23 +38,13 @@ package object types {
      */
     def fromType1(tt: TlaType1): CellT = {
```
shonfeder, line 39: I've a general clarifying question: iiuc, it looks like the only way to construct a `CellT` is via a `TlaTyp1` (`fromTypeTag` is just a wrapper around getting the `TlaType1` from a type tag). So we not just define `CellT` as a case class that wraps a `TlaType1` directly, rather than defining a case class of the trait?

shonfeder: A spitball idea: would it make any sense to define `case class CellT[T <: TlaType1](tt: T)` so that we'd be able to take advantage of static analysis when we'd narrowed down to a specific type?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateDecoder.scala
```diff
@@ -212,6 +156,61 @@ class SymbStateDecoder(solverContext: SolverContext, rewriter: SymbStateRewriter
       throw new RewriterException("Don't know how to decode the cell %s of type %s".format(cell, cell.cellType), NullEx)
   }
 
+  private def decodeFunToTlaEx(arena: Arena, cell: ArenaCell, funT1: FunT1): TlaEx = {
+    // a function is represented with the relation {(x, f[x]) : x \in S}
+    val relation = arena.getCdm(cell)
+
+    def isInRelation(pair: ArenaCell): Boolean = {
```
Kukovec, line 163: You can pattern match `funT1` at the start for readability: 
```suggestion
  private def decodeFunToTlaEx(arena: Arena, cell: ArenaCell, funT1: FunT1): TlaEx = {
    // a function is represented with the relation {(x, f[x]) : x \in S}
    val relation = arena.getCdm(cell)
    val FunT1(domT, cdmT) = funT1

    def isInRelation(pair: ArenaCell): Boolean = {
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateDecoder.scala
```diff
@@ -212,6 +156,61 @@ class SymbStateDecoder(solverContext: SolverContext, rewriter: SymbStateRewriter
...
+        seen + (keyEx -> valueEx)
+      }
+    }
+
+    val pairT = TupT1(funT1.arg, funT1.res)
+    val pairs = arena
+      .getHas(relation)
+      .filter(isInRelation)
+      .foldLeft(Map[TlaEx, TlaEx]())(decodePair)
+      .map(p => tla.tuple(p._1, p._2).as(pairT))
```
Kukovec, line 209:
```suggestion
      .map { case (k,v) => tla.tuple(k, v).as(pairT) }
```

Kukovec (approved):
Looks OK, though stylistically, I'd replace pattern matches with anonymous args like
```scala
case F(TupT1(_ @_*)) => ...
case F(a @ FunT1(_,_)) => ...
```
with 
```scala
case F(_ : TupT1) => ...
case F(a : FunT1) => ...
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/types/package.scala
konnov: `CellT` has several case classes that do not fit into `TlaType1` directly. These case classes with disappear in the future refactorings. Ultimately, we should use `TlaType1` only.

konnov:
> > Are you sure it works all the time? I had an impression that some of these types can get erased when you use the notation `_: T`.
> 
> Erasure is only ever a problem for parametric types.

OK. I will keep it in mind for the future refactorings. Going through all the changes in this PR is a bit of pain.

PR #1700, opened 2022-05-02
Update docs on docker-based installation

### LINKED ISSUE #1657: Docker installation docs need updating
Thanks to feedback from @tesnimab, there are some ways we should improve the documentation around install via docker:

- explain what APALACHE_HOME is (probably by linking to docs)
- correct the docker image tag name (using old tag name)
- note that docker is not the recommended way to run Apalache, due to limitations in accessing and interacting with the users environment.

shonfeder:
Closes #1657

This small fix addresses some shortcomings in the docs for the
docker installation. Thanks to feedback from @tesnimab for pointing
out the issues.


shonfeder:
Auto merge enabled, so please don't approve if you'd like to see some changes.

konnov (approved):
Looks good. Just looking at our docs, perhaps, we could make one more step and build native packages with https://docs.oracle.com/en/java/javase/14/docs/specs/man/jpackage.html

Issue #1577, opened 2022-03-29
Stop forking tests


thpani:
We should remove `Test / fork := true` from `build.sbt`: https://github.com/informalsystems/apalache/pull/1553#discussion_r834524465

This will require some additional work:
- Without forking, sbt runs tests from the project root (instead of subproject directories). This breaks some tests with relative paths to `.tla` files in unit tests.
- [Forked tests run sequentially](https://www.scala-sbt.org/1.x/docs/Testing.html#Forking+tests). There are some unit tests that fail when run in parallel.

In summary:
- [x] Remove `Test / fork := true` #1705
- [x] Fix path-dependent unit tests #1705

shonfeder:
We think maybe the pending work item dependent on SANY can be factored out into its own issue and this closed.

thpani:
Closed by #1705 after factoring out #1761

PR #1705, opened 2022-05-03
Stop forking JVMs for unit tests

thpani:
Towards #1577

Stop forking JVMs for unit tests.
This enables unit tests to run in parallel (forked tests [run sequentially by default](https://www.scala-sbt.org/1.x/docs/Testing.html#Forking+tests)).

This fixes some paths, because tests now run from the project root (instead of subproject directories).

Due to [this race in SANY](https://github.com/tlaplus/tlaplus/issues/688) we still need to run some unit tests sequentially.


shonfeder (approved):
Nice!

PR #1621, opened 2022-04-12
Implement rewriting subseteq with forall quantification in Keramelizer

p-offtermatt:
- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [ ] ~~Documentation added for any new functionality~~
- [x] Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality

PR #1716, opened 2022-05-05
Remove SetInclusionRule{,WithArrays}

### LINKED ISSUE #1710: Remove SetInclusionRule and SetInclusionRuleWithArrays
Since `S \subseteq T` is rewritten by the Keramelizer in #1621, we can safely remove the two rewriting rules, which are now redundant. We also have to remove `\subseteq` as an acceptable operator in `KeraLanguagePred`.

@rodrigo7491 what do you think about that?
@thpani in case you have free refactoring cycles.

thpani:
Keramelizer rewrites `S \subseteq T` since https://github.com/informalsystems/apalache/pull/1621.

* Remove `\subseteq` as an acceptable operator from `KeraLanguagePred`.
* Remove `SetInclusionRule` which is now redundant.

There are still two delegations to `\subseteq` for `arrays` in `CherryPick` and `SetInRuleWithArrays`, thus we keep `SetInRuleWithArrays`.

This PR also includes a partial attempt at removing the delegation in `SetInRuleWithArray`:

* Rewrite `f \in [S -> T]` in Keramelizer

However, this does not make the rule obsolete, because we don't inline nullary LET-INs. Thus a membership test like in https://github.com/informalsystems/apalache/issues/1735 will still reach this rule.

Closes #1710


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
```diff
@@ -90,8 +90,8 @@ class SetInRuleWithArrays(rewriter: SymbStateRewriter) extends SetInRule(rewrite
           nextState = rewriter.rewriteUntilDone(nextState.setRex(tla.appFun(funCell.toNameEx, funsetDomElem.toNameEx)))
           val funAppRes = nextState.asCell
           val powSetDom = nextState.arena.getDom(funsetCdm)
-          val subsetEq = tla.subseteq(funAppRes.toNameEx, powSetDom.toNameEx)
-          nextState = rewriter.rewriteUntilDone(nextState.setRex(subsetEq))
+          val quant = tla.forall(tla.name("x"), funAppRes.toNameEx, tla.in(tla.name("x"), powSetDom.toNameEx))
```
thpani, line 93: Not sure if it's sound to not use a fresh name here?
If not, how would I create one?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
```diff
@@ -828,8 +828,7 @@ class CherryPick(rewriter: SymbStateRewriter) {
           cdm.cellType match {
             case _: PowSetT =>
               val powSetDom = nextState.arena.getDom(cdm)
-              val appFunResInDom = tla.subseteq(powSetDom.toNameEx, appFunRes.toNameEx)
-              nextState = rewriter.rewriteUntilDone(nextState.setRex(appFunResInDom))
+              nextState = rewriter.lazyEq.subsetEq(nextState, powSetDom, appFunRes)
```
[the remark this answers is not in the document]
thpani: Igor suggested going through `LazyEquality` here. @konnov?

Kukovec (approved):

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
thpani: Renaming in [83bbac4](https://github.com/informalsystems/apalache/pull/1716/commits/83bbac46be1236ee55ee957c89c8269465351204)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
```diff
@@ -90,8 +91,9 @@ class SetInRuleWithArrays(rewriter: SymbStateRewriter) extends SetInRule(rewrite
           nextState = rewriter.rewriteUntilDone(nextState.setRex(tla.appFun(funCell.toNameEx, funsetDomElem.toNameEx)))
           val funAppRes = nextState.asCell
           val powSetDom = nextState.arena.getDom(funsetCdm)
-          val subsetEq = tla.subseteq(funAppRes.toNameEx, powSetDom.toNameEx)
-          nextState = rewriter.rewriteUntilDone(nextState.setRex(subsetEq))
+          // rename to obtain a unique name for the bound variable
+          val quant = renaming(tla.forall(tla.name("x"), funAppRes.toNameEx, tla.in(tla.name("x"), powSetDom.toNameEx)))
+          nextState = rewriter.rewriteUntilDone(nextState.setRex(quant))
```
konnov, line 96: I think my previous comment got deleted. I think the proper way of doing that is by rewriting `f \in [S -> SUBSET T]` into `DOMAIN f = S /\ \A x \in S: \A y \in f[x]: y \in T` in `Keramelizer`. Since it is about rewriting expressions, we should move it in preprocessing.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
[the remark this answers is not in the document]
thpani: Now I'm confused. AFAICT `CherryPick` isn't used by any arrays rules?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
thpani: So is `f \in [S -> SUBSET T]` not in KerA+?

konnov: we have it in KerA+. But I had this idea for some time that we should translate expressions over functions sets and powersets in the preprocessor, so we do not have to deal with them in the rewriting rules: https://github.com/informalsystems/apalache/milestone/56

If you think that it would be easier to handle this case in `Keramelizer`, go ahead.

thpani: I asked because you said above
> I think the proper way of doing that is by rewriting ... into ... in Keramelizer

So is the sentiment that we should do it in preprocessing, but not in Kereamelizer?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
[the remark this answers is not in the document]
rodrigo7491: It is, e.g., in [FunAppRuleWithArrays](https://github.com/informalsystems/apalache/blob/fd57ed590fb892731f7284590b88b604d6514082/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/FunAppRuleWithArrays.scala#L50-L53).

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
konnov: We have `Keramelizer` and `ExprOptimizer`. `Keramelizer` was designed for general reductions, whereas `ExprOptimizer` was designed for sound but ad hoc simplifications. Your case looks like a fit for `Keramelizer`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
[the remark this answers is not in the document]
thpani: Yeah sorry, missed the `case arraysEncoding` above

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
thpani: Okay. Still, shouldn't that be an additional preprocessor then? It seems confusing to have a class `Keramelizer` with its Scaladoc saying `a simplifier from TLA+ to KerA+` rewrite an expression that is already KerA+.

thpani: Also, for my own understanding: why is it better to have this in Keramlizer than the rewrite rule?

konnov: I mean, KerA+ is not set in stone. We have removed some expressions from it in the past. It would be better to handle it in the preprocessor than in a rewrite rule, since preprocessing is purely about TLA-to-TLA transformations, whereas rewriting rules are producing SMT constraints and cells. So having a purely TLA-to-TLA transformation in a rewriting rule usually means that this transformation should be moved up in the pipeline. Normally, TLA-to-TLA transformations do not depend on the SMT, whereas rewriting rules heavily depend on the encoding.

konnov: > Okay. Still, shouldn't that be an additional preprocessor then? It seems confusing to have a class `Keramelizer` with its Scaladoc saying `a simplifier from TLA+ to KerA+` rewrite an expression that is already KerA+.

In principle, you are right. I wanted to do some systematic refactoring of the preprocessing passes, e.g, see #480. I would do that in Keramelizer at the moment and come back to a more principled solution later, when we understand what we want from the TLA-to-TLA framework.

thpani: Everything about moving this into Keramelizer seems to be orthogonal to this PR. I'd like to understand what's wrong with the proposed solution. Are you concerned about the name generation or the quantifier? The OOPSLA rule also expands this expression into a conjunction over the function relation.

CC @rodrigo7491

konnov: Yes, I am concerned about name generation in the rewriting rule.

thpani: Good point! @rodrigo7491, do you see any (performance) problem with rewriting this in Keramelizer as suggested by Igor?
One thing to keep in mind is that also `Apalache!SelectInSet` goes through [SetInRuleWithArrays.scala](https://github.com/informalsystems/apalache/pull/1716/files/83bbac46be1236ee55ee957c89c8269465351204#diff-5624951d59d02e27a1efb359513855166e609fcfd97e454c67a72d539fa246f0)

rodrigo7491: From what I understand we will get a quantified expression in the end, regardless of where the rewriting happens, so I don't think it should be a problem. I don't get your point about the `SelectInSet` IR op though @thpani.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/CherryPick.scala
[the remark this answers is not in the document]
thpani: It seems infeasible to remove this or rewrite it in Keramelizer, as we call `pick`, e.g., in `CardinalityConstRule`. We also don't have proper name generation here, so we need to delegate to \subseteq and keep `SetInclusionRuleWithArrays` for now.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
thpani: > I don't get your point about the `SelectInSet` IR op though @thpani.

Ah, I just meant that `SelectInSet` won't be rewritten by `Keramelizer`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
thpani: Ugh, so since we're not inlining nullary LET-INs, rewriting this in Keramelizer actually won't allow us to get rid of this case. Something like the spec in https://github.com/informalsystems/apalache/issues/1735 would still need this rewrite rule.

@konnov Should I still include the rewriting in Keramelizer, even though it's orthogonal to this PR?

konnov: You are right. Feel free to create two PRs instead of one. We should redesign the preprocessing to make it all work as we expected here.

PR #1718, opened 2022-05-06
Update z3

shonfeder:
Closes #1702, closes #1660, closes #751

Issue #751, opened 2021-04-19
Apalache build fails on Apple silicon


andrey-kuprianov:
## Description

Apalache fails to build on Apple ARM due to missing support from Z3 side:

```
Tests in error: 
 at.forsyte.apalache.tla.assignments.TestSymbTransPass: Unsupported CPU architecture: aarch64
 at.forsyte.apalache.tla.assignments.TestManualAssignments: Could not initialize class com.microsoft.z3.Native
```

## System information

- Apalache version 0.15.2:
- OS: Mac OS
- JDK version: openjdk 11.0.10 2021-01-19 LTS

konnov:
I just relabelled it as an enhancement, as we could not foresee Apple introducing one more CPU architecture.

Just curious, does it work in docker on M1?

andrey-kuprianov:
Just checked: as a workaround, running Apalache via docker on Apple M1 works fine:)
https://apalache.informal.systems/docs/apalache/installation/docker.html

andrey-kuprianov:
Sorry, I was too fast to celebrate. Some phases work fine; but it seems that as soon as it comes to running Z3, it segfaults.

andrey-kuprianov:
Here is a more precise info when running different Docker images: 

**v.0.9.0**
Non-deterministically halts at different places; I also got a segfault:
```
  > After preprocessing: UniqueRenamer                            I@08:35:21.478
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000401798c3a0, pid=44, tid=47
#
# JRE version: OpenJDK Runtime Environment (9.0+12) (build 9.0.4+12-Debian-4)
# Java VM: OpenJDK 64-Bit Server VM (9.0.4+12-Debian-4, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# J 3147 c1 org.bitbucket.inkytonik.kiama.output.PrettyPrinter$$Lambda$248.apply(Ljava/lang/Object;)Ljava/lang/Object; (8 bytes) @ 0x000000401798c3a0 [0x000000401798c380+0x0000000000000020]
```

**v.0.8.3**
It starts checking the steps, but then halts

**v.0.8.2, v.0.8.0**
Sometimes they work fine, and produce the counterexample. But the counterexamples are non-deterministic, which was supposed to be fixed in #318 in v.0.8.0.

Sometimes I get again segfaults:
```
[...]

PR #1721, opened 2022-05-06
Note support for aarch64 in release notes

thpani:
Note support of AArch64 through #751 and #1718 in the release notes.

⚠️ Set to auto-merge.


shonfeder (approved):
Good call! thanks for adding it :)

PR #1725, opened 2022-05-09
Introduce an optimization in `CherryPick`

konnov:
This PR refactors the `Oracle` interface, which we use to pick an arbitrary value from a sequence of elements or from a set. It also introduces a new oracle `ZipOracle` that compresses several values into a single one. By using `ZipOracle` in `CherryPick`, we can achieve a speed up on some benchmarks.

Below is a table for a small subset of benchmarks. We probably have to do large scale experiments later.

We have tried various configuration options:

 - `new`: using ZipOracle
 - `old`: not using ZipOracle
 - `bef`: using `--tuning-options=search.invariant.mode=before`
 - `aft`: using `--tuning-options=search.invariant.mode=after`
 - `nd`: using `--discard-disabled=0`

For some of the benchmarks, we ran 10-20 experiments with `hyperfine`, to
compute the average running times. The running times for such experiments look
like `t ± d`, that is, the expected running time lies in the time interval
`[t - d, t + d]`. These experiments demonstrate fluctuations of Z3.

## BMC up to 10 steps

|           | new/bef | old/bef | new/aft |old/aft | new/aft/nd | old/aft/nd |
| --------- | ------- | ------- | ------- | ------ | ---------- | ---------- |
| perf      |280 ± 25 |340 ± 104| 28 ± 4  | 30 ±12 |     14 ± 2 |    15 ± 3  |
| perf2     |9.35± 0.6|10.8± 1.2|8.2 ± 1.5|8 ± 1.2 |10.16 ± 4.7 |    9 ± 1   |
| EWD998-4  |183 ± 16 |   210   | 97 ± 7  |    106 | 95 ±  8    |     94     |
| ChangR-4  |  2952   |         |  2829   |        |      998   |     1105   |

## Inductive invariants

|           | new/bef | old/bef | new/aft |old/aft | new/aft/nd | old/aft/nd |
| --------- | ------- | ------- | ------- | ------ | ---------- | ---------- |
| EWD998-40 |154 ± 13 | 151 ± 7 |460 ± 66 | 478±75 | 441 ± 87   | 480 ± 45   |
| Bakery-3  | 32 ± 3  | 29 ± 5  | 20 ± 4  |19 ±1.6 | 17.5 ± 2   | 19.3 ± 3.4 |
 


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/Oracle.scala
```diff
@@ -12,6 +15,14 @@ import at.forsyte.apalache.tla.lir.TlaEx
  */
 trait Oracle extends Serializable {
 
+  /**
+   * The number of values that this oracle is defined over: `0..(size - 1)`.
```
shonfeder, line 19: I don't understand what `0..(size - 1)` is indicating here. Is it supposed to be an enumeration of all the values? Does that mean it's only defined over integers?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * {{{ZipOracle}}} is an optimization of {{{Oracle}}}. It groups several values of the background oracle together, in
+ * order to reduce the number of constraints. It is up to the user of {{{ZipOracle}}} to make sure that the grouped
+ * values may be treated as equivalent.
+ *
+ * @param backOracle
+ *   the background oracle whose values are grouped together
+ * @param groups
+ *   a list of groups over the values of the background oracle
```
shonfeder, line 17: It's not clear to me in what sense this is zipping anything. Is it just a general grouping? Or is it establishing pairwise groupings? Having something to clarify what is zippy in the comment would help with readability for me.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+class ZipOracle(backOracle: Oracle, groups: List[Set[Int]]) extends Oracle {
+  override def size: Int = groups.size
+
+  override def whenEqualTo(state: SymbState, position: Int): TlaEx = {
+    assert(position < groups.size)
+    val conds = groups(position).map(index => tla.fromTlaEx(backOracle.whenEqualTo(state, index))).toList
+    tla.or(conds: _*).as(BoolT1())
+  }
+
```
shonfeder, line 27: The name of this method doesn't make sense to me, given what I can understand of the code. maybe I just lack needed context?

shonfeder (commented):
a couple questions, but probably just due to me lacking context...

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
```diff
@@ -0,0 +1,32 @@
...
+ * {{{ZipOracle}}} is an optimization of {{{Oracle}}}. It groups several values of the background oracle together, in
+ * order to reduce the number of constraints. It is up to the user of {{{ZipOracle}}} to make sure that the grouped
+ * values may be treated as equivalent.
+ *
+ * @param backOracle
+ *   the background oracle whose values are grouped together
+ * @param groups
+ *   a list of groups over the values of the background oracle
+ */
+class ZipOracle(backOracle: Oracle, groups: List[Set[Int]]) extends Oracle {
```
rodrigo7491, line 19: Shouldn't `ZipOracle` override `caseAssertions` is a similar fashion to `SparseOracle`?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
```diff
@@ -0,0 +1,32 @@
+package at.forsyte.apalache.tla.bmcmt.rules.aux
+
+import at.forsyte.apalache.tla.bmcmt.SymbState
+import at.forsyte.apalache.tla.bmcmt.smt.SolverContext
+import at.forsyte.apalache.tla.lir.TypedPredefs.BuilderExAsTyped
+import at.forsyte.apalache.tla.lir.convenience.tla
+import at.forsyte.apalache.tla.lir.{BoolT1, TlaEx}
+
+/**
+ * {{{ZipOracle}}} is an optimization of {{{Oracle}}}. It groups several values of the background oracle together, in
```
rodrigo7491, line 10: What is the reason for using three curly brackets?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
rodrigo7491, line 17: "groups of indices"? I could only understand this parameter by looking at the usage of `ZipOracle` in `CherryPick`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
```diff
@@ -0,0 +1,32 @@
...
+ *   a list of groups over the values of the background oracle
+ */
+class ZipOracle(backOracle: Oracle, groups: List[Set[Int]]) extends Oracle {
+  override def size: Int = groups.size
+
+  override def whenEqualTo(state: SymbState, position: Int): TlaEx = {
+    assert(position < groups.size)
+    val conds = groups(position).map(index => tla.fromTlaEx(backOracle.whenEqualTo(state, index))).toList
+    tla.or(conds: _*).as(BoolT1())
+  }
```
rodrigo7491, line 26: This a cool idea. It was not immediately obvious to me how this would reduce the burden on the solver, since we still have to call `whenEqualTo` for every index, but I see now that this leads to less ITEs. Maybe mention this in a comment?

test/tla/PickPerf2.tla
```diff
@@ -0,0 +1,40 @@
...
+
+Next ==
+  /\\/ 1 \in S /\ S' = { 2 } \union S /\ UNCHANGED T
+    \/ 2 \in S /\ S' = { 3 } \union S /\ UNCHANGED T
+    \/ 3 \in S /\ S' = { 1 } \union S /\ UNCHANGED T
+    \/ 4 \in T /\ T' = { 5 } \union T /\ UNCHANGED S
+    \/ 5 \in T /\ T' = { 6 } \union T /\ UNCHANGED S
+    \/ 6 \in T /\ T' = { 1 } \union T /\ UNCHANGED S
+  /\ R' = S' \union T'
+  /\ state' \in 0..4
```
rodrigo7491, line 28: Would it make any difference to have it as `state \in 0..3`?

rodrigo7491 (commented):
Overall looks good to me. Besides the minor comments below, I would like to understand two points:

1) I found the move of comments to be exclusively on the parent class unexpected. Is this something we should aim to have in the whole codebase?
2) About `PickPerf` and `PickPerf2`, should we add them to the integration tests? I found it odd to have them in `test/tla`, but not use them for tests. If we are only interested in performance checks maybe they belong somewhere else.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/Oracle.scala
Kukovec: Oracles always return an integer value (in a given range). The interpretation of the value depends on the kind of oracle, but typically, `n` means e.g. choose the `n`-th `has` edge in the arena.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/Oracle.scala
```diff
@@ -35,7 +46,25 @@ trait Oracle extends Serializable {
...
+  def caseAssertions(state: SymbState, assertions: Seq[TlaEx], elseAssertions: Seq[TlaEx] = Seq()): TlaEx = {
+    size match {
+      case 0 => state.arena.cellTrue().toNameEx
+
+      case 1 => assertions.head
+
+      case _ =>
+        // iteCases is a sequence of tuples, with the fst and snd elements of each tuple being the "if" and "else" cases of an ite.
+        // If elseAssertions is not empty, each tuple has its fst element from assertions and its snd form elseAssertions.
+        // If elseAssertions is empty, each tuple has its fst element from assertions and its snd defaults to "ValEx(TlaBool(true))".
```
Kukovec, line 58:
```suggestion
        // iteCases is a sequence of tuples, with the 1st and 2nd elements of each tuple being the "if" and "else" cases of an ite.
        // If elseAssertions is not empty, each tuple has its 1st element from assertions and its 2nd form elseAssertions.
        // If elseAssertions is empty, each tuple has its 1st element from assertions and its 2nd defaults to "ValEx(TlaBool(true))".
```

Kukovec (approved):

test/tla/PickPerf2.tla
konnov: Yes, one of the cases below has `state' = 4`. If we write `state' \in 0..3`, this case will be excluded

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/Oracle.scala
konnov: Yes, oracles are always enumerating over integers. As Jure wrote, these are basically the indices of the edges. How they are mapped to SMT, depends on the implementation.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov, line 10:
```suggestion
 * [[ZipOracle]] is an optimization of [[Oracle]]. It groups several values of the background oracle together, in
```

konnov: The curly brackets stand for code formatting: https://docs.scala-lang.org/style/scaladoc.html

test/tla/PickPerf2.tla
rodrigo7491: I meant `state` no primed, and looking at it again `state \in 0..2` would yield the same I think.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov: But you are right. I have changed the reference to double square brackets

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov: It zips constraints. Essentially, instead of introduces instances of `k` records, it introduces one instance that can apply to `k` values.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov: Not only it decreases the number of conditions over the oracle values, it also introduces significantly fewer cells and constraints. So it has further implications than just a few ITEs.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov: It's the condition saying that the oracle value is equal to a constant.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov: `ZipOracle` is using `whenEqualTo`, because the values should go over the groups, not the original values.

konnov:
> I found the move of comments to be exclusively on the parent class unexpected. Is this something we should aim to have in the whole codebase?

Yes, we had a discussion about this. I think it makes sense, as the comments of the overriding methods may get out of sync quick.

test/tla/PickPerf2.tla
rodrigo7491: I see your point. Using `state \in 0..2` might work in a similar way but it does have the effect you intend.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
rodrigo7491: Sorry to open this again, but is the assumption that `ZipOracle` will always be used properly? As is, `ZipOracle` will default to `Oracle`'s behavior for `caseAssertions`, which should not be the case from what I understand. Shouldn't `backOracle` be called?

rodrigo7491: Right now it makes no difference, as we have the same body for all non-mock oracles, which you nicely refactored, but it may not always be the case. I'm trying to anticipate problems, but maybe I'm just overthinking it.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/Oracle.scala
shonfeder: I see. Thank you for clafifying!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
shonfeder: I see. Thank you! I do think adding that sentence could help with documentation here, but not essential.

shonfeder (approved):

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov: I think problems may arise, if we mix oracles. If we always use `ZipOracle`, as `CherryPick` does, this should be fine.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/ZipOracle.scala
konnov: I have added the following two sentences:

> In this sense, it compresses several oracle values into one. As a consequence, [[CherryPick]] pick constructs significantly fewer constants and constraints.

PR #1729, opened 2022-05-10
New LevelFinder implementation

p-offtermatt:

~~- [] Documentation added for any new functionality~~

~~- [] Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality~~

Was previously on another branch, but `git rebase` was convoluted due to being several dozen commits behind, so I created a separate branch.


tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevelFinder.scala
```diff
@@ -0,0 +1,108 @@
...
+import _root_.at.forsyte.apalache.tla.lir.oper.TlaActionOper
+import _root_.at.forsyte.apalache.tla.lir.oper.TlaTempOper
+
+class TlaLevelFinder(module: TlaModule) {
+  private val consts = Set(module.constDeclarations.map(_.name): _*)
+  private val vars = Set(module.varDeclarations.map(_.name): _*)
+  private val defs = Map(module.operDeclarations.map { d => d.name -> d }: _*)
+  private var levelCacheMap: Map[String, TlaLevel] = Map()
+  var levelCacheGetFun: String => Option[TlaLevel] = (name: String) => levelCacheMap.get(name)
+  var levelCacheSetFun = (name: String, level: TlaLevel) => levelCacheMap += name -> level
```
konnov, line 12: Why not call these functions simply `get` and `set`

konnov (approved):
In principle, looks good. What was the main motivation for merging the two classes into one?

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaLevelFinder.scala
p-offtermatt: With the old structure, it was very hard to evaluate the level of an expression - the expression level finder depended on the map of names, which was populated in the declaration level finder. It never mattered, since the only call to the level finding was for declarations, but for the temporal encoding I need to find the level of individual expressions

p-offtermatt: > Why not call these functions simply `get` and `set`

I think it's important that it doesn't `get` or `set` the level of an expression, but rather the cache, that's the reason for `levelCacheGet` and `levelCacheSet`. And since it is a function object that can be manipulated, I added Fun at the end - it seems clearer to do `levelCacheGetFun = _ => TlaLevelState` rather than `levelCacheGet = _ => TlaLevelState`, but it's up to preference

PR #1730, opened 2022-05-10
Tune `search.invariant.mode` depending on length

konnov:
As we have found in #1725, the mode specified with `search.invariant.mode` dramatically changes the running times, depending on whether we are checking an inductive invariant or running bounded model checking.

Since the names `before` and `after` are not very helpful, we rename them to `inLoop` and `atOnce`. The old names `before` and `after` are kept for backward compatibility (e.g., with scripting), though they are removed from the documentation.

Currently blocked by failing integration tests; tried to understand the failure by writing a TLA+ spec in #1756.


docs/src/apalache/tuning.md
```diff
@@ -20,53 +20,44 @@ The following options are supported:
...
-   transition numbers. For instance,
+2. __Timeouts__: `search.smt.timeout=<seconds>` defines the timeout to the SMT solver in seconds. The default value
+   is `0`, which stands for the unbounded timeout. For instance, the timeout is used in the following cases: checking if
+   a transition is enabled, checking an invariant, checking for deadlocks. If the solver times out, it reports '
+   UNKNOWN', and the model checker reports a runtime error.
+
+3. __Invariant mode__: `search.invariant.mode=(inLoop|atOnce)` defines the moment when the invariant is checked. In
+   the `atOnce` mode, all transitions are first translated, one of them is picked non-deterministically and then the
+   invariant is checked. Hence, they are checked "all at once". Although this mode reduces the number of SMT queries, it
+   usually requires more memory than the `inLoop` mode. In the `inLoop` mode, the invariant is checked for every enabled
```
thpani, line 31: I think this needs some reference to the transition explorer to be understandable.
E.g., "is picked non-deterministically" is only meaningful in context.

docs/src/apalache/tuning.md
```diff
@@ -20,53 +20,44 @@ The following options are supported:
...
+   transition independently. The
+   `inLoop` mode may drastically reduce memory consumption, but it may take longer than the `atOnce` mode, provided that
+   Apalache has enough memory. When the specification has `t` symbolic transitions and `--length=k`, Apalache performs
+   up to `t * k` SMT queries in the `inLoop` mode and up to `t` SMT queries in the `atOnce` mode. As a result, the
+   default mode depends on the value of `k`:
+   when `k > 1`, the default mode is `atOnce`; when `k <= 1`, the default mode is
+   `inLoop`.
```
thpani, line 38:
```suggestion
   * If `k > 1`, the default mode is `atOnce`.
   * If `k <= 1`, the default mode is `inLoop`.
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala
```diff
@@ -225,7 +225,7 @@ class SeqModelChecker[ExecutorContextT](
           trex.recover(snapshot.get)
         } else {
           // when --all-enabled is on, the transition has not been assumed
-          if (params.invariantMode == InvariantMode.BeforeJoin) {
+          if (params.invariantMode == CheckInvariantInLoop()) {
             if (isNext) {
               checkInvariantsForPreparedTransition(isNext, no, transitionInvs, transitionActionInvs)
             }
```
thpani, line 231: There's still a reference to `search.invariant.mode=before` below (ln 263)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/search/ModelCheckerParams.scala
```diff
@@ -43,8 +49,16 @@ class ModelCheckerParams(
...
-    if ("after" == tuningOptions.getOrElse("search.invariant.mode", "before")) AfterJoin else BeforeJoin
+  var invariantMode: ModelCheckerParams.InvariantMode = {
+    // we have experimentally found that the mode "atOnce" is more efficient for bounded model checking,
+    // whereas the mode "inLoop" is more efficient for invariant checking
+    val defaultMode = if (stepsBound <= 1) "inloop" else "atonce"
+    if (Set("after", "atonce").contains(tuningOptions.getOrElse("search.invariant.mode", defaultMode).toLowerCase)) {
+      ModelCheckerParams.CheckInvariantAtOnce()
+    } else {
+      ModelCheckerParams.CheckInvariantInLoop()
+    }
```
thpani, line 60: Since the argument here is a string, should we also test in the else branch and throw on an invalid value?

konnov: I am divided here. On one hand, you are right, we should indicate an error. On the other hand, I would prefer no to throw in case of an optimization option. Maybe log an warning?

thpani: > On the other hand, I would prefer no to throw in case of an optimization option

Why not? Obviously the tool then does something other than the user intended.
We do know that users often miss warnings in the log ;)

Kukovec (approved):

shonfeder:
What should we do in this pr? Is it still relevant and needing merge conflict resolution or should we close it?

PR #1733, opened 2022-05-10
Add ADR 019 proposing improvement of changelog

### LINKED ISSUE #1538: Adopt changelog practice that will avoid merge conflicts in release notes
Making PRs that require additions to the release notes tend to involve resolving merge conflicts.

We can fix this by representing `UNRELEASED.md` as a directory structure instead. We can also consider the `unclog` tool used by some other teams internally, tho I'm not entirely convinced that this problem causes for the overhead of a CLI tool in our dev practice and CI pipeline. I'll get the read of the team in a meeting this week.

- [ ] write adr proposing approach
- [ ] write script to generate changelog from entries
- [ ] wire script into CI

shonfeder:
Closes #1538

This ADR evaluates all the options we've discussed to remove friction
from merge conflicts in the changelog. It also sets out the solution we
see to have agreed upon in our meetings. But not this is only a proposal
at this stage, and the decision can be altered if we come up with a
better approach.

Please leave comments or suggest change if you see ways to improve upon
the plan or reasons to select a different choice.

### DIFF
--- a/docs/src/adr/0190-harmonize-changelog.md
+++ b/docs/src/adr/0190-harmonize-changelog.md
@@ -1,0 +1,203 @@
+---
+authors: Shon Feder
+last revised: 2022-05-06
+---
+
+# ADR-019: Harmonize changelog management
+
+**Table of Contents**
+
+- [Summary (Overview)](#summary)
+- [Context (Problem)](#context)
+- [Options (Alternatives)](#options)
+- [Solution (Decision)](#solution)
+- [Consequences (Retrospection)](#consequences)
+
+## Summary
+
+In the context of maintaining a changelog to communicate salient changes about
+our software, facing the friction caused by frequent merge conflicts in our
+`UNRELEASED` changelog, we decided to find a lightweight, conflict free,
+changelog process in order to achieve a smoother development process. We have
+accepted the development cost and learning curve required by adopting a new
+process.
+
+## Context
+
+We maintain a [changelog](https://keepachangelog.com/en/1.0.0/)
+
+> to make it easier for users and contributors to see precisely what notable
+> changes have been made between each release (or version) of the project.
+
+Any changeset that introduces observable changes to the behavior of our software
+should include additions to the changelog.  Currently, [additions are made](https://github.com/informalsystems/apalache/blob/60e6a14d451cb67c93dfd29cffd0cc0eb9d7922d/CONTRIBUTING.md#changelog) by
+updating the `UNRELEASED.md` file. The changes recorded there are then added to
+the `CHANGES.md` file during our automated release process.
+
+This process creates a race condition over `UNRELEASED.md` between any
+concurrent pull requests. This consequently results in developers constantly
+having to resolve merge conflicts. This busy-work slows down development and
+adds no value.
+
+## Options
+
+We now enumerate and consider the various options we've conceived for addressing
+this problem.
+
+With the exception of option (0), all of the following options would resolve the
+problem of merge conflicts and could be integrated into our existing automated
+deployment pipeline.  However each option has different costs w/r/t
+dependencies, processes, and learning curve.
+
+All of the following options will likely require ongoing maintenance, even
+option (0): this is because we already have a set of scripts that manage our
+changelog entries in order to support our fully automated releases. So this is
+not a space where we are introducing automation for the first time, but is
+instead a situation where we are changing an existing automated system.
+
+### 0. Don't change anything
+
+We could do nothing. Developers would continue to resolve changelog conflicts manually.
+
+#### Pros
+
+- Requires no development time on a solution.
+- Would require no additional maintenance.
+
+#### Cons
+
+- We'll continue having developer slow-down and busy work when merge conflicts arise.
+
+### 1. Use one of the automated changelog generators
+
+There is a [superabundance](https://github.com/search?q=changelog+auto) of tools
+for automatically generating changelogs. These tools extract changelogs from
+various different sources such as commit messages, issues, or pull requests.
+    
+#### Pros
+
+- Requires no new development
+- Would eliminate the need to add changelog entries in a separate file
+
+#### Cons
+
+- Developers have to learn the annotation format required by the tool, and they
+  have to remember to deploy it in their commits/PRs/issues.
+- Infrastructure maintainers have to learn how to configure and maintain the tool.
+- We pickup a dependency external to both the company and our project
+- Changelogs are meant to communicate a very specific kind of information to a
+  specific audience. Git commit messages, pull requests, and issues are for
+  communicating quite different information to an entirely different audience.
+  Requiring that these communication channels be overloaded risks degrading the
+  quality of communications in all the conflated channels.
+  - However, it's possible we could work out some conventions to add empty
+    commits or designate special PR labels to work around this interference?
+
+
+### 2. Use Unclog
+
+[@thanethomson](https://github.com/thanethomson) has written the [unclog][] utility, which is used in both
+[informalsystems/tendermint-rs][] and [informalsystems/ibc-rs][]. The tool
+stores changelog entries in a directory structure, with each entry in its own
+file, ensuring merge conflicts can't arise. 
+
+Entries are added via the CLI. E.g.:
+
+```sh
+unclog add --id some-new-feature \
+  --issue 23 \
+  --section breaking-changes \
+  --message "Some *new* feature"
+```
+
+If the `--messages` argument is omitted, it will open your configured editor for
+authoring the message.
+
+The tool has subcommands to generate and update the `CHANGES.md`, and it supports
+a variety of options via its TOML configuration.
+
+[unclog]: https://github.com/informalsystems/unclog
+[informalsystems/tendermint-rs]: https://github.com/informalsystems/tendermint-rs
+[informalsystems/ibc-rs]: https://github.com/informalsystems/ibc-rs
+
+#### Pros
+
+- Requires no new development
+- It is developed internally to informal
+
+#### Cons
+
+- Adds cognitive and procedural overhead to adding changelog entries
+- Would add a rust dependency to our devenv
+- Requires external contributors to either install a CLI tool or figure out a
+  relatively complex file and folder structure to add changelog entries
+
+### 3. Use a custom git merge driver
+
+If we wrote a [custom git merge driver](https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver) to work on the simple changelog format we
+could continue our current practice, and just fall back to the custom merge
+driver in case of merge conflicts.
+
+#### Pros
+
+- Allows us to keep our current process of simply updating a markdown file,
+  with which, moreover, most devs are already familiar
+
+#### Cons
+
+- Could require external dependencies to be installed, depending on
+  implementation.
+- It's not possible to configure custom merge drivers for github, so we'd either
+  need to develop a github action or bot to monitor for merge conflicts in the
+  changelog and merge them automatically, or devs would be back to having to
+  resolve merge conflicts locally.
+
+### 4. Write some lightweight tooling
+
+We could author some simple tooling that would capture the core behavior of
+`unclog`, using files in directories, but neither require external dependencies
+nor knowledge of a new CLI.
+
+#### Pros
+
+- No new tooling to learn
+- No external dependencies to integrate
+
+#### Cons
+
+- Would require some development time (I estimate approx. 1 day, see the design
+  sketch below for details on the implementation)
+- Would require contributors to understand the folder structure
+
+## Solution
+
+We have opted for option 4. Despite the added cost of development, we think the
+simplicity of use and maintenance will offset the development cost.
+
+### Design
+
+We will introduce a new directory tree to the project:
+
+```sh
+.unreleased/
+├── breaking-changes/
+├── bug-fixes/
+├── documentation/
+└── features/
+```
+
+Each change will be added as a single markdown file in the appropriate
+directory.
+
+Two `sbt` targets will be added to facilitate our continuous deployment:
+
+- `releaseNotes`, merges the unreleased notes into a single new file of the expected format in our `target` directory. This can be used for upload to github releases.
+- `changeLog` updates `CHANGES.md` by prepending the output of `releaseNotes`. It then removes all notes
+  from the `.unreleased` directory so that the next iteration of development starts from a clean slate.
+
+## Consequences
+
+TBD
+<!-- Records the results of the decision over the long term.
+     Did it work, not work, was changed, upgraded, etc.
+-->


docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+├── documentation/
+└── features/
+```
+
+Each change will be added as a single markdown file in the appropriate
+directory.
+
+Two `sbt` targets will be added to facilitate our continuous deployment:
+
+- `releaseNotes`, will compile to the unreleased notes into the expected format
```
shonfeder, line 172:
```suggestion
- `releaseNotes`, will compile the unreleased notes into the expected format in a new file our `target` directory. This can be used for upload to github releases
```

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+- `changeLog` will depend on the value of `releaseNotes` and will update
+  `CHANGES.md` by prepending the new notes, and then clear out all the notes
+  from the `.unreleased` directory.
```
shonfeder, line 175:
```suggestion
  from the `.unreleased` directory so that they clear for the next iteration of development.
```

konnov (approved):
Nice! I really like the directory structure by unclog. I understand the worry of pulling the whole cargo in the java/sbt build process. Would it be different if `unclog` has proper packages? I guess, the main problem is that it is not java, right?

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+#### Cons
+
+- Adds cognitive and procedural overhead to adding changelog entries
+- Would add a rust dependency to our devenv
+- Requires external contributors to either install a CLI tool or figure out a
+  relatively complex file and folder structure to add changelog entries
+
+### 3. Use a custom git merge driver
+
+If we wrote a custom merge driver to work on the simple changelog format we
```
thpani, line 116:
```suggestion
If we wrote a [custom git merge driver](https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver) to work on the simple changelog format we
```

konnov:
I had a quick look at unclog. Personally, I would love to use it. Given that it is a public project and it is developed at Informal, I see a really good reason for using it. However, if we have to pull a completely different dev environment to do that, this does not sound right.

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+- It's not possible to configure custom merge drivers for github, so we'd either
+  need to develop a github action or bot to monitor for merge conflicts in the
+  changelog and merge them automatically, or devs would be back to having to
+  resolve merge conflicts locally.
+
+### 4. Write some lightweight tooling
+
+We could author some simple tooling that would capture the core behavior of
+`unclog`, using files in directories, but neither require external dependencies
+nor knowledge of a new CLI.
```
thpani, line 138: > but neither require external dependencies

That depends very much on the implementation?

> nor knowledge of a new CLI

I don't get this part. If we introduce new tooling, we would need to learn a new CLI? Or did you mean relative to `unclog`?

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### Pros
+
+- No new tooling to learn
+- No external dependencies to integrate
```
thpani, line 143: See above.

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### Cons
+
+- Would require some development time (=< 1 day)
```
thpani, line 147: I would like to understand more about the proposed implementation, but this estimate seems overly optimistic to me.
Shouldn't we also account for time needed to learn about the new tooling?

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+- Would require contributors to understand the folder structure
```
thpani, line 148:
```suggestion
#### Cons

- Requires some development time (=< 1 day)
- Requires contributors to understand the folder structure
- Adds an additional project the Apalache team needs to maintain.
```

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+```sh
+.unreleased/
+├── breaking-changes/
+├── bug-fixes/
+├── documentation/
+└── features/
+```
+
+Each change will be added as a single markdown file in the appropriate
+directory.
```
thpani, line 168: What is a "change" in this context? A markdown bullet list?

thpani: It is sometimes beneficial to keep some logical order among changelog entries, it would be good to support this in the future.

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+
+Two `sbt` targets will be added to facilitate our continuous deployment:
+
+- `releaseNotes`, will compile the unreleased notes into the expected format in a new file our `target` directory. This can be used for upload to github releases
```
thpani, line 172:
```suggestion
- `releaseNotes`, merges the unreleased notes into a single new file of the expected format in our `target` directory. This can be used for upload to github releases.
```

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+- `changeLog` will depend on the value of `releaseNotes` and will update
+  `CHANGES.md` by prepending the new notes, and then clear out all the notes
+  from the `.unreleased` directory so that they clear for the next iteration of development.
```
thpani, line 175: > will depend on the value of `releaseNotes`

I don't understand this part.

thpani, line 175:
```suggestion
- `changeLog` updates `CHANGES.md` by prepending the output of `releaseNotes`. It then removes all notes
  from the `.unreleased` directory so that the next iteration of development starts from a clean slate.
```

docs/src/adr/019adr-changelog-management.md
p-offtermatt, line 147: We should not forget that even if this is simple, it will also at some point likely require maintenance work by us, so it's not just the up-front cost we'd pay

thpani: Now that I've read the whole document, this should probably refer down to `### Design`.

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+#### Pros
+
+- Requires no new development
+- It is developed internally to informal
+
+#### Cons
+
+- Adds cognitive and procedural overhead to adding changelog entries
+- Would add a rust dependency to our devenv
+- Requires external contributors to either install a CLI tool or figure out a
```
p-offtermatt, line 111: I don't know much about the rust stack; is there a way to keep the installation very simple, e.g. precompiled binaries for common platforms?

p-offtermatt (approved):

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+#### Pros
+
+- Allows us to keep our current process of simply updating a markdown file,
+  with which, moreover, most devs are already familiar
+
+#### Cons
+
+- Could require external dependencies to be installed, depending on
+  implementation.
+- It's not possible to configure custom merge drivers for github, so we'd either
```
thpani, line 129: Did you test this? libgit2 which seems to drive the merge on github [seems to support this](https://github.com/libgit2/libgit2/pull/3564#issuecomment-217581428).

thpani (commented):
Thanks Shon, this is a really helpful writeup of our options!

I'm curious if you tested whether Github prohibits custom merge drivers? libgit2, which drives the merge on github, [has support for it](https://github.com/libgit2/libgit2/pull/3564#issuecomment-217581428). I only found some stackoverflow posts saying Github didn't support this 2 years ago.

If option 3 is not possible, I agree with the decision for option 4.

I also left some stylistic suggestions.

docs/src/adr/019adr-changelog-management.md
thpani: Now that I've read the whole document, this should probably refer down to `### Design`.

docs/src/adr/019adr-changelog-management.md
shonfeder: Isn't this true of all the options? That is, won't all require some maintenance one way or another, whether it be updated configs, or getting bugs fixed, etc.?

docs/src/adr/019adr-changelog-management.md
shonfeder: That would keep the installation simple for those platforms, but this adds even more overhead for the setup, since no such precompiled are currently available. afaiu, rust libs dynamically link against glibc, so it doesn't provide the same sort of highly portable binaries we'd get from Go, or a JAR file (modulo JVM req).

We could package it for each platform, but that's an even heavier lift.

For the sake of reducing setup speed in our CI, we could precompile a binary for the particular architecture we use (we do this for emojitsu). But this won't help developers or external contributors much.

Still, installing from cargo is not difficult, it's just one extra step.

Moreover, as @konnov has pointed out, it should be possible to just maintain the entries by hand in the directory structure, as we'd do any how.

The format is a bit finicky (for some reason requiring an issue number to start every file name) and they have to be formatted as a single bullet point, but it's not too terribly confusing if we document it clearly.

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+changelog process in order to achieve a smoother development process. We have
+accepted the development cost and learning curve required by adopting a new
+process.
+
+## Context
+
+We maintain a [changelog](https://keepachangelog.com/en/1.0.0/)
+
+> To make it easier for users and contributors to see precisely what notable
+> changes have been made between each release (or version) of the project.
```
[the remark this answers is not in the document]
shonfeder: I read this as a single sentence and don't think the colon reads correctly. But maybe lowercasing the leading `T` in `to` of the blockquote would make that clearer?

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+
+Any changeset that introduces observable changes to the behavior of our software
+should include additions to the changelog.  Currently, additions are made by
+updating the `UNRELEASED.md` file. The changes recorded there are then added to
+the `CHANGES.md` file during our automated release process.
```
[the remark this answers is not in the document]
shonfeder: Nothing required as part of this that isn't already part of our release pipeline.

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+  communicating quite different information to an entirely different audience.
+  Requiring that these communication channels be overloaded risks degrading the
+  quality of communications in all the conflated channels.
+  - However, it's possible we could work out some conventions to add empty
+    commits or designate special PR labels to work around this interference?
+
+
+### 2. Use Unclog
+
+@thanethompson has written the [unclog][] utility, which is used in both
```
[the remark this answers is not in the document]
shonfeder: Thanks for the fix!

docs/src/adr/019adr-changelog-management.md
shonfeder: I was thinking an entry describing a change would just be a paragraph or two of markdown. Different entries should be different files. No point in having bullet points inside the files afaict, that seems confusing to me.

Good point about order. I think we can just assemble the log entries in lexigraphic order of the file names. So that can be used to control the ordering within a section.

docs/src/adr/019adr-changelog-management.md
shonfeder: This `sbt` terminology. Each `sbt` target (in thise, a `taskKey`) has a `.value` attribute that represents whatever is accomplished by the task. In this case, it will be the path to file with the assembled notes for the release. I can make that clearer.

shonfeder: This works!

shonfeder: Resolved from your fix. Thanks.

docs/src/adr/019adr-changelog-management.md
shonfeder: Right, there's no problem to install custom merge drivers on github runners in CI. But github doesn't allow configuration of custom merge drivers on the auto merge functionality it uses to update PRs. And without that, we'll have the added overhead of writing a bot/action that monitors our pending PRs and then updates them with the custom merge strategy installed.

Otherwise we're just back to devs having to fix these locally. While it would save a bit of time having a custom merge driver for these files rather than having to do it manually, it doesn't really clear up the flow the way we want.

That said, I really do like this solution, and I prototyped a very rough merge driver for this last weekend for fun. I think I may pursue that as a side project regardless. Personally, I still am inclined to think that keeping changes in a single markdown file is the nicest way to record them :D

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### Pros
+
+- Requires no new development
+- It is developed internally to informal
+
+#### Cons
+
+- Adds cognitive and procedural overhead to adding changelog entries
+- Would add a rust dependency to our devenv
```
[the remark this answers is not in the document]
shonfeder: Yeah, the current packaging is currently from source and installed with cargo. We could of course package it up, probably make a nix package for it? Tho I can't even estimate how long that would take.

Honestly, requiring a dependency on cargo is not the most arduous thing in the world, it just seems kind of heavyhanded to me given how simple the problem is here. Like we're 70% there with `find .changelog/unreleased -iname '*.md' -exec cat {} \;`

shonfeder:
I've tried adding a changelog update manually and building the changelog file with `unclog`. It mostly worked.

I did notice at least two constraint imposed by `unclog` that I don't prefere:

- It requires the names of the files holding change entries to start with numbers (meant to reference an issue/pr), and I'm not crazy about this overhead (since the pr/issue has to be mentioned in the change entry anyhow).
- It requires every release to have a summary. This seems to introduce an unnecessary manual step imo. It may be a sensible constraint for projects with a much slower release cycle, but doesn't make sense with our weekly release cadence.

I have not been able to get the `release` functionality of unclog to work however (see https://github.com/informalsystems/unclog/issues/31). This is where all of the non-trivial logic lies, but the problem is just a file lookup issue, so I'm guessing it should be easy to fix.

I expect all three of these minor friction points could be fixed with small patches to `unclog`. So it comes down a question of where we want to put our energy and what kinds of overhead we feel comfortable taking in terms of dependencies and process.

Tho I'm still partial to the the simpler solution proposed, I'm on a fence here. Pretty OK with going whatever direction you all prefer.

One thing I can note: I would guess that the overhead of adding cargo into our CI (or building custom binaries/packages) and getting the tool installed on all of the devs computers and getting familiar with the CLI will take around the same time as scripting out a simple solution for our use case.

On the otherhand, we might want something similar for tnt or other projects, and then having unclog already integrated into our workflows could be a benefit.

WDYT?

shonfeder:
On further thought: if everyone else is ok with using unclog and we can get the `release` bug fixed, that might be the most expedient solution, and it’s probably also a good idea to support using the tooling across teams, so we can benefit from sharing our needs and pooling our solutions.

docs/src/adr/019adr-changelog-management.md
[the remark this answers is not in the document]
thpani: Aaah, that makes sense. Yes then please
```suggestion
We maintain a [changelog](https://keepachangelog.com/en/1.0.0/)

> to make it easier for users and contributors to see precisely what notable
> changes have been made between each release (or version) of the project.
```

docs/src/adr/019adr-changelog-management.md
thpani: > I was thinking an entry describing a change would just be a paragraph or two of markdown. [...] No point in having bullet points inside the files afaict, that seems confusing to me.

I'm just wondering how this will render in the combined changelog file, but we can leave this for implementation.

docs/src/adr/019adr-changelog-management.md
thpani: > Without that functionality, we'll have the added overhead of writing a bot/action that monitors our pending PRs and then updates them with the custom merge strategy installed.

Yeah no, that seems too complicated...

> Personally, I still am inclined to think that keeping changes in a single markdown file is the nicest way to record them

I agree!

thpani:
> It requires the names of the files holding change entries to start with numbers (meant to reference an issue/pr), and I'm not crazy about this overhead (since the pr/issue has to be mentioned in the change entry anyhow).

So `unclog`'s solution to avoiding merge conflicts is to simply prefix each entry file with an issue id? It seems like we can do this ourselves...

One thing I really don't want is ending up in a state where one needs a local Java+Scala+TypeScript+Rust dev env to work on Apalache (+OCaml if you want to run integration tests).

docs/src/adr/019adr-changelog-management.md
thpani: > won't all require some maintenance one way or another

I guess the current solution does not 😅
Maybe we should list it as an option 😄

p-offtermatt: > Isn't this true of all the options? That is, won't all require some maintenance one way or another, whether it be updated configs, or getting bugs fixed, etc.?

True! I'd however expect the burden to be less in the case of preexisting solutions (e.g. for unclog, other teams also use it, so a similar amount of maintenance serves more users... though communication may then become a larger overhead).

docs/src/adr/019adr-changelog-management.md
p-offtermatt: Okay, precompiling probably seems not very beneficial, thanks for the insight.

shonfeder: It is a good thing to consider if we find a CI slowdown as a result of this dependency! It would also help users, of course. It's just that maintaining packages for bunch of different systems is quite a bit of work (afaik) and since it's not already already done, that work would probably fall on us.

We **could** probably drop this dependency into our nix flake configuration. That might be the best option here from a devenv perspective. I'll include a note about that.

p-offtermatt:
> One thing I really don't want is ending up in a state where one needs a local Java+Scala+TypeScript+Rust dev env to work on Apalache (+OCaml if you want to run integration tests).

Agreed. If unclog is very isolated in the process and one could also work without it by doing some manual fiddling with changelog files, then this shoud however not be an issue in my opinion. I'm still leaning towards using more unified tooling as a company, i.e. using unclog, rather than writing something simple just for us.

docs/src/adr/019adr-changelog-management.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+Each change will be added as a single markdown file in the appropriate
+directory.
+
+Two `sbt` targets will be added to facilitate our continuous deployment:
+
+- `releaseNotes`, merges the unreleased notes into a single new file of the expected format in our `target` directory. This can be used for upload to github releases.
+- `changeLog` updates `CHANGES.md` by prepending the output of `releaseNotes`. It then removes all notes
+  from the `.unreleased` directory so that the next iteration of development starts from a clean slate.
```
bugarela, line 174: If we are still maintaining a version controlled `CHANGES.md`, wouldn't it still be prone to conflicts on concurrent PRs? If I understand correctly, this will just make it easier to fix the conflict (in an automated way), but then I believe the git merge driver would be a better option.

shonfeder:
> So unclog's solution to avoiding merge conflicts is to simply prefix each entry file with an issue id? It seems like we can do this ourselves...

The real  trick is just that each changelog entry (or perhaps a small group of entries) goes in its own file. The issue id specifically might help keep the file names unique, tho I expect there's a vanishing small chance of conflict even without the issue numbers: just having short descriptive file names would seem to do the trick I'd think? Or each file name could be preceded with a contributors user name, or the branch that introduced the change. The file names just get thrown away in any case. We can definitely do this bit ourselves.

> One thing I really don't want is ending up in a state where one needs a local Java+Scala+TypeScript+Rust dev env to work on Apalache (+OCaml if you want to run integration tests).

I'm sympathetic to this (while also being guilty of introducing the dependency on mdx XD ). But I also think the current taboo against pulling different language dependency is an unfortunate situation and something we should aim to overcome, and which nix does help us overcome to a pretty large extent... I guess I'm just very very on the fence here, lol.

bugarela (approved):

thpani (approved):

docs/src/adr/019adr-changelog-management.md
shonfeder: The current design just extends our exiting CI automation, so I'd argue it's really not adding another project requiring specific maintenance.

docs/src/adr/019adr-changelog-management.md
shonfeder: I think there's not anythinb to learn here? Dev's need to know where the directories would be located, by they won't need to run or use any tools.

shonfeder:
After our discussion today, it seems everyone is fairly ambivalent. I think the preference for a lightweight solution specific to this repo was more held by a slight majority of those who voiced a preference, but we agreed I could just make a final call here.

Given the fact that the `release` functionality is not working, and the shared hesitation about bringing in cargo as dev dependency, I'm gonna call this in favor of option 4.

Thanks all for the careful review and insightful considerations!

PR #1734, opened 2022-05-11
Add .metals and .vscode to gitignore

p-offtermatt:
Folders that should not be pushed to the repo, containing local data for Metals and VSCode


konnov (approved):
LGTM

PR #1744, opened 2022-05-12
Add separate CI workflow to check for compiler warnings

### LINKED ISSUE #1741: Run CI tests without `APALACHE_FATAL_WARNINGS`
We're currently running unit and integration tests in CI with `APALACHE_FATAL_WARNINGS` set.

This sometimes makes all test jobs fail early, e.g, due to wrong formatting or compiler warnings, but not because the actual tests are broken. See for example: https://github.com/informalsystems/apalache/runs/6389502799?check_suite_focus=true

Ideally, we'd be alerted of both formatting errors *and* failing test on the first CI run.
Otherwise, especially with integration tests, there's considerable round-trip time for a second CI run.

I wonder if we should split off a separate build job to be run with `APALACHE_FATAL_WARNINGS` and disable it on unit/integration test jobs. @shonfeder WDYT?

shonfeder:
Closes #1741

Removes the fatal warnings configuration from all our other CI jobs, and instead
runs a separate job that just compiles the code (including test code) with
fatal warnings enabled. This should provide fast feedback for compiler warnings
without interfering with other tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,9 +38,35 @@ jobs:
       - name: Check formatting
         run: make fmt-check
 
-  unit-tests:
+  compiler-warnings:
     env:
       APALACHE_FATAL_WARNINGS: true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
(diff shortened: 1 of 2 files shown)


thpani (approved):
Wow that was super quick! 🤩
LGTM, thank you! 👍

PR #1749, opened 2022-05-13
Introduce dedicated exit codes

### LINKED ISSUE #1392: [FEATURE] Designated error code when a type checking error occurs
As highlighted in #1391, it would be good to return an exit code different from the error code report by the model checker. The related code is here: 

https://github.com/informalsystems/apalache/blob/73ea58304bc97c652dd6de911dd0c2de0e9ffa70/mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala#L244-L249

The exit codes can be found here: https://github.com/informalsystems/apalache/blob/unstable/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/ExitCodes.scala

We should stay compatible with TLC on the error code, as much as possible.

thpani:
Closes #1392 

Exit with dedicated error codes for
- type-checking errors,
- parser errors, and
- evaluation errors (e.g., due to unsupported language constructs and operators).

To this extent, we:
- Add exit code `120` (the range 80–149 is unused by TLC) for type-checking errors.
- Refactor `Pass` and its implementations to return an `Either` encapsulating the exit code or pass result.

&nbsp;


konnov (approved):
Great PR! Exactly what we needed!

UNRELEASED.md
```diff
@@ -10,6 +10,11 @@
          * Some bug fix, see #124
 
      DO NOT LEAVE A BLANK LINE BELOW THIS PREAMBLE -->
+### Breaking changes
+
+ * Dedicated exit codes for type-checking errors, parser errors, and evaluation
+   errors (e.g., due to unsupported language constructs or operators), see #1749
```
shonfeder, line 16:
```suggestion
 * Introduced dedicated exit codes for type-checking errors, parser errors, and evaluation errors (e.g., due to unsupported language constructs or operators), see #1749
```

mod-infra/src/main/scala/at/forsyte/apalache/infra/ExitCodes.scala
```diff
@@ -0,0 +1,57 @@
+package at.forsyte.apalache.infra
+
+import tlc2.output.EC
+
+/**
+ * Exit codes that are reported by Tool.
```
shonfeder, line 6: Should `Tool` be linked or else maybe write "by the CLI tool" or something? Depending on whether this is referring to the Scala object or the program entrypoint more generally?

mod-infra/src/main/scala/at/forsyte/apalache/infra/ExitCodes.scala
```diff
@@ -0,0 +1,57 @@
...
+ *   - 150–153 and 255 for syntax errors and system-level errors.
+ *
+ * In addition, Apalache uses exit code
+ *   - 120 for type-checking errors.
+ *
+ * @author
+ *   Igor Konnov, Thomas Pani
+ */
+object ExitCodes {
+  type TExitCode = Int
```
shonfeder, line 24: What's gained by having a type alias for `Int`? And why not just call it `ExitCode`?

mod-infra/src/main/scala/at/forsyte/apalache/infra/ExitCodes.scala
```diff
@@ -0,0 +1,57 @@
...
+ * that exit codes should not replace richer error explanations.
+ *
+ * TLC currently specifies exit codes
+ *   - 0 for success,
+ *   - 10–14 for (various kinds of) property violations,
+ *   - 75–77 for "failures" (e.g., properties that are tautologies, or primed variables in invariants), and
+ *   - 150–153 and 255 for syntax errors and system-level errors.
+ *
+ * In addition, Apalache uses exit code
+ *   - 120 for type-checking errors.
```
shonfeder, line 18: Seems like this is now missing some of the exit codes we use in Apalache?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/PassChainExecutor.scala
```diff
@@ -35,7 +36,8 @@ class PassChainExecutor(val options: WriteablePassOptions, passes: Seq[Pass]) ex
     }
 
     passes.zipWithIndex
-      .foldLeft(Option[TlaModule with TlaModuleProperties](new TlaModule("empty", Seq()) with TlaModuleProperties)) {
+      .foldLeft[Either[TExitCode, TlaModule with TlaModuleProperties]](Right[TExitCode,
+              TlaModule with TlaModuleProperties](new TlaModule("empty", Seq()) with TlaModuleProperties)) {
```
shonfeder, line 40: Just a suggestion, but maybe this could be cleaned up with some temp variables? It seems a bit hairy to me now. One possibility:

```suggestion
  val emptyModule = new TlaModule("empty", Seq()) with TlaModuleProperties)
  val init = Right[TExitCode,TlaModule with TlaModuleProperties](emptyModule)
  // ...
      .foldLeft[Either[TExitCode, TlaModule with TlaModuleProperties]](init) {
```

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
```diff
@@ -184,16 +184,16 @@ object Tool extends LazyLogging {
...
-      errCode
-    }
+    result.fold(errorCode => {
+          logger.info(msgIfFail)
+          errorCode
+        },
+        module => {
+          logger.info(msgIfOk(module))
+          ExitCodes.OK
+        })
```
shonfeder, line 196: Have to say, I feel like a case analysis using `match` would be clearer here:

```suggestion
    result match {
      case Left(errorCode) => {
          logger.info(msgIfFail)
          errorCode
        }
      case Right(module) => {
          logger.info(msgIfOk(module))
          ExitCodes.OK
        })
```

My sense is that the `fold` cominator only aids readability when you have some other combinators to feed into it, or you are ignoring the values. Otherwise you just end up with an explicit case analysis only it's positional instead of using pattern matching to make it explicit what value on the lhs gets eliminated by what expression on the rhs.

WDYT?

shonfeder (changes requested):
Great improvement!

I have one suggestion that I think helps with readability, and a few questions.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
shonfeder: This is the one place I'd really prefer to see a change, unless you feel strongly that it isn't an improvement.

UNRELEASED.md
thpani: Good point!
Not sure if I prefer past tense indicative or imperative with these though 🤣

mod-infra/src/main/scala/at/forsyte/apalache/infra/ExitCodes.scala
thpani: The problem I have here, is to properly `[[link]]` in Scaladoc, the name must be visible/imported. But then the compiler complains about the unused import.

I actually hit this a couple of times now, any ideas how to resolve this?
I'd rather not silence a compiler warning just to have a relatively innocent link in a comment.

Maybe this is a compromise?
```suggestion
 * Exit codes that are reported by `Tool`.
```

mod-infra/src/main/scala/at/forsyte/apalache/infra/ExitCodes.scala
thpani: > What's gained by having a type alias for `Int`?

a) I believe that type names are important in conveying meaning, e.g., a return type of `TExitCode` is more meaningful than `Int`.
b) Exit codes range from 0..255, they are definitely not `Int`s ;)

> And why not just call it `ExitCode`?

I want to use both the type and the object with unprefixed names. If Scala 2 allowed top-level type declarations (as does Scala 3), the type would anyhow live next to the `object` and be named differently to avoid a name clash.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
thpani: > only it's positional instead of using pattern matching to make it explicit what value on the lhs gets eliminated

To me personally, `errorCode` and `module` make it pretty obvious which case is which, and the fold is syntactically less cluttered. I don't have a strong preference here, so I'm going include your suggestion.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/PassChainExecutor.scala
thpani: Good point, this is a lot more readable!

mod-infra/src/main/scala/at/forsyte/apalache/infra/ExitCodes.scala
thpani: I think I got it now, reworded in 3841307ae.

mod-infra/src/main/scala/at/forsyte/apalache/infra/ExitCodes.scala
shonfeder: Yep, that seems to make sense to me! It’d a a small thing, but I think this makes it read a bit better.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
shonfeder: I can see what you’re saying here. the perceived readability probably just comes from familiarity on my end. I guess I still think this parses easier given the rest of our codebase, but wouldn’t object if you d coded to change it back. :)

shonfeder (approved):
LGTM!

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
thpani: I think this really depends on what you're reading the code for. Clearly marking `Left` and `Right` has its value, and it makes the type of `result` obvious without a type annotation, so I'm quite happy to keep your suggestion :)

Issue #1752, opened 2022-05-13
Non-determinism in function sets


konnov:
The following test is failing, though it should not:

```tla
TestFunFromFunSet ==
    \* Introduce cells via LET-IN, so they do not get optimized.
    LET x == 1 IN
    LET y == 2 IN
    LET S == { 0, x, y - 1 } IN
    LET T == 3..6 IN
    \* This will get negated when checking the invariant.
    \* Note that x = y - 1, but the model checker does not statically deduce it.
    \* Hence, a sound implementation must ensure that the equal elements of S
    \* are mapped on the same elements of T.
    \A f \in [ S -> T ]:
        f[x] = f[y - 1]
```

shonfeder:
Closed by #1753

PR #1753, opened 2022-05-13
Fix 1752 by de-duplicating sets

konnov:
This PR fixes the tricky issue discovered in #1752, which is related to the underlying representation of sets. Basically, the function domain may point to two cells `c_1` and `c_2` that are equal in an SMT model. In this case, the function relation would contain two pairs `(c_1, c_3)` and `(c_2, c_4)`. As a result, the function may sometimes return `c_3` and sometimes return `c_4`, when applied to `c_1 = c_2`.

To fix this bug, we have to de-duplicate the set representation. As we have been already doing something similar in `CardinalityRule` and `FoldSetRule`, I have extracted deduplication into `SetOps.dedup` and refactored the two rules to use it. Having deduplication, it was trivial to fix `CherryPick` to pick functions in a sound way.


rodrigo7491 (commented):
The changes look sound to me, but the fact that the integration tests are taking more than one hour needs to be investigated. Also, I imagine you meant to link #1752 in the PR's description.

konnov:
It's weird that it takes so long.

shonfeder:
The integration test on which things were hanging appears to be

[check TestSets.tla reports no error (array-encoding)](https://github.com/informalsystems/apalache/blob/unstable/test/tla/cli-integration-tests.md#check-testsetstla-reports-no-error-array-encoding)

konnov:
I think `SetOps.dedup` should be fixed for the arrays encoding. It is now used in the implementation of `Cardinality` and `FoldSet`.

konnov:
Independently of a fix, I would really love to understand why it takes z3 more than 6 hours to solve this instance (with the incorrect encoding). This may help us understanding the performance bottlenecks.

rodrigo7491:
@konnov Sure, let's have a look at it.

konnov:
@rodrigo7491 and I have debugged it. The bug was not about the arrays encoding at all. Actually, I have introduced an exponentially large expression by accident, so any recursive code over the expression was stuck forever.

It's amazing that we have caught that via the integration tests! 🎉

rodrigo7491 (approved):
LGTM

Issue #1766, opened 2022-05-16
Wire the changelog management plugin into the continuous deployment pipeline


shonfeder:
Factoring out the third work item in #1538 

Depends on #1765

PR #1768, opened 2022-05-17
Add plugin to harmonize changelog management

### LINKED ISSUE #1765: Add change log management sbt plugin
Factoring out the second work item from #1538

## Plan 

- [x] Set up an (internal) SBT plugin, which I'll call `changeling`
- [x] Define `changelingDirectory: TaskKey[File]` which
  - [x] depends on a `changelingKinds: SettingKey[Seq[String]]` that established the
    supported kinds of changes and order in which these should be reported in
    the rendered changelog, and
  - [x] ensures an `.unreleased` directory exists at the root of our repo, with a
    subdirectory for each supported kind of change.
  - [x] has the value of the path to the `.unreleased` directory.
- [x] Define `changelingReleaseNotes : TaskKey[File]` which
  - [x] depends on the value of `changelingDirectory`
  - [x] maps the contents of the `.unchanged` directory into a markdown file,
    rendering the unreleased changelog notes into the usual expected form for
    release notes
  - [x] has the value of the path to the produced release notes file
- [x] Define `changelingChangelog : TaskKey[File]` which
  - depends on 
    - [x] the value of `changelingKinds`
    - [x] the value of `changelingDirectory`
    - [x] the value of `changelingReleaseNotes`
    - [x] the value of `changelingChangelogName : SettingKey[File]`, configuring
      the location of the main changelog file
  - [x] prepends `changelingReleaseNotes.value` onto the `changelingChangelogName.value`
  - [x] removes all the unchanged entries from the `.unrleased` directory
- [x] Enable the pluging in apalache's SBT config

shonfeder:
Closes #1765

This introduces a little SBT plugin to improve our changelog management process. It implements the solution decided in https://github.com/informalsystems/apalache/tree/4934bc96a759c4c613938aceeb7adae0f3641e1e/docs/src/adr/019adr-harmonize-changelog.md

A good chunk of the line count is due to documentation of the general design and motivation for the plugin, so I won't reiterate that here.

Documentation on how to add changes with the new system and integration with the CI/CD system will be introduced in the followup PR closing #1766.

The commits are crafted pretty deliberately, and so reviewing by commit should make things fairly easy to go over.


project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+    val leafDirs = children.map(normalizeFileName.andThen(childOfBase))
+    IO.createDirectories(leafDirs)
+    base
+  }
+
+  /**
+   * Render the directory unreleased directory structure as a markdown file
+   *
+   * @param changeKinds
+   *   The supported kinds of cnhages, establishes order of change sections
```
thpani, line 197:
```suggestion
   *   The supported kinds of changes, establishes order of change sections
```

project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+   * @param version
+   *   The version to be released, used as heading of realease notes
+   * @param unreleasedDir
+   *   The directory from which to read the unreleased change entries
+   * @param relaseNotes
+   *   The file into which the release notes will be written
+   * @returns
+   *   The TrelaseNotes
```
thpani, line 205:
```suggestion
   *   The file containing the rendered release notes
```

project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+  def concatFiles(front: File, target: File): Unit = {
+    val content = s"${IO.read(front)}\n${IO.read(target)}"
+    IO.write(target, content)
+  }
+
+  /**
+   * Move all the contents of the cange directories in `unrelasedDir` into a "trash" dir in `resourceDir`
+   *
+   * @return
+   *   the trash directory crated
```
thpani, line 256:
```suggestion
   *   the trash directory created
```

project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+   */
+  def cleanChanges(resourceDir: File, unreleasedDir: File): Unit = {
+    // So we can recover dirs if needed
+    val trashDir = resourceDir / "changeling-trash" / LocalDate.now.toString
```
thpani, line 260: I'm not clear yet how this would work when cutting a release. Are we committing the trashed files back into `main`?

Also, if we keep the trash, shouldn't we keep it by version, not by date?

project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+    val lineBreaks = "(\\r\\n?|\\n)"
+    val linesBreaksRemoved = s.replaceAll(lineBreaks, " ").trim()
+    s"- ${linesBreaksRemoved}"
+  }
+
+  // Heading -> Filename
+  private val normalizeFileName: String => String = _.replace(' ', '-').toLowerCase()
+  // Filename -> Heading
+  private val deNormalizeFileName: String => String = _.replace('-', ' ').capitalize
+  private val readFile: File => String = IO.read(_)
```
thpani, line 279: You're calling `IO.read()` directly in all but one place; maybe this can go?

project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+ * =Changeling=
+ *
+ * Changeling is an SBT pluging to make allow simple maintenance of a merge-conflict free changelog.
+ *
+ * ==Motivation==
+ *
+ * Merge conflicts arise when the same lines in the same files are edited
+ *
+ * "simultaneously" in two different branches. The basic idea is to reduce the chance of merge conflicts by adding each
+ * entry of an unreleased change into the changelog via a separate file.
```
thpani, line 13:
```suggestion
 * Merge conflicts arise when the same lines in the same files are edited
 * "simultaneously" in two different branches. The basic idea is to reduce the chance of merge conflicts by adding each
 * entry of an unreleased change into the changelog via a separate file.
```

project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+ * |                       \------ Change entry 4
+ * |
+ * \------- 0.0.1
+ * |--------- Features
+ * |            |------ Change entry 1
+ * |            \------ Change entry 2
+ * |
+ * \--------- Bug fixes
+ * |------ Change entry 3
+ * \------ Change entry 4
```
thpani, line 69: I think this needs some reformatting; `Features` and `Bug fixes` should go under `0.0.1`, change entries 3+4 under `Bug fixes`, right?

I also like the `├──` and `└` below a bit better :)

project/sbt-changeling/Changeling.scala
```diff
@@ -0,0 +1,280 @@
...
+ * - Change entry 1
+ * - Change entry 2
+ *
+ * ### Bug fixes
+ *
+ * - Change entry 3
+ * - Change entry 4
+ * }}}
+ *
+ * The obvious, high-level document model is a tree of depth 4:
```
thpani, line 49: I've only heard "depth" applied to individual nodes, I think the entire tree has a height.
Also, I believe the convention is to count the path length, not the vertices.

```suggestion
 * The obvious, high-level document model is a tree of height 3:
```

thpani (commented):
I like this a lot!
This looks like the leanest solution to our problem and it should be quite easy to maintain 🎉😃 

I'm still unsure how the trashing of an obsolete `.unreleased` would work when cutting a release; see my question below.

Otherwise, just a few typos in the comments.

konnov (approved):
Looks good to me. I would leave the decision to @thpani, as he had closer look.

project/sbt-changeling/Changeling.scala
shonfeder: I think the auto formatter broke my ascii art. I’ll see if I can fix it.

project/sbt-changeling/Changeling.scala
shonfeder: This is just a safeguard for when people run it on their local machines, just in case they, for instance, configure the .unreleased due to point to something important. It’s not meant to be kept or inspected at all, and plays no role in releases.

thpani: sg, this was my only blocking concern

thpani (approved):

project/sbt-changeling/Changeling.scala
shonfeder: Yeah, it’s probably not needed. I like to enable the function chaining in the one place it’s used, but that’s not really a huge readability win. I’ll ditch this. Thanks.

shonfeder:
Thanks for all the fixes, @thpani. Sorry for the mess of typos! 🤦

project/sbt-changeling/Changeling.scala
shonfeder: I fixed the alignment, but will leave the different ascii art. This one is meant to reflect the conceptual structure describing the document model, and the unicode chars are from the `tree` command, representing the concrete directory structure. Thanks again for the careful review!

thpani: Makes sense, thanks!

PR #1769, opened 2022-05-17
Add printer for tla identifiers

p-offtermatt:
- ~~[ ] Documentation added for any new functionality~~
- ~~[ ] Entry added to [UNRELEASED.md](./UNRELEASED.md) for any new functionality~~

This is needed for the tableau encoding, where in a TLA spec, variables stand for expressions.
The names of the variables should reflect which expression they stand for.
This printer takes an expression and outputs a string representing it which is usable as an identifier in TLA, i.e. it contains only alphanumeric characters and '_'

### DIFF
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
@@ -442,7 +442,7 @@ object UsableAsIdentifierPrinter extends Printer {
         throw new NotImplementedError("Printing Let-In expressions as identifiers is not supported")
       case _ =>
         throw new NotImplementedError(
-            s"Printing expression as identifier is not supported for expression ${p_ex.toString()}"
+            s"Printing expression as identifier is not supported for expression ${p_ex.toString()} of type ${p_ex.getClass()}"
         )
     }
   }
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
@@ -302,3 +302,148 @@ object SimplePrinter extends Printer {
 object FullPrinter extends Printer {
   override def apply(p_ex: TlaEx): String = ""
 }
+
+object UsableAsIdentifierPrinter extends Printer {
+  val leftBracket = "d"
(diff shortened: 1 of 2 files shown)


tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
```diff
@@ -302,3 +302,148 @@ object SimplePrinter extends Printer {
 object FullPrinter extends Printer {
   override def apply(p_ex: TlaEx): String = ""
 }
+
+object UsableAsIdentifierPrinter extends Printer {
```
thpani, line 306: I think it would be good to include a Scaladoc comment with an example.
It's a bit difficult to see from the code how this prints.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
```diff
@@ -302,3 +302,148 @@ object SimplePrinter extends Printer {
...
+  val leftBracket = "d"
+  val rightBracket = "b"
```
thpani, line 308: Are these just magic constants? Maybe we should use some ASCII constant like `_COLON_` below? Either way, it would be good to have a comment explaining what they represent.

thpani (commented):
Interesting beast. Is there some length limit to identifiers?

I briefly skimmed the code and left some suggestions.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/io/Printer.scala
p-offtermatt: Yes - they are a weird way to print brackets, which maybe is not optimal.
I originally chose this for brevity, since identifiers get quite long, and something like _LEFTBRACKET_ is very long, but maybe it's too obscure to have these

thpani: Might be fine, but we should check that we don't produce name clashes if these constants are alphabetical characters.

p-offtermatt:
I discussed this with Jure, and this printer doesn't seem like the best way to approach this.
The length of identifiers is in fact limited.

Jure instead suggested to auto-generate names and map them to strings via a function and modify the counterexample writer to evaluate the function, which makes sense to me. I'll close this for now

PR #1773, opened 2022-05-17
Print oracle/verifier module in Nitpicker

thpani:
Add convenience methods to Nitpicker that print the oracle/verifier module if no counterexample is found (i.e., if there is a problem with one of the encodings).


konnov (approved):
Looks great!

Issue #1798, opened 2022-05-21
Strange counterexample with function with Int domain


nano-o:
Consider the following:
```
----------------------------- MODULE Test -----------------------------
EXTENDS Integers
VARIABLES
    \* @type: Int -> Int;
    f
Init == f = [[i \in Int |-> 0] EXCEPT ![0] = 42]
Next == UNCHANGED f
Inv == f[0]=42
=============================================================================
```
Run Apalache with options `check --inv=Inv --length=0`.
This fails with a strange counterexample:
```
---------------------------- MODULE counterexample ----------------------------

EXTENDS TestGuess

(* Constant initialization state *)
ConstInit == TRUE

(* Initial state *)
State0 == f = SetAsFun({})

(* The following formula holds true in the last state and violates the invariant *)
InvariantViolation == ~(f[0] = 42)

================================================================================
```

nano-o:
And with `--smt-encoding=array` Apalache crashes:
```
PASS #12: BoundedChecker                                          I@16:20:45.591
(Please report an issue at: [https://github.com/informalsystems/apalache/issues],java.lang.ClassCastException: class com.microsoft.z3.Expr cannot be cast to class com.microsoft.z3.ArrayExpr (com.microsoft.z3.Expr and com.microsoft.z3.ArrayExpr are in unnamed module of loader 'app'))
Unhandled exception                                               E@16:20:45.817
java.lang.ClassCastException: class com.microsoft.z3.Expr cannot be cast to class com.microsoft.z3.ArrayExpr (com.microsoft.z3.Expr and com.microsoft.z3.ArrayExpr are in unnamed module of loader 'app')
        at at.forsyte.apalache.tla.bmcmt.smt.Z3SolverContext.mkSelect(Z3SolverContext.scala:227)
        at at.forsyte.apalache.tla.bmcmt.smt.Z3SolverContext.toExpr(Z3SolverContext.scala:698)
        at at.forsyte.apalache.tla.bmcmt.smt.Z3SolverContext.toExpr(Z3SolverContext.scala:672)
        at at.forsyte.apalache.tla.bmcmt.smt.Z3SolverContext.assertGroundExpr(Z3SolverContext.scala:313)
        at at.forsyte.apalache.tla.bmcmt.smt.RecordingSolverContext.assertGroundExpr(RecordingSolverContext.scala:157)
[...]

PR #1802, opened 2022-05-23
Encode infinite sets in the arrays encoding

rodrigo7491:

This PR encodes infinite sets as constant SMT arrays in the arrays encoding. This fixes a problem first mentioned by @konnov when checking a version of [EWD998](https://github.com/tlaplus/Examples/tree/master/specifications/ewd998). It is also the cause of the exception thrown by the arrays encoding mentioned in #1798.


konnov (approved):
Nice! There is definitely something about supporting functions of infinite domains in the arrays encoding. However, we should be real careful about not letting cells like that to propagate into the rules that expect bounded sets, e.g., what happens if we do `DOMAN f \union { 1 }`?

shonfeder (approved):
LGTM

thpani:
Is this ready to be merged?

rodrigo7491:
This PR fixes the immediate crash problem reported on #1798. Further improvements to the handling of infinite sets will be done at a later stage.

PR #1804, opened 2022-05-23
Extend Apalache, Integers in Nitpicker specs

thpani:
Nitpicker-produced specs sometimes contain unary `-` and `Apalache!SetAsFun`.

Include `EXTENDS Apalache, Integers` in the docs and printed oracle/verifier specs, so they properly run through `apalache-mc`.


rodrigo7491 (approved):
LGTM!

PR #1809, opened 2022-05-24
Add the simulation command

konnov:
This PR adds an experimental command `simulate`. It is basically a special mode of `check`, where an enabled transition is randomly picked instead of considering all transitions at once. This mode is somewhat similar to the option `-simulate` of TLC.

We will experiment with this mode and let the MBT team experiment with it. If we find this mode useful, we will add detailed documentation for it.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala
```diff
@@ -52,12 +54,23 @@ class SeqModelChecker[ExecutorContextT](
...
+        makeStep(isNext = true, checkerInput.nextTransitions)
+      }
+
+      searchState.onRunDone()
+      if (searchState.canContinue) {
+        trex.recover(constSnapshot)
+        logger.info("----------------------------")
+        logger.info(s"Symbolic runs left: ${searchState.nRunsLeft}")
+      }
     }
```
Kukovec, line 74: I had to read this bit very carefully. Please add a comment that `searchState.canContinue` is now defined differently, and terminates the loop in 1 step for non-simulation, because it's not clear just from this code diff.

Kukovec (approved):

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala
thpani: I agree. We should add a similar comment to the nested `if (searchState.canContinue)`.

thpani (approved):

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala
konnov: Added a few comments

PR #1810, opened 2022-05-24
Implement temporal encoding

p-offtermatt:
There is a temporary solution for how the temporal encoding is called; documentation will be added as soon as that is final
Same as documentation, this will wait for the final version.

This is the first version of the temporal encoding. I decided it makes the most sense to get feedback on the code even before all the details are done, to decrease the size of the individual pull requests.
The new pass is not invoked yet, so it shouldn't surprisingly affect any users.

Regarding the commit history: I'm not sure what exactly went wrong, but I think I messed up a rebase. In any case, the history will be squashed/condensed into a single commit, so no need to worry about it. For reviewing, I suggest not looking at any individual commits and just taking a look at the changed files.


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,443 @@
...
+class TableauEncoder(
+    module: TlaModule,
+    gen: UniqueNameGenerator,
+    loopEnc: LoopEncoder,
+    tracker: TransformationTracker)
+    extends LazyLogging {
+  val levelFinder = new TlaLevelFinder(module)
+  val varNamesToExStrings = new HashMap[String, String]()
+
+  def encodeVarNameMapping(modWithPreds: ModWithPreds): ModWithPreds = {
```
p-offtermatt, line 34: It just generates a useless expression (that is removed by the Inliner) for now, but the expression will be used in a later iteration to write nicer counterexamples.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestTemporalEncoder.scala
```diff
@@ -0,0 +1,180 @@
...
+
+  private val tableauEncoder =
+    new TableauEncoder(
+        modWithPreds.module,
+        new UniqueNameGenerator(),
+        loopEncoder,
+        new IdleTracker(),
+    )
+
+  // ad hoc builder for well-typed temporal expressions (everything is typed boolean)
```
p-offtermatt, line 52: once there exists a typed generator, this ad-hoc implementation could be replaced, to enable more extensive tests

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,444 @@
...
+    val newModule = new TlaModule(curModWithPreds.module.name, newDecls)
+
+    curModWithPreds.setModule(newModule)
+  }
+
+  def insertAt[T](elem: T, seq: Seq[T], pos: Int): Seq[T] = {
+    seq.take(pos) ++ List(elem) ++ seq.drop(pos)
+  }
+
+  def getAuxVarForTempOper(oper: TlaTempOper, nodeIdentifier: String): TlaVarDecl = {
```
konnov, line 129: This also looks like a private method. Also, add a comment?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/utils.scala
```diff
@@ -0,0 +1,86 @@
...
+}
+
+object DeclUtils {
+  import at.forsyte.apalache.tla.pp.temporal.utils.builder
+
+  /**
+   * Takes decl, ex and returns newDecl with the same name as decl, with its body extended like this: newDecl == decl /\
+   * ex
+   */
+  def conjunctExToOperDecl(ex: TlaEx, decl: TlaOperDecl, tracker: TransformationTracker): TlaOperDecl = {
```
konnov, line 74:
```suggestion
  def andInDecl(ex: TlaEx, decl: TlaOperDecl, tracker: TransformationTracker): TlaOperDecl = {
```

In general, I keep longer-living objects on the left-hand side. So I would prefer this order of the arguments: `tracker`, `decl`, and `ex`

konnov (commented):
Great work! Added a bunch of comments :)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  /**
+   * Transforms init into
+   *
+   * newInit ==
+   *
+   * /\ init
+   *
+   * /\ inLoop = FALSE
```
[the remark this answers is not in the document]
p-offtermatt: InLoop = true in state i means state i-1 is the first state on the loop.
It's the way it's done in the paper (see p22 of https://lmcs.episciences.org/2236)

Note that the behaviour like this neatly disallows single-states loops unless they are allowed by having stuttering.

If conversely InLoop being true at state i means state i is the first state of the loop, you could immediately end the execution, since the copy vars are equal to the vars by definition.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,267 @@
...
+
+  /**
+   * Adds a loop variable to the loopOK expression by transforming loopOK into
+   *
+   * newLoopOK ==
+   *
+   * /\ oldLoopOK
+   *
+   * /\ foo = loop_foo
+   */
```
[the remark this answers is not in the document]
p-offtermatt: LoopOK answers the question "Does the execution right now encode a proper loop?"
So it ensures that a) the loop has been started and b) the saved copies of the variables have the same value as the variables right now, e.g. the loop is closed

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/utils.scala
p-offtermatt: Interesting. My intuition is to have the "primary actors" on the left-hand side, and "background actors" on the right-hand side.
E.g. `AdditionThatAlsoLogs(summand1, summand2, logger)`
If you prefer tracker, decl, ex nonetheless, I'll adjust :)

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,267 @@
...
+class LoopEncoder(tracker: TransformationTracker) extends LazyLogging {
+  import LoopEncoder.NAME_PREFIX
+
+  val boolTag = Typed(BoolT1)
+
+  val inLoopDecl = TlaVarDecl(LoopEncoder.IN_LOOP_NAME)(boolTag)
+  val inLoop = builder.declAsNameEx(inLoopDecl)
+  val inLoopPrime = builder.primeVar(inLoopDecl)
+
+  /** For each variable foo, creates a declaration of an auxiliary variable loop_foo */
```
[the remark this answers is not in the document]
p-offtermatt: I like __saved_ the most, so I made it this for now.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,267 @@
...
+   *
+   * loopOK ==
+   *
+   * /\ InLoop
+   *
+   * /\ foo = loop_foo
+   *
+   * /\ bar = loop_bar
+   */
+  def createLoopOKPredicate(
```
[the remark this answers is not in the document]
p-offtermatt: But let me know if I misunderstood your point!

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,444 @@
...
+class TableauEncoder(
+    module: TlaModule,
+    gen: UniqueNameGenerator,
+    loopEnc: LoopEncoder,
+    tracker: TransformationTracker)
+    extends LazyLogging {
+  val levelFinder = new TlaLevelFinder(module)
+  val varNamesToExStrings = new HashMap[String, String]()
+
+  def encodeVarNameMapping(modWithPreds: ModWithPreds): ModWithPreds = {
```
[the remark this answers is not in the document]
p-offtermatt: This is essentially a filler method, it will be refactored by the more informative counterexample writer, I don't think it's worth documenting since it will be gone in the next PR.

For context, it generated the table you saw in the temporal encoding a while ago, e.g.
```
GenVarNamesToExFun ==
  { { "loop_t_9", "☐(∀i$10 ∈ Corr: (pc[i$10] ≠ 'AC'))" },
    { "loop_t_8",
      "(∀i$9 ∈ Corr: (pc[i$9] = 'V0')) ⇒ ☐(∀i$10 ∈ Corr: (pc[i$10] ≠ 'AC'))" },
    { "t_8",
      "(∀i$9 ∈ Corr: (pc[i$9] = 'V0')) ⇒ ☐(∀i$10 ∈ Corr: (pc[i$10] ≠ 'AC'))" },
    { "t_9", "☐(∀i$10 ∈ Corr: (pc[i$10] ≠ 'AC'))" } }
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,444 @@
...
+                "Found a null expression of temporal level; this should not be possible")
+          case NameEx(_) =>
+            throw new IrrecoverablePreprocessingError(
+                "Found a name expression of temporal level.  After inlining no such name expressions should be left in the predicate!")
+          case ValEx(_) =>
+            throw new IrrecoverablePreprocessingError(
+                "Found a value expression of temporal level. This should not be possible.")
+          case LetInEx(_, _) =>
+            throw new IrrecoverablePreprocessingError(
+                "Expect to find no LET-IN expressions. They should have been rewritten by the inliner")
```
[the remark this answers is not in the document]
p-offtermatt: I inline unary letin in temporal properties now, see https://github.com/informalsystems/apalache/pull/1810/commits/20a40cd166873aa4873e9a10e6f7bb7e453256de

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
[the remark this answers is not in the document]
p-offtermatt: They also use a loopOK predicate, albeit called differently:
see p166, fig 5. They `define` `looped := saved & state = loop`
This will need to be reevaluated in each step too, as far as I understand.
It is exactly the same as my predicate: `LoopOK = InLoop /\ state = loop_state`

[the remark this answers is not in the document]
p-offtermatt: Maybe it's confusing because I called it a predicate? LoopOk is a 0-parameter boolean valued expression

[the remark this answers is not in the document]
konnov: Oh, I think I have though about the start of the loop, and you are closing the loop here. Sorry about the confusion. It's all good.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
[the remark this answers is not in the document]
p-offtermatt: Will be fixed in https://github.com/informalsystems/apalache/pull/1823

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/utils.scala
thpani: I agree with Philip on this; I'd find it confusing if `tracker` was the first parameter here.

tla-pp/src/test/scala/at/forsyte/apalache/tla/pp/TestTemporalEncoder.scala
```diff
@@ -0,0 +1,209 @@
...
+import org.scalacheck.Gen.oneOf
+import org.scalacheck.Prop
+import org.scalacheck.Prop.AnyOperators
+import org.scalacheck.Prop.forAll
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatestplus.junit.JUnitRunner
+import org.scalatestplus.scalacheck.Checkers
+
+@RunWith(classOf[JUnitRunner])
+class TestTemporalEncoder extends AnyFunSuite with Checkers {
```
Kukovec, line 31:
```suggestion
class TestTableauEncoder extends AnyFunSuite with Checkers {
```
?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+ */
+@Singleton
+class TableauEncoder(
+    module: TlaModule,
+    gen: UniqueNameGenerator,
+    loopEnc: LoopEncoder,
+    tracker: TransformationTracker)
+    extends LazyLogging {
+  val levelFinder = new TlaLevelFinder(module)
+  val varNamesToExStrings = new HashMap[String, String]()
```
Kukovec, line 32: For clarity, I'd prefer `var x = immutable.HashMap` over `val x = mutable.HashMap`.

Kukovec: Also, why does this need to be mutable in the first place?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+
+    // replace formula by formula == loopOK => formula_init, and move formula to the end of the spec so it is after loopOK
+    val newFormulaDecl = new TlaOperDecl(
+        formula.name,
+        formula.formalParams,
+        builder.impl(
+            builder.appOp(builder.declAsNameEx(curModWithPreds.loopOK), Seq.empty[TBuilderInstruction]: _*),
+            builder.declAsNameEx(initExVarDecl),
+        ),
+    )(formula.typeTag)
```
Kukovec, line 98:
```suggestion
    val newBody = builder.impl(
            builder.appOp(builder.declAsNameEx(curModWithPreds.loopOK)),
            builder.declAsNameEx(initExVarDecl),
        ),
    val newFormulaDecl = formula.copy(body = newBody)
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+   */
+  def temporalsToInvariants(
+      modWithPreds: ModWithPreds,
+      formulas: Seq[TlaOperDecl]): ModWithPreds = {
+
+    formulas.foldLeft(modWithPreds)(singleTemporalToInvariant)
+  }
+
+  /**
+   * Adds a variable that stores the initial value of the given expression ex. Create a new variable Init_ex and modify
```
Kukovec, line 47:
```suggestion
   * Adds a variable that stores the initial value of the given Boolean expression ex. Create a new variable Init_ex and modify
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+  /**
+   * Encodes each of a sequence of temporal formulas.
+   * @see
+   *   [[encodeFormula]]
+   */
+  def temporalsToInvariants(
+      modWithPreds: ModWithPreds,
+      formulas: Seq[TlaOperDecl]): ModWithPreds = {
+
+    formulas.foldLeft(modWithPreds)(singleTemporalToInvariant)
```
Kukovec, line 43: Because there's quite a bit of indirection it's hard to summarize as a single `suggestion`, but it's the same sentiment as before: instead of iteratively updating `mod` in a loop, it would make more sense, at least to me, to compute collections of all additions/overrides (e.g. separate collections of `Init_ex = ex` to be added to Init, and a `UNCHANGED << Init_ex1, ..., Init_ex2 >>` to be added to Next, as well as the implications for the formulas) and construct the final module once at the end.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+
+  /**
+   * Adds a variable that stores the initial value of the given expression ex. Create a new variable Init_ex and modify
+   * init and next:
+   *
+   * newInit == oldInit /\ Init_ex = ex
+   *
+   * newNext == oldNext /\ UNCHANGED << Init_ex >>
+   */
+  def addInitExVar(modWithPreds: ModWithPreds, ex: TlaEx, exName: String): (ModWithPreds, TlaVarDecl) = {
```
Kukovec, line 54: See comment on `temporalsToInvariants`

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+    )
+
+    val curModWithPreds = modWithPreds.setPredicates(newInit, newNext, modWithPreds.loopOK)
+    (curModWithPreds.prependDecl(exVarDecl), exVarDecl)
+  }
+
+  /**
+   * Encodes a given formula, using the Tableau encoding by adjusting init, next, loopOK and the given formula.
+   */
+  def singleTemporalToInvariant(
```
Kukovec, line 80: See comment on `temporalsToInvariants`

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+
+  /**
+   * Takes a temporal operator and a name, and generates a new variable declaration declaring the auxiliary "unrolling"
+   * variable for that temporal operator.
+   */
+  private def createUnrollingVar(oper: TlaTempOper, nodeIdentifier: String): TlaVarDecl = {
+    val nameSuffix = oper match {
+      case TlaTempOper.box     => TableauEncoder.BOX_SUFFIX
+      case TlaTempOper.diamond => TableauEncoder.DIAMOND_SUFFIX
+    }
```
Kukovec, line 113: AFAICT, this should produce a warning due to being underspecified for other `TlaTempOper`. You might need to add a default case with a comment to the effect of "impossible by construction".

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+                "Found a null expression of temporal level; this should not be possible")
+          case NameEx(_) =>
+            throw new IrrecoverablePreprocessingError(
+                "Found a name expression of temporal level.  After inlining no such name expressions should be left in the predicate!")
+          case ValEx(_) =>
+            throw new IrrecoverablePreprocessingError(
+                "Found a value expression of temporal level. This should not be possible.")
+          case LetInEx(_, _) =>
+            throw new IrrecoverablePreprocessingError(
+                "Expect to find no LET-IN expressions. They should have been rewritten by the inliner")
```
Kukovec, line 154:
```suggestion
                "Expected to find no LET-IN expressions. They should have been rewritten by the inliner.")
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+   */
+  def encodeSyntaxTreeInPredicates(
+      modWithPreds: ModWithPreds,
+      curNode: TlaEx): (ModWithPreds, TlaEx) = {
+    levelFinder.getLevelOfExpression(Set.empty, curNode) match {
+      case TlaLevelTemporal =>
+        curNode match {
+          case NullEx =>
+            throw new IrrecoverablePreprocessingError(
+                "Found a null expression of temporal level; this should not be possible")
```
Kukovec, line 145:
```suggestion
                "Found a null expression of temporal level; this should not be possible.")
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+            throw new IrrecoverablePreprocessingError(
+                "Found a name expression of temporal level.  After inlining no such name expressions should be left in the predicate!")
+          case ValEx(_) =>
+            throw new IrrecoverablePreprocessingError(
+                "Found a value expression of temporal level. This should not be possible.")
+          case LetInEx(_, _) =>
+            throw new IrrecoverablePreprocessingError(
+                "Expect to find no LET-IN expressions. They should have been rewritten by the inliner")
+          case OperEx(oper, args @ _*) =>
+            var curModWithPreds = modWithPreds
```
Kukovec, line 156: Do you really need a mutable `var`?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,424 @@
...
+
+            val nodeIdentifier = TableauEncoder.NAME_PREFIX + gen.newName()
+            varNamesToExStrings.addOne(nodeIdentifier, curNode.toString().replace("\"", "\'"))
```
Kukovec, line 159: Where do the `\"`s come from?

Kukovec: Also, see earlier comment on making `varNamesToExStrings` immutable.

Kukovec (commented):
I'm not 100% done yet, missing about 2/3 of `TableauEncoder`, but my brain power is dropping. You can look at the current comments while I review the rest tomorrow. 

Though `TableauEncoder` by itself is complex enough that it should have had its own PR.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/TemporalPassImpl.scala
```diff
@@ -0,0 +1,133 @@
...
+    val levelFinder = new TlaLevelFinder(module)
+
+    val temporalFormulas = temporalProperties
+      .map(invName => {
+        module.declarations.find(_.name == invName)
+      })
+      .filter(invOption =>
+        invOption match {
+          case Some(inv: TlaOperDecl) if inv.formalParams.isEmpty =>
+            // either a state invariant, or an action invariant
```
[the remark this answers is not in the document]
p-offtermatt: Yes, that was copied from some other pass

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/TemporalPassImpl.scala
```diff
@@ -0,0 +1,133 @@
...
+            throw new TlaInputError(message, Some(initDecl.body.ID))
+          }
+          initDecl
+        case None =>
+          val message = (s"init predicate named `${init}` not found")
+          throw new TlaInputError(message)
+        case _ =>
+          val message = (s"Expected to find a predicate named `${init}` but did not")
+          throw new TlaInputError(message)
+      }
```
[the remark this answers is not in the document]
p-offtermatt: No, not really. This was more or less only as I didn't quite know whether this would all be guaranteed to be fine, but I never ran into any issues while testing, so I've removed all checking that init and next are fine, assuming another pass before this handles it if not

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/TemporalPassImpl.scala
```diff
@@ -0,0 +1,149 @@
...
+      val loopEncoder = new LoopEncoder(tracker)
+
+      val loopModWithPreds =
+        loopEncoder.addLoopLogic(module, initDecl, nextDecl)
+
+      // let-in expressions are hard to handle in temporal formulas,
+      // so they are inlined here
+      val inliner = new Inliner(tracker, renaming, keepNullary = false)
+      val scope = Map[String, TlaOperDecl](temporalFormulas.map(operDecl => (operDecl.name, operDecl)): _*)
+      val transformation = inliner.transform(scope)
```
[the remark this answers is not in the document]
p-offtermatt: I assume there's a proper inliner pass before this (the inliner here is merely because
I don't want to keep nullary in temporal formulas, but they are fine outside of this),
so I think such dependencies will already be gone.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,272 @@
...
+   * For each variable foo, creates a declaration of an auxiliary variable __saved_foo __saved_foo stores the value of
+   * variable foo at the start of the loop.
+   */
+  def createAllVarCopiesInLoop(originalVariables: Seq[TlaVarDecl]): Seq[TlaVarDecl] = {
+    originalVariables.map(varDecl => createVarCopyVariableInLoop(varDecl))
+  }
+
+  def createVarCopyVariableInLoop(varDecl: TlaVarDecl): TlaVarDecl = {
+    TlaVarDecl(s"${NAME_PREFIX}${varDecl.name}")(varDecl.typeTag)
+  }
```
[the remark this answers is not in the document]
p-offtermatt: I tried this locally and it gives me an error: "could not find implicit value for parameter typeTag: at.forsyte.apalache.tla.lir.TypeTag"

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,272 @@
...
+  def addLoopLogicToInit(
+      variables: Seq[TlaVarDecl],
+      loopVariables: Seq[TlaVarDecl],
+      init: TlaOperDecl): TlaOperDecl = {
+
+    val initWithLoopVarInits =
+      variables
+        .zip(loopVariables)
+        .foldLeft(init)({ case (curInit, (varDecl, loopVarDecl)) =>
+          addLoopVarToInit(varDecl, loopVarDecl, curInit)
```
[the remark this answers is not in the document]
p-offtermatt: > Is addLoopVarToInit really necessary as a method? AFAICT, this is the only place it's invoked, and it is itself a 1-liner.

It's also invoked in the TableauEncoder. But you're right, I think this function is not complicated enough nor invoked often enough to merit not being inlined.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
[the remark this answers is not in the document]
Kukovec: Interesting, I guess we never rewrote `copy` for non-`OperDecl`s when we added tags. Keep it like it was and ignore this comment then.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,272 @@
...
+  /**
+   * Adds a loop variable to the loopOK expression by transforming loopOK into
+   *
+   * newLoopOK ==
+   *
+   * /\ oldLoopOK
+   *
+   * /\ foo = __saved_foo
+   */
+  def addVariableToLoopOK(
```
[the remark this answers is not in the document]
p-offtermatt: You're right. I tend to write skeletons for things like this while writing the code before I know how complex they'll be, I guess this tells me I need to integrate an inliner pass in my dev flow

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/LoopEncoder.scala
```diff
@@ -0,0 +1,272 @@
...
+      )
+
+    val nextOrUnchangedWithInLoopUpdate = andInDecl(inLoopUpdate, next, tracker)
+
+    /*
+     * /\ __saved_foo' = IF (InLoop' = InLoop) THEN __saved_foo ELSE foo
+     */
+    variables.zip(loopVariables).foldLeft(nextOrUnchangedWithInLoopUpdate) { case (curNext, (varDecl, loopVarDecl)) =>
+      addLoopVarToNext(varDecl, loopVarDecl, curNext)
+    }
```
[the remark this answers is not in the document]
p-offtermatt: Just intuitively, it seems the latter is better, so I changed it

Kukovec (changes requested):
Despite the number of comments, it's good work overall, just feels a bit more imperative than the rest of our codebase (somewhat unsurprising, given your Java background). 
Note that very few, if any, parts were straight-up incorrect, but I would still like to push you more towards the immutable-style.
Well documented for the most part (tough I think the nomenclature in the comments is a bit out of date, since we keep changing our mind w.r.t what the names of the introduced variables should be).

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
p-offtermatt: There could be TlaStrs in the expression. Escaping the quotes avoids terminating strings prematurely, e.g. if this is used in a json. Alternatively, could replace them with escaped quotes (`\\\"`); not sure what to prefer here

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/ScopedBuilder.scala
```diff
@@ -219,4 +219,11 @@ class ScopedBuilder
     funDef(e, tuple(elems: _*), times(sets: _*))
   }
 
+  /**
+   * A name expression referring to the TlaDecl
+   */
+  def declAsNameEx(decl: TlaDecl): TBuilderInstruction = {
```
Kukovec, line 225: Just a note, you don't want this method for all `TlaDecl`, in particular not for `TlaOperDecl`. If you're referencing declarations, you'd need `appDecl` (see current builder, not ported yet), because it's easy to accidentally write `A`, where you needed to have a `A()`.

p-offtermatt:
I'm rewriting `encodeSyntaxTreeInPredicates` in the TableauEncoder in order to not update ModWithPreds constantly, but raher collect and conjunct later. However, this means it has to have a lot of return types (it returns the stuff to be added to init, next, loopOK, and the expression variable for the current node). Is it better style to have `encode(...): (Seq[TlaEx], Seq[TlaEx], Seq[TlaEx], TlaEx)` or to introduce an extra object that stores these, e.g. encode(...): `(Preds, TlaEx)`, where `Preds` is something like
```
class Preds(
    val init: TlaOperDecl,
    val next: TlaOperDecl,
    val loopOK: TlaOperDecl)
```
? I think the many return types look weird, and since so many have the same type, it makes it a bit obscure which component should be what

Kukovec:
> ? I think the many return types look weird, and since so many have the same type, it makes it a bit obscure which component should be what

I'd probably introduce something like `Preds` as well (though `case class`), or alternatively use `State`, but I think the latter might be a bit overkill here.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,423 @@
...
+                def innerOp(args: TBuilderInstruction*): TBuilderInstruction = innerOpTmp(args)
+
+                /* __temporal_curNode_unroll' \in BOOLEAN */
+                val auxVarUpdateAssignmentEx = inBoolSet(builder.prime(auxVarEx))
+
+                /* __temporal_curNode <=> A /\ __temporal_curNode' */
+                val nodeVarUpdateConditionEx = builder.equiv(
+                    nodeVarEx,
+                    outerOp(
+                        argExs(0),
```
Kukovec, line 258: You should make a note as to why args is nonempty and this won't throw. Also, possibly use `head`.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,423 @@
...
+              newFormulas
+        ))
+  }
+
+  /**
+   * Encodes a given formula, using the Tableau encoding by adjusting init, next, loopOK and the given formula.
+   */
+  def singleTemporalToInvariant(formula: TlaOperDecl): (Seq[TlaVarDecl], PredExs, TlaVarDecl) = {
+
+    var (varDecls, preds, (formulaEx)) = encodeSyntaxTreeInPredicates(formula.body)
```
Kukovec, line 90:
```suggestion
    val (varDecls, preds, formulaEx) = encodeSyntaxTreeInPredicates(formula.body)
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,423 @@
...
+    // create a new variable that stores whether the formula evaluated to true in the first state
+    // this is necessary because a temporal formula on a sequence of states should be satisfied
+    // if it holds in the first state, so we need to remember this information
+    val exVarDecl = new TlaVarDecl("__" + formula.name + "_init")(Typed(BoolT1))
+    val exVar = builder.varDeclAsNameEx(exVarDecl)
+
+    // __foo_init = [inital evaluation of foo]
+    val initExVarEx = builder.eql(
+        exVar,
+        builder.useTrustedEx(formulaEx),
```
Kukovec, line 101:
```suggestion
        formulaEx,
```
It's already a `TBuilderInstruction` by construction

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,423 @@
...
+    )
+
+    // UNCHANGED << __foo_init >>
```
Kukovec, line 104:
```suggestion
    // UNCHANGED __foo_init
```

Kukovec (approved):
LGTM, with a few syntax nits. One thing I'd prefer though is if the methods (e.g. in `TableauEncoder`) were in ascending dependency order (= independent on top), so I didn't have to skip forward when a method was invoking another method defined (and described) below.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -0,0 +1,423 @@
...
+                                                    outerOp      innerOp
+                 */
+
+                /*
+                box and diamond differ in the inner and outer operators,
+                and whether the InLoop variable is negated in the conjunction
+                 */
+                val (outerOpTmp, innerOpTmp, innerInLoopEx) = oper match {
+                  case TlaTempOper.box     => (builder.and _, builder.or _, builder.not(loopEnc.inLoopPrime))
+                  case TlaTempOper.diamond => (builder.or _, builder.and _, loopEnc.inLoopPrime)
```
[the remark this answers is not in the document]
Kukovec: > This gives me the following error: "missing argument list for method and in trait BoolBuilder Unapplied methods are only converted to functions when a function type is expected. You can make this conversion explicit by writing `and _` or `and(_)` instead of `and`."

You can fix this by annotating the LHS IIRC.

PR #1770, opened 2022-05-17
ADR020 on improving arena edges

konnov:
In this ADR020, we propose to refactor arenas:

 - so different sets may share edges, and
 - statically known edges propagate without introducing SMT constraints.

This should improve the overall performance of the model checker and will make powersets and function sets much more efficient.

PR #1812, opened 2022-05-24
Add domain comparison to function equality in the arrays encoding

### LINKED ISSUE #1811: Comparison problem with functions with an empty domain in the arrays encoding
## Description

Comparing a function with an empty domain with another function may currently lead to an unsound result in the arrays encoding. This is due to the fact that functions are encoded as constant SMT arrays with a predefined default value, e.g., zero for integers. If the function with the non-empty domain has its range as a singleton set containing the default value for the type in question the equality will be evaluated to true. A potential solution is to compare the function's domains, in addition to the function's themselves. This will add a single SMT equality between arrays (encoding the domains) per comparison.

This is another find of Nitpicker (see #1588).

## Input specification

Simplified spec from the output of Nitpicker:
```
 ------ MODULE Verifier --------
EXTENDS Apalache, Integers

VARIABLES
  \* @type: (Seq(Str) -> <<Str>>);
  result,
  \* @type: Bool;
  found

Init ==
  /\ result \in {[x \in {<<"s1">>} |-> <<"s2">>]}
  /\ result = [x \in {} |-> <<"s3">>]
  /\ found \in BOOLEAN

 Next ==
   UNCHANGED <<result, found>>

 NotFound ==
   ~found
 ======================
```

Version of the spec with integers:
```
 ------ MODULE Verifier --------
EXTENDS Apalache, Integers

VARIABLES
  \* @type: (Int -> Int);
  result,
  \* @type: Bool;
  found

Init ==
  /\ result \in {[x \in {1,2,3} |-> 0]}
  /\ result = [x \in {} |-> 9]
  /\ found \in BOOLEAN

 Next ==
   UNCHANGED <<result, found>>

 NotFound ==
   ~found
 ======================
```

`apalache-mc check --smt-encoding=arrays Verifier.tla`

## Expected behavior

```
Found a deadlock.
The outcome is: Deadlock
Checker has found an error
```

rodrigo7491:

Add SMT equality check of function domains when comparing functions. Closes #1811.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/LazyEquality.scala
```diff
@@ -424,10 +424,17 @@ class LazyEquality(rewriter: SymbStateRewriter)
       case `arraysEncoding` =>
         // In the arrays encoding we only cache the equalities between the elements of the functions' ranges
         // This is because the ranges consist of pairs of form <arg,res>, thus the domains are also handled
+        // TODO: stop caching unnecessary equalities, similar to SetInRuleWithArrays.basicIn?
```
konnov, line 427: This will be resolved in #1770. So it probably does not need additional treatment here.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/LazyEquality.scala
```diff
@@ -424,10 +424,17 @@ class LazyEquality(rewriter: SymbStateRewriter)
...
         val leftElems = state.arena.getHas(leftRel)
         val rightElems = state.arena.getHas(rightRel)
         cacheEqConstraints(state, leftElems.cross(rightElems)) // Cache all the equalities
         eqCache.put(leftFun, rightFun, EqCache.EqEntry())
+        // For the rare case in which one function has an empty domain, we need to be extra careful
+        // See https://github.com/informalsystems/apalache/issues/1811
+        val leftDom = state.arena.getDom(leftFun)
+        val rightDom = state.arena.getDom(rightFun)
+        rewriter.solverContext.assertGroundExpr(tla.eql(leftDom.toNameEx, rightDom.toNameEx))
```
konnov, line 436: I don't think this is sound, as this constraint unconditionally requires the two domains to be equal. This condition should be a part of the function equality.

konnov (commented):
I believe this fix needs a fix :)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/LazyEquality.scala
rodrigo7491: Good point. `LazyEquality` don't make it easy to add constraints to the cached equalities, so I think the best way to go about it is to assert `funEq => domEq`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/LazyEquality.scala
rodrigo7491: Good to know, thanks!

konnov (approved):
Looks good!

PR #1820, opened 2022-05-25
Handle undefined results from model evaluation

### LINKED ISSUE #1819: Undefined solver evaluation leading to spurious counter-example
## Description

Z3's model evaluation might return `UNDEF` for some queries, which can make the decoder generate spurious counter-examples (among other problems), since Apalache currently equates `UNDEF` to false. So far this was only observed in nested set membership in the arrays encoding, see the spec below. This bug was found by Nitpicker (#1588).

## Input specification

```
------ MODULE Oracle --------
EXTENDS Apalache, Integers

VARIABLES
  \* @type: Set(Set(Int));
  witness,
  \* @type: Bool;
  found

Init ==
  /\ witness \in {
    {{-1, 0, 5}, {0}},
    {{9}, {-2, -3}},
    {{1, 2}, {-1}}
    }
  /\ found \in BOOLEAN

Next ==
  UNCHANGED <<witness, found>>

NotFound ==
  ~found
======================
```

`apalache-mc check --inv=NotFound --smt-encoding=arrays Oracle.tla`

## Expected behavior

Correct counter-example, currently the following is being generated:
```

EXTENDS Oracle

ConstInit == TRUE

(* Initial state *)
State0 == found = TRUE /\ witness = {{0}}

InvariantViolation == found

```

## Additional context

We are getting `witness = {{0}}` and not `witness = {{-1, 0, 5}, {0}}` because Z3's set membership check for `{-1, 0, 5}` via the model generated returns undefined, the checked expression is below.
```
(let ((a!1 (store (store (store ((as const (Array Int Bool)) false) (- 1) true)
                         0
                         true)
                  5
                  true)))
  (select (store (store ((as const (Array (Array Int Bool) Bool)) false)
                        a!1
                        true)
                 (store ((as const (Array Int Bool)) false) 0 true)
                 true)
          a!1))

```
Asserting this directly on Z3 gives a SAT result, so we should probably make a `check-sat` query for undefined results, instead of returning false.

rodrigo7491:

This PR makes makes the solver context query Z3 if a model evaluation returns an undefined result. See the linked issue for details. Closes #1819.

### DIFF
--- a/.unreleased/bug-fixes/1819.md
+++ b/.unreleased/bug-fixes/1819.md
@@ -1,0 +1,1 @@
+Fix nested set membership in the arrays encoding, see #1819
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
@@ -331,8 +331,16 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext {
     val (smtEx, _) = toExpr(ex)
     val z3expr = z3solver.getModel.eval(smtEx, true)
     if (z3expr.isBool) {
-      val isTrue = z3expr.getBoolValue.equals(Z3_lbool.Z3_L_TRUE)
-      ValEx(if (isTrue) TlaBool(true) else TlaBool(false)) // in undefined case, just return false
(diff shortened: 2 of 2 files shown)


rodrigo7491:
@konnov I will merge this now, because I think the updated treatment of `UNDEF` is an improvement in any case. I opened an issue in the Z3 repo (https://github.com/Z3Prover/z3/issues/6053) to see if this is indeed a bug in the solver.

PR #1823, opened 2022-05-25
Make counterexample writer more informative

p-offtermatt:
Draft will be ready for review when #1810 is reviewed

An annotated counterexample with this approach looks like this:
https://gist.github.com/p-offtermatt/03f62963715887af33420d580af85940


tla-io/src/main/scala/at/forsyte/apalache/io/lir/ItfCounterexampleWriter.scala
```diff
@@ -49,11 +49,20 @@ class ItfCounterexampleWriter(writer: PrintWriter) extends CounterexampleWriter
     val mappedStates = state0 :: states.drop(2).map(_._2)
     // construct the root JSON object
     val rootMap: mutable.LinkedHashMap[String, ujson.Value] = mutable.LinkedHashMap()
+
+    val metaInformation: List[(String, ujson.Value)] =
+      List[(String, ujson.Value)](
+          "format-description" -> "https://apalache.informal.systems/docs/adr/015adr-trace.html",
```
thpani, line 55: I think this can be made a `Map`?

```suggestion
    val metaInformation: Map[String, ujson.Value] =
      Map[String, ujson.Value](
          "format-description" -> "https://apalache.informal.systems/docs/adr/015adr-trace.html",
```

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * This class produces counterexamples in the Informal Trace Format.
+ *
+ * @param writer
+ *   a print writer to use
+ * @author
+ *   Philip Offtermatt
+ */
+package object NameReplacementMap {
```
thpani, line 13: I think this doccomment belongs to another class?

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
```diff
@@ -0,0 +1,15 @@
...
+  var NameReplacementMap = new HashMap[String, String]()
```
thpani, line 14:
```suggestion
  var nameReplacementMap = new HashMap[String, String]()
```

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -41,12 +43,42 @@ class PrettyWriter(
   private def prettyWriteDoc(doc: Doc): Unit = writer.write(pretty(doc, layout.textWidth).layout)
 
   def write(mod: TlaModule, extendedModuleNames: List[String] = List.empty): Unit =
-    prettyWriteDoc(toDoc(mod, extendedModuleNames))
+    prettyWriteDoc(modToDoc(mod, extendedModuleNames))
+
+  /**
+   * Pretty-prints the given decl twice: Once as valid TLA, and once (before the TLA expression) in a comment, where
+   * names of NameExs are replaced by their mapping in the NameReplacementMap. The output looks, for example, like this:
+   *
```
thpani, line 51: Could you include an example of the `NameReplacementMap` that produces the example below?

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
thpani, line 13: Do we assume the PrettyWriter to be a singleton?
Otherwise, we might need multiple `NameReplacementMap`s?
In short, I'm not sure this should be a `package object` rather than simply a `class`.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -72,13 +104,24 @@ class PrettyWriter(
...
+  def modToDoc(mod: TlaModule, extensionModuleNames: List[String] = List.empty): Doc = {
     moduleNameDoc(mod.name) <>
       moduleExtendsDoc(extensionModuleNames) <>
-      lsep((mod.declarations.toList.map(toDoc)) :+ moduleTerminalDoc, line)
+      lsep((mod.declarations.toList.map(declToDoc(_))) :+ moduleTerminalDoc, line)
   }
 
-  def toDoc(parentPrecedence: (Int, Int), expr: TlaEx): Doc = {
+  /**
+   * writes a provided ex as a doc. The parent precedence is needed to determine whether and how to wrap with braces.
```
thpani, line 114:
```suggestion
   * Writes the provided expr as a doc. parentPrecedence determines whether and how to wrap with braces.
```

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -41,12 +43,42 @@ class PrettyWriter(
...
+  }
 
   // Declarations have a trailing empty line
-  def write(decl: TlaDecl): Unit = prettyWriteDoc(toDoc(decl) <> line <> line)
+  def write(decl: TlaDecl): Unit = {
+    prettyWriteDoc(declToDoc(decl) <> line <> line)
+  }
 
-  def write(expr: TlaEx): Unit = prettyWriteDoc(toDoc((0, 0), expr))
+  def write(expr: TlaEx): Unit = prettyWriteDoc(exToDoc((0, 0), expr, (x: String) => x))
```
thpani, line 81: I think we can just use `Predef.identity` here?

```suggestion
  def write(expr: TlaEx): Unit = prettyWriteDoc(exToDoc((0, 0), expr, Predef.idenitity[String])
```

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -32,6 +32,8 @@ class PrettyWriter(
     extends PrettyPrinter with TlaWriter {
   override val defaultIndent: Int = layout.indent
 
+  val defaultNameResolver = (x: String) => x
```
thpani, line 35: This is referred to at some points, but I think it's never used, and the argument to `write` is explicitly defined?

tla-io/src/main/scala/at/forsyte/apalache/io/lir/ItfCounterexampleWriter.scala
thpani: Or rather a mutable `scala.collection.mutable.Map`?

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
thpani: Also, wrap what? Expressions?

tla-io/src/main/scala/at/forsyte/apalache/io/lir/ItfCounterexampleWriter.scala
```diff
@@ -49,11 +49,20 @@ class ItfCounterexampleWriter(writer: PrintWriter) extends CounterexampleWriter
...
+          "description" -> "Created by Apalache on %s".format(Calendar.getInstance().getTime),
+      ) ++ (if (NameReplacementMap.NameReplacementMap.isEmpty)
+              List()
+            else
+              List("variables-to-expressions" -> NameReplacementMap.NameReplacementMap))
```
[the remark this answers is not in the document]
shonfeder: Actually, just using a map as Thomas suggested seems like an ever better option.

shonfeder (commented):
I had a few suggestions in progress concurrent with Thomas's review, which I'm submitting now.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
p-offtermatt: I updated the comment, does this make it clearer? It's supposed to be globally accessible, since I use it to propagate information from the temporal encoding pass to the counterexample writer at the very end. It also doesn't make sense to have multiple NameReplacementMaps (in what I'm using it for) - if there are multiple PrettyWriters, they should all access the same one

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
```diff
@@ -0,0 +1,15 @@
...
+}
```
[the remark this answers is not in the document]
p-offtermatt: I want this to be a globally accessible store of mappings [String -> String], which can be modified in the temporal pass, and then read from in the counterexample writer.

Do you think there's a better way to do this? I was looking at e.g. `convenience.scala`, which keeps a globally defined object also like this, but maybe it's not the best way

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
p-offtermatt: I'm not sure I prefer it - it's longer and (to me) slightly less clear about the intent (since I'd have to convince myself identity is indeed simply the identity function).

thpani: In this case I'd still prefer to name it, e.g.,

```suggestion
  private def identity(x: String): String = x
  def write(expr: TlaEx): Unit = prettyWriteDoc(exToDoc((0, 0), expr, identity)
```

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
```diff
@@ -0,0 +1,15 @@
+package at.forsyte.apalache.io.lir
+
+import scala.collection.mutable.HashMap
+
+/**
+ * This class produces counterexamples in the Informal Trace Format.
```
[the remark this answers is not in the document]
thpani: Marking unresolved, I think there's still the `@param writer` to clean up.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
thpani: I think Shon had a similar [comment below](https://github.com/informalsystems/apalache/pull/1823#discussion_r902699251), so I will follow up there.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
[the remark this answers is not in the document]
thpani: Following up from my similar [comment above](https://github.com/informalsystems/apalache/pull/1823#discussion_r902688042): So is the issue here that if this was just an instance of `HashMap` we would need to pass it around a lot to get it to the necessary places?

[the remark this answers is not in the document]
p-offtermatt: Yes, exactly

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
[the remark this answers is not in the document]
p-offtermatt: My bad, I missed that! Thanks for pointing out

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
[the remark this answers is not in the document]
thpani: Looks like a hack, but idk how to fix it. Maybe @shonfeder or @Kukovec have an idea.

thpani (approved):
Having `NameReplacementMap` as a package object looks like a hack to me, but I don't want to block on this. We can also fix it in a follow-up.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/NameReplacementMap.scala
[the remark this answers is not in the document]
shonfeder: Right. I hadn’t even considered the global mutable state aspect of this, but that does mean techdebt and could kick of problems with future developments. Even with pending work. I’ll take a closer look here in a bit and see if I can think of anything.

shonfeder (changes requested):
Just dropping a token so I can double check for ways to avoid the global mutable hashmap.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -41,12 +41,55 @@ class PrettyWriter(
...
+   * Here, the NameReplacementMap used was
+   * {{{
+   *  "__temporal_t_1" -> "♢(x = 11)",
+   *  "__temporal_t_1_unroll" -> "♢(x = 11)_unroll"
+   * }}}
+   *
+   * Here, the nameReplacementMap used was
+   * {{{
+   *  "__temporal_t_1" -> "♢(x = 11)",
+   *  "__temporal_t_1_unroll" -> "♢(x = 11)_unroll"
```
shonfeder, line 68: Seems like a duplicate here:'

```suggestion
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -140,6 +141,8 @@ class TableauEncoder(
                 "Expected to find no LET-IN expressions. They should have been rewritten by the inliner.")
           case OperEx(oper, args @ _*) =>
             val nodeIdentifier = TableauEncoder.NAME_PREFIX + gen.newName()
+            NameReplacementMap.store = NameReplacementMap.store
```
shonfeder, line 144: Since these assignments are mutable, I wouldn't think that the assignment to `store` is necessary. Is it?

shonfeder:
> > Just dropping a token so I can double check for ways to avoid the global mutable hashmap.
> 
> Maybe in the same way that passes get their parameters, like `InlinePassImpl` is given an `IncrementalRenaming`? The issue I see with this is getting the map from the BoundedModelCheckerPass to the CounterexampleWriter; I had a cursory look and it appears nontrivial, the calls are quite nested

I'm just spitballing here, but would it be possible to avoid using a map for this at all and instead extend [TlaDecl](https://github.com/informalsystems/apalache/blob/1bcbec5d24c34d36b4c7dcbc776bce921aec15b8/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDecl.scala#L8-L10)

https://github.com/informalsystems/apalache/blob/1bcbec5d24c34d36b4c7dcbc776bce921aec15b8/tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaDecl.scala#L8-L10

with something like

```scala
/**
 * A declaration, e.g., of a variable, constant, or an operator. Technically, this class should be called TlaDef, as we
 * are dealing with TLA+ definitions, see Specifying Systems, Ch. 17.3. Unfortunately, there are variable declarations
 * and operator definitions...
 */
abstract class TlaDecl(implicit val typeTag: TypeTag) extends Identifiable with Serializable with TypeTagged[TlaDecl] {
  def name: String
  def docRepresentation: Option[String]
}
```

So you still get the partial function `TlaDecl -> Option[String]`, but it is given by an association carried with each `d : TlaDecl` rather than having to carry around an external map to track this. WDYT?

konnov:
> WDYT?

We have been keeping the IR data structures minimal for a purpose. The AST of TLC is quite baroque because they have been adding all kind of data that might be useful one day to it. In contrast to that, we decided to keep the minimal representation and carry maps for additional data. Sometimes, it is annoying to deal with the maps, but they compose much better.

shonfeder:
Then I guess either using DI or instantiating the map at some point during instance creation is the only option. I'm not sure where the best place to do that is, but pulling in global mutable state for this purpose seems like it should be avoided.

p-offtermatt:
> Then I guess either using DI or instantiating the map at some point during instance creation is the only option. I'm not sure where the best place to do that is, but pulling in global mutable state for this purpose seems like it should be avoided.

I tried to do this locally for a bit, but anything I came up with is, in my opinion, much worse than the global mutable store.
As far as I can tell, I'd have to pipe the NameReplacementMap through the bounded model checker pass.
If everyone's okay with this, I'll merge this for now and put an issue to search for another way to solve this. Again, as far as I can tell, it requires some relatively invasive refactoring

konnov:
I think it is fine to have a "global" store, as long as the store is passed as a class parameter and instantiated by the dependency injection framework. If we do it this way, the class is still easily testable and mockable. Usually the problems arise with static singletons.

shonfeder:
> I think it is fine to have a "global" store, as long as the store is passed as a class parameter and instantiated by the dependency injection framework. If we do it this way, the class is still easily testable and mockable. Usually the problems arise with static singletons.

@konnov what would prevent this causing bugs once we are running e.g. multiple instances in the server mode?

shonfeder (commented):
Removing my change request so that this can be merged if other agree the existing implementation is OK. I'd not prefer introduction of the global mutable state for this, but won't block.

p-offtermatt:
Retired in favor of https://github.com/informalsystems/apalache/pull/1874

PR #1838, opened 2022-05-31
Output an example trace by default and save all runs in `simulate`.

### LINKED ISSUE #1837: Always write an example after `check`, `simulate`, and `test`
It often happens that people are running the commands `check` and `simulate` without having an invariant specified. It looks like an intention is just to get an execution that was explored by the model checker. We only output such an execution, if a deadlock is found. It's a bit cumbersome to output a single example. That would require us to define a false invariant and use it with the command.

It's very easy to just dump an execution in the current context.

konnov:
Closes #1837. This PR adds two UX features:
 
 - always output an example trace in `example{0,}.*`, if no counterexample is found,
 - add the flag `--save-runs` to the command `simulate`, which saves example for each symbolic run.


.unreleased/features/1838.md
```diff
@@ -0,0 +1 @@
+Always output an example trace, add --save-runs to save examples for each run of `simulate`
```
shonfeder, line 1:
```suggestion
Always output an example trace, add `--save-runs` flag to save examples for each run of `simulate`, see #1838
```

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
```diff
@@ -319,7 +320,7 @@ object Tool extends LazyLogging {
     runAndExit(
         executor,
         _ => "No example found",
-        "Checker has found an example. Check counterexample.tla.",
+        "Found a violation of the postcondition. Check failure.tla.",
```
shonfeder, line 323: Why are we renaming counterexample to `failure`? This doesn't seem easier to understand to me, and seems to be an inaccurate name given the contents of the file.

test/tla/cli-integration-tests.md
```diff
@@ -2864,15 +2864,23 @@ EXITCODE: ERROR (120)
 If we run with the `--out-dir` flag
 
 ```sh
-$ apalache-mc check --out-dir=./test-out-dir --length=0 Counter.tla | sed 's/[IEW]@.*//'
+$ apalache-mc check --out-dir=./test-out-dir --write-intermediate=0 --length=0 Counter.tla | sed 's/[IEW]@.*//'
```
shonfeder, line 2867: Why the need to add this flag?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala
```diff
@@ -46,6 +46,15 @@ class SeqModelChecker[ExecutorContextT](
...
 
   override def run(): CheckerResult = {
+    // output an example of a run, if the context is satisfiable
+    def outputRun(): Unit = {
+      trex.sat(params.smtTimeoutSec) match {
+        case Some(true) =>
+          notifyExample(checkerInput.rootModule, trex.decodedExecution(), searchState.nRunsLeft)
+        case _ => ()
+      }
+    }
```
shonfeder, line 56: Doesn't look like there's any reason for this function to be defined inside the `run` function. Maybe make it a private helper function? (Should help with performance, i'd think, also keeps `run` from adding unneeded lines?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala
```diff
@@ -106,6 +122,23 @@ class SeqModelChecker[ExecutorContextT](
...
+   * Notify all listeners that an example should be processed.
+   *
+   * @param rootModule
+   *   The checked TLA+ module.
+   * @param trace
+   *   The counterexample trace.
+   * @param index
+   *   Number of found error (likely [[SearchState.nRunsLeft]]).
+   */
+  private def notifyExample(
```
shonfeder, line 135: Otherwise it sounds like the example itself is what is being notified.
```suggestion
  private def notifyOfExample(
```

shonfeder (approved):
A few questions and typos. Overall looks good to me.

Integration tests are failing tho.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
konnov: It has been always hard to explain why counterexamples are "counter"-examples. Also, the word is too long :)

 Now I wanted to add examples and this makes things even more confusing. I don't like the word failure either, but as Jure pointed out, "error" would not be technically correct. Perhaps, a "violation" would be a better term.

konnov: ok, then I will change it to "violation" and merge this PR later

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/DumpCounterexamplesModelCheckerListener.scala
```diff
@@ -2,50 +2,46 @@ package at.forsyte.apalache.tla.bmcmt
...
-   * @param trace
-   *   The counterexample trace.
-   * @param invViolated
-   *   The invariant violation to record in the counterexample. Pass
-   *   - for invariant violations: the negated invariant,
-   *   - for deadlocks: `ValEx(TlaBool(true))`,
-   *   - for trace invariants: the applied, negated trace invariant (see [[SeqModelChecker.applyTraceInv]]).
-   * @param errorIndex
-   *   Number of found error (likely [[SearchState.nFoundErrors]]).
-   */
```
thpani, line 31: This has been merged, but is there any reason why this whole doccomment was removed? I spent quite some time figuring out the interface for this.

konnov: Does not it simply repeat the comment from `ModelCheckerListener`? If not, sorry about that!

PR #1842, opened 2022-06-01
Fix function cherry picking relation propagation

### LINKED ISSUE #1840: Spurious counter-examples with functions of records in the arrays encoding
## Description

The arrays encoding is not fully propagating functions' domain information to the decoder to enable correct counter-example generation. Currently, the decoder can erroneously filter out elements that cannot be compared via SMT equality, leading to counter-examples as the one shown below. This problem was found by Nitpicker (#1588).

## Input specification

```
------ MODULE Oracle --------
EXTENDS Apalache, Integers

VARIABLES
  \* @type: ([b: Bool, s: Str] -> (Str -> Bool));
  witness,
  \* @type: Bool;
  found

Init ==
  /\ witness \in {
      [x1 \in {[b |-> FALSE, s |-> "s1"]} |-> [y1 \in {"s2"} |-> FALSE]],
      [x2 \in {[b |-> TRUE,  s |-> "s3"]} |-> [y2 \in {"j", "uAY"} |-> TRUE]]
    }
  /\ found \in BOOLEAN

Next ==
  UNCHANGED <<witness, found>>

NotFound ==
  ~found
======================
```

```

EXTENDS Oracle

ConstInit == TRUE

(* Initial state *)
State0 == found = TRUE /\ witness = SetAsFun({})

InvariantViolation == found

```

apalache-mc check --inv=NotFound --smt-encoding=arrays Oracle.tla

## Expected behavior

A valid witness in the counter-example.

rodrigo7491:

This PR makes cherry picking of functions equate the function's domain, for which SMT constraints are generated, to the domain components of the function's relation, for which no SMT constraints are generated. Without the newly added lazy equality constraints, equivalent but not identical records are treated differently by the decoder, leading to spurious counter-examples.

The PR also contain a small fix to the use of SMT equality in set membership, which ensures that lazy equality is used for sets and functions, which can themselves contain records.

Closes #1840.

### DIFF
--- a/.unreleased/bug-fixes/1840.md
+++ b/.unreleased/bug-fixes/1840.md
@@ -1,0 +1,1 @@
+Fix spurious counter-example generation with functions of records in the arrays encoding, see #1840
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRuleWithArrays.scala
@@ -143,7 +143,7 @@ class SetInRuleWithArrays(rewriter: SymbStateRewriter) extends SetInRule(rewrite
 
   private def supportsSMTEq(cellType: CellT): Boolean = {
     cellType match {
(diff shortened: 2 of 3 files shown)


konnov (commented):
Looks good. We should probably revisit this code, when we have new records and variants stabilized.

konnov:
I think it is fine to merge it as is

Issue #936, opened 2021-07-29
Propose prioritization rubric


shonfeder:
Discussed in retro meeting today.

We'd like a set, and systematic way to triage issues. We want to make sure that we deal with critical issues ASAP, that we track and address important issues in a timely fashion, and that we remember, but deprioritize low-importance issues.

I'm meant to do a bit of research and propose a standard method for this at our planning meeting next week.

konnov:
Shall we remove this one?

shonfeder:
The need for some sort of schema for prioritization came up today in our planning meeting. @gabrielamafra and I are doing some work towards this.

shonfeder:
We discussed this again in our weekly planning meeting. Here are the relevant part of the meeting notes:

```
- Prioritization and planning of incoming work
  - Weight?
    - Fib nums?
    - Times?
  - Priority? 
    - what's the relation between prioritization and weight?
      - need to have a sense for amount/time or work to evaluate whether we can afford to invest in it.
  - Seems that we need a shared vocabulary
  - How do we decide what's important?
      - Igor current high priority issues:
          - user reports
              - if makes sense
              - if can be quickly fixed
          - bottlenecks that are blocking development
  - Thomas: Need to also bear in mind what we are deprioritizing
      - smaller issues/fixes can impact ability to deliver strategic goals
      - Gabriela: it's hard to see when/if we are gradually undermining strategic efforts
```

shonfeder:
I'll propose a rubric, just as a foil to get started with, by our next MondApalache.

PR #1846, opened 2022-06-01
Add impact prompt in issue templates

shonfeder:
Followup to #936

This adds prompts to issue templates so we can gather the information
needed to effectively priority their issue requests.

-------


konnov (approved):
Looks great!

PR #1855, opened 2022-06-07
Shai: Add support for loading models

### LINKED ISSUE #1114: Add RPC for loading a model
Work on this will involve adding the RPC calls for these the following methods, specified in [RFC10](https://github.com/informalsystems/apalache/blob/unstable/docs/src/adr/010rfc-transition-explorer.md)

```scala
  def loadModel(spec: String, aux: List[String] = List()): Either[LoadErr, UninitializedModel]
```

It will include:

- [x] Defining the data types
- [x] Adding the proto specs for the RPC call
- [x] Calling from the server module into specification loading code
- [x] Adding unit and integration tests

shonfeder:
Closes #1114 

## Main addition

This adds the RPC `loadModel` to Shai. This allows clients to send a spec (and
optional auxiliary modules) to receive the spec back as the JSON encoding of our
IR.

It also loads the `TlaModule` resulting from a successful parse into the
connection map. This will be used in subsequent work (e.g., #1958) to enable
clients to interactively explore the specified model.

## Supporting work

This change is facilitated by addition of the following to the `TransExplorerService`:

- A semaphore scoped over all methods in the service, allowing us to synchronize
  atomic invocation of SANY.
- Helper methods to abstract getting and updating connections stored in the map.

## Reviewing

See individual commits for details and to see the evolution of the work, but
it's probably easiest to review all at once, because the code took many shapes
as I worked out what changes were needed to Apalache's CLI and Pass system to
enable this feature.


shai/src/main/scala/at/forsyte/apalache/shai/RpcServer.scala
```diff
@@ -15,7 +15,10 @@ object RpcServer extends ServerMain {
     // A thread safe mutable reference to the active connections
     // See https://zio.dev/version-1.x/datatypes/concurrency/ref/
     connections <- Ref.make(Map[UUID, Conn]())
-  } yield new TransExplorerService(connections)
+    // Ensure atomic access to the parser
+    // See https://github.com/informalsystems/apalache/issues/1114#issuecomment-1180534894
+    parserSemaphore <- Semaphore.make(permits = 1)
```
konnov, line 20: Not sure about whether we need a semaphore here, which in general allows `permits` threads to acquire a resource (in this code `permits=1`), or just a [mutex/lock](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html), which is designed for a single-thread access.

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
```diff
@@ -40,10 +68,75 @@ class TransExplorerService(connections: Ref[Map[UUID, Conn]]) extends ZioTransEx
...
+        } yield LoadModelResponse.Result.Spec(json)
+      case Left(err) =>
+        ZIO.succeed(LoadModelResponse.Result.Err(err.getMessage()))
+    }
+  } yield LoadModelResponse(result)
+
+  private def parseSpec(spec: String, aux: Seq[String]): Result[Either[Throwable, TlaModule]] =
+    // Obtain permit on the semaphore protecting access to the parser, ensuring the parser is not
+    // run by more than one thread at a time.
+    parserSemaphore.withPermit(ZIO.effectTotal(try {
```
konnov, line 100: I would rather use [Lock#tryLock](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html#tryLock(long,%20java.util.concurrent.TimeUnit)) with a large timeout (several minutes), in case SANY gets stuck somewhere for some reason. Although I am not sure what to do if SANY really gets stuck. Perhaps hide it in a thread that is killed by a timeout?

konnov (approved):
Looks good to me! Of course, the interaction with the parser is the most sensitive point here. We probably do not need to think about all possible cases for the first version, but it would be good to turn these concurrency considerations into issues for future work.

shai/src/main/scala/at/forsyte/apalache/shai/RpcServer.scala
thpani, line 20: Somewhat orthogonal, but I'm wondering if we shouldn't sequentialize parsing once and for all in `SanyImporter`?

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
```diff
@@ -40,10 +68,75 @@ class TransExplorerService(connections: Ref[Map[UUID, Conn]]) extends ZioTransEx
...
+  def loadModel(req: LoadModelRequest): Result[LoadModelResponse] = for {
+    parseResult <- parseSpec(req.spec, req.aux)
+    result <- parseResult match {
+      case Right(module) =>
+        for {
+          _ <- updateConnection(req.conn)(_.setModel(module))
+          json = jsonOfModule(module)
+        } yield LoadModelResponse.Result.Spec(json)
+      case Left(err) =>
+        ZIO.succeed(LoadModelResponse.Result.Err(err.getMessage()))
```
thpani, line 93: This is probably my ignorance of the ZIO effect system, but why is only the `Left` branch wrapped in `ZIO.succeed`?

thpani (commented):
I have a preliminary question, because I have a hard time understanding the overall use of the effect system here.

Looking at the type alias `Result[T] = ZIO[ZEnv, Status, T]`:

* The failure type here is a `grpc.Status`, which encodes both success and failure codes.
* Similarly, the success type parameter `T` is set to `LoadModelResponse`, which holds a tagged union of either a JSON module on success or a string error on failure. It's also set to `Either[Throwable, TlaModule]`.

It looks like there's three levels of failures:
* at the effect-system level
* at the success type level (`Either`)
* at the level of the returned protobuf (`LoadModelResponse`)

Can you give an explanation of when to use which?

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
```diff
@@ -40,10 +68,75 @@ class TransExplorerService(connections: Ref[Map[UUID, Conn]]) extends ZioTransEx
...
+      uuid <-
+        try {
+          reqConn match {
+            case Some(conn) => ZIO.succeed(UUID.fromString(conn.id))
+            case None       => throw new IllegalArgumentException()
+          }
+        } catch {
+          case _: IllegalArgumentException =>
+            // TODO log for invalid conn ID: https://github.com/informalsystems/apalache/issues/1849
+            ZIO.fail(Status.INVALID_ARGUMENT)
```
thpani, line 125: Can we immediately return a `ZIO.fail` here?

shai/src/main/scala/at/forsyte/apalache/shai/RpcServer.scala
shonfeder: We are using ZIO for concurrency, not scala/java primatives. ZIO doesn't have a special mutex construct, but instead just considers mutex as a special case of a semaphore when `n = 1` (all concurrency using the ZIO scheduler is cooperative, so the preemptive behavior that would, iiuc, otherwise differentiate a binary semphore from mutex is irrelevant). See https://github.com/zio/zio/blob/b9bfff0f505c392df46b720489b25e550ce80bb5/core/shared/src/main/scala/zio/Semaphore.scala#L29-L35

If you see a more suitable synchronization mechanism available in https://zio.dev/version-1.x/datatypes/concurrency/ I'm to switch it out. But otherwise just consider this the idiomatic way of making a mutex in ZIO :)

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
shonfeder: We can add a timeout, but not with lock try. Every effect can have a timeout added in ZIO. See https://zio.dev/version-1.x/overview/overview_basic_concurrency#timeout

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
```diff
@@ -40,10 +68,75 @@ class TransExplorerService(connections: Ref[Map[UUID, Conn]]) extends ZioTransEx
...
+   *
+   * @param req
+   *   the request to load a model, including the root module spec and any auxiliary modules
+   */
+  def loadModel(req: LoadModelRequest): Result[LoadModelResponse] = for {
+    parseResult <- parseSpec(req.spec, req.aux)
+    result <- parseResult match {
+      case Right(module) =>
+        for {
+          _ <- updateConnection(req.conn)(_.setModel(module))
```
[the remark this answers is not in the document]
shonfeder: We are adding a connection to a map, which will replace the previous value of the key with the new value for that key. I.e., in 

```scala
  private def updateConnection(reqConn: RequestConn)(f: Conn => Conn): Result[Conn] = for {
    conn <- getConnection(reqConn)
    updatedConn = f(conn)
    _ <- addConnection(updatedConn)
  } yield updatedConn
```

We get from the map, then update the value, then set the new value by adding the connection back in.

The only thing that affects mutable state here is where we run the `update` method on the `connections` ref:

```scala
  private def addConnection(c: Conn): Result[Unit] = connections.update(_ + (c.id -> c))
```

If you think of any suggestions to improve the code here, I'd be happy to hear them! Also happy to sync up quickly to discuss :)

shonfeder:
> Looking at the type alias `Result[T] = ZIO[ZEnv, Status, T]`:
> 
>     * The failure type here is a `grpc.Status`, which encodes both success and failure codes.
> 
>     * Similarly, the success type parameter `T` is set to `LoadModelResponse`, which holds a tagged union of either a JSON module on success or a string error on failure. It's also set to `Either[Throwable, TlaModule]`.
> 
> 
> It looks like there's three levels of failures:
> 
>     * at the effect-system level
> 
>     * at the success type level (`Either`)
> 
>     * at the level of the returned protobuf (`LoadModelResponse`)
> 
> 
> Can you give an explanation of when to use which?

So the `Either` used here is only for internal purposes, and is not exposed as part of the API. There are only two levels of error exposed to users of this module:

- ZIO `Status` errors are for problems with the grpc connection, so the equivalent of 400 or 500 http errors. These are protocol-level errors. As you can see, in this code we have promoted inclusion of valid of connection IDs to the protocol level. A failure here means the interaction cannot move forward. I.e., a failure at this level means the RPC system is being used incorrectly (or is faulty).
- The protobuf level errors are for communicating success/failure data to users in the context of successful RPC calls. So if you successfully execute a "parse spec" RPC but the result is that the spec is syntactically invalid, you receive an error results in the protobuf message sent backs as a response.

The use of `Either` you notice just represents a possibly erroneous computation internal to the server, and any `Left` here is handled internally. In this case, we transform it into a protobuf message error reporting on the parsing failure:

https://github.com/informalsystems/apalache/blob/bbc4f7686649b1ab37e98542ee62963ed7be23bf/shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala#L90-L91

Does this help clarify? (If so, i'll add it to the readme for the project.)

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
thpani: I meant in the `None` branch above, instead of throwing and catching `IllegalArgumentException`.

shonfeder: Ah. Yeah we can do, I just figured this way had less repetition and gave us just two exit points, rather than three. iiuc, you think just returning the same fail on line 120 would read better?

shai/src/main/scala/at/forsyte/apalache/shai/RpcServer.scala
shonfeder: We could just use a ref here: https://zio.dev/version-1.x/datatypes/concurrency/ref

That should give us unique access, it's just that also also exposes the update operations that we don't need. Probably would look cleaner tho. WDYT?

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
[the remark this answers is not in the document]
thpani: Uff, I missed that the updated connection will have the same id.
Can we rename `addConnection` to `addOrUpdateConnection` then?
Also, `updateConnection` only works as intended if its parameter `f` does not change the connection id (otherwise it would *add* a connection)?

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
thpani: For me yes, but resolve it as you see fit :)

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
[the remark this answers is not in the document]
shonfeder: How about changing it to `setConnection`? Would this work for you, since "setting" often the has the sense of adding something new if not already there? It doesn't do an "update" in the sense that `updateConnextion` does.

Good point on the possible misuse of the update function. I'll add a requirement on the `updateConenction` to enforce a condition that the id of the connection doesn't change.

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
shonfeder: Timeout added in 49948dd7cb90696110e2d2f7136773c1e0de1fd4

shai/src/main/scala/at/forsyte/apalache/shai/TransExplorerService.scala
shonfeder: Changed this.

thpani (approved):
Makes sense, thanks for the sync and async explanations 👍 

LGTM

shonfeder:
Thanks for the great questions and review, @thpani! I added documentation to record the salient points from our discussion in 6b9eaf4706e48d88b4ff38b5502d8cc261992afb

PR #1863, opened 2022-06-21
Incorporate comments

p-offtermatt:
Followup to the comments of #1810 , which was mistakenly merged before comments were incorporated.


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -87,7 +85,7 @@ class TableauEncoder(
    */
   def singleTemporalToInvariant(formula: TlaOperDecl): (Seq[TlaVarDecl], PredExs, TlaVarDecl) = {
 
-    var (varDecls, preds, (formulaEx)) = encodeSyntaxTreeInPredicates(formula.body)
+    var (varDecls, preds, formulaEx) = encodeSyntaxTreeInPredicates(formula.body)
```
Kukovec, line 88:
```suggestion
    val (varDecls, preds, formulaEx) = encodeSyntaxTreeInPredicates(formula.body)
```

p-offtermatt:
I generally prefer having utility methods at the end of a file, and other methods ordered top-down by which is invoked first. My reasoning is that generally reading `SpecificUtility()` should have less priority than reading `MainFunction()`, so it should come later. I still usually put tiny utility functions at the top, since they are easy to ignore there anyways.

I've changed the ordering, and I prefer the old one, but not strongly so, so I'll go with the suggestion.

PR #1879, opened 2022-06-24
Add support for actions under temporal operators

p-offtermatt:
Fixes buggy behaviour when actions are inside temporal operators by subtly changing the encoding.
It necessitates a change from the way the encoding is specified in the paper, which is necessary due to the technicalities of TLA+.
Essentially, double-priming is not allowed, while in LTL stacking next operators is no issue, so the subtle changes account for this.

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
@@ -15,6 +15,7 @@ import at.forsyte.apalache.tla.typecomp.TBuilderInstruction
 import at.forsyte.apalache.tla.lir.transformations.TransformationTracker
 import at.forsyte.apalache.tla.lir.transformations.standard.DeclarationSorter
 import at.forsyte.apalache.tla.lir.oper.TlaOper
+import at.forsyte.apalache.tla.lir.oper.ApalacheOper
 
 /**
  * Encodes temporal formulas as invariants.
@@ -32,6 +33,21 @@ class TableauEncoder(
   val levelFinder = new TlaLevelFinder(module)
 
   private def inBoolSet(element: TBuilderInstruction): TBuilderInstruction = builder.in(element, builder.booleanSet())
+
+  /* replaces all occurences of foo := bar with foo = bar */
+  private def rewriteAssignmentsAsEquality(ex: TlaEx): TlaEx = {
+    ex match {
+      case OperEx(oper, args @ _*) =>
+        oper match {
+          case ApalacheOper.assign =>
+            OperEx(TlaOper.eq, args: _*)(ex.typeTag)
+          case _ =>
+            OperEx(oper, args.map(rewriteAssignmentsAsEquality): _*)(ex.typeTag)
+        }
+      case _ =>
+        ex
+    }
+  }
 
   /**
    * Creates PredExs for a given propositional operator application of the form OperEx(oper, argExs)(typeTag). The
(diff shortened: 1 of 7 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
```diff
@@ -33,6 +34,21 @@ class TableauEncoder(
 
   private def inBoolSet(element: TBuilderInstruction): TBuilderInstruction = builder.in(element, builder.booleanSet())
 
+  /* replaces all occurences of foo := bar with foo = bar */
```
konnov, line 37: could you add a comment on why it is sound?

konnov (approved):
Looks good. I am not sure about `rewriteAssignmentsAsEquality` though.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
p-offtermatt: You mean, why this is okay to do?
This is inside a temporal property, not in an actual action. Leaving these as assignments leads the transition finder to see an assignment on the right-hand side of another assignment.
E.g. consider this
```
CoolAction == x := 2

TemporalProperty == <>[CoolAction]_x
```

Due to the details of the encoding, a variable will be produced that gets assigned like this:
```
_temporal__CoolAction := (x := 2)
```

Then the assignment (x := 2) should be written as an equality, and that's what this function does

thpani, line 37: Could you add a comment here why this is necessary?

thpani (approved):
👍

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala
thpani: Just saw that this is similar to Igor's comment.
I think it's also valuable to mention why this is done at all.

konnov: I mean, usually you would have `x' := 2`, that is `x` with a prime. Would you keep the prime?

p-offtermatt: see https://github.com/informalsystems/apalache/blob/6d5a5280aa66b7f3584188ea810f40c6447411f7/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala#L388

p-offtermatt: There is a short explanation where this is invoked: https://github.com/informalsystems/apalache/blob/6d5a5280aa66b7f3584188ea810f40c6447411f7/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/temporal/TableauEncoder.scala#L388

do you think this is sufficient?

thpani: Oh yes, that makes sense. ty

p-offtermatt: Clarified this a bit more. (TLDR: everything is kept the same, except := is replaced by =)

PR #1887, opened 2022-06-28
Refactor SanyParserPassImpl

shonfeder:
A light refactor of the `execute` method in the `SanyParserPassImpl`.
I prepared this initially to support #1114, but realized it wasn't quite
the right way to meet the needs of the server.

Still, I think it's a nice enough cleanup to be worth integrating.

The refactor just breaks the execution logic into three chunks with some
helper functions and uses a for comprehension over `Either` to sequence
the possible error cases.

I recommend reviewing using the [split view of the diff](https://github.com/informalsystems/apalache/pull/1887/files?diff=split&w=0).

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/passes/SanyParserPassImpl.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/passes/SanyParserPassImpl.scala
@@ -30,48 +30,62 @@ class SanyParserPassImpl @Inject() (
 
   override def name: String = "SanyParser"
 
+  private def loadFromJson(filename: String): PassResult = {
+    try {
+      val moduleJson = UJsonRep(ujson.read(new File(filename)))
+      val modules = new UJsonToTla(Some(sourceStore))(DefaultTagReader).fromRoot(moduleJson)
+      modules match {
+        case rMod +: Nil => Right(rMod)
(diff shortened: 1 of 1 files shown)


tla-io/src/main/scala/at/forsyte/apalache/tla/imp/passes/SanyParserPassImpl.scala
```diff
@@ -30,48 +30,62 @@ class SanyParserPassImpl @Inject() (
...
+      rootModule <-
+        if (filename.endsWith(".json")) {
+          loadFromJson(filename)
+        } else {
+          loadFromFile(filename)
+        }.flatMap(sortDeclarations)
 
-    Right(rootModule)
+      _ <- saveLoadedModule(rootModule)
+    } yield rootModule
```
Kukovec, line 88: Since you're already using `for`, you could just split this as
```suggestion
  for {
      rootModule <-
        if (filename.endsWith(".json")) {
          loadFromJson(filename)
        } else {
          loadFromFile(filename)
        }
      sorted <- sortDeclarations(rootModule)
      _ <- saveLoadedModule(sorted)
    } yield sorted
```

Kukovec (approved):
LGTM, feel free to ignore the suggestion if you prefer it the current way.

PR #1891, opened 2022-06-28
Two tests for variants

konnov:
Experimental PR that migrates two tests to the variant type. This works even better than I expected.

Also, deleted two old test files that are not used in the integration tests.


test/tla/Bug1682.tla
```diff
@@ -1,20 +1,44 @@
 ---- MODULE Bug1682 ----
 
-EXTENDS Integers
+EXTENDS Integers, Variants
+
+(*
+ @typeAlias: MESSAGE =
+      RequestVote({ term: Int, from: Str })
+    | Vote({ from: Str, to: Str, term: Int })
+    | AppendEntries({ term: Int, from: Str }) ;
```
Kukovec, line 9: Not sure I like the `({ ... })` double delimiter syntax in practice.

test/tla/TwoPhaseTyped_typedefs.tla
```diff
@@ -0,0 +1,7 @@
+----------------------- MODULE TwoPhaseTyped_typedefs ----------------
+(*
+ @typeAlias: MESSAGE = Commit(NIL) | Abort(NIL) | Prepared(RM);
```
Kukovec, line 3: This makes me wonder, should we reserve a `Unit`-like type for stuff like this, so it isn't ad-hoc implemented?

Kukovec (approved):

test/tla/Bug1682.tla
konnov: You can have any type there, `{ ... }` is a record

test/tla/TwoPhaseTyped_typedefs.tla
Kukovec: Yes, but what I'm saying is we should create a list of reserved words (not just for variants, globally), and `Unit` or `NIL` or something alike should be on it. You can't expect people to figure out this idiom from a GH issue.

Kukovec: Followed by static analysis for reserved words.

konnov: We will write about it in the howto and the tutorial.

PR #1908, opened 2022-07-04
introducing an end point for the typed builder

konnov:
I have tried to use the new typed builder and it was not obvious to me how to do that. After talking to @Kukovec, I've introduced a helper object `at.forsyte.apalache.tla.types.tla`, similar to the one we have in the old builder. Additionally, I have changed the API documentation, to make it more accessible to the end user (us). If you look at the test, you will see that there are two ways to build an expression:

 - via an implicit conversion, which needs the target type specified, or
 - via an implicit class that enriches the builder state with the method `build`.

We should understand, whether to keep the both, or to prefer one way to another.


tla-types/src/test/scala/at/forsyte/apalache/tla/types/TestTypesApi.scala
```diff
@@ -0,0 +1,42 @@
+package at.forsyte.apalache.tla.types
```
Kukovec, line 1: This is more or less a redundant test suite, given the tests in `typecmp`.

Kukovec (approved):

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/package.scala
```diff
@@ -1,15 +1,70 @@
...
+    def build: T = {
+      builderState.run(TBuilderContext.empty)._2
+    }
+  }
+
+  /**
+   * Apply the `build` method to a sequence.
+   *
+   * @param builderStates
+   *   a sequence of
```
thpani, line 57:
```suggestion
   *   a sequence of [[TBuilderInternalState]]
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/package.scala
```diff
@@ -1,15 +1,70 @@
...
+   * @tparam T
+   *   the type of a data structure to build, e.g., `TlaEx` or `TlaOperDecl`.
+   * @return
+   *   the sequence of constructed data structured of type `T`
```
thpani, line 61:
```suggestion
   *   the sequence of constructed data structures of type `T`
```

thpani (approved):
👍

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/package.scala
```diff
@@ -1,15 +1,70 @@
...
+   *
+   * @param builderState
+   *   the internal state of the builder, which captures a data structure made so far
+   * @tparam T
+   *   the type of a data structure to build, e.g., `TlaEx` or `TlaOperDecl`.
+   */
+  implicit class BuildViaMethod[T](builderState: TBuilderInternalState[T]) {
+    def build: T = {
+      builderState.run(TBuilderContext.empty)._2
+    }
```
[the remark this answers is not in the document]
konnov: > Just call the implicit explicitly. Or even

I think I tried the first version, and Scala did not like it. The second one requires an implicit conversion, right? Are you sure it would work with the restriction of at most one implicit conversion?

tla-types/src/test/scala/at/forsyte/apalache/tla/types/TestTypesApi.scala
konnov: I agree that it is not as exhaustive. I have introduced this test as a minimal test of the API, in contrast to the exhaustive tests in `typecmp`.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/package.scala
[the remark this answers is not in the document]
konnov: The second way works, to my surprise :)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/package.scala
```diff
@@ -1,15 +1,70 @@
...
+   * An implicit conversion via a class that works as [[build]], but via a method call to `.build()`.
+   *
+   * @param builderState
+   *   the internal state of the builder, which captures a data structure made so far
+   * @tparam T
+   *   the type of a data structure to build, e.g., `TlaEx` or `TlaOperDecl`.
+   */
+  implicit class BuildViaMethod[T](builderState: TBuilderInternalState[T]) {
+    def build: T = {
+      builderState.run(TBuilderContext.empty)._2
```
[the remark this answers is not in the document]
konnov: yep, I have added a doc here too.

konnov:
Damn. I forgot to commit the latest changes and merged... @shonfeder do you know what to do in this case? Create another PR?

shonfeder:
@konnov yeah, another PR is safest. the only other alternative would be to temporarily remove the branch protections on `unstable` to allow admins to push directly.

PR #1930, opened 2022-07-06
Add manual page on variants

### LINKED ISSUE #1923: Add variants to the language manual

konnov:
Closes #1923. Add the manual page on variants.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -54,9 +54,6 @@
 # TLA+ Language Manual for Engineers
 
 - [Introduction](./lang/index.md)
-- [Apalache extensions](./lang/apalache-extensions.md)
-  - [Apalache module](./lang/apalache-operators.md)
-  - [Variants](./lang/variants.md)
 - [The standard operators of TLA+](./lang/standard-operators.md)
     - [Booleans](./lang/booleans.md)
     - [Control And Nondeterminism](./lang/control-and-nondeterminism.md)
@@ -69,6 +66,9 @@
     - [Tuples](./lang/tuples.md)
     - [Sequences](./lang/sequences.md)
     - [Bags]()
+- [Apalache extensions](./lang/apalache-extensions.md)
+  - [Apalache module](./lang/apalache-operators.md)
+  - [Variants](./lang/variants.md)
 - [User-defined operators](./lang/user-operators.md)
     - [Top-level operator definitions](./lang/user/top-level-operators.md)
     - [LET-IN definitions](./lang/user/let-in.md)
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -54,6 +54,9 @@
 # TLA+ Language Manual for Engineers
 
 - [Introduction](./lang/index.md)
+- [Apalache extensions](./lang/apalache-extensions.md)
+  - [Apalache module](./lang/apalache-operators.md)
+  - [Variants](./lang/variants.md)
 - [The standard operators of TLA+](./lang/standard-operators.md)
     - [Booleans](./lang/booleans.md)
     - [Control And Nondeterminism](./lang/control-and-nondeterminism.md)
@@ -72,7 +75,6 @@
     - [Higher-order operators definitions](./lang/user/higher-order-operators.md)
     - [Anonymous operator definitions](./lang/user/lambdas.md)
     - [Local operator definitions](./lang/user/local-operators.md)
-- [Apalache operators](./lang/apalache-operators.md)
 - [Modules, Extends, and Instances]()
 
 # Idiomatic TLA+
--- a/docs/src/lang/variants.md
+++ b/docs/src/lang/variants.md
@@ -186,8 +186,7 @@ an invariant violation and will lead to a spurious counterexample.
 Note that `tagName` is an identifier in this notation. In this type, `b` is a
 type variable that captures other options in the variant type.
 
-**Effect:** The variant constructor returns a new value that keeps the
-pair `(tagName, associatedValue)`.
+**Effect:** The variant constructor returns a new value of the variant type.
 
 **Determinism:** Deterministic.
 
--- a/docs/src/lang/variants.md
+++ b/docs/src/lang/variants.md
@@ -191,6 +191,32 @@ Beer(malt, strength) == Variant("Beer", [ malt |-> malt, strength |-> strength ]
 
 ----------------------------------------------------------------------------
 
+<a name="variantTag"></a>
+### Variant tag
+
+**Notation:** `VariantTag(variant)`
+
+**LaTeX notation:** same
+
+**Arguments:** One argument: a variant constructed via `Variant`.
+
+**Apalache type:** `(tagName(a) | b => Str)`, for some types `a`
+and `b`. Note that `tagName` is an identifier in this notation.
+
+**Effect:** This operator simply returns the tag attached to the variant.
+
+**Determinism:** Deterministic.
+
+**Errors:** No errors.
+
+**Example in TLA+:**
+
+```tla
+VariantTag(Variant("Water", [ sparkling |-> sparkling ])) = "Water"
+```
+
+----------------------------------------------------------------------------
+
 <a name="variantFilter"></a>
 ### Variant filter
 
--- a/docs/src/lang/variants.md
+++ b/docs/src/lang/variants.md
@@ -3,7 +3,7 @@
 **Warning:** This feature is currently under testing. To enable variants
 support in Apalache, pass the option `--features=rows`.
 
-[[Back to all operators]](./apalache-extensions.md)
+[[Back to Apalache extensions]](./apalache-extensions.md)
 
 [Variants][] (also called *tagged unions* or *sum types*) are useful, when you want to combine
 values of different shapes in a single set or a sequence.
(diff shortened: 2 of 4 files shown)


docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+constructors for each option right away:
+
+```tla
+\* @typeAlias: FRUIT = Apple(Str) | Orange(Bool);
+\*
+\* @type: Str => FRUIT;
+Apple(color) == Variant("Apple", color)
+
+\* @type: Bool => FRUIT;
+Orange(seedless) == Variant("Orange", seedless)
```
thpani, line 76: Nitpicking here, but do you consider this interleaved `\*` idiomatic?
I've seen it used below as well, but I find the grouping confusing, because then `Orange` is offset by an empty line.

```suggestion
\* @typeAlias: FRUIT = Apple(Str) | Orange(Bool);

\* @type: Str => FRUIT;
Apple(color) == Variant("Apple", color)

\* @type: Bool => FRUIT;
Orange(seedless) == Variant("Orange", seedless)
```

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+### Variant tag
+
+**Notation:** `VariantTag(variant)`
+
+**LaTeX notation:** same
+
+**Arguments:** One argument: a variant constructed via `Variant`.
+
+**Apalache type:** `(tagName(a) | b => Str)`, for some types `a`
+and `b`. Note that `tagName` is an identifier in this notation.
```
thpani, line 204: Is the grouping here correct? I'd expect

```suggestion
**Apalache type:** `(tagName(a) | b) => Str`, for some types `a`
and `b`. Note that `tagName` is an identifier in this notation.
```

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+\* @type: (Str, Int) => DRINK;
+Beer(malt, strength) == Variant("Beer", [ malt |-> malt, strength |-> strength ])
+```
+
+----------------------------------------------------------------------------
+
+<a name="variantTag"></a>
+### Variant tag
+
+**Notation:** `VariantTag(variant)`
```
thpani, line 197: Looks like we don't support this in `Variants.tla`? Should we mention that?

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+\*   | Beer({ malt: Str, strength: Int });
+\*
+\* @type: Bool => DRINK;
+Water(sparkling) == Variant("Water", [ sparkling |-> sparkling ])
+
+\* @type: (Str, Int) => DRINK;
+Beer(malt, strength) == Variant("Beer", [ malt |-> malt, strength |-> strength ])
+
+LET Drinks == { Water(TRUE), Water(FALSE), Beer("Radler", 2) } IN
+\E d \in VariantFilter("Beer", Drink):
```
thpani, line 255:
```suggestion
LET Drinks == { Water(TRUE), Water(FALSE), Beer("Radler", 2) } IN
\E d \in VariantFilter("Beer", Drinks):
```

thpani (commented):
Looks good so far, I think there's some minor bugs.

I don't think this should be placed above standard TLA+ operators in the index.

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+
+```
+Apple(Str) | a
+```
+
+In this type, we know that whenever a value is tagged with "Apple" it should be
+of the string type. However, we know nothing about other options. Most of the
+time, we want to define variants that are sealed, that is, we know all
+available options. In the Apalache terms, it means that our type would look
+like follows:
```
shonfeder, line 60:
```suggestion
like the following:
```

shonfeder, line 60:
```suggestion
available options. Suppose we wanted to reason about different kinds of fruit, 
but wanted to limit our model to only comparing apples and oranges.
In Apalache, the type for a value that could be either an apple or an orange, but nothing else, 
would be as follows:
```

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+\* @typeAlias: DRINK =
+\*     Water({ sparkling: Bool })
+\*   | Beer({ malt: Str, strength: Int });
+\*
+\* @type: Bool => DRINK;
+Water(sparkling) == Variant("Water", [ sparkling |-> sparkling ])
+
+\* @type: (Str, Int) => DRINK;
+Beer(malt, strength) == Variant("Beer", [ malt |-> malt, strength |-> strength ])
+```
```
shonfeder, line 99: Framing this as  "losing the field names" seems a bit confusing to me. Could it be framed just as: "the value in a variant can be a record"? E.g.,

```suggestion
Variants can wrap records, for when we want to represent compound data with named fields:

```tla
\* @typeAlias: DRINK =
\*     Water({ sparkling: Bool })
\*   | Beer({ malt: Str, strength: Int });
\*
\* @type: Bool => DRINK;
Water(sparkling) == Variant("Water", [ sparkling |-> sparkling ])

\* @type: (Str, Int) => DRINK;
Beer(malt, strength) == Variant("Beer", [ malt |-> malt, strength |-> strength ])
```
```

File: docs/src/SUMMARY.md [hunk 7]
```diff
@@ -54,6 +54,9 @@
 # TLA+ Language Manual for Engineers
 
 - [Introduction](./lang/index.md)
+- [Apalache extensions](./lang/apalache-extensions.md)
+  - [Apalache module](./lang/apalache-operators.md)
+  - [Variants](./lang/variants.md)
 - [The standard operators of TLA+](./lang/standard-operators.md)
     - [Booleans](./lang/booleans.md)
```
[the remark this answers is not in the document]
konnov: Logically speaking, yes, you are right. However, I wanted Apalache sections to be more visible. People are starting to reference our manual, without referring to the tool. So I want readers see that this is not a merely TLA+ manual :)

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
+# Variants
+
+**Warning:** This feature is currently under testing. To enable variants
+support in Apalache, pass the option `--features=rows`.
+
+[[Back to all operators]](./apalache-extensions.md)
```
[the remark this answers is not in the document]
konnov: We have the back links in other sections too. I find them useful, as it is easy to get lost in the manual structure.

docs/src/lang/variants.md
konnov: Makes sense, though we have to make this works with SANY.

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+LET drink == ... IN
+LET nonFree ==
+    IF VariantTag(drink) = "Water"
+    THEN VariantGetUnsafe("Water", drink).sparkling
+    ELSE VariantGetUnsafe("Beer", drink).strength > 0
+IN
+...
+```
+
+In general, you should avoid using `VariantGetUnsafe`, as it is type unsafe :-)
```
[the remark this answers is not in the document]
konnov: Right. It will produce no error, just a value whose value we do not know.

docs/src/SUMMARY.md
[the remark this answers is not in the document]
thpani: That's a noble goal, but here I think it breaks the flow of a novice reader.

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+```tla
+  \E e \in ApplesAndOranges:
+    /\ e.tag = "Apple"
+    /\ e.color /= "green"
+```
+
+This idiom is quite common in untyped TLA+. [Tagged unions in Paxos][] is
+probably the most illuminating example of this idiom. Unfortunately, it is way
+too easy to make a typo in the tag name, since it is a string, or simply access
+a wrong record without checking its tag.
```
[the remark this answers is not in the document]
Kukovec: Unfortunately, backticks don't work well in "suggest" mode...

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+                 [ malt |-> "Non-alcoholic", strength |-> 0])).strength
+```
+
+In the above example, we unpack `water` by using the tag name `"Beer"`.  Since
+`water` is actually tagged with `"Water"`, the operator falls back to the
+default case and returns the record `[ malt |-> "Non-alcoholic", strength |->
+0]`.
+
+**Type-unsafe get.** Sometimes, using `VariantFilter` and `VariantGetOrElse`
+is a nuisance, when we know exactly the value type. In this case, we can bypass
```
Kukovec, line 137:
```suggestion
is a nuisance, when we know the exact value type. In this case, we can bypass
```

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+
+**Notation:** `Variant(tagName, associatedValue)`
+
+**LaTeX notation:** same
+
+**Arguments:** Two arguments: the tag name (a string literal) and a value
+(a TLA+ expression).
+
+**Apalache type:** `(Str, a) => tagName(a) | b `, for some types `a` and `b`.
+Note that `tagName` is an identifier in this notation.
```
Kukovec, line 169: Take some time to explain where `b` comes from in all these examples. A new reader will likely be confused by it.

docs/src/lang/variants.md
```diff
@@ -0,0 +1,351 @@
...
+**Arguments:** Two arguments: the tag name (a string literal) and a variant
+constructed via `Variant`.
+
+**Apalache type:** `(Str, tagName(a) | b) => a`, for some types `a` and
+`b`. Note that `tagName` is an identifier in this notation.
+
+**Effect:** The operator `VariantGetUnsafe` unconditionally returns some value
+that is compatible with the type of values tagged with `tagName`. If `variant`
+is tagged with `tagName`, the returned value is the value that was wrapped via
+the `Variant` constructor. Otherwise, it is some value of proper type. As such,
```
Kukovec, line 319:
```suggestion
the `Variant` constructor. Otherwise, it is some arbitrary value of proper type. As such,
```

Kukovec (approved):

docs/src/lang/variants.md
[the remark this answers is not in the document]
konnov: Added an example.

PR #1934, opened 2022-07-07
Support loading auxiliary modules from Source

shonfeder:
Supports #1114

The server API is able must support load a model by receiving a string
representing the root module and a sequence of strings representing any
auxiliary modules needed due to imports or refinement. This changeset extends
our SanyParser wrapper to support loading auxiliary modules supplied as a
sequence of `scala.io.Source`s.

-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
@@ -102,8 +102,8 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
   }
 
   /**
-   * Load a TLA+ specification from a text source. This method creates a temporary file and saves the source's contents
-   * into it, in order to call SANY.
+   * Load a TLA+ specification from a text source, possibly including auxiliary modules needed for imports. This method
+   * creates temporary files and saves the sources contents into it, in order to call SANY.
    *
    * @param source
    *   the text source for the root module
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
@@ -106,34 +106,41 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
    * into it, in order to call SANY.
    *
    * @param source
-   *   the text source
+   *   the text source for the root module
+   * @param aux
+   *   the text sources for any auxiliary modules
    * @return
    *   the pair (the root module name, a map of modules)
    */
-  def loadFromSource(source: Source): (String, Map[String, TlaModule]) = {
-    val moduleName = moduleNameOfSource(source) match {
-      case Some(n) => n
-      case None    => throw new SanyImporterException(s"No module name found in given source")
-    }
-
+  def loadFromSource(source: Source, aux: Seq[Source] = Seq()): (String, Map[String, TlaModule]) = {
     val tempDir = Files.createTempDirectory("sanyimp").toFile
-    val temp = new File(tempDir, moduleName + ".tla")
     try {
-      // write the contents to a temporary file
-      val pw = new PrintWriter(temp)
-      try {
-        Try(source.getLines()) match {
-          case Success(lines) => lines.foreach(line => pw.println(line))
-          case Failure(e)     => throw new FileNotFoundException("Source not found: " + e.getMessage)
-        }
-      } finally {
-        pw.close()
-      }
-      loadFromFile(temp)
+      val rootModule = saveTlaFile(tempDir, source)
+      aux.foreach(saveTlaFile(tempDir, _))
+      loadFromFile(rootModule)
     } finally {
-      temp.delete()
       tempDir.delete()
     }
+  }
+
+  private def saveTlaFile(dir: File, source: Source): File = {
+    val moduleName = moduleNameOfSource(source) match {
+      case Some(n) => n
+      case None    => throw new SanyImporterException(s"No module name found in source for module")
+    }
+
+    val temp = new File(dir, moduleName + ".tla")
+    // write the contents to a temporary file
+    val pw = new PrintWriter(temp)
+    try {
+      Try(source.getLines()) match {
+        case Success(lines) => lines.foreach(line => pw.println(line))
+        case Failure(e)     => throw new FileNotFoundException("Source not found: " + e.getMessage)
+      }
+    } finally {
+      pw.close()
+    }
+    temp
   }
 
   private def throwOnError(specObj: SpecObj): Unit = {
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
@@ -17,7 +17,7 @@ import scala.util.{Failure, Success, Try}
  * This is the entry point for parsing TLA+ code with SANY and constructing an intermediate representation.
  *
  * @author
- *   Igor Konnov, Thomas Pani
+ *   Igor Konnov, Thomas Pani, Shon Feder
  */
 class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) extends LazyLogging {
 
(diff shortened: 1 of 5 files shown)


shonfeder:
I've enabled auto merge. So please only approve if you don't want to see any changes at all :) :pray:

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -106,34 +106,41 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
    * into it, in order to call SANY.
```
thpani, line 106: Should we extend this explanation with the treatment of auxiliary modules?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -106,34 +106,41 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
...
+      val rootModule = saveTlaFile(tempDir, source)
+      aux.foreach(saveTlaFile(tempDir, _))
+      loadFromFile(rootModule)
+    } finally {
+      tempDir.delete()
+    }
+  }
+
+  private def saveTlaFile(dir: File, source: Source): File = {
     val moduleName = moduleNameOfSource(source) match {
```
thpani, line 127: I think `moduleNameOfSource` (L95) can be made private?
(Commenting here because GH won't let me comment in unchanged code...)

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -106,34 +106,41 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
    * into it, in order to call SANY.
    *
    * @param source
-   *   the text source
+   *   the text source for the root module
+   * @param aux
+   *   the text sources for any auxiliary modules
```
thpani, line 111: Does auxiliary mean import dependencies?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -106,34 +106,41 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
...
+    } finally {
+      tempDir.delete()
+    }
+  }
+
+  private def saveTlaFile(dir: File, source: Source): File = {
     val moduleName = moduleNameOfSource(source) match {
       case Some(n) => n
-      case None    => throw new SanyImporterException(s"No module name found in given source")
+      case None    => throw new SanyImporterException(s"No module name found in source for module")
```
thpani, line 129: Can we pass the file / module name here? Otherwise it might be hard to debug.

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/passes/SanyParserPassImpl.scala
```diff
@@ -58,10 +58,10 @@ class SanyParserPassImpl @Inject() (
     Right(modules.get(rootName).get)
   }
 
-  private def loadFromTlaString(content: String): PassResult = {
+  private def loadFromTlaString(content: String, aux: Seq[String]): PassResult = {
     val (rootName, modules) =
       new SanyImporter(sourceStore, annotationStore)
-        .loadFromSource(Source.fromString(content))
+        .loadFromSource(Source.fromString(content), aux.map(Source.fromString(_)))
```
thpani, line 64: Why don't we pass the `String` directly? What's the benefit of wrapping it in a `scala.io.Source`?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -106,34 +106,41 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
...
-      try {
-        Try(source.getLines()) match {
-          case Success(lines) => lines.foreach(line => pw.println(line))
-          case Failure(e)     => throw new FileNotFoundException("Source not found: " + e.getMessage)
-        }
-      } finally {
-        pw.close()
+      Try(source.getLines()) match {
+        case Success(lines) => lines.foreach(line => pw.println(line))
+        case Failure(e)     => throw new FileNotFoundException("Source not found: " + e.getMessage)
```
thpani, line 138: Looks like this will never happen (as we only wrap `String`s into `Source`s) and could be avoided if we just directly passed a `String`.

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -106,34 +106,41 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
...
+   *   the text sources for any auxiliary modules
    * @return
    *   the pair (the root module name, a map of modules)
    */
-  def loadFromSource(source: Source): (String, Map[String, TlaModule]) = {
+  def loadFromSource(source: Source, aux: Seq[Source] = Seq()): (String, Map[String, TlaModule]) = {
+    val tempDir = Files.createTempDirectory("sanyimp").toFile
+    try {
+      val rootModule = saveTlaFile(tempDir, source)
+      aux.foreach(saveTlaFile(tempDir, _))
```
thpani, line 119: Do we need any error handling for when the same file is passed multiple times?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
thpani: Maybe there should be a test for that.

thpani (commented):
I'm not super convinced that wrapping `String` into `scala.io.Source` is necessary here; looks like it complicates the code more than necessary. But maybe I missed the rationale on some previous PR?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
thpani: Maybe there should be a test for that.

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
shonfeder: Yep’

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/passes/SanyParserPassImpl.scala
shonfeder: The preexisting method was written to take a source, and I was respecting that design design decision. The benefit of using a source, I assume, is that you can load the data incrementally when feasible, and don't need to read it all into memory.

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
shonfeder: This code wasn't added by PR, just FYI, it was only moved.

We are currently only converting source's to strings, but if this was used to stream data from stdin, for example, then this error could arise, no?

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/passes/SanyParserPassImpl.scala
shonfeder: Right, in his case we already have a string. We can change the underlying method in `SanyParser` to only take a string if we think there's no reason to support loading from sources. I was trying to disturb the existing design as little as possible :)

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
shonfeder: I added the first line of the source to the error message in fe66642038af698a4809607988860bb653d40f8d

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
shonfeder: I moved this so we have this safety check when we first call getLines now, and we jus use try's everywhere, so i think it's a bit cleaner?

fe66642038af698a4809607988860bb653d40f8d

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
shonfeder: Added a check for this in fe66642038af698a4809607988860bb653d40f8d

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/passes/SanyParserPassImpl.scala
thpani: Also true 😆

thpani (approved):
Thank you for the thorough responses!

LGTM!

shonfeder:
Thanks for the careful review!

PR #1936, opened 2022-07-07
Introduce VariantTag and remove VariantUnwrap

### LINKED ISSUE #1927: Add VariantTag to Variants.tla
With the current set of operators over variants it is impossible to extract the tag name. We should add such an operator. Then we could make `VariantGetUnsafe` more meaningful:

```tla
IF VariantTag(variant) == "Foo" THEN VariantGetUnsafe("Foo") ELSE 0
```

### LINKED ISSUE #1928: Remove VariantUnwrap in Variants.tla
We added this operator as the base case for `VariantMatch`. Since we have removed `VariantMatch`, there is no meaningful use case for `VariantUnwrap` any longer.

konnov:
Closes #1927 and closes #1928. Instead of doing PRs where I would remove `VariantUnwrap` and then introduce `VariantTag` in the same places again, I just renamed `VariantUnwrap` into `VariantTag`, fixed the type signature and the rewriting rule. It touches a number of files, but most of the changes are mechanical. I had to give up on rewriting tags with a special sort `TAG`, as we now have to returns tags as strings.

### DIFF
--- a/src/tla/Variants.tla
+++ b/src/tla/Variants.tla
@@ -41,24 +41,18 @@ VariantFilter(__tagName, __S) ==
     { __d \in { __e \in __S: __e.tag = __tagName }: __d.value }
 
 (**
- * In cases where `variant` allows for one value,
- * extract the associated value and return it.
- * The type checker must enforce that `variant` allows for one option.
+ * Get the tag name that is associated with a variant.
  *
- * @param `tagValue` the tag attached to the variant
  * @param `variant` a variant that is constructed with `Variant(...)`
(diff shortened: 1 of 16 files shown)


Kukovec (approved):
LGTM, though I would suggest using the new builder wherever possible, so we have less work porting later.

konnov:
I knew you would say that. However, the PR would blow up :)

Kukovec:
> I knew you would say that. However, the PR would blow up :)

How? You add 64 LOC, so there's at most 64 calls to the builder (and practically probably < 10)

PR #1937, opened 2022-07-07
Typed Builder implementation Part 10.5: Unsafe builders as instances

### LINKED ISSUE #1935: Change unsafe builder inheritance to instantiation
Following the discussion under https://github.com/informalsystems/apalache/pull/1932, we have decided to change the way safe builders access the methods of unsafe builders. Concretely, instead of 
```
trait XBuilder extends UnsafeXBuilder {
...
}
```
we should have 
```
trait XBuilder {
val unsafeXBuilder = new UnsafeXBuilder
...
}
```
where Unsafe builders are package-private.

Kukovec:

closes #1935 

Also includes standardization of Sacladocs and a fix to shadowing.


Kukovec:
Also includes standardization of Sacladocs and a fix to shadowing.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/BoolBuilder.scala
```diff
@@ -10,37 +10,40 @@ import at.forsyte.apalache.tla.typecomp.BuilderUtil._
...
+    binaryFromUnsafe(p, q)(unsafeBuilder.impl)
 
-  /** p <=> q */
-  def equiv(p: TBuilderInstruction, q: TBuilderInstruction): TBuilderInstruction = binaryFromUnsafe(p, q)(_equiv)
+  /** {{{p <=> q}}} */
+  def equiv(p: TBuilderInstruction, q: TBuilderInstruction): TBuilderInstruction =
+    binaryFromUnsafe(p, q)(unsafeBuilder.equiv)
 
-  /** \A x \in set: p */
+  /** {{{\A x \in set: p}}} `x` must be a variable name */
```
thpani, line 33: Not marking all the commas here, because it's consistent within the file.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/SetBuilder.scala
```diff
@@ -13,80 +13,97 @@ import scalaz.Scalaz._
...
+   * { e: pairs[0]._1 \in pairs[0]._2 , ..., pairs[n]._1 \in pairs[n]._2 } }}} `pairs` must be nonempty, and all vars
+   * must be unique variable names
+   */
+  def map(e: TBuilderInstruction, pairs: (TBuilderInstruction, TBuilderInstruction)*): TBuilderInstruction =
+    boundVarIntroductionVariadic(unsafeBuilder.map)(e, pairs: _*)
+
+  /**
+   * {{{
+   * { e: pairs[0] \in pairs[1] , ..., pairs[n-1] \in pairs[n] } }}} `pairs` must have even, positive arity, and all vars
+   * must be unique variable names
```
thpani, line 60: Will the linebreak here render correctly?

tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/TestFunBuilder.scala
```diff
@@ -585,4 +659,30 @@ class TestFunBuilder extends BuilderTest {
 
   }
 
+  test("debug") {
```
thpani, line 662: Is this a left-over? I can't tell from the test name what this does.

thpani (approved):
LGTM

I like the side conditions in the Scaladoc.
They would be even more useful if we tagged them with `@param`?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/FunBuilder.scala
```diff
@@ -55,22 +64,23 @@ trait FunBuilder extends UnsafeFunBuilder {
   // APP overload //
   //////////////////
 
-  /** f[x] for any Applicative f */
+  /** {{{f[x]}}} for any Applicative `f` */
```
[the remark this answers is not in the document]
Kukovec: Actually, `Applicative` is a class of operators, not just the English term. Though it should probably be `[[Applicative]]`

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/SetBuilder.scala
```diff
@@ -13,80 +13,97 @@ import scalaz.Scalaz._
...
-
-  /** { mapExpr: x1 \in set1 , ..., xN \in setN }, must have at least 1 var-set pair */
-  def mapMixed(e: TBuilderInstruction, varSetPairs: TBuilderInstruction*): TBuilderInstruction = {
-    require(varSetPairs.size % 2 == 0)
-    val asPairs = varSetPairs.grouped(2).toSeq.map { case Seq(a, b) =>
+    boundVarIntroductionTernary(unsafeBuilder.filter)(x, set, p)
+
+  /**
+   * {{{
+   * { e: pairs[0]._1 \in pairs[0]._2 , ..., pairs[n]._1 \in pairs[n]._2 } }}} `pairs` must be nonempty, and all vars
```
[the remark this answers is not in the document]
Kukovec: Yes, without it the auto-format joins `pairs[n]._2 } }}}` into `pairs[n]._2 }}}}` and then takes the first 3 as `{{{ _ }}}` termination, which doesn't render the trailing `}`. `\}` doesn't help.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/TestFunBuilder.scala
Kukovec: Oh, good catch, forgot to remove that.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/BaseBuilder.scala
```diff
@@ -10,31 +10,32 @@ import at.forsyte.apalache.tla.typecomp.unsafe.UnsafeBaseBuilder
...
+  } yield unsafeBuilder.appOp(opEx, argsExs: _*)
 
-  /** CHOOSE x: p */
+  /** {{{CHOOSE x: p}}} `x` must be a variable name */
   def choose(x: TBuilderInstruction, p: TBuilderInstruction): TBuilderInstruction =
-    boundVarIntroductionBinary(_choose)(x, p)
+    boundVarIntroductionBinary(unsafeBuilder.choose)(x, p)
 
-  /** CHOOSE x \in set: p */
+  /** {{{CHOOSE x \in set: p}}}, `x` must be a variable name */
```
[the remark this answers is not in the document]
Kukovec: Comma won't work, `}}}` line-breaks, so you'd get
```
CHOOSE x \in set : p
, `x` ...
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/SetBuilder.scala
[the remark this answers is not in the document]
Kukovec: Screenshot 2022-07-08 at 14 47 16

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
```diff
@@ -25,7 +25,7 @@ object BuilderUtil {
 
   /** Removes elem from the scope, as the scope only contains types of free variables */
   def markAsBound(elem: TlaEx): TBuilderInternalState[Unit] = State[TBuilderContext, Unit] { mi: TBuilderContext =>
-    require(elem.isInstanceOf[NameEx])
+    require(elem.isInstanceOf[NameEx], s"elem = $elem must be a variable name.")
```
shonfeder, line 28: So say that `elem` were incorrectly an integer expression, am I correct in understanding that this error would read `"elem = TlaInt(1) must be a variable name"`? I don't think I'd now how to parse that out of context. I think i'd be looking for a bad assignment or something.

Would it be clearer if the message were just something like:

```suggestion
    require(elem.isInstanceOf[NameEx], s"$elem must be a variable name.")
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
```diff
@@ -38,8 +38,10 @@ object BuilderUtil {
     )(variable: TBuilderInstruction,
       set: TBuilderInstruction,
       expr: TBuilderInstruction): TBuilderInstruction = for {
+    usedBefore <- allUsed
     setEx <- set
-    usedInSet <- allUsed // variable may not appear as bound or free in set
+    usedInSetOrBefore <- allUsed // variable may not appear as bound or free in set
```
shonfeder, line 43: Why does `allUsed` get bound to two different variables here?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
```diff
@@ -38,8 +38,10 @@ object BuilderUtil {
...
+    usedInSet = usedInSetOrBefore -- usedBefore
```
shonfeder, line 44: This doesn't  always end up empty? Looks to me like we are taking the difference of the same set and itself, giving that both sets come from the same `allUsed`. Am I missing something?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
```diff
@@ -91,18 +93,23 @@ object BuilderUtil {
...
-          // variable_i is shadowed iff boundAfterVar \subseteq usedInSet \union boundAfterBodyEx
-          boundAfterVar <- allBound
-        } yield
-          if (boundAfterVar.subsetOf(usedInSet.union(boundAfterBodyEx))) {
+          // we delay marking as bound, to not interfere with other variable-set pairs
+        } yield {
+          require(varEx.isInstanceOf[NameEx], s"varEx = $varEx must be a variable name.")
+          // variable_i is shadowed iff boundVar \in usedInSet \union boundAfterBodyEx
+          val boundVar = varEx.asInstanceOf[NameEx].name
+          if (usedInSet.union(boundAfterBodyEx).contains(boundVar))
```
shonfeder, line 106: These changes don't seem like they can actually be related to the stated purpose of this PR. iiuc, the PR says it is about making a pretty simple design change, that should mostly be mechanical, but these changes look like pretty subtle adjustments to the business logic of key operations. 

If I am reading this right, could you please refactor this PR into to separate parts? One that makes the design/api change about how the unsafe builders are accessed and one that makes the adjustments to the business logic and errors etc? 

It's a quite large PR, and I think factoring it out into those different parts would make review much easier and more useful.

shonfeder (changes requested):
Sorry to be a pain, but if I'm reading the changes in this PR right, it seems like this should be broken into two different changesets, so I'm requesting that change.  I think that will let us provide a much more useful review on the important changes (the alteration to the construction logic) while just rubber stamping the more trivial change about namespacing etc.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
Kukovec: Because it's a getter of the internal state, which was changed in between by evaluating `set`.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
Kukovec: Yes, the value of `allUsed` dynamically changes with the internal state.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
thpani: I think it makes sense to point this out in the Scaladoc of the shorthands, and to explain it in a short comment here.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
thpani: See above, maybe a comment helps.

shonfeder:
> Maybe we can explain the shorthands (allUsed, ...) a bit better in code, as requested by

So, imo, the fix for this is to make it clear from the name that this is dynamic. E.g., call it `getAllUnused`. IMO, current naming makes this (and similar) look like they should be static values.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
shonfeder: Ah yeah. I had totally missed that comment. My bad.

Ideally these would be separate PRs imo, assuming the overhead is not too great. Otherwise, I agree that updating the pr title description will help clarify for reviewers.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
Kukovec: I can make it `getAllUsed`, np

Kukovec:
> > So, imo, the fix for this is to make it clear from the name that this is dynamic. E.g., call it `getAllUnused`. IMO, current naming makes this (and similar) look like they should be static values.
> 
> I like this suggestion!

Ok, I've made a new issue for this (#1956). I'll close this PR as is and add it in the next iteration.

PR #1957, opened 2022-07-11
Typed Builder implementation Part 11: `LET-IN` 

### LINKED ISSUE #1955: Implement `TypedBuilder` for `LetInEx`
Also resolves TODOs in `TestApalacheOper`.

### LINKED ISSUE #1956: Rename `BuilderUtil.allUsed` and `BuilderUtil.allBound` with a `get-` prefix
Some methods in `BuilderUtil` are getters of internal state, which means that calling them at different time yields different results, e.g.
```scala
for {
A1 <- allUsed
...
A2 <- allUsed // May be different from A1
} yield ...

```
We should rename them with a `get-` prefix and add more expansive documentation, to make it more intuitive.

Kukovec:

closes #1955 
closes #1956 

### DIFF
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
@@ -38,16 +38,16 @@ object BuilderUtil {
     )(variable: TBuilderInstruction,
       set: TBuilderInstruction,
       expr: TBuilderInstruction): TBuilderInstruction = for {
-    usedBefore <- allUsed
+    usedBefore <- getAllUsed
     setEx <- set
-    usedInSetOrBefore <- allUsed // variable may not appear as bound or free in set
+    usedInSetOrBefore <- getAllUsed // variable may not appear as bound or free in set
     usedInSet = usedInSetOrBefore -- usedBefore
     exprEx <- expr
-    boundAfterExpr <- allBound // variable may not appear as bound in expr
+    boundAfterExpr <- getAllBound // variable may not appear as bound in expr
     varEx <- variable
     _ <- markAsBound(varEx)
     // variable is shadowed iff boundAfterVar \subseteq usedInSet \union boundAfrerExpr
-    boundAfterVar <- allBound
+    boundAfterVar <- getAllBound
   } yield {
     val ret = unsafeMethod(varEx, setEx, exprEx)
     if (boundAfterVar.subsetOf(usedInSet.union(boundAfterExpr))) {
@@ -66,11 +66,11 @@ object BuilderUtil {
     )(variable: TBuilderInstruction,
       expr: TBuilderInstruction): TBuilderInstruction = for {
     exprEx <- expr
-    boundAfterExpr <- allBound // variable may not appear as bound in expr
+    boundAfterExpr <- getAllBound // variable may not appear as bound in expr
     varEx <- variable
     _ <- markAsBound(varEx)
     // variable is shadowed iff boundAfterVar \subseteq boundAfterExpr
-    boundAfterVar <- allBound
+    boundAfterVar <- getAllBound
   } yield {
     val ret = unsafeMethod(varEx, exprEx)
     if (boundAfterVar.subsetOf(boundAfterExpr)) {
@@ -88,14 +88,14 @@ object BuilderUtil {
     )(ex: TBuilderInstruction,
       varSetPairs: (TBuilderInstruction, TBuilderInstruction)*): TBuilderInstruction = for {
     bodyEx <- ex
-    boundAfterBodyEx <- allBound // variables may not appear as bound in bodyEx
+    boundAfterBodyEx <- getAllBound // variables may not appear as bound in bodyEx
     pairs <- varSetPairs.foldLeft(Seq.empty[(TlaEx, TlaEx)].point[TBuilderInternalState]) {
       case (cmp, (variable, set)) =>
         for {
           seq <- cmp
-          usedBefore <- allUsed
+          usedBefore <- getAllUsed
           setEx <- set
-          usedInSetOrBefore <- allUsed // variable_i may not appear as bound or free in set_i
+          usedInSetOrBefore <- getAllUsed // variable_i may not appear as bound or free in set_i
           usedInSet = usedInSetOrBefore -- usedBefore
           varEx <- variable
           // we delay marking as bound, to not interfere with other variable-set pairs
@@ -112,12 +112,18 @@ object BuilderUtil {
     _ <- buildSeq(pairs.map(pa => markAsBound(pa._1)))
   } yield unsafeMethod(bodyEx, pairs)
 
-  /** Convenience shorthand to access the set of used names. */
-  def allUsed: TBuilderInternalState[Set[String]] =
+  /**
+   * Convenience shorthand to access the set of used names. May return different values at different places, depending
+   * on the internal state.
+   */
+  def getAllUsed: TBuilderInternalState[Set[String]] =
     gets[TBuilderContext, Set[String]] { _.usedNames }
 
-  /** Convenience shorthand to access the set of bound names. */
-  def allBound: TBuilderInternalState[Set[String]] =
+  /**
+   * Convenience shorthand to access the set of bound names. May return different values at different places, depending
+   * on the internal state.
+   */
+  def getAllBound: TBuilderInternalState[Set[String]] =
     gets[TBuilderContext, Set[String]] { ctx => ctx.usedNames -- ctx.freeNameScope.keySet }
 
   /**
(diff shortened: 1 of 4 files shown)


thpani (changes requested):
I'll echo @shonfeder's previous concerns (https://github.com/informalsystems/apalache/pull/1937#pullrequestreview-1033155210) that this should really be 2 separate, orthogonal PRs.
Can you please cherry-pick those two commits into different branches and open separate PRs?

I'd also appreciate some context in the PR description about what a PR is trying to achieve and how (also useful for future reference, if we have to do archeology).

shonfeder:
I agree with @thpani that it would have be nicer in principle to split this into two separate PRs, and I would support blocking on that PR refactor if this were a large PR. But since it is pretty small anyhow I don't think it's worth the added friction here. 

> I don't really consider it worthwhile to go through all of the overhead of turning one-click refactoring (https://github.com/informalsystems/apalache/issues/1956) into a PR. It also seems like a waste of time for reviewers to have to go through 2 context-switches instead of 1 for reviewing

IMO, the overhead to open a PR should be trivial enough that this isn't much of a concern, and I think we can really improve the quality and velocity of PR reviews by separating automatic changes from the nuanced changes to business logic (and by separating distinct logical changes). Just my 2 cents there.

thpani:
> I agree with @thpani that it would have been nice in principle to split this into two separate PRs, and I would support blocking on that PR refactor if this were a large PR. But since it is pretty small anyhow I don't think it's worth the added friction here.

I'm not concerned at all if small fixes sneak in once in a while, which is why I reviewed #1937 despite the same issue. I'm more concerned about repeated PRs that could've been split, and where reviewing would've been much easier if they had been split.
I shouldn't have blocked this though, if someone else wants to review.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
```diff
@@ -66,11 +66,11 @@ object BuilderUtil {
     )(variable: TBuilderInstruction,
       expr: TBuilderInstruction): TBuilderInstruction = for {
     exprEx <- expr
-    boundAfterExpr <- allBound // variable may not appear as bound in expr
+    boundAfterExpr <- getAllBound // variable may not appear as bound in expr
```
shonfeder, line 69: Much clearer, thank you!

Just to confirm, the call to the `variable` instruction below doesn't update the state?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
```diff
@@ -38,16 +38,16 @@ object BuilderUtil {
     )(variable: TBuilderInstruction,
       set: TBuilderInstruction,
       expr: TBuilderInstruction): TBuilderInstruction = for {
-    usedBefore <- allUsed
+    usedBefore <- getAllUsed
     setEx <- set
```
shonfeder, line 42: So, iiuc, this acccess to `set` is also changing the state, right? Then wouldn't it be clearer to have this something like:

```suggestion
    setEx <- getNextSet
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
```diff
@@ -112,12 +112,18 @@ object BuilderUtil {
...
     gets[TBuilderContext, Set[String]] { _.usedNames }
 
-  /** Convenience shorthand to access the set of bound names. */
-  def allBound: TBuilderInternalState[Set[String]] =
+  /**
+   * Convenience shorthand to access the set of bound names. May return different values at different places, depending
+   * on the internal state.
+   */
+  def getAllBound: TBuilderInternalState[Set[String]] =
     gets[TBuilderContext, Set[String]] { ctx => ctx.usedNames -- ctx.freeNameScope.keySet }
```
shonfeder, line 127: Very helpful. Thank you!

tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/TestHybrid.scala
```diff
@@ -219,6 +219,221 @@ class TestHybrid extends BuilderTest {
...
+    type TParam1 = (TlaType1, DeclParamT)
+    type TParamN = (TlaType1, Seq[DeclParamT])
+
+    val parameterTypeGen: Gen[TlaType1] = for {
+      n <- Gen.choose(1, 5)
+      ts <- Gen.listOfN(n, tt1gen.genPrimitive)
+    } yield ts match {
+      case head :: Nil  => head
+      case head :: tail => OperT1(tail, head)
+      case Nil          => IntT1 // impossible, since 1 <= n <= 5, but the compiler doesn't know and complains
```
shonfeder, line 234: appreciate the message here! I suggest throwing an exception here with the comment as the message instead of returning the type. That will ensure that if this ever does get hit for some reason we'll get the expected error rather than returning an int arbitrarily :)

Kukovec:
> [...] I'm more concerned about repeated PRs [...] where reviewing would've been much easier if they had been split.

That's a fair concern, but I honestly don't think it applies here. The time it takes to review changes to `BuilderUtil` is the time it takes to scroll to the end of the file. This is primarily a PR that implements https://github.com/informalsystems/apalache/issues/1955.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/TestHybrid.scala
```diff
@@ -219,6 +219,221 @@ class TestHybrid extends BuilderTest {
...
+
+    checkRun[TParam1](run(mkWellTyped1, forceScopeException1, isExpected1))
+    checkRun[TParamN](run(mkWellTypedN, forceScopeExceptionN, isExpectedN))
+
+    // throws on n = 0
+    assertThrows[IllegalArgumentException] {
+      builder.letIn(builder.int(0))
+    }
+
+  }
```
shonfeder, line 435: This is a very big test, imo, and i cannot  tell by reading this part what properties or behaviors are actually being tested. I know that the test is testing `let-in`, but other than than I can't tell what kind of things are mean to hold here. 

I suggest either refactoring the test into a few distinct property tests that clearly communicate and exemplify what is being tested or adding comments here that explain what is being tested in each chunk. The former is much more preferable, IMO, but I'd be fine with just the latter.

shonfeder (changes requested):
Looking good! I have a few questions, and at least two suggestions I'd like to see addressed before merging.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
Kukovec: All `TBuilderInstruction`s "change state". The reason the parameters are named the way they are is to keep the intuition that, e.g. `builder.exists(x, S, p)` creates a `OperEx(exists, y, T, q)`-like expression, as not to bleed implementation details into the interface.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/TestHybrid.scala
Kukovec: That's fair, I'll do that.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
Kukovec: It does, they all do.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/BuilderUtil.scala
shonfeder: This makes sense. So the set/get prefix can be saved just for the utilities that inspect the environment or update it, not for the operators that create the expressions. Thanks for explaining.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/ScopedBuilder.scala
```diff
@@ -171,6 +171,41 @@ class ScopedBuilder
...
+    usedAfterDecl <- getAllUsed
+    usedInDecl = usedAfterDecl -- usedBeforeDecl
+    bodyEx <- body
+    boundAfterBody <- getAllBound // decl may not appear as bound in body
+    boundInBody = boundAfterBody -- usedAfterDecl
+    declName <- name(declEx.name, declEx.typeTag.asTlaType1()) // puts name in scope w/ type
+    _ <- markAsBound(declName)
+  } yield {
+    if ((usedInDecl ++ boundInBody).contains(declEx.name))
+      throw new TBuilderScopeException(s"Declaration name ${declEx.name} is shadowed in ${declEx.body} or $bodyEx.")
```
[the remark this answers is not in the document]
shonfeder: I think that’s worth doing here.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/TestHybrid.scala
Kukovec: I've split the tests and added comments, most notably to the method `runLetTests`.

shonfeder (approved):
LGTM. Thanks!

PR #1959, opened 2022-07-11
Set a unique java.io.tmpdir for each call to SANY

shonfeder:
This should provide a workaround for the concurrency bug noted
https://github.com/tlaplus/tlaplus/issues/688, while we wait on an upstream fix.

Note that this only allows concurrent executions of SANY in parallel processes.
AFAICT, the deeper issue noted at
https://github.com/informalsystems/apalache/issues/1114#issuecomment-1180534894
will still prevent concurrent runs of SANY within the same JVM.

-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
@@ -70,7 +70,6 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
 
     // Set a unique tmpdir to avoid race-condition in SANY
     // TODO: RM once https://github.com/tlaplus/tlaplus/issues/688 is fixed
-    //       see
     System.setProperty("java.io.tmpdir", sanyTempDir().toString())
 
     // call SANY
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
@@ -68,8 +68,14 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
     // Resolver for filenames, patched for wired modules.
     val filenameResolver = SanyNameToStream(libraryPaths)
 
+    // Set a unique tmpdir to avoid race-condition in SANY
+    // TODO: RM once https://github.com/tlaplus/tlaplus/issues/688 is fixed
+    //       see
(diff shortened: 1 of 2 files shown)


tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -68,8 +68,14 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
     // Resolver for filenames, patched for wired modules.
     val filenameResolver = SanyNameToStream(libraryPaths)
 
+    // Set a unique tmpdir to avoid race-condition in SANY
+    // TODO: RM once https://github.com/tlaplus/tlaplus/issues/688 is fixed
+    //       see
```
thpani, line 73: I think the second line should either go or reference an Apalache issue?

```suggestion
    // TODO: RM once https://github.com/tlaplus/tlaplus/issues/688 is fixed
```

thpani (approved):
Just want to type this out, to wrap my head around it:

If loading from string sources, this will create `$TMPDIR/sanyimpXXX` and write the string sources to that directory. Then it sets `java.io.tmpdir` to `$TMPDIR/sanyimpYYY` (where `YYY` != `XXX`) and calls SANY, which writes its temporary files to the latter directory.

We're not doing the first part if loading directly from files, right?

If this is all accurate, feel free to merge :)

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
konnov: It looks like [createTempDirectory](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory(java.lang.String,%20java.nio.file.attribute.FileAttribute...)) does not need a pattern, just a prefix.

thpani: > It looks like [createTempDirectory](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory(java.lang.String,%20java.nio.file.attribute.FileAttribute...)) does not need a pattern, just a prefix.

If this is in response to my comment above, I meant `XXX` and `YYY` to be wildcards, not patterns to `createTempDirectory`.

shonfeder:
That is all accurate! One temp
dir is always created for SANY to use. And one is only created if we need to write strings into files to load from, and the two dots are different.

Kukovec (approved):

PR #1977, opened 2022-07-13
Introduce new syntax for type aliases

### LINKED ISSUE #1843: Change the syntax for type aliases
Our type checker uses the same syntax for type aliases and uninterpreted types, e.g., `PROC_SET`. It often leads to confusion, especially, when the users make a typo in the alias name.

We should introduce different syntax for uninterpreted types and type aliases. For instance, uninterpreted types should be in upper case, whereas type aliases should be in camel case.

konnov:
Closes #1843. As we have found, the users are confused by missing type aliases, which default to uninterpreted types. This PR introduces new syntax for type aliases:

```tla
\* @typeAlias: newAlias = Set(Int);
\* @type: Set($newAlias);
Foo == { 0 }
```

The benefit of having special syntax for type aliases and references is that the type checker can now complain about missing type aliases. Compare `TestAliasOld.tla` to `TestAliasNew.tla` and `TestAliasNewMissing.tla`.

The documentation is postponed for a follow-up PR.


tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
```diff
@@ -96,7 +95,13 @@ class TypeCheckerTool(annotationStore: AnnotationStore, inferPoly: Boolean, useR
         annotations.filter(_.name == StandardAnnotations.TYPE_ALIAS).foreach { annot =>
           annot.args match {
             case Seq(AnnotationStr(text)) =>
-              aliases += parseTypeAlias(decl.name, text)
+              val (alias, assignedType) = parseTypeAlias(decl.name, text)
+              if (!ConstT1.isAliasReference(alias)) {
+                val msg =
+                  s"Operator ${decl.name}: Deprecated syntax in type alias $alias. Use camlCase of Type System 1.2."
```
shonfeder, line 101: Why do we have enforce anything about the case of the letters if we are requiring a `$` to reference the alias? Shouldn't the case be irrelevant then?

shonfeder (commented):
One question posed about the limit on cases.

I also have a proposal for an alternative solution: why not require users declare uninterpreted types up front with an annotation? It seems that this be a more robust solution, as it would also prevent typos in uninterpreted types. 

As an aside, I happen to think dollar signs are ugly :stuck_out_tongue_closed_eyes:

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
konnov: In principle, yes, but we still have the all-caps syntax for old aliases and uninterpreted types. I just wanted some disambiguation here.

konnov:
> One question posed about the limit on cases.

@shonfeder, I do not see a question. Is it lost in the Github review process? :)

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
Kukovec, line 101:
```suggestion
                  s"Operator ${decl.name}: Deprecated syntax in type alias $alias. Use camelCase of Type System 1.2."
```

tla-io/src/main/scala/at/forsyte/apalache/io/typecheck/parser/DefaultType1Parser.scala
```diff
@@ -285,6 +295,22 @@ object DefaultType1Parser extends Parsers with Type1Parser {
 
   // a type constant or an alias name, e.g., BAZ
   private def typeConst: Parser[ConstT1] = {
-    acceptMatch("type constant", { case IDENT(name) if (name.toUpperCase() == name) => ConstT1(name) })
+    acceptMatch("type constant", { case IDENT(name) if name.toUpperCase() == name => ConstT1(name) })
+  }
+
+  // old ALIAS_SYNTAX, or newSyntaxInCamelCase
+  private def oldOrNewAlias: Parser[ConstT1] = {
```
Kukovec, line 302: It looks like we still match the old alias syntax and just show a deprecation message instead of failing. Shouldn't we just pick 1 syntax and go with that, instead of some pseudo backwards-compatibility? We're not even in alpha, we shouldn't be beholden to old syntax.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeAliasSubstitution.scala
```diff
@@ -1,32 +1,38 @@
...
-}
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.typecheck.TypingInputException
 
 /**
  * A substitution from constant names to types. It is very similar to Substitution. However, `ConstSubstitution` is
- * meant to replace constant types, e.g., ENTRY, with a concrete type, whereas `Substitution` replaces variables. We use
- * `ConstSubstitution` to implement aliases.
+ * meant to replace constant types, e.g., ENTRY or $entryInCamlCase, with a concrete type, whereas `Substitution`
+ * replaces variables. We use `ConstSubstitution` to implement rewrite type aliases.
```
Kukovec, line 9:
```suggestion
 * replaces variables. We use `ConstSubstitution` to rewrite type aliases.
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeAliasSubstitution.scala
```diff
@@ -1,32 +1,38 @@
...
-    ConstSubstitution.mk(context)(tp)
+    TypeAliasSubstitution.mk(context)(tp)
   }
 
-  def closure(): ConstSubstitution = {
+  /**
+   * Recursively apply type aliases to all other type aliases in [[context]].
+   *
+   * @throws TypingInputException
+   *   if found undefined type aliases.
```
Kukovec, line 26:
```suggestion
   *   if undefined type aliases are found.
```

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeAliasSubstitution.scala
```diff
@@ -133,7 +142,16 @@ object ConstSubstitution {
           if (fun.isDefinedAt(name)) {
             (fun(name), true)
           } else {
-            (tp, false)
+            // The behavior for an OLD_TYPE_ALIAS is to understand it as an uninterpreted type.
+            // Since in the old syntax, we cannot distinguish between uninterpreted types and type aliases,
+            // we cannot do much here.
```
Kukovec, line 147: I wrote the previous comment, and up until this line it seemed to be the case, from reading the docs, that we support both syntaxes. In particular https://github.com/informalsystems/apalache/blob/ecd5b7188c211b8b7e3f6440f7ecefda4daa420d/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeAliasSubstitution.scala#L8
mentions `ENTRY` or `$entryInCamlCase`.

But these lines imply that we don't support old aliases anymore. This is a bit confusing and it should be much clearer whether we support old aliases or not (and in what way). Semantically, I don't think treating caps types as uninterpreted means that we support "old aliases" (and the Scala variable names should reflect this).

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
```diff
@@ -43,8 +44,18 @@ class TestTypeCheckerTool extends AnyFunSuite with BeforeAndAfterEach with EasyM
...
+    System.getenv("APALACHE_HOME") match {
+      // Warn if environment variable APALACHE_HOME is not set
+      case null =>
+        logger.error("Not running from fat JAR and APALACHE_HOME is not set.")
+        logger.error("Set APALACHE_HOME to a directory where Apalache has been checked out.")
+        throw new IllegalStateException("Missing APALACHE_HOME to run the tests")
+
+      case apalacheHome: String =>
+        Source.fromFile(s"$apalacheHome/tla-types/src/test/resources/$name.tla")
+    }
```
Kukovec, line 58: I get complaints about unrelated changes added into PRs, so I'm going to start complaining about unrelated changes added to other people's PRs. Consider creating a separate issue for this and factoring this into a separate PR.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
```diff
@@ -98,19 +98,69 @@ case object StrT1 extends TlaType1 {
...
 
   override def toString: String = name
 
   override def usedNames: Set[Int] = Set.empty
+}
+
+/**
+ * A companion object for [[ConstT1]].
+ */
+object ConstT1 {
```
Kukovec, line 121: Not something I necessarily want now, but consider combining this class with `ModelValueHandler` (more accurately, moving methods from that class to here).
We don't want 2 unrelated classes handling utility for uninterpreted literals.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
```diff
@@ -98,19 +98,69 @@ case object StrT1 extends TlaType1 {
...
+   * Does this type represent an uninterpreted type, e.g., PROCESS. Outside of the type parser and the type checker,
+   * this method should always return true.
+   *
+   * @param name
+   *   type name
+   * @return
+   *   true iff the type name represents an uninterpreted type.
+   */
+  def isUninterpreted(name: String): Boolean = {
+    name.forall(c => c.isUpper || c.isDigit || c == '_')
```
Kukovec, line 133: Why don't you use regex here?

Kukovec: Also, this is incorrect, because you can start your value name with `[0-9]` and the forall test would pass.
The regex is `[A-Z_][A-Z0-9_]` (see `ModelValueHandler.matchRegex`)

tla-io/src/main/scala/at/forsyte/apalache/io/typecheck/parser/DefaultType1Parser.scala
konnov: Yes, we will keep the old syntax for the same transition period, as we do for the old records. We should give the users a bit of time to switch to the new syntax.

tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
konnov: It is related, as I had to run the tests, and they were not runnable in Idea, as we had two different solutions to dealing with resources.

tla-io/src/main/scala/at/forsyte/apalache/io/typecheck/parser/DefaultType1Parser.scala
Kukovec: Why though? They're going to have to change it eventually anyway. It'll be just as much of an issue for users then as it would be now, except we get the added downside of maintenance burden in the meantime.

shonfeder:
> > One question posed about the limit on cases.
> 
> @shonfeder, I do not see a question. Is it lost in the Github review process? :)

You answered my question already! Thanks :)

tla-io/src/main/scala/at/forsyte/apalache/io/typecheck/parser/DefaultType1Parser.scala
konnov: Because we don't want angry users? 🤷‍♂️

tla-io/src/main/scala/at/forsyte/apalache/io/typecheck/parser/DefaultType1Parser.scala
```diff
@@ -259,6 +264,11 @@ object DefaultType1Parser extends Parsers with Type1Parser {
     }
   }
 
+  // a reference to an alias
+  private def aliasReference: Parser[TlaType1] = {
+    DOLLAR() ~> newAlias ^^ { n => n }
```
shonfeder, line 269: I believe `_ ^^ { n => n}`` is equivalent to `_.map(x => x)` in the context of the parser combinators, which means it's a no-op, and should be equivalent to:

```suggestion
    DOLLAR() ~> newAlias
```

shonfeder (approved):
Outside of my general reservations and general agreement with Jure's suggestions, LGTM.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
shonfeder: fwiw, I don't see the value in applying this unneeded constraint.

tla-io/src/main/scala/at/forsyte/apalache/io/typecheck/parser/DefaultType1Parser.scala
Kukovec: Why would they be angry now, but not in, say, 1 month when this would become the new standard?

shonfeder:
> If we start requiring type declarations for uninterpreted types, this will make life of specification writers even harder.

I don't not see how requiring a single line delcaring an uninterpreted type makes the life of a specification writer than requiring them to prefix every instance of an alias with `$`. The latter seems more onerous.

Moreover, the slight inconvenience of adding a single line declaration seems more than compensated by helping spec writers ensure they don't introduce spurious errors with typos in their uninterpreted types.

Perhaps I am missing something?

konnov:
> I don't not see how requiring a single line delcaring an uninterpreted type makes the life of a specification writer than requiring them to prefix every instance of an alias with $. The latter seems more onerous.

The main problem is that annotations in comments are still fragile, so I am trying to minimize the number of annotations in general. The reason for having `$` is that the type checker can now complain about a missing type alias. With the current implementation, the type checker silently assumes that the user meant an uninterpreted type. So yes, the user can make an error in a reference to a type alias, or completely forget to write `$`, but in this case they will get a type checker error. This is the most critical point for me here.

konnov:
> Moreover, the slight inconvenience of adding a single line declaration seems more than compensated by helping spec writers ensure they don't introduce spurious errors with typos in their uninterpreted types.

The way modules work in TLA+, this will introduce more confusion than it would help. You will have to be careful about explicitly declaring an uninterpreted type across different modules. We are already doing that with type aliases, so it is not impossible. However, I do not see any value of doing that for uninterpreted types. If you make a typo in the name of an uninterpreted type, you will get a type checking error.

konnov:
Just to clarify the context here. We have four kinds of identifiers in the type grammar:

 - Uninterpreted types that are currently written in `ALL_CAPS`.
 - Variable names that match the regex `a[0-9]+|[a-z]`.
 - Aliases, which currently coincide with uninterpreted types.
 - Type constructors such as `Set` and `Seq`.

So my proposal here is to disambiguate aliases by:

 - Enforcing caml case starting with a small letter in their names, so they do not conflict with uninterpreted types and type constructors.
 - Refer to type aliases via `$`, so this syntax does not conflict with type variables.

I also do not like the dollar syntax. However, this syntax is not really surprising. Type aliases are simply substituted in the type checker, they are not types of their own.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeAliasSubstitution.scala
konnov: What I meant by "we cannot do much here" is that we cannot detect an error. I think I will fix the comment.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
konnov: Fixing the regular expression. I will open an issue about ModelValueHandler.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
konnov: I have opened a follow-up issue: https://github.com/informalsystems/apalache/issues/1992#issue-1307695067

konnov:
@Kukovec I think I took all of your comments into account. Would you like to approve it now?

tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/TypeAliasSubstitution.scala
```diff
@@ -1,32 +1,38 @@
...
-}
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.typecheck.TypingInputException
 
 /**
  * A substitution from constant names to types. It is very similar to Substitution. However, `ConstSubstitution` is
- * meant to replace constant types, e.g., ENTRY, with a concrete type, whereas `Substitution` replaces variables. We use
- * `ConstSubstitution` to implement aliases.
+ * meant to replace constant types, e.g., ENTRY or $entryInCamlCase, with a concrete type, whereas `Substitution`
+ * replaces variables. We use `ConstSubstitution` to rewrite type aliases.
```
thpani, line 9: Also need to rename the class here :)
(Unfortunately I can't mark the unchanged line above to make a suggestion)

```
 * A substitution from constant names to types. It is very similar to [[Substitution]]. However,
 * [[TypeAliasSubstitution]] is meant to replace constant types, e.g., ENTRY or $entryInCamlCase, with a concrete type,
 * whereas [[Substitution]] replaces variables. We use [[TypeAliasSubstitution]]` to rewrite type aliases.
```

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
```diff
@@ -98,19 +98,71 @@ case object StrT1 extends TlaType1 {
 }
 
 /**
- * An uninterpreted type constant.
+ * <p>An uninterpreted type constant such as PROC_NAME.</p>
+ *
+ * <p>Inside the type checker, this class may also represent references to type aliases. Do not use this feature outside
+ * of the type checker.</p>
```
thpani, line 104: I think Scaladoc infers the markup here

```suggestion
 * An uninterpreted type constant such as PROC_NAME.
 *
 * Inside the type checker, this class may also represent references to type aliases. Do not use this feature outside
 * of the type checker.
```

thpani (approved):

Kukovec (approved):

konnov:
I think I have implemented all the comments. Thanks for the reviews! Will merge it now.

PR #1981, opened 2022-07-14
Typed Builder implementation part 13.2: Clean up generators and lifter methods

### LINKED ISSUE #1966: Clean up generators and lifter methods in `BuilderTest`
Many tests define their own `Gen[TParam]`, despite the number of different `TParam` instances being relatively low overall. 
We should collect predefined generators in a sub-object within `BuilderTest`, then make all generators explicit, to avoid collisions.
Similarly, we should predefine lifter methods used to instantiate `toSeq` in `expectEqTyped`.

Kukovec:

closes #1966

Major changes:
 - All `Gen[TParam]` instances removed from individual tests and collected in `BuilderTest.Generators`
 - Ad-hoc implementations of `toSeq` values passed to `expectEqTyped` replaced with standard implementations in `BuilderTest.ToSeq`
 - `checkRun` now takes explicit generators; all generators are defined as explicit

Minor changes:
 - Normalized builder methods/generators to use `Int` instead of `BigInt` (inputs come from users, not TLA sources)
 - Changed `notSupportedByModelChecker` tests to use the standard framework.


tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/LiteralAndNameBuilder.scala
```diff
@@ -16,7 +16,7 @@ trait LiteralAndNameBuilder {
   private val unsafeBuilder = new UnsafeLiteralAndNameBuilder
 
   /** {{{i : Int}}} */
-  def int(i: BigInt): TBuilderInstruction = unsafeBuilder.int(i).point[TBuilderInternalState]
+  def int(i: Int): TBuilderInstruction = unsafeBuilder.int(i).point[TBuilderInternalState]
```
konnov, line 19: What is the reason for this change? We have put some effort to support big integers, instead of just integers. There is a practical need for that, as Cosmos and Ethereum contracts operate over big integers. If we use the typed builder to import specs in JSON (as we are planning to do), this interface will become a problem.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/unsafe/UnsafeApalacheBuilder.scala
```diff
@@ -14,7 +14,7 @@ class UnsafeApalacheBuilder extends ProtoBuilder {
 
   // We borrow the LiteralBuilder to make TLA integers from Scala ints
   private val intBuilder = new UnsafeLiteralAndNameBuilder
-  private def mkTlaInt: BigInt => TlaEx = intBuilder.int
+  private def mkTlaInt: Int => TlaEx = intBuilder.int
```
konnov, line 17: Again, I don't see a reason for switching back to 32-bit integers.

konnov (commented):
Most of it looks fine, just refactoring. However, I do not like that we are going to make 32-bit integers the default instead of big integers.

tla-types/src/main/scala/at/forsyte/apalache/tla/typecomp/subbuilder/LiteralAndNameBuilder.scala
Kukovec: This had more to do with generators and tests, where (`a until b`) didn't work with big integers and had to be cast. But you're right, I was thinking about how it's not a big deal since the builder was only ever called programmatically, but I suppose we need this if we want to use the builder in JSON importing. I'll change this back.

konnov (approved):
This looks good now!

PR #1982, opened 2022-07-15
Highlight some TLA+ built-in operators

### LINKED ISSUE #478: [FEATURE] Add syntax highlighting to manual
While mdbook is nice, one disadvantage compared to the markdown display on github is that the TLA+ syntax highlighting is lost.

This might not be possible without a 3rd party plugin to highlight.js, some previous work is at https://github.com/highlightjs/highlight.js/pull/1658

thpani:
Follow-up to the syntax-highlighter in #1972:

Define some built-in TLA+ operators.
Also defines an alias language specifier `'tla'` inside the language spec, so we don't have to do it manually.

Closes #478


konnov (approved):
This looks absolutely amazing :)

PR #1983, opened 2022-07-15
 Typed Builder implementation part 13.3: Standardize early tests

### LINKED ISSUE #1975: Port early `ScopedBuilder` tests to the new testing framework
Some of the tests from #1634 still use ad-hoc testing instead of the `BuilderTest` framework and should be standardized.

Kukovec:

closes #1975 


tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/BuilderTest.scala
```diff
@@ -51,9 +51,24 @@ trait BuilderTest extends AnyFunSuite with BeforeAndAfter with Checkers with App
...
     def minIntGen(min: Int) = Gen.choose(min, min + 10)
 
+    val intGen: Gen[Int] = minIntGen(-5)
     val positiveIntGen: Gen[Int] = minIntGen(1)
     val nonnegativeIntGen: Gen[Int] = minIntGen(0)
 
+    val boolGen: Gen[Boolean] = Gen.oneOf(true, false)
+
+    // empty strings mess up regex matching for uninterpreted literals
+    val strGen: Gen[String] = Gen.alphaStr.suchThat(_.nonEmpty)
```
thpani, line 61: Where is the regex matching this comment is talking about?
Could you add a pointer to that place?

thpani, line 61: I think we can name this more aptly

```suggestion
    val nonEmptyStrGen: Gen[String] = Gen.alphaStr.suchThat(_.nonEmpty)
```

thpani (approved):
I haven't read all the tests in detail, but LGTM

tla-types/src/test/scala/at/forsyte/apalache/tla/typecomp/BuilderTest.scala
Kukovec: `ModelValueHandler.matchRegex` which governs uninterpreted literals is defined as `([...]+)_OF_([...]+)`, so `_OF_` doesn't get parsed as `"": ConstT1("")`, but is treated as a string value. 
The question is, do we want to change this to allow empty uninterpreted literal identifiers and type? In principal, the backend should be fine with it, the problem is you can't explicitly declare the empty type:
```
Null == "_OF_" \* Should be OK

A == LET
  \* @type: Set(); <-- should fail 
  S == {Null}
  IN ...
```

PR #1990, opened 2022-07-18
Rename `unstable` to `main` throughout codebase

### LINKED ISSUE #1012: [FEATURE] Rename `unstable` branch to `trunk`
The unstable branch is pretty stable now, and not really an accurate reflection of the quality of the code. It'll take a bit of work to fix this everywhere, but probably worth it just for the sake of communicating the reality of our maturity to possible users and contributors.

shonfeder:
Closes #1012

**NOTE**: This PR is into the new branch named `main`, which will be made the default once this PR is is approved by the team.

## Overview

The trunk of our development has been called `unstable` for a long time.
This is a holdover from before we adopted a trunk-based development
model, and from before we had CI/CD running for the trunk. Now, the
trunk is really not very unstable, and stability is given by versioned
releases, rather than by maintaining long-lived branches in parallel.

After this change, we should be able to reconfigure the default branch.

## Notable changes

- Renamed all mention of the `unstable` branch to `main`
- Updated docs on installing from docker 
 
## Reviewing

With the exception of the update to the docker installation instructions and the changelog entry (both isolated in their own commits), the change effected here was purely mechanical, with the exception of the changelog entry. It couldn't hurt to have some sanity checks that I've not missed anything or altered anything I shouldn't have.

---

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,4 +5,4 @@
 - [ ] Documentation added for any new functionality
 - [ ] Entries added to [./unreleased/][unreleased] for any new functionality
 
-[unreleased]: https://github.com/informalsystems/apalache/tree/unstable/.unreleased
+[unreleased]: https://github.com/informalsystems/apalache/tree/main/.unreleased
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -2,12 +2,12 @@
 
 ## [./main.yml](./main.yml)
 
-- Triggered on pull requests into `unstable`.
+- Triggered on pull requests into `main`.
 - Our primary build and test workflow.
 
 ## [./deploy.yml](./deploy.yml)
 
-- Triggered by pull requests into `unstable`.
+- Triggered by pull requests into `main`.
 - Used for any artifacts that we deploy into production environments. Currently,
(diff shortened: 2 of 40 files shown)


Kukovec (approved):
Question, what happens to all the PRs in flight? Does GH know to change "XYZ wants to merge N commits into `unstable` from `branch`" or do they break?

konnov (approved):
Good idea! I don't see any major issues with that, except that we would have to manually retarget active pull requests against `trunk`.

thpani (commented):
I support moving away from `unstable`, but isn't `trunk` a very Subversion-y name?
Both Git and GitHub use `main` as the default branch name.

thpani:
We also already use `main` on the Cloud and TNT repos.

shonfeder:
> We also already use `main` on the Cloud and TNT repos.

Ah good point. When the change away from master was happing, I configured my git config to use trunk, and I prefer it cause it fits with the branching metaphor, but forgot that it is less common. I’ll change this to `main`.

shonfeder:
Changed this to target `main` and to switch out the development branch name to `main`.

thpani (approved):
👍

PR #1997, opened 2022-07-19
Add sbt-unicode dependency and build scaladoc in CI

### LINKED ISSUE #1911: Generate scaladoc of unstable
As we discussed today, we are lacking internal developer's documentation. Although the methods in the new code are usually well-documented, there is no good overview of each package. Additionally, the documentation is written for someone who would maintain the code, not use the code as an API. That is, there is an implicit assumption that the developers of a package have to read its codebase, even if they do not want to change it.

We should improve this situation. The most economic way is probably to write the developer's documentation in javadoc itself. We can draw some inspiration from Java's javadoc, e.g., [collections](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html).

The next question is how we should publish the api doc. We can add a target to `sbt`. In my experience, however, developers would not call this target just to read the docs. IMO, a more effective way is to publish the javadoc of unstable somewhere, e.g., at github pages. What do you think @shonfeder?

shonfeder:
Closes #1911

## Overview

We want to begin running the scaladoc build in our CI and enable devs to easily
build and view the scaladoc rendering of our API docs. This change gets us part
way there. Further work to streamline building and viewing the docs will follow,
along with developer documentation.

## Changes

- Adds the `sbt-unidoc` plugin to generate a single site from all our docs
- Fixes a number of scaladoc errors
- Runs unidoc in our CI builds

## Reviewing

Probably easiest to go by commit, as most of the scaladoc error fixes can
probably be skimmed. If anyone can advise on how to get some of the stubborn
references working correctly, I'd appreciate the tips.

To build the docs locally to see what they look like, you can run

``` sh
sbt unidoc
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -244,7 +244,7 @@ lazy val apalacheCurrentPackage = taskKey[File]("Set the current executable apal
 
 // Define the main entrypoint and uber jar package
 lazy val root = (project in file("."))
-  .enablePlugins(UniversalPlugin, sbtdocker.DockerPlugin, ChangelingPlugin)
+  .enablePlugins(UniversalPlugin, sbtdocker.DockerPlugin, ChangelingPlugin, ScalaUnidocPlugin)
   .dependsOn(distribution)
   .aggregate(
       // propagate commands to these sub-projects
@@ -261,6 +261,9 @@ lazy val root = (project in file("."))
   )
   .settings(
       testSettings,
+      // TODO: uncomment to enable building unidoc for for all test and src code
+      // Generate scaladoc for both test and src code
+      // ScalaUnidoc / unidoc / unidocConfigurationFilter := inConfigurations(Compile, Test),
       // Package definition
       Compile / packageBin / mappings ++= Seq(
           // Include theese assets in the compiled package at the specified locations
--- a/mod-infra/src/main/scala/at/forsyte/apalache/infra/exceptions.scala
+++ b/mod-infra/src/main/scala/at/forsyte/apalache/infra/exceptions.scala
@@ -16,7 +16,7 @@ class PassExecException(message: String) extends Exception(message)
 class PassOptionException(message: String) extends Exception(message)
 
 /**
- * An error that has been adapted by the [[ExceptionAdaptor]]
+ * An error that has been adapted by a [[ExceptionAdapter]]
  * @param err
  *   the [[ErrorMessage]] into which the originating error was adapted
  */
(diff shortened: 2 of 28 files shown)


build.sbt
```diff
@@ -261,6 +261,9 @@ lazy val root = (project in file("."))
   )
   .settings(
       testSettings,
+      // TODO: uncomment to enable building unidoc for for all test and src code
+      // Generate scaladoc for both test and src code
+      // ScalaUnidoc / unidoc / unidocConfigurationFilter := inConfigurations(Compile, Test),
```
shonfeder, line 266: This will configure unidoc to build scaladoc for all our tests and the source code (assuming this what we want). But there are many failures in the tests, so I'd like to first get the scaladoc build for our src into CI to prevent additional breakage from being added, and then work on cleaning up the problems in the tests.

konnov:
> Add sbt-unicode dependency and build scaladoc in CI

Just to make sure, it is `sbt-unidoc`, not `sbt-unicode`?

konnov (approved):
That's great! I think it is the right approach to broken references: First replace them with `...` and then gradually repair them.

mod-infra/src/main/scala/at/forsyte/apalache/infra/exceptions.scala
```diff
@@ -16,7 +16,7 @@ class PassExecException(message: String) extends Exception(message)
 class PassOptionException(message: String) extends Exception(message)
 
 /**
- * An error that has been adapted by the [[ExceptionAdaptor]]
+ * An error that has been adapted by a [[ExceptionAdapter]]
```
thpani, line 19:
```suggestion
 * An error that has been adapted by an [[ExceptionAdapter]]
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -12,7 +12,7 @@ import at.forsyte.apalache.tla.lir.values._
  * After Apalache's type-checking, we can rewrite some expressions to simpler forms. For example, the (after
  * type-checking) vacuously true `x \in BOOLEAN` is rewritten to `TRUE` (as `x` must be a `BoolT1`).
  *
- * We currently perform the following simplifications (for type-defining sets TDS, see [[isTypeDefining]]):
+ * We currently perform the following simplifications:
```
thpani, line 15: We use the abbreviation TDS below, so we should keep this here:

```suggestion
 * We currently perform the following simplifications (for a definition of type-defining sets (TDS), see the private method `isTypeDefining`):
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/SetMembershipSimplifier.scala
```diff
@@ -24,6 +24,7 @@ import at.forsyte.apalache.tla.lir.values._
  * @author
  *   Thomas Pani
  */
+// For type-defining sets TDS, see the private method `isTypeDefining`
```
thpani, line 27:
```suggestion
```

thpani (approved):
Thank you for doing all the cleanup of Scaladoc errors! 💐

LGTM

PR #1999, opened 2022-07-19
Include sha256sums of release artifacts

### LINKED ISSUE #1998: Include artifact checksums in releases
At [Modelator](https://github.com/informalsystems/modelator), we are working on [a PR to pull the JAR files from Github releases](https://github.com/informalsystems/modelator/pull/223).

We want to use checksums for integrity verification.

We can manually compute the checksum. but having it as part of each Apalache release would be nice :slightly_smiling_face:

rnbguy:
Closes #1998 


script/release-publish.sh
```diff
@@ -47,6 +47,12 @@ ZIPF_NO_VER="target/universal/apalache.zip"
 TGZF="target/universal/apalache-${VERSION}.tgz"
 TGZF_NO_VER="target/universal/apalache.tgz"
 
+cd target/universal
+sha256sum apalache.* > sha256sum.txt
+cd -
+
+SHA256F="target/universal/sha256sum.txt"
+
```
shonfeder, line 55: Could this be simplified a bit by using the variables defined so dir changes aren't needed? E.g.,

```suggestion
SHA256F="target/universal/sha256sum.txt"
sha256sum "$ZIPF_NO_VER" "$TGZF_NO_VER" > "$SHA256F"
```

What do you think?

script/release-publish.sh
```diff
@@ -60,4 +66,5 @@ hub release create \
     --attach="$ZIPF" --attach="$ZIPF_NO_VER" \
     --attach="$TGZF" --attach="$TGZF_NO_VER" \
     --message="$TAG_NAME" --message="$body" \
+    --attach="$SHA256F" --message="$(cat \"$SHA256F\")" \
```
shonfeder, line 69: attachments and messages together:

```suggestion
    --attach="$TGZF" --attach="$TGZF_NO_VER" --attach="$SHA256F" \
    --message="$TAG_NAME" --message="$body" --message="$(cat \"$SHA256F\")" \
```

Incidentally, why do we need to he hash both as an attachment and as a message in the release notes? (just curious)

shonfeder (approved):
Thanks for the PR! :D

script/release-publish.sh
rnbguy: 1. Attachments and messages together look good. :+1:
2. I have seen some releases in the wild to be like this.

Example: https://github.com/cosmos/gaia/releases/latest
They include `build_report` in the release body.

Probably people prefer copying into clipboard from the browser rather than from a separate downloaded file.

rnbguy: I also wanted to ask your opinion if I should change `--message="$(cat \"$SHA256F\")"` to `checksums=$(cat "$SHA256F")` and `--message="$checksums"`.

script/release-publish.sh
rnbguy: I avoided doing this because the output will include the given path, not just `apalache.tgz` or `apalache.zip`.

```
$ sha256sum target/universal/apalache.zip target/universal/apalache.tgz 
7dd22045cea14cedbfccfc48f7a3e09233126f270f34f4e5f6667e55622e286d  target/universal/apalache.zip
c18b054f228a4bd7e3a96960b316c04d3255274c8cdf921a136d2b7e37062685  target/universal/apalache.tgz
```

Do you have any other ideas to avoid this?

script/release-publish.sh
shonfeder: I'm OK with the inline cat here. But would be ok with factoring that into a temp variable too if you preferred.

I'm not sure of what the benefit is here tho. You have the checksum available either way, but including it in the release message seems to clutter human-readible communications with machine readable info. I'd prefer we just upload the file and not clutter or release notes.

script/release-publish.sh
shonfeder: Ah, I see. I'm OK with keeping it as is then.

thpani: Makes sense, but should we also report hashes of the the versioned bundles?

thpani: I'd still clean up the shafile path by moving the declaration of `SHA256F` to the top as suggested by Shon. Perhaps running in a subshell is the most concise:

```suggestion
SHA256F="target/universal/sha256sum.txt"
(cd target/universal && sha256sum apalache.* > "$SHA256F")
```

script/release-publish.sh
thpani: > but including it in the release message seems to clutter human-readible communications with machine readable info

Agreed, I also can't think of a compelling reason to have it inline.
I'd leave it out for now, we can always paste it on if we discover a use case.

script/release-publish.sh
shonfeder: I like this!

shonfeder (changes requested):
Looks like just a couple changes pending here, and then I think we are good to go!

script/release-publish.sh
rnbguy: We can also parse the `sha256sum.txt` and put it as markdown tables. Example, [tlaplus release](https://github.com/tlaplus/tlaplus/releases/tag/v1.7.2).

Nonetheless, it is not that important to me. I will be totally ok if you decide to keep `sha256sum.txt`` as a release artifact.

script/release-publish.sh
```diff
@@ -47,6 +47,8 @@ ZIPF_NO_VER="target/universal/apalache.zip"
 TGZF="target/universal/apalache-${VERSION}.tgz"
 TGZF_NO_VER="target/universal/apalache.tgz"
 
+SHA256F="target/universal/sha256sum.txt"
+(cd target/universal && sha256sum apalache.* > "$SHA256F")
```
thpani, line 51: I think the glob here will only match the unversioned `$TGZF_NO_VER` and `$ZIPF_NO_VER` archives, not the versioned `$TGZF` and `$ZIPF`.

Not sure though what the glob for the other two would look like, as this also contains build directories at this point, right?

shonfeder: That's a good point. I had assumed that was intended, but now that I think of it I'm not sure why we *wouldn't* want to include hashes for all the archives. 

I think the following should do the trick:

```suggestion
(cd target/universal && sha256sum {apalache*.tgz,apalache*.zip} > "$SHA256F")
```

shonfeder: Actually, we don't even have to be that precise, and could get away here just with globing for `apalache`.

thpani (approved):
I think we have consensus on the suggestions, so I've committed them.
Think this is good to merge. @shonfeder PTAL?

shonfeder:
I added two fixes to things we broke with our suggestions after some local testing. Good to go imo!

shonfeder (approved):

thpani:
> I added two fixes to things we broke with our suggestions after some local testing. Good to go imo!

Ugh, I should've looked over the code more carefully after applying the suggestions 🙈 
Thanks for keeping an eye on the changes, @shonfeder!

rnbguy:
Awesome! Checksums in next release :slightly_smiling_face:

Thanks for the careful suggestions.

shonfeder:
No worries, @thpani! I think this came through nicely, IMO. A real team effort :D 

Thanks for the PR and for adding this improvement, @rnbguy!

Issue #1174, opened 2021-12-14
Replace untyped option maps with statically typed configuration objects


shonfeder:
We are currently using a `Map[String, Any]` to store most of the application configuration. E.g., see

https://github.com/informalsystems/apalache/blob/98306abe0b7d2ffb913b50e9fd5dacd772712f10/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/WriteablePassOptions.scala#L23-L24

When values are fetched from the map, we need to do a runtime cast, which has also required carrying around typing information which could otherwise be erased:

https://github.com/informalsystems/apalache/blob/98306abe0b7d2ffb913b50e9fd5dacd772712f10/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/WriteablePassOptions.scala#L38-L57

This makes the application configuration significantly more brittle than it needs to be, as it it subject to runtime errors due to illtyped configuration values or missing keys.

Afaict, we should be able to replace the untyped map with a class that provides static guarantees for the correct typing of the configuration values, and that any components that depend on configuration receive all values they need.

If we push all such configuration to the `ConfigManager` added in #1160, we'll also get the added benefit of being able to load and dump the full application configuration programmatically. The former should support automation, and the latter reproducibility, benchmarking, and debugging (e.g., the bug reports we generate could report a dump of the entire application configuration, which could then be loaded up trivially in regression tests).

---

We need to thread the statically specified options through our DI framework Guice. Here's the plan:

- [x] #2038 after which the class `ApalacheConfig` will be used specify the values that can be be supplied to the options from the various configuration sources.
- [x] #1177
- [x] #2040
- [x] #2066
- [x] #2050

shonfeder:
Next up, we need to thread the statically specified options through our DI framework Guice. Here's the plan:

- [ ] #2038 after which the class `ApalacheConfig` will be used specify the values that can be be supplied to the options from the various configuration sources.
- [ ] A class `ConfigurableOptions` will be defined with a method `configure:
      ApalacheConfig => Try[Unit]` that is expected to update any needed options
      based on the `ApalacheConfig` -- this will replace setting values in the
      current untyped map based on the CLI params.
- [ ] Options classes `FooOptions` will be defined as subclasses
      `ConfigurableOptions` extended with mixins for each additional option
      class needed for the given mode of execution, drawing from the already
      defined `infra.options` traits.
- [ ] For each pass, we will bind the corresponding `FooOptions` to a
[...]

Issue #2011, opened 2022-07-25
Add SMT encoding with sets as uninterpreted constants and functions as arrays


rodrigo7491:
Initial evaluation of the arrays encoding indicates that SMT arrays provide significant performance gains when encoding TLA+ functions, but can cause a slowdown when compared to the `oopsla19` encoding when encoding TLA+ sets. In light of this, we should add a new `funArrays` encoding, in which SMT arrays are used solely to encoded TLA+ functions.

Having two competing array encodings is intended to be a temporary situation, with the encoding shown to be more efficient by large scale evaluation intended to be kept and the other encoding intended to be removed.

PR #2018, opened 2022-07-26
Move SMTEncoding into `infra.passes.options`

shonfeder:
This is required for #1174, since we'll need the SMTEncoding type for
the pass options declarations, but we cannot introduce a dependency from
infra -> bmcmt, as there is already a dependency going the other way.

Along the way, I also made some small cleanups to

- the naming the way the case class was declared (to use a more
- idiomatic pattern, following examples like, inter alia,
  https://nrinaudo.github.io/scala-best-practices/definitions/adt.html)
- remove the unneeded literal identifiers ``
- move string conversion into the class (keeping the closely related
  logic together)


mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
```diff
@@ -72,8 +68,8 @@ class CheckCmd(name: String = "check", description: String = "Check a TLA+ speci
 
     logger.info("Tuning: " + tuning.toList.map { case (k, v) => s"$k=$v" }.mkString(":"))
 
-    executor.passOptions.set("general.tuning", tuning)
-    executor.passOptions.set("checker.nworkers", nworkers)
+    executor.passOptions.set("general.tuning", tuning) // XXX
+    executor.passOptions.set("checker.nworkers", nworkers) // XXX
```
thpani, line 72: What's `XXX`? 🙃

thpani (approved):
LGTM. I like that we're getting rid of the literal `` comparisons!

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -0,0 +1,28 @@
...
+    final case object Oopsla19 extends SMTEncoding {
+      override def toString: String = "oopsla19"
+    }
+    final case object Arrays extends SMTEncoding {
+      override def toString: String = "arrays"
+    }
+
+    val ofString: String => SMTEncoding = {
+      case "arrays"        => Arrays
+      case "oopsla19"      => Oopsla19
```
rodrigo7491, line 24: Having arrays capitalized is fine, but `Oopsla19` is a bit odd. Not sure what would be a good workaround though.

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/TestSymbStateRewriterApalacheGen.scala
```diff
@@ -1,5 +1,6 @@
 package at.forsyte.apalache.tla.bmcmt
 
+import at.forsyte.apalache.infra.passes.options.SMTEncoding
```
rodrigo7491, line 3: Why is this import added here? This is happens in many other test files as well.

rodrigo7491 (approved):
Looks good! This will affect #2011, but I can make a refactoring there.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
thpani: Hmmm true, we have it in all caps in other places, e.g. `TestArenaWithOOPSLA19`.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
shonfeder: That marks these stringly typed settings for eradication :dagger: ;) -- but it should not be in this PR.

shonfeder:
> Looks good! This will affect #2011, but I can make a refactoring there.

I hope it wont mean much work there? btw, that's just an issue, did you mean to link to a PR, or is it that there's work in a branch that hasn't been pushed yet?

PR #2023, opened 2022-07-28
Move SourceOptions into infra.passes.options

shonfeder:
This is just a simple code organization change consistent with #2018 and the
ongoing work on #1174.

---

- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] Rebase onto `main` once #2020 is merged.

PR #2024, opened 2022-07-28
Add static representation of configuration options

shonfeder:
Contributes towards #1174

## Context

This adds traits with values to represent all the options that we are
currently setting in the `at.forsyte.apalache.tla.tooling.opt.Cmd*`
classes. None of those options are actually transferred over here.

The proposed design uses traits to specify all the options, with the idea that
various subprograms can depend on mixins of the traits to statically declare the
configurations they respond to.

## Reviewing

This changeset just prepares the representation of the options, with the aim of
obtaining feedback and clearing the way for the changeogver. The followup PR
will remove the `PassOptions` class, replacing it with objects that implement
the configuration traits needed by the respective subcommands.

- [x] Rebase once #2023 is merged in

### DIFF
--- a/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
+++ b/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
@@ -1,47 +1,99 @@
-package at.forsyte.apalache.infra.passes
+package at.forsyte.apalache.infra.passes.options
 
-/** Collects the passoptions supported for configuring Apalache's various passes */
-object options {
+import at.forsyte.apalache.tla.lir.Feature
+import java.nio.file.Path
 
-  /** Defines the data sources supported in the [[PassOptions]] */
-  sealed abstract class SourceOption
+/**
+ * Collects the passoptions supported for configuring Apalache's various passes
+ *
+ * @author
+ *   Shon Feder
+ */
 
-  object SourceOption {
+/** Defines the data sources supported */
+sealed abstract class SourceOption
 
-    /** Data to be loaded from a file */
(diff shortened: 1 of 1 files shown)


shonfeder:
Note this PR is quite small, and reviewing should hold off until the pending PRs it branches off of are merged in and this branch is rebased.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -1,47 +1,99 @@
...
-   * Defines the SMT encoding options supported
+   * Data supplied as a string
    *
-   * @author
-   *   Shon Feder
+   * @param content
+   *   the principle data source
+   * @param aux
+   *   auxiliary data sources
    */
```
shonfeder, line 30: This was just in the wrong place.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -1,47 +1,99 @@
-package at.forsyte.apalache.infra.passes
+package at.forsyte.apalache.infra.passes.options
 
-/** Collects the passoptions supported for configuring Apalache's various passes */
-object options {
```
shonfeder, line 4: Realized we can declare the package just by putting it in the file and giving the package name, so no need for the object.

shonfeder (commented):
@thpani this is ready for review now!

I suggest setting the diff to ignore whitespace, as it cuts the diff down to just about 60 lines.

thpani (approved):
LGTM

PR #2025, opened 2022-07-28
Fix documentation links

shonfeder:
Followup to recent documentation work that had broken scaladoc links which we
didn't catch because CI wasn't failing on the scaladoc warnings.

There are a couple cases where I couldn't fix the references due to one of two
things:

- The links were to private methods
- The links were to overloaded method and the reference was therefore ambiguous.
  I spent about 20 mins trying to figure out how to make the links unambiguous,
  but couldn't find a solution that worked.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,8 @@ jobs:
         run: make fmt-check
 
   scaladoc-check:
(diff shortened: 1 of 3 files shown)


thpani (approved):
LGTM

PR #2026, opened 2022-07-29
Add build targets to build and view API docs

### LINKED ISSUE #1911: Generate scaladoc of unstable

shonfeder:
Closes #1911

## Context

We've agreed as a team that we should invest in more thorough API documentation.
Improving the ease with which we can build and view the docs is important for
this.

## Changes

- Add sbt task `browseApiDocs` to build the docs, then open the `index.html` in
  the default browser (on MacOS and Linux, or print the file location on Windows).
- Add makefile task to run the task, and start watching and autobuilding the
  docs on file updates.
- Add a section on documentation to `CONTRIBUTING.md`

## Reviewing

It would be helpful if this could be tried on a Mac to ensure that the macos
case works as expected.

-------


CONTRIBUTING.md
```diff
@@ -283,6 +283,26 @@ writeup][writeup]
 For installation and configuration in vanilla emacs, see
 https://scalameta.org/metals/docs/editors/emacs.html
 
+## Documentation
+
+We maintain two main kinds of documentation: tutorials and manuals meant as user
+documentation and API documentation meant for developers.
```
thpani, line 289: Should we also mention RFCs/ADRs (which seem to fall somewhere in between)?

thpani (approved):
Confirmed working on macOS.

One question though: when the root view opens, I see `zio` as one of the packages? Any ideas why it ends up there?
Screen Shot 2022-07-29 at 14 25 30

shonfeder:
Yeah, I'm not sure why ZIO is showing up there! I'll investigate, but will probably address in a separate PR, as it's not likely related to these changes.

CONTRIBUTING.md
shonfeder: Ah yeah. Good idea.

shonfeder: how do you find https://github.com/informalsystems/apalache/pull/2026/commits/1a4b77c1623bb4cf480766d4f477d0cf5619db1e ?

thpani: Excellent, just a word fix in the comment below.

CONTRIBUTING.md
```diff
@@ -283,6 +283,26 @@ writeup][writeup]
 For installation and configuration in vanilla emacs, see
 https://scalameta.org/metals/docs/editors/emacs.html
 
+## Documentation
+
+We maintain three principle kinds of documentation:
+
+- Tutorials and manuals meant as general user documentation can be found in [./docs/src](./docs/src).
+- ADRs and RFCs documenting design and technical decisions are meant for developers
+  and expert users and be found in [./docs/src/adr](./docs/src/adr)
```
thpani, line 292: I think this needs either a modal verb or declination?

```suggestion
- ADRs and RFCs documenting design and technical decisions are meant for developers
  and expert users and can be found in [./docs/src/adr](./docs/src/adr)
```

```suggestion
- ADRs and RFCs documenting design and technical decisions are meant for developers
  and expert users and are found in [./docs/src/adr](./docs/src/adr)
```

Issue #2015, opened 2022-07-26
Wrap or refactor `check` command so it can be treated as a function frominput configuration params to a checker result.


shonfeder:
I suspect this would really be helped by 

- [x] #1177
- [x] #1174

shonfeder:
Closed by #2119

Issue #1177, opened 2021-12-15
Integrate all CLI options into the configuration management system


shonfeder:
With #1160, it should be quite simple to add new configurations through the configuration system. We should expose all configurations that are currently configurable via CLI via the configuration files too.

In particular, the tuning options are a good fit for this.

I propose making this work item dependent on #1174: once we have moved the configuration parameters into a class, it will be trivial to drop the class into the the class targeted by PureConfig.

Suggested by @konnov https://github.com/informalsystems/apalache/pull/1160/files#r769408346

thpani:
Just leaving a note to document that we currently have two kinds of boolean commandline flags:

- binary flags, e.g., `--debug` and `--smtprof`, and
- tri-state flags, e.g., `--profiling` and `--write-intermediate`

The latter are 3-state to enable overriding env vars + config options. These should probably be unified as part of this effort.

shonfeder:
The problem @thpani identified should be fixed by https://github.com/informalsystems/apalache/pull/2041

PR #2038, opened 2022-08-05
Prepare the ApalacheConfig trait to communicate the override values of the options traits from `ApalacheCommand`

shonfeder:
Advances #1174, #1177, and #2015 

This implements the first step in the plan described at https://github.com/informalsystems/apalache/issues/1174#issuecomment-1205981013

## Overview

We'll need to integrate the new options system into our DI via Guice (#1174) and we'll want want to be able configure options via our various supported configuration sources (such as the local and global configuration files, #1177). We'll also be able to use the `ConfiManagers` deserialization for decoding the JSON dict of param values supplied via the `CmdExecutor` service's RPC calls. 

This takes us a step in that direction by making each parsed command also hold an attribute with the final `ApalacheConfig` derived by loading all available configs and then overriding with the final CLI args.

In followup work, we will read from this attribute to supply the `ApalacheConfig` that specifies the program options during the invocation of each command's `run` (#2015). 

## Reviewing

It's a small changeset, but reviewing the two commits separately is probably still helpful.
 
-------


thpani (approved):
👍 LGTM

Issue #2019, opened 2022-07-27
Unable to tell where `SeqModelChecker`/`TransitionExecutorImpl` hangs.


thpani:
## Description

I've encountered a slow `apalache-mc check` run.
I'm unable to determine from the detailed log where the slowdown between the last two lines originates:

```
[...]
20:36:26.222 [main] DEBUG a.f.a.t.b.t.TransitionExecutorImpl - Step #9, transition #19
20:36:26.222 [main] DEBUG a.f.a.t.b.t.TransitionExecutorImpl - Translating to SMT...
20:36:29.479 [main] INFO  a.f.a.t.b.SeqModelChecker - State 8: Checking 1 action invariants
20:36:29.479 [main] DEBUG a.f.a.t.b.SeqModelChecker - State 8: Checking action invariant 0
20:36:29.692 [main] DEBUG a.f.a.t.b.t.TransitionExecutorImpl - Step #9, transition #20
20:36:29.692 [main] DEBUG a.f.a.t.b.t.TransitionExecutorImpl - Translating to SMT...
20:36:30.141 [main] INFO  a.f.a.t.b.SeqModelChecker - State 8: Checking 1 action invariants
20:36:30.141 [main] DEBUG a.f.a.t.b.SeqModelChecker - State 8: Checking action invariant 0
20:36:30.356 [main] INFO  a.f.a.t.b.t.TransitionExecutorImpl - Step 9: picking a transition out of 21 transition(s)
21:26:19.278 [main] ERROR a.f.a.t.b.DumpFilesModelCheckerListener$ - Check the trace in: ...
```

## Input specification

Cannot share here, ping me for info.

## The command line parameters used to run the tool

`check --inv=XXX --length=9 --discard-disabled=false --no-deadlock --tuning-options-file=array.properties`

where `array.properties` contains `search.transitionFilter=(0->.*|[1-7]->.*|8->4|9->.*|[1-9][0-9]+->.*)`

## Expected behavior

More log messages, or a log message that explains a call to the SMT solver.

_Note to self: this is job id `d08a6d33-0821-4523-8a2d-de5e19bd0216:{4,9}`._

thpani:
I tried to figure this out on my own reading the code, to no avail. Maybe @konnov can help here.

thpani:
This is a performance regression introduced by producing an example trace (#1838) described in #2039. Closing this as duplicate.

PR #2047, opened 2022-08-08
Make example trace output optional via `--save-runs`

### LINKED ISSUE #2039: Performance regression outputting an example trace
We added default output of an example trace in #1838. It is implemented by the call to `trex.sat()` below.

https://github.com/informalsystems/apalache/blob/eb92cc88883ca2134c06ba6ab644b8d60585d4bf/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala#L95

This may cause a significant performance regression, as witnessed by #2019, where producing the example trace takes one hour.

In addition, when Apalache is run with `--discard-disabled`, it might overshoot the actually feasible trace length. This essentially causes an unsat query in the `trex.sat()` call that takes substantial time. Also, the reason for the delayed behavior is not obvious from the tool output:

```
$ apalache-mc check --features=rows --length=9 --inv=Mutex --discard-disabled=false --no-deadlock --tuning-options='search.transitionFilter=(0->.*|[1-6]->.*|7->3|[8-9]->.*|[1-9][0-9]+->.*)' test/tla/MC_LamportMutexTyped.tla
...
13:12:43.344 [main] DEBUG a.f.a.t.b.t.TransitionExecutorImpl - Step #9, transition #5
13:12:43.345 [main] DEBUG a.f.a.t.b.t.TransitionExecutorImpl - Translating to SMT...
13:12:43.442 [main] INFO  a.f.a.t.b.t.TransitionExecutorImpl - Step 9: picking a transition out of 6 transition(s)
<<note the delay here>>
13:16:34.672 [main] INFO  a.f.a.t.b.p.BoundedCheckerPassImpl - The outcome is: NoError
13:16:34.672 [main] DEBUG a.f.a.i.p.PassChainExecutor - PASS #13: BoundedChecker [OK]
```

thpani:
We introduced example trace output in #1838. As described in #2039 and #2019, this creates a unexpected performance bottleneck if the SMT call for constructing the trace takes a long time.

Oddly enough, writing the trace was made optional for `simulate` via the `--save-runs` flag   (defaulting to `false`) in #1838, but always enabled for `check`.

With this PR, we lift `--save-runs` to `check`, making example trace output optional also for `check`.

We also make some UX improvements that
* fix the case where the last symbolic run of `simulate` was output twice,
* add log output around the example trace output, and
* adjust example output log levels from `ERROR` to `INFO`.

Fixes #2039

### DIFF
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
@@ -57,13 +57,13 @@ class CheckCmd(name: String = "check", description: String = "Check a TLA+ speci
         default = "")
 
   var saveRuns: Boolean =
-    opt[Boolean](name = "save-runs", description = "save an example trace for each symbolic run, default: false",
+    opt[Boolean](name = "output-traces", description = "save an example trace for each symbolic run, default: false",
         default = false)
 
   def collectTuningOptions(): Map[String, String] = {
     val tuning =
       if (tuningOptionsFile != "") loadProperties(tuningOptionsFile) else Map[String, String]()
-    overrideProperties(tuning, tuningOptions) ++ Map("search.saveRuns" -> saveRuns.toString)
+    overrideProperties(tuning, tuningOptions) ++ Map("search.outputTraces" -> saveRuns.toString)
   }
 
   def run() = {
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
@@ -63,7 +63,7 @@ class CheckCmd(name: String = "check", description: String = "Check a TLA+ speci
   def collectTuningOptions(): Map[String, String] = {
     val tuning =
       if (tuningOptionsFile != "") loadProperties(tuningOptionsFile) else Map[String, String]()
-    overrideProperties(tuning, tuningOptions)
+    overrideProperties(tuning, tuningOptions) ++ Map("search.saveRuns" -> saveRuns.toString)
   }
 
   def run() = {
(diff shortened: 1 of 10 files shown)


mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
```diff
@@ -56,6 +56,10 @@ class CheckCmd(name: String = "check", description: String = "Check a TLA+ speci
     opt[String](name = "view", description = "the state view to use with --max-error=n, default: transition index",
         default = "")
 
+  var saveRuns: Boolean =
+    opt[Boolean](name = "save-runs", description = "save an example trace for each symbolic run, default: false",
+        default = false)
```
Kukovec, line 61: I feel like the name "save-runs" isn't really correct, at least in the non-simulation case. Would it make sense to rename the flag in both cases to e.g. `enable-traces` or `tracing`? Also, the description here is not really suitable for `check` IMO ("each" is confusing here).

thpani: > I feel like the name "save-runs" isn't really correct, at least in the non-simulation case

Why do you think that's the case? It's doing the same thing, right?

> Also, the description here is not really suitable for check IMO ("each" is confusing here).

Do you think "each" suggests that there are multiple?

I think Igor's idea was that we do not use the word "trace" with users.

thpani: How about `--dump-traces`?

thpani:
Had to do some wiring in in [f3a5b2b](https://github.com/informalsystems/apalache/pull/2047/commits/f3a5b2b2d1e08d012f95182d64b781c8ec4c05f1).
I find it weird that this is wired in via tuning options, but we can fix that in a follow-up PR.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
Kukovec: Maybe `output-`, rather than `dump-`.

Kukovec (approved):

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
thpani: Renamed in [63494ad](https://github.com/informalsystems/apalache/pull/2047/commits/63494ad4fffe6cba3443127ed9ea3566338d4c63)

Issue #2013, opened 2022-07-26
Shai: Expose `check` functionality via an RPC call


shonfeder:
Requires: #2020

## Context

In a meeting today with @tesnimab, @hvanz, and @thpani  it became clear that one of the
highest value features we can provide for MBTs immediate work is an RPC call to
the server to expose the functionality of the `check` CLI command. 

## Work

The principle work anticipated to complete his feature:

- [x] #2015
- [x] #2121
- [x] Define a protobuf message to obtain all needed input for running `check` 
- [x] Define a protobuf message to provide all needed output from the results of the
      `check`.
- [x] Implement methods to support the generated gRPC calls
- [x] Add tests for the new RPC calls
- [x] Add Chai support for the subcommand service -- tracked in https://github.com/informalsystems/apalache-chai/issues/21

## Risks

It is possible we will hit some concurrency bugs with the checker in it's
interaction with z3, but this is a risk we would be confronting sooner than
later in any case.

Issue #1624, opened 2022-04-13
Handling of `version` by backuity/clist


thpani:
We are currently parsing CLI arguments using [backuity/clist](https://github.com/backuity/clist).
We recently had some problems with how the `version` command is handled, see #1497, #1505, and #1611:

* We cannot parse access CLI arguments, in particular `--debug`, when Apalache is run with these commands (#1497, #1505). However, we use `version` as a no-op command, e.g. in integration tests.
* The built-in commands `version` and `help` are not printed in the usage / `--help` output (#1611).

This is an architectural shortcoming of `clist`, which does not internally represent the built-ins `version` and `help` as `Command`s.
In particular, for accessing flags, it returns an `Option[Command]` from parsing, and `None` on the built-ins.

We should file an upstream bug or consider another CLI parser.

konnov:
> We should file an upstream bug or consider another CLI parser.

Yep, `clist` has annoying behavior. Perhaps, there is a newer better CLI parser for scala.

shonfeder:
Definitely room for improvement here, i think the key question is around prioritization.

shonfeder:
The most actively developed and attractive (IMO) CLI parsing libs I've found so far are

- https://github.com/bkirwi/decline
- https://github.com/alexarchambault/case-app

Scallop also looks pretty good, and it has more stars, but it hasn't been updated quite as recently (could just be cause it's mature and stable tho):

- https://github.com/scallop/scallop

Finally, `scopt` is the most heavily stared library I've seen, but the API doesn't look wonderful to me:

- https://github.com/scopt/scopt

shonfeder:
`case-app` could be desirable as it could enable automating much of the bookkeeping required to integrate with the configuration system.

konnov:
This should be fairly easy to fix with LLMs now. I would also prefer a Java library that has a history of long-term maintenance.

thpani:
The well-maintained JVM/Java options seem to be:

- https://github.com/remkop/picocli
- https://commons.apache.org/proper/commons-cli/changes.html

konnov:
`picocli` looks especially interesting, as they encourage the projects to include their source code, which aims at having small footprint.

PR #2065, opened 2022-08-10
Wire CLI args into configuration system

### LINKED ISSUE #1177: Integrate all CLI options into the configuration management system

shonfeder:
Closes #1177

## Overview

To advance #2013 and friends, we need to make the configuration of various pass executor pipelines

1. easy to reason about, and
2. thread-safe, so that concurrent RPC requests can be handled safely

This work sets us up to address (1) by solving #1177, which further prepares the way for #2073.

## Changes

- The `ApalacheConfig` class has been expanded and reorganized to give a single specification of all the configurable values that determine program behavior, along with their default values.
- The `ConfigManager` class has been cleaned up to improve the loading and overriding logic
- All CLI parsers have been reworked to communicate all their configured values through an instance of `ApalacheConfig`.
  - This involved ensuring that all CLI options are `Option`s, so that the configuration merging logic can differentiate when a value set by the CLI should override something set in one of the other configuration sources.

## Reviewing

The changeset is mostly simpler than the count would imply: many line are just listing on configuration parameters, and about 100 of the removed lines are work that has been deferred to #2074. See my comments on the PR for info.

While reviewing, it's probably simplest to start with the `options.scala` file, followed by the changes to the `ConfigManager.scala`, and then the `ApalacheCommand.scala`: all other changes are correlates of those.

Unfortunately, I think the commits aren't well organized to make review by commit very helpful in this case.

For a high-level description of these changes and where they are headed, see [ADR 22](https://github.com/informalsystems/apalache/blob/cf8cd12f30c708f593ea510037002ac2ed3efbd4/docs/src/adr/022adr-unification-of-configs-and-options.md).

## Followup

- I will be following up this change set with a short ADR documenting the configuration system and what motivated its introduction. See  #2075.


mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/AbstractCheckerCmd.scala
```diff
@@ -5,46 +5,55 @@ import org.backuity.clist.{arg, opt}
...
+    val cfg = super.toConfig()
+    cfg.copy(
+        common = cfg.common.copy(file = Some(file)),
+        checker = cfg.checker.copy(config = config, cinit = cinit, init = init, next = next, inv = inv,
+            temporal = temporal, length = length),
+    )
+  }
   override def setCommonOptions(): Unit = {
+    // TODO: rm once OptionProvider is wired in
+    val cfg = configuration.left.map(err => new ConfigurationError(err)).toTry.get
```
shonfeder, line 43: This is an intermediate step, using the derived configuration to set all values in the option map. In a follow up, we will remove the option map and replace it with a statically verifiable class of options.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/ServerCmd.scala
```diff
@@ -1,17 +1,12 @@
...
-
 import com.typesafe.scalalogging.LazyLogging
 import at.forsyte.apalache.shai
 
 class ServerCmd
     extends ApalacheCommand(name = "server", description = "Run in server mode (in early development)")
     with LazyLogging {
 
-  // Dummy file used for application (esp. output) configuration
-  def file = new File("server")
```
shonfeder, line 13: This dummy file is no longer needed now that the `OutputManager` can fall back to the `routine` config value.

test/tla/cli-integration-tests.md
```diff
@@ -3414,7 +3414,7 @@ $ rm -rf ./test-out-dir ./test-run-dir
 ### configuration management: read run-dir from local `.apalache.cfg`
 
 ```sh
-$ echo "run-dir: ./configured-run-dir" > .apalache.cfg
+$ echo "common.run-dir: ./configured-run-dir" > .apalache.cfg
```
shonfeder, line 3417: Note that this introduces a change to the configuration format, which moves al the previous top-level config values under the `common` key.

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
```diff
@@ -23,25 +24,8 @@ private object Converters {
   implicit val overrideFileReader = ConfigReader.fromString[File](catchReadError(expandedFilePath(_).toFile()))
   // PureConfig's optF will convert None values to appropriate configuration errors
   implicit val featureReader = ConfigReader.fromString[Feature](optF(Feature.fromString))
-}
+  implicit val featureWriter = ConfigWriter.toString[Feature](_.toString)
 
-/**
```
shonfeder, line 28: This trait is no longer needed now that every CLI spec is mapped into an instance of `ApalacheConfig`

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -64,53 +229,3 @@ object SMTEncoding {
     case oddEncodingType => throw new IllegalArgumentException(s"Unexpected SMT encoding type $oddEncodingType")
   }
 }
-
```
shonfeder, line 67: This bits can safely be ignored during review. Followup WIP the implements the options classes can be seen in #2074.

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
```diff
@@ -50,67 +34,74 @@ trait CliConfig {
...
-      // `withFallback` supplies configuration sources that only apply if the preceding configs aren't set
-      .withFallback(globalConfig.optional)
+    val localConfig: ConfigObjectSource =
+      cfg.common.configFile.orElse(findLocalConfig(Paths.get(".").toAbsolutePath())) match {
+        case Some(cfgFile) => ConfigSource.file(cfgFile)
+        case None          => ConfigSource.empty
+      }
+
+    // Derive lightbend `Config` objects from the default and primary config
+    // case classes.
```
thpani, line 80: What are the default and primary config case classes?

thpani (approved):
Phew, that's some heavy lifting 😄 

LGTM!
I'm not too excited about the added logic in the `*Cmd`s' `toConfig` methods, but I also can't think of a better alternative...

shonfeder:
>  I'm not too excited about the added logic in the *Cmds' toConfig methods, but I also can't think of a better alternative...

Ditto! :laughing: 

I take some consolation from thinking that it "only" replaces and expands the logic that had previously been located in the `ConfigManager` class ;)  I also have some hope that if, as per #1624, we end up changing the CLI library and opting for https://github.com/alexarchambault/case-app, we might be able to automate much of that conversion logic away.

Thanks for reviewing!

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -1,26 +1,191 @@
...
-    writeIntermediate: Boolean = false,
-    profiling: Boolean = false,
-    features: Seq[Feature] = Seq())
+/**
+ * The basic interface of classes that specify application configurations
+ */
+sealed trait Config[T] {
+
+  /**
+   * Produces a copy of the config group `T` with all it's attributes (attributes nested config groups) set to `None`
```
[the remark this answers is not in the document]
shonfeder: Good point. I added a note to clarify this term a few lines above in 571d14aae0294e415f26bdd0a8e4ae2ea7c6b15e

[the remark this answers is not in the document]
shonfeder: There were several words missing! Thanks: 7e5b558876d6c182fa0b83a6a0edc6e1e1018775

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -1,26 +1,191 @@
...
+   *   A file from which a local configuration is to be read
+   * @param writeIntermediate
+   *   Whether or not to write intermediate data into the `runDir`
+   * @param profiling
+   *   Whether or not to write general profiling data into the `runDir`
+   * @param features
+   *   Enable experimental features
+   */
+  case class Common(
+      routine: Option[String] = Some("UNCONFIGURED-ROUTINE"),
```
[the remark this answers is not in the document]
shonfeder: That's a good question. I was thinking at the time that this fallback would let us have some useful output from the output manager in absence of a properly configured "routine", but then isn't it just better to raise an error so the routine can be configured correctly? I think so! 56accb5e6f50e599705e7ae4c7008d120ed54067

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -1,26 +1,191 @@
...
+    implicit def noneCase[T]: Case.Aux[Option[T], Option[T]] = at(o => o.flatMap(_ => None))
+    // Takes a `Config[T]` to an empty version of the config (with all fields set to `None`)
+    implicit def configCase[T <: Config[T]]: Case.Aux[T, T] = at(cfg => cfg.empty)
+  }
+
+  /**
+   * The common configurations shared by all of Apalache's modes of execution
+   *
+   * @param routine
+   *   The subcommand or process being executed
```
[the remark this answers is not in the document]
shonfeder: I was thinking this was a more general category, as we could have configured operations that don't correspond to subcommands. E.g., every session in the server could be its own routine, and each RPC call could be a routine, but maybe it is enough to just identify this commands.  It's true that the only use case currently is for commands, and probably best not to overly generalize: e11cf93fcd4b11a8c69ddc2d531b70bf03f0d510

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -1,26 +1,191 @@
...
+   * @param file
+   *   The file from which input data can be read
```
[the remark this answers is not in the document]
shonfeder: Yeah, it's overly general. It came from just picking up the naming convention used in the CLI parameters. I agree `inputfile` is clearer db2ee1e036e270d0d570f273aa7a0f29b49cd0e6

tla-io/src/main/scala/at/forsyte/apalache/io/ConfigManager.scala
shonfeder: They are the named variable commented here: 

- The "primary config" is derived from the `cfg` param supplied to this function (see the `@param cfg` in the scaladoc above).
- The default is derived from the default values specified in `ApalacheConfig`, invoked by calling applying the case class.

thpani: Oh, I missed that this refers to the doccomment on `load`. All good!

PR #2074, opened 2022-08-15
Replace the `PassOptions` map with `OptionsGroup` classes

### LINKED ISSUE #2040: Implement class `OptionsProvider` for dependency injection
By implementing Guice's provider trait we can get late binding of pass options inserted into the DI cascade.

- [x] Options classes `FooOptions` will be defined as a subclasses
      `ConfigurableOptions` extended with mixins for each additional option
      class needed for the given mode of execution, drawing from the already
      defined `infra.options` traits.
- [x] A `Provider[FooOptions]` will be implemented to allow the options object
      to be constructed.
- [ ] For each pass, we will [bind a
      `Provider[FooOptions]`](https://github.com/google/guice/wiki/ProviderBindings)
      the Guice module -- this will replace the current uniform binding of
      `WriteableOptions` to `PassOptions`.
- [ ] After creating the executor in the `run` method of each `*Cmd`, we will
      call `options.configure(this)`

shonfeder:
Closes #2040, #2015, #1174 

## Context

To improve the maintainability of our code base along the lines first proposed in #1174 and elaborated in [ADR-022](https://github.com/informalsystems/apalache/blob/main/docs/src/adr/022adr-unification-of-configs-and-options.md), we have decided to replace the untyped map of options with a statically typed family of case classes. While we have been considering work of this sort for some time, it was finally motivated by the need to implement the gRPC service exposing our CLI functionality to RPC calls, as per #2013 and related.

## Changes

- Remove `PassOptions` and `WriteablePassOptions`
- Add `DerivedPredicates` class for communication of computed specification predicates between passes (this effectively isolates the one part of the mutable `PassOptions` that is still needed given our current pass execution architecture.
- Move the initial extraction of configuration data from TLC configs into the passes package, so we can use it during configuration time. There was a portion that must remain in `ConfigurationPassImpl`, as it requires inspecting a parsed `TlaModule`.

## Reviewing

This PR is much larger than I'd prefer, but many of the changes are quite mechanical. If desired, I could factor out some standalone portions. Only bear in mind that reviewers would likely need to compare those extracted bits with their use here to grok their fitness for purpose. I'd be happy to sync up to give an overview of the changes and point out the parts where careful attention is most needed.


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/NormalizedNames.scala
```diff
@@ -26,21 +25,6 @@ object NormalizedNames {
   // the names of the options that capture the critical specification pieces
   val STANDARD_OPTION_NAMES = Seq("init", "cinit", "next", "inv", "temporal")
 
-  /**
```
shonfeder, line 29: This was moved into the `DerivedPredicates` class

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
```diff
@@ -85,31 +84,36 @@ class CheckCmd(name: String = "check", description: String = "Check a TLA+ speci
...
+    }
+
+    newCfg
   }
 
   def run() = {
-    // TODO: rm once OptionProvider is wired in
-    val cfg = configuration.left.map(err => new ConfigurationError(err)).toTry.get
+    val cfg = configuration.get
+    val options = OptionGroup.WithCheckerPreds(cfg).get
```
shonfeder, line 108: Currently we're just letting the error handling here fall back to the existing exception handlers, but since these both return `Try`'s, and the methods end with an `Either`, we should be able to move to ADT based error handling for the run functions. I opened an issue for followup: https://github.com/informalsystems/apalache/issues/2106

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/AbstractCheckerCmd.scala
```diff
@@ -22,11 +21,11 @@ abstract class AbstractCheckerCmd(val name: String, description: String)
...
-  var inv: Option[String] =
-    opt[Option[String]](name = "inv", default = None, description = "the name of an invariant operator, e.g., Inv")
-  var temporal: Option[String] =
-    opt[Option[String]](name = "temporal", default = None,
-        description = "the name of a temporal property, e.g. Property")
+  var inv =
+    opt[Option[List[String]]](name = "inv", default = None,
+        description = "the name of an invariant operator, e.g., Inv")
+  var temporal = opt[Option[List[String]]](name = "temporal", default = None,
+      description = "the name of a temporal property, e.g. Property")
```
shonfeder, line 28: @konnov Note that this now technically makes it possible for users of the CLI to pass in lists of invariants  or temporal properties. This possibility seemed to be naturally implied by the way we use these internally, but I'm not sure if there's a reason we want to *avoid* exposing this functionality to users. What do you think?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/InlinePassImpl.scala
```diff
@@ -25,15 +25,15 @@ class InlinePassImpl @Inject() (
 
   override def execute(module: TlaModule): PassResult = {
     // Since PrimingPass now happens before inlining, we have to add InitPrime and CInitPrime as well
-    val initName = options.getOrElse[String]("checker", "init", "Init")
-    val cinitName = options.getOrElse[String]("checker", "cinit", "CInit")
+    val initName = derivedPreds.init
+    val cinitName = derivedPreds.cinit.getOrElse("CInit") // TODO: why default on None here?
```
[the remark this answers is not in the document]
shonfeder: This is just an aside: we can leave the comment to resolve later, as the logic here does not change from what it was, so this isn't blocking for review or anything.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/AbstractCheckerCmd.scala
konnov: Right. I think we always had this possibility internally, e.g., a TLC config allows for multiple invariants. However, we were stuck with the arguments parser, which did not let that behavior.

shonfeder: Ok nice! Then I'll update the CLI docs here to communicate this functionality to the users. I'll also update the manual with info on this, but will do that in a followup, since this is big enough :)

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -317,7 +344,7 @@ object OptionGroup {
         features <- common.features.toTry("common.features")
         outDir <- common.outDir.toTry("common.outDir")
         profiling <- common.profiling.toTry("common.profiling")
-        smtprof <- common.debug.toTry("common.smtprog")
+        smtprof <- common.smtprof.toTry("common.smtprog")
```
thpani, line 347: This a typo?
```suggestion
        smtprof <- common.smtprof.toTry("common.smtprof")
```

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -160,22 +170,33 @@ object Config {
...
+      init: Option[String] = None,
+      inv: Option[List[String]] = None,
+      next: Option[String] = None,
       length: Option[Int] = Some(10),
       maxError: Option[Int] = Some(1),
       noDeadlocks: Option[Boolean] = Some(false),
       nworkers: Option[Int] = Some(1),
       smtEncoding: Option[SMTEncoding] = Some(SMTEncoding.OOPSLA19),
-      temporal: Option[String] = None, // TODO SHould be list?
+      temporal: Option[List[String]] = None,
```
thpani, line 184: `temporal` sounds like it's a boolean, should this be named `temporalProperties` or something?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -357,80 +384,147 @@ object OptionGroup {
     } yield Typechecker(inferpoly)
   }
 
+  /** Options used to rack specification predicates */
+  case class Predicates(
+      behaviorSpec: BehaviorSpec,
+      cinit: Option[String],
+      invariants: List[String],
+      temporal: List[String],
```
thpani, line 392: See above

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -357,80 +384,147 @@ object OptionGroup {
...
+          val msg = s"  > $basename: IO error when loading the TLC config: ${e.getMessage}"
+          throw new PassOptionException(msg)
+
+        case e: TlcConfigParseError =>
+          val msg = s"  > $basename:${e.pos}:  Error parsing the TLC config file: ${e.msg}"
+          throw new PassOptionException(msg)
+      }
+    }
+
+    // Overrdie TLCConfig from CLI/Config args, reporting the resulting value
```
thpani, line 476:
```suggestion
    // Override TLCConfig from CLI/Config args, reporting the resulting value
```

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -357,80 +384,147 @@ object OptionGroup {
...
+  case class Predicates(
+      behaviorSpec: BehaviorSpec,
+      cinit: Option[String],
+      invariants: List[String],
+      temporal: List[String],
+      tlcConfig: Option[(TlcConfig, File)],
+      view: Option[String])
+      extends OptionGroup
+
+  object Predicates extends Configurable[Config.Checker, Predicates] {
```
thpani, line 397: I think this needs a bit of doc on how it handles TLC configs.
Found it hard to follow based on the code alone.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/VCGenPassImpl.scala
```diff
@@ -1,40 +1,41 @@
...
 
   override def name: String = "VCGen"
 
   override def execute(tlaModule: TlaModule): PassResult = {
     val newModule =
-      options.get[List[String]]("checker", "inv") match {
-        case Some(invariants) =>
+      derivedPreds.invariants match {
+        case List() =>
+          logger.info("  > No invariant given. Only deadlocks will be checked")
```
thpani, line 30: Shouldn't this check for `--no-deadlocks`?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/DerivedPredicates.scala
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * ==Motivation==
+ *
+ * When checking a spec, there is a set of predicates which are singled out for special treatment, such the predicate
+ * which initializes state variables and the predicate that computes state transitions. We call these "specification
+ * predicates".
+ *
+ * A few specification predicates are determined purely on the basis of configuration input. However, most specification
+ * predicates can only be finalized after inspecting the given `TlaModule` through a sequence of several passes. We call
+ * specification predicates that cannot be determined based on configuration input alone "derived predicates". We an
```
thpani, line 16:
```suggestion
 * specification predicates that cannot be determined based on configuration input alone "derived predicates". We can
```

thpani (approved):
Wow, so cool to see the new options in place 😻 
Super impressed by how patiently you hacked your way through the undergrowth!

Just a handful of questions below, nothing blocking.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
shonfeder: oh thanks!

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
shonfeder: Yes, i think that makes sense. I started from just picking up whatever names had been used in the option map, but I agree that's clearer.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/PrimingPassImpl.scala
```diff
@@ -1,20 +1,24 @@
...
-import at.forsyte.apalache.tla.lir.UntypedPredefs._
 import com.google.inject.Inject
 import com.typesafe.scalalogging.LazyLogging
 
 /**
  * PrimingPass adds primes to the variables in state initializers and constant initializers.
  */
-class PrimingPassImpl @Inject() (options: PassOptions, tracker: TransformationTracker, writerFactory: TlaWriterFactory)
+class PrimingPassImpl @Inject() (
+    derivedPreds: DerivedPredicates,
```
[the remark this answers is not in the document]
shonfeder: I'm not sure I follow your question, but if you're asking why this is called `derivedPreds` here, when we aren't actually deriving them or anything, its just so that we have consistent naming in all the passes that access this data structure.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/VCGenPassImpl.scala
shonfeder: Maybe! Perhaps I'll add that as a followup tho, as this just inherits the current logic.

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/passes/ConfigurationPassImpl.scala
```diff
@@ -102,176 +106,70 @@ class ConfigurationPassImpl @Inject() (
...
-      to: WriteablePassOptions): Unit = {
-    for (name <- NormalizedNames.STANDARD_OPTION_NAMES) {
-      from
-        .get[Any]("checker", name)
-        .collect { case value => to.set("checker." + name, value) }
+  private def setDerivedPredicates(): Unit = {
+    val (init, next) = options.predicates.behaviorSpec match {
+      case InitNextSpec(init, next) => (init, next)
+      // Should we change data structure to make this state unrepresentable?
+      case _ => throw new Exception("TODO: should be impossible, since this arises only with no TLCC config")
```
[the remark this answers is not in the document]
shonfeder: Thanks for the fix. I see this also shouldn't be a TODO :D

shonfeder:
Thank you for the many careful reviews on the intermediate changes, @thpani, and for making it thru this last beast :)

I'll take all your suggestions and then merge in :)

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/PrimingPassImpl.scala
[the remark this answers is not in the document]
thpani: I meant that the only kind of predicates at this point will be `DerivedPredicates` at this point, right?
If so, it doesn't add much – rather, might be confusing – to distinguish it by name. But that's maybe just my preference.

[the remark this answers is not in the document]
shonfeder: Do you think this would be better as

```suggestion
    specPredicats: DerivedPredicates,
```

?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/AbstractCheckerCmd.scala
shonfeder: https://github.com/informalsystems/apalache/issues/2117

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
shonfeder: Good call! Most likely I will be the one most helped by this in the immediate future :D 

See 66951c6d935270f31abbe3242b1e8e51ad9d8f3f

test/tla/cli-integration-tests.md
```diff
@@ -1889,11 +1891,12 @@ EXITCODE: OK
 
 When a configuration file does not exist, the tool should error.
 
+NOTE: We truncate the output to avoid printing the file, making the test
+indifferent to the execution environment (including in docker).
```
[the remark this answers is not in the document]
shonfeder: It prints the absolute file name in any environment, which is generally not the same between different computers. I've elaborated here!z

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/VCGenPassImpl.scala
shonfeder: https://github.com/informalsystems/apalache/issues/2118

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/passes/PrimingPassImpl.scala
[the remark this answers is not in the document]
shonfeder: I will lave this for potential followup improvement. I don't think it will cause confusion as is.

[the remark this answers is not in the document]
thpani: Totally, we can keep it as-is!

PR #2079, opened 2022-08-17
Organize website decision records

bugarela:
Hello :octocat: 

We are missing several ADRs from the website's menu, so I'm linking all of them there. In this process, I fixed some small numbering problems.

I also found a problem in how the ADR template's way of defining authors and revision dates is rendered. It currently looks like this:

image

So I've updated the template and existing entries to use a markdown table which renders like this: 

image

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -89,6 +89,7 @@
 
 # Design Documents
 
+- [RFC 001: types and type annotations](./adr/001rfc-types.md)
 - [ADR-002: types and type annotations](./adr/002adr-types.md)
 - [ADR-003: transition executor (TRex)](./adr/003adr-trex.md)
 - [ADR-004: code annotations](./adr/004adr-annotations.md)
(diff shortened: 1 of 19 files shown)


docs/src/SUMMARY.md
```diff
@@ -97,9 +98,16 @@
 - [ADR-007: restructuring](./adr/007adr-restructuring.md)
 - [ADR-008: exceptions](./adr/008adr-exceptions.md)
 - [ADR-009: outputs](./adr/009adr-outputs.md)
+- [RFC-010: Implementation of Transition Exploration Server](./adr/010rfc-transition-explorer.md)
```
shonfeder, line 101: I don't think this should be linked into the manual. It's a purely internal technical doc and doesn't add value to our user documentation IMO.

docs/src/SUMMARY.md
```diff
@@ -97,9 +98,16 @@
...
 - [ADR-011: alternative SMT encoding using arrays](./adr/011adr-smt-arrays.md)
+- [ADR-012: Adopt an ADR Template](./adr/012adr-adopt-adr-template.md)
```
shonfeder, line 103: Ditto for this, I don't think it's relevant for users.

docs/src/SUMMARY.md
```diff
@@ -97,9 +98,16 @@
...
+- [ADR-013: Configuration Management Component](./adr/013adr-configuration.md)
```
shonfeder, line 104: Also this one.

docs/src/SUMMARY.md
```diff
@@ -97,9 +98,16 @@
...
+- [ADR-014: Precise type inference for records and variants](./adr/014adr-precise-records.md)
 - [ADR-015: ITF: informal trace format](./adr/015adr-trace.md)
 - [ADR-016: ReTLA: Relational TLA](./adr/016adr-retla.md)
 - [PDR-017: Checking temporal properties](./adr/017pdr-temporal.md)
 - [ADR-018: Inlining in Apalache](./adr/018adr-inlining.md)
+- [ADR-019: Harmonize changelog management](./adr/019adr-harmonize-changelog.md)
```
shonfeder, line 110: And this one.

docs/src/SUMMARY.md
```diff
@@ -97,9 +98,16 @@
...
 - [ADR-020: Improving membership in arenas](./adr/020adr-arenas.md)
+- [RFC-021: Prioritization of Work](./adr/021rfc-prioritization.md)
```
shonfeder, line 112: And this one! :D

docs/src/SUMMARY.md
```diff
@@ -97,9 +98,16 @@
...
+- [ADR-022: Unify Configuration Management and "Pass Options"](./adr/022adr-unification-of-configs-and-options.md)
```
shonfeder, line 113: This one is probably valuable for users to have access to. Thank you for including it!

shonfeder (approved):
Thank you for all the fixes!

I think most of the omitted ADRs should not be included in the published user docs (and they were omitted intentionally with two exceptions, iirc), because afaict they can't possibly be of value to the users for whom the manual is written. WDYT?

shonfeder:
We talked about this in the standup, and I'm happy to have these added into the published manual docs, as I can understand the reasoning and motive for including them, even if it doesn't perfectly align with my preferred style :D

PR #2080, opened 2022-08-18
Add specification of the OptionGroup classes and traits

shonfeder:
Works towards #2040

## Context

This PR factors out just the addition of the `OptionGroup` classes and related
traits into its own PR. There is ongoing work to in #2074 to integrate the
options provider defined here into the rest of the code base, but this hunk can
be factored out for a more digestable PR review.

You can see how these the `OptionGroups` are being used by taking a peak at #2074.

## Changes

- Add case classes to represent required sets of options, "option groups", with
  the groupings and namings closely following the existing `PassOption` options
  map. (If we want, we can refactor to adjust the names and hierarchy much
  easier once this in place.)
- Add traits that specify combinations of options and classes to construct those
  combinations.
  

### DIFF
--- a/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
+++ b/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
@@ -2,9 +2,24 @@ package at.forsyte.apalache.infra.passes.options
 
 import java.io.File
 import at.forsyte.apalache.tla.lir.Feature
+import com.google.inject
+import scala.util.{Failure, Success, Try}
+import at.forsyte.apalache.infra.PassOptionException
 
 /**
  * The components of this package specify the configurations and options used to configure Apalache
+ *
+ * Configurations are represented by extension of the [[Config]] trait. Configurations are derived via injective maps
+ * from configuration sources (souch as CLI arguments or .cfg files) to config classes. As such, each value in a config
+ * class is an `Option`, where a value of `None` indicates that value was left unconfigured.
+ *
+ * Options are represeneted by extensions of the [[OptionGroup]] trait. Option groups are typically derived via
+ * surjective maps from from [[Config]]s. As such, for every field in the option group, there must be a value given in
+ * the originating config.
+ *
+ * See
+ * [[https://github.com/informalsystems/apalache/blob/main/docs/src/adr/022adr-unification-of-configs-and-options.md ADR022]]
+ * for motivation and details.
  *
  * @author
  *   Shon Feder
@@ -86,8 +101,8 @@ object Config {
       debug: Option[Boolean] = Some(false),
       smtprof: Option[Boolean] = Some(false),
       configFile: Option[File] = None,
-      writeIntermediate: Option[Boolean] = None,
-      profiling: Option[Boolean] = None,
+      writeIntermediate: Option[Boolean] = Some(false),
+      profiling: Option[Boolean] = Some(false),
       features: Option[Seq[Feature]] = Some(Seq()))
       extends Config[Common] {
 
@@ -145,12 +160,9 @@ object Config {
       config: Option[String] = None,
       discardDisabled: Option[Boolean] = Some(true),
       cinit: Option[String] = None,
-      // TODO Set default here once ConfigurationPassImpl is fixed
-      init: Option[String] = None,
-      // TODO Should be list?
-      inv: Option[String] = None,
-      // TODO Set default here once ConfigurationPassImpl is fixed
-      next: Option[String] = None,
+      init: Option[String] = None, // TODO Set proper default here once ConfigurationPassImpl is fixed
+      inv: Option[String] = None, // TODO Should be list?
+      next: Option[String] = None, // TODO Set proper default here once ConfigurationPassImpl is fixed
       length: Option[Int] = Some(10),
       maxError: Option[Int] = Some(1),
       noDeadlocks: Option[Boolean] = Some(false),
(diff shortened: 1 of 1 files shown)


mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -234,3 +249,298 @@ object SMTEncoding {
...
+          debug = debug,
+          features = features,
+          outDir = outDir,
+          profiling = profiling,
+          runDir = common.runDir, // Remains optional
+          smtprof = smtprof,
+          writeIntermediate = writeIntermediate,
+      )
+    }
+  }
```
shonfeder, line 333: I'm not pleased with how much typing was involved in all this, but I couldn't think of a cleaner way to get the desired guarantees (i.e., of a fully configured set of options for passes) without doing some real non-trivial type astronaut stuff with shapeless.

I'm open to other ideas tho!

The upside is that at least the logic here is dead simple.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -234,3 +249,298 @@ object SMTEncoding {
...
+  trait HasTypechecker extends HasIO {
+    val typechecker: Typechecker
+  }
+
+  // This is the maximal interface, that should always be the greatest upper
+  // bound on all combinations of option groups
+  trait HasAll extends HasTypechecker with HasChecker
+
+  //////////////////
+  // Constructors //
```
shonfeder, line 473: These constructors is another place where I wish we could get away with a bit less boilerplate, but I don't know how it could be done.

docs/src/adr/022adr-unification-of-configs-and-options.md
```diff
@@ -3,7 +3,7 @@ authors: Shon Feder
 last revised: 2022-08-15
 ---
 
-# ADR-003: Unify Configuration Management and "Pass Options"
+# ADR-022: Unify Configuration Management and "Pass Options"
```
shonfeder, line 6: Just happened to notice this weird typo will filling in the docs.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -234,3 +246,295 @@ object SMTEncoding {
...
+      } yield Common(
+          command = command,
+          debug = debug,
+          features = features,
+          outDir = outDir,
+          profiling = profiling,
+          runDir = common.runDir, // Remains optional
+          smtprof = smtprof,
+          writeIntermediate = writeIntermediate,
+      )
```
thpani, line 328: Feels like a lot of this boilerplate could be avoided if we use reflection – have you considered that?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -234,3 +246,295 @@ object SMTEncoding {
...
+  case class Common(
+      command: String,
+      debug: Boolean,
+      features: Seq[Feature],
+      outDir: File,
+      profiling: Boolean,
+      runDir: Option[File],
+      smtprof: Boolean,
+      writeIntermediate: Boolean)
+      extends OptionGroup
```
thpani, line 299: So at this point adding new options becomes a bit expensive with boilerplate in multiple places. I'm wondering if we cannot recycle the `Config` objects and dynamically ensure that the required options are set?

thpani (commented):
From a typed config perspective this looks awesome!

I'm a bit concerned about the amount of boilerplate we're introducing here. In particular, the work needed for adding new options/flags seems to become non-trivial at this point.
My main questions boil down to:

1. Should we maybe revisit what guarantees we want from this? Looks like we could get away with recycling `Config` objects for options and dynamically ensure that required fields are set?
2. Feels like some of the boilerplate could be avoided by using reflection?
3. A technical question why we need a Guice provider?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
shonfeder: I have not. Could you give any pointers to documentation or examples how to achieve this kind of thing with reflection?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
shonfeder: Right. I wanted to be sure people were ok with this before doing most of the work, which is why I asked in our stand ups :)

In the most common case, 3 entries are required:

- The cli arg
- The the mapping to the config class
- The mapping to the option class

That should be about 5 changes all together, including adding the fields and maps. It is a fair bit of boilerplate.

keep in my that with the current system, 3 entries are also required to add a cli arg that can be set by configuration:

- cli arg
- mapping to config
- mapping to option map

Plus validation logic is needed at any access of non-optional values and we have to cast them to the right types at access time. So while I’m not thrilled about the amount of boiler plate, afaict this doesn’t seem to make matters worse? I *think* it’s about the same amount of overhead, only it makes it possible to reason about and get static guarantees. 

It does make the amount of overhead more apparent tho, because it puts it all in one place :)

I’ve considered reusing the config and have these questions around it:

- then do we give up on static guarantees around which subset of the configs are used? Or are you thinking we’d use the same interface pattern, still?
- How do we specify the required options and provide safe access to them without writing approximately the same amount of boilerplate? Or are you suggesting we just make each access of required options have responsibility for possibly crashing/throwing?
- We would lose the ability to introduce an abstraction barrier between options provided and config values supplied. We’re only taking advantage of this in a few minutes laces currently (with input config group and when m dropping the config file in the options), but to me it makes sense that the shape of the configurations users can supply will not in general be isomorphic to the shape of the options we need to run process. What are your thoughts on this trade off?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -2,10 +2,25 @@ package at.forsyte.apalache.infra.passes.options
...
 
 /**
  * The components of this package specify the configurations and options used to configure Apalache
  *
+ * Configurations are represented by extension of the [[Config]] trait. Configurations are derived via injective maps
+ * from configuration sources (souch as CLI arguments or .cfg files) to config classes. As such, each value in a config
+ * class is an `Option`, where a value of `None` indicates that value was left unconfigured.
+ *
+ * Options are represeneted by extensions of the [[OptionGroup]] trait. Option groups are typically derived via
+ * surjective maps from from [[Config]]s. As such, for every field in the option group, there must be a value given in
```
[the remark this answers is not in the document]
shonfeder: The OptionGroup is the co domain. That’s meant to be indicated by the “derived via”: we map out of the Config into an OptionGroup.

Is there a concise rephrasing that you think would make this clearer?

shonfeder:
> A technical question why we need a Guice provider?

It’s the only way I’ve been able to find that will allow us to insert the data derived from our configs into the dependency injection chain. See the accompanying PR linked for what that looks like.

Basically the current OptionMap is an ad hoc collection of guice providers, with each key being a provider of its value.

I’m open to other ideas if you know of a better way!

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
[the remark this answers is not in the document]
thpani: Maybe I was just confused because in the similar paragraph about `Config` above you explicitly stated the codomain (so I assumed it's not implicitly obvious).

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
thpani: I guess my concern stems from (a) the assumption that all the entries would be in the same place / file, and (b) a lack of understanding of how the configs will eventually be wired through. Maybe it's already obvious how to introduce these entries if one actually wants to access an option in code?

thpani: > I _think_ it’s about the same amount of overhead

Not sure it's the same amount overhead; onboarding for a mutable map is easier than having to understand a custom class structure (not saying that I prefer the mutable map, though ;) )

> * then do we give up on static guarantees around which subset of the configs are used? Or are you thinking we’d use the same interface pattern, still?

Anything preventing us from using it?

> * How do we specify the required options and provide safe access to them without writing approximately the same amount of boilerplate? Or are you suggesting we just make each access of required options have responsibility for possibly crashing/throwing?

My understanding was that even non-required options have some kind of default value. I would assume that only fully populated `Config` objects reach the actual option-consuming code.

> * We would lose the ability to introduce an abstraction barrier between options provided and config values supplied. We’re only taking advantage of this in a few minutes laces currently (with input config group and when m dropping the config file in the options), but to me it makes sense that the shape of the configurations users can supply will not in general be isomorphic to the shape of the options we need to run process. What are your thoughts on this trade off?

That's a good point. Currently we're not making much use of it? Also sounds to me like such "composite" options should be computed values, and the logic for them could reside inside the configuration classes?

thpani:
> It’s the only way I’ve been able to find that will allow us to insert the data derived from our configs into the dependency injection chain. See the accompanying PR linked for what that looks like.

I looked at the PR but couldn't fully figure out why it's needed.
Is it that we cannot statically build the configuration at the point that we're setting up the Guice bindings? If so, it seems that we're just doing it in the wrong order?

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
shonfeder: > Not sure it's the same amount overhead

I meant the overhead in terms of lines of code to write. The *think* the cognitive overhead here more or less follows the general tradeoff with static vs. dynamic typing: this approach moves some complexity into the types, but removes some complexity from the logic around values (e.g., you don't need validation and type casting logic at time of access).

shonfeder:
Thanks to sync chat with @thpani, we have two ideas for improvement:

- Inject the constructed options class for binding by Guice directly by supplying a class parameter, which might avoid needing a provider.
- Derive the OptionGroup automatically by converting to and from a lightbend `config` object.

shonfeder:
Turns out deriving `OptionGroup`s from `Config`s via `PureConfig` will be more invasive than I expected. We are doing a few value conversion when generating the `OptionGroup` values, and the shape of the two top-level classes is not isomorphic. This change is still feasible, but will require some changes to the CLI, some changes to `ConfigManager` and some changes to `OutputManager`. 

This is something we should pursue, I think, but can probably best left as a followup refinement so we don't get too much WIP piled up.

So I'll just fix the concrete improvements and merge this in, and save the WIP for the automatic derivation for future work.

shonfeder:
I'll remove the provider in the followup PR, assuming that turns out to work as expected!

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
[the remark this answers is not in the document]
shonfeder: Ah yeah, that makes sense. I made the parallel more consistent. Thanks.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -2,10 +2,25 @@ package at.forsyte.apalache.infra.passes.options
...
 import at.forsyte.apalache.tla.lir.Feature
+import com.google.inject
+import scala.util.{Failure, Success, Try}
+import at.forsyte.apalache.infra.PassOptionException
 
 /**
  * The components of this package specify the configurations and options used to configure Apalache
  *
+ * Configurations are represented by extension of the [[Config]] trait. Configurations are derived via injective maps
+ * from configuration sources (souch as CLI arguments or .cfg files) to config classes. As such, each value in a config
```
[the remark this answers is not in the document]
shonfeder: I rephrased and added a line about the various places maps get defined.

shonfeder:
Thanks again for the careful review! :pray:

PR #2081, opened 2022-08-18
Update sonatype resolver method

shonfeder:
This addresses the deprecation warning https://github.com/sbt/librarymanagement/pull/404/files
which is currently creating some noise in our CI logs.


shonfeder:
The fix is making things more broken, and is anyhow totally inessential.

PR #2084, opened 2022-08-19
improve the warning about large sets

konnov:
This is a small change in the warning about expanding power sets and function sets. The previous warning is annoying, especially when `SUBSET S` is applied to `S` of cardinality of 1 or 2.

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExpansionMarker.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExpansionMarker.scala
@@ -35,8 +35,6 @@ class ExpansionMarker @Inject() (tracker: TransformationTracker) extends TlaExTr
       val tag = ex.typeTag
       if (shallExpand) {
         // Expand the set as well as the underlying set!
-        logger.warn(s"The set S = $ex will be expanded, producing 2^Cardinality(S) constraints.")
-        logger.warn("Expansion may significantly slow down the solver, e.g., when Cardinality(S) >= 10.")
         OperEx(ApalacheOper.expand, OperEx(op, transform(true)(underlyingSet))(tag))(tag)
       } else {
         // Do not expand the set itself, but expand the underlying set!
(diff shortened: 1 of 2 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExpansionMarker.scala
```diff
@@ -35,7 +35,8 @@ class ExpansionMarker @Inject() (tracker: TransformationTracker) extends TlaExTr
       val tag = ex.typeTag
       if (shallExpand) {
         // Expand the set as well as the underlying set!
-        logger.warn(s"The set $ex will be expanded. This will blow up the solver.")
+        logger.warn(s"The set S = $ex will be expanded, producing 2^Cardinality(S) constraints.")
+        logger.warn("Expansion may significantly slowdown the solver, e.g., when Cardinality(S) >= 10.")
```
thpani, line 39:
```suggestion
        logger.warn("Expansion may significantly slow down the solver, e.g., when Cardinality(S) >= 10.")
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExpansionMarker.scala
```diff
@@ -46,7 +47,10 @@ class ExpansionMarker @Inject() (tracker: TransformationTracker) extends TlaExTr
       val tag = ex.typeTag
       if (shallExpand) {
         // Expand everything, including the function set.
-        logger.warn(s"The set $ex will be expanded. This will blow up the solver.")
+        logger.warn(
+            s"The function set [ S -> T ] = $ex will be expanded, producing Cardinality(T)^Cardinality(S) constraints.")
+        logger.warn(
+            "Expansion may significantly slowdown the solver, e.g., when Cardinality(T) >= 4 or Cardinality(S) >= 5.")
```
thpani, line 53:
```suggestion
            "Expansion may significantly slow down the solver, e.g., when Cardinality(T) >= 4 or Cardinality(S) >= 5.")
```

thpani, line 53: Given that this depends on both base and exponent, should it be "and"?

```suggestion
            "Expansion may significantly slowdown the solver, e.g., when Cardinality(T) >= 4 and Cardinality(S) >= 5.")
```

thpani: On a second thought, seems like you want to indicate that things slow down above ~1000 constraints. That's somewhat lost by just giving an example; should we specify that number explicitly?

```suggestion
            "Expansion may significantly slowdown the solver, e.g., when Cardinality(T)^Cardinality(S) >= 1000.")
```

thpani (approved):
LGTM

Given that we already print thresholds in the warning, does it make sense to only log these warnings when above the thresholds?

konnov:
I will move the first message to the point where the powerset is actually expanded. Since we do not have an implementation for function sets, it will stay where it is.

PR #2091, opened 2022-08-23
Introduce --features=imprecise.records and test type annotations

### LINKED ISSUE #1926: Show an error when variants or new records are used without --features=rows
To tell the users that they are using the experimental feature without enabling it.

### LINKED ISSUE #1943: Switch to new records and make the old records optional
Introduce a feature flag `--features=imprecise.records` to optionally enable the old records, make --features=rows the default

- [x] Introduce the optional feature `--features=imprecise.records`
- [x] Make `--features=rows` the default setting

konnov:
Closes #1926. Closes #1943.

- [x] Following #1943, This PR introduces a new feature flag `imprecise.records`, which is the opposite of `rows`.
- [x] As requested in #1926, we have to warn the users when they are using the old type annotations with `rows` and when they are using new type annotations with `imprecise.records`. This check is now done in `TypeCheckerTool`.
- [x] To make the check in `TypeCheckerTool`, I had to refactor `ToEtcExpr`. It does not receive `AnnotationStore` anymore, but just a map from identifiers to types. This nicely isolates annotation processing and additional checks in `TypeCheckerTool`.
- [x] The feature `rows` is the default now, that is, it is only kept for transitional compatibility with the documentation.
- [x] The feature `imprecise.records` is not the default anymore, it has to be explicitly enabled.

- [ ] Documentation added for any new functionality. Postponed for another PR.

### DIFF
--- a/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/ApalacheCommand.scala
+++ b/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/ApalacheCommand.scala
@@ -38,8 +38,8 @@ abstract class ApalacheCommand(name: String, description: String) extends Comman
       useEnv = true)
   var features = opt[Option[Seq[Feature]]](default = None,
       description = {
-        val featureDescriptions = Feature.all.map(f => s"  ${f.name}: ${f.description}")
-        ("a comma-separated list of experimental features:" :: featureDescriptions).mkString("\n")
+        val featureDescriptions: Seq[String] = Feature.all.map(f => s"  ${f.name}: ${f.description}")
+        ("a comma-separated list of experimental features:" +: featureDescriptions).mkString("\n")
       })
 
   /**
--- a/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
+++ b/tla-types/src/main/scala/at/forsyte/apalache/tla/typecheck/TypeCheckerTool.scala
@@ -1,11 +1,11 @@
 package at.forsyte.apalache.tla.typecheck
 
 import at.forsyte.apalache.io.annotations.store.AnnotationStore
-import at.forsyte.apalache.io.annotations.{AnnotationStr, StandardAnnotations}
+import at.forsyte.apalache.io.annotations.{Annotation, AnnotationStr, StandardAnnotations}
 import at.forsyte.apalache.io.typecheck.parser.{DefaultType1Parser, Type1ParseError}
 import at.forsyte.apalache.tla.lir
-import at.forsyte.apalache.tla.lir._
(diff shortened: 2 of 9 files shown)


shonfeder (approved):
LGTM, except I don't like the feature name.

thpani:
Is this up for review? It's still marked as draft.

konnov:
Now all tests are passing, it is ready

konnov:
> Why is the name of the new feature flag so unusual and not at all resemble or relate to the rows flag if it is its opposite? Do you plan to have many features call imprecise.foo?
>
> How about instead calling it no-rows, or something?

I wanted to call it by the property of the old records. Not sure, many people would get that `no-rows` is worse than `rows` :)

shonfeder:
IMO, we shouldn't worry about "better than" communication in the flag name, since we enable the "best" by default and people should only change the flags if they know what they are doing. I think we should also keep in mind that you implicitly designing the conventions that will be used by feature flags introduces in the future.

I suggest we optimize for clear communication in the flag names, in particular for communicating how feature flags relate to each-other and what the syntax means. These are the aspects I'm worried will cause confusion:

- What does the `.` in the flag indicate? Is there some sort of name-space? Why not separate by `-` if it's just a space? 
- How would a user know that `imprecise.records` has anything at all to do with `rows`, let alone that it is the opposite?

I have two suggestions to address these concerns: 

- afaik, we generally use `-` to separate words in our CLI and config system (see `run-dir`, `out-dir`, `tuning-options` etc.). The only time I see `.` used is for namespacing (e.g., with `smt.random_seed` or `search.invariant.mode` in tuning parameters). I think we should be consistent with this convention. I also suggest we not introduce a name-space under `imprecise` ;P
- Since these two strings toggle on or off the same feature, I think they should be named in a way that indicates this clearly. I don't see the value of "call it by the property of the old records" in this case, especially since the "property" is just indicative of a flaw in the previous system, afaiu. If you think it is really essential this reference "records" then maybe we should also change the name of the "rows" feature?

Putting these two lines of reasoning together, I suggest we find an option like one of these three:

- Use `.` separated name-spacing: 
  - `rows.enabled` vs. `rows.disabled` with the current string `rows` being the same as `rows.enabled`.
  - `record-typing.precise` vs. `record-typing.imprecise` with `rows` being the same as the first.
- Use `-` to separate words, and refer to records, but and introduce a symmetrical name for the row typing: `imprecise-records` vs. `precise-records` with `rows` being an alias for the latter

konnov:
@shonfeder your reasoning makes sense to me. The users will get an error message, when they are using the old records, which will tell them to either update their annotations or enable `--no-rows`. Let me sleep over it.

konnov:
Renamed the feature to `--no-rows` and enabled auto-merge

shonfeder (approved):
LGTM. Thanks!

PR #2094, opened 2022-08-24
Add debug logging of adapted exceptions

shonfeder:
The `ExceptionAdapter` currently throws away the stack trace of
exceptions it intercepts, and only the message is preserved. As a
result, bug reports and logs are less useful than they could be.

This small change just adds a debug level of the full exception before
passing it on to the `AdaptedException` structure for reporting by the
tool. The new traceback output is included in the debug output we
generate for filing bug reports, but are not shown in the user's console
under normal operation.


thpani (approved):
This looks incredibly useful, thanks for the change!

konnov (approved):
That's a very small but extremely neat addition :)

shonfeder:
Thanks for the reviews :)

Issue #2100, opened 2022-08-24
Update language manual with new record types


konnov:
The page on [records](https://apalache.informal.systems/docs/lang/records.html) needs an update

Issue #2099, opened 2022-08-24
Update idiom 15 with row types


konnov:
[Idiom 15](https://apalache.informal.systems/docs/idiomatic/003record-sets.html) refers to old record types. We have to update this page.

PR #2101, opened 2022-08-24
Update ADR002, tutorial, and howto

### LINKED ISSUE #2093: Update the manual with `no-rows`
Once #2091 is merged, we have to update the manual with an explanation about `--features=imprecise.records`. Also, we need a short description of how to migrate from imprecise records to row records.

konnov:
Closes #2093.

This PR contains updates regarding the transition to Type Systems 1.2:

 - [x] ADR002
 - [x] The Snowcat tutorial
 - [x] The HOWTO on types

The manual contains two other documents that are using the old syntax for record types. They are postponed for #2099 and #2100.

### DIFF
--- a/docs/src/HOWTOs/howto-write-type-annotations.md
+++ b/docs/src/HOWTOs/howto-write-type-annotations.md
@@ -528,12 +528,12 @@ System 1][] (TS1) as follows:
 ### Case 1: plain records
 
 Many specifications are using plain records. For instance, they do not assign
-records of different domains to the same variable. Nor they mix records of
+records of different domains to the same variable. Nor do they mix records of
 different domains in the same set. Plenty of specifications fall into this
 class.
 
 For example, check [Recipe 2](#recipe2). In this recipe, the variable `chan` is
-always carrying a record of the domain `{ "val", "rdy", "ack" }`.
+always carrying a record with the domain `{ "val", "rdy", "ack" }`.
 
 In this case, all you have to do is to replace the old record types of the form
 `[ field_n: T_1, ..., field_n: T_n ]` with the new record types of the form `{
--- a/docs/src/HOWTOs/howto-write-type-annotations.md
+++ b/docs/src/HOWTOs/howto-write-type-annotations.md
@@ -1,8 +1,13 @@
 # How to write type annotations
 
-**Revision:** July 22, 2022
+**Revision:** August 24, 2022
 
 **Important updates:**
+
+ - Version 0.29.0: The new syntax for records and variants is enabled by
+   default (previously, enabled with `--features=rows`). For the transition
+   period, the old type syntax can be activated with `--features=no-rows`.
+   See [Recipe 9](#recipe9) on transitioning to Type System 1.2.
 
  - Version 0.25.10: This HOWTO introduces new syntax for type aliases. See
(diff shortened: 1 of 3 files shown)


docs/src/HOWTOs/howto-write-type-annotations.md
```diff
@@ -492,6 +495,69 @@ Attacks(queens,i,j)
...
+   annotations over records and variants, whose shape is only
+   partially-defined.  For example, `{ foo: Int, bar: Bool, a }` defines a
+   record type that has at least two fields (that is, `foo` of type `Int` and
+   `bar` of type `Bool`), but may have more fields, which are captured with the
+   row variable `a`.
+
+### Case 1: plain records
+
+Many specifications are using plain records. For instance, they do not assign
+records of different domains to the same variable. Nor they mix records of
```
shonfeder, line 531:
```suggestion
records of different domains to the same variable. Nor do they mix records of
```

docs/src/HOWTOs/howto-write-type-annotations.md
```diff
@@ -492,6 +495,69 @@ Attacks(queens,i,j)
...
+different domains in the same set. Plenty of specifications fall into this
+class.
+
+For example, check [Recipe 2](#recipe2). In this recipe, the variable `chan` is
+always carrying a record of the domain `{ "val", "rdy", "ack" }`.
```
shonfeder, line 536:
```suggestion
always carrying a record with the domain `{ "val", "rdy", "ack" }`.
```

docs/src/adr/002adr-types.md
```diff
@@ -25,25 +25,40 @@ when writing TLA+ specifications. This document is filling this gap.
...
-        // warning: the following type will be removed in Type System 1.2
-      | '[' field ':' T ',' ...',' field ':' T ']'
+        // operators
       | '(' T ',' ...',' T ')' '=>' T
+        // constant types (uninterpreted types)
       | typeConst
+        // type variables
       | typeVar
+        // groups
       | '(' T ')'
```
shonfeder, line 59: What are "groups"?

docs/src/adr/002adr-types.md
```diff
@@ -270,23 +278,26 @@ captures all interesting cases that occur in practice. Obviously, this type
...
+
+```tla
+{[type |-> "1a", bal |-> 1], [type |-> "2a", bal |-> 2, val |-> 3]}
+```
+
+This pattern often occurs in practice, e.g., see [Paxos][].  The type of that
+set will be `Set([type: Str, bal: Int, val: Int])`, which is probably not what
+you expected, but it is the best type we can actually compute without having
+algebraic datatypes in TLA+. It also reminds the user that one better tests the
+field `type` carefully.
```
shonfeder, line 292:
```suggestion
algebraic datatypes in TLA+. It also reminds the user that one must test the
field `type` carefully.
```

shonfeder (approved):
LGTM!

Just a few minor suggestions.

docs/src/adr/002adr-types.md
konnov: well, these are just normal parentheses, which are normally use to change associativity. I will change it to something about associativity

PR #2110, opened 2022-08-26
Deduplicate assembly of custom TLA modules

### LINKED ISSUE #2097: Introduce the Options.tla module
@shonfeder has written a nice explanation of option types in the tlaplus group:

https://groups.google.com/g/tlaplus/c/XiiJ2zg_f3U/m/RJbHjVx1AQAJ

It probably makes sense to extract this example into a module called like `Options.tla`. This module could also introduce the default value for the `UNIT` type, which seems to be a common question too.

shonfeder:
Towards #2097

This replaces the hard-coded copying of our custom TLA modules into the
tla2sany directory with a general solution for any `.tla` files in our
`src/tla` directory.


konnov (approved):
That's a great idea! The old way was kind of tedious :)

PR #2112, opened 2022-08-29
Fix cell type in DomainRuleWithArrays

rodrigo7491:
- [ ] ~[Entries added to `./unreleased/`][changelog format] for any new functionality~

This PR propagates the use of `RecRowT1` in `DomainRule` to `DomainRuleWithArrays`.


Kukovec (approved):

konnov (approved):
Do I see it right that this is simply the case that did not account for row records? Perhaps, we should go over other arrays rules and see if they did not account for row records.

rodrigo7491:
Exactly. Not having the row records in `apply` led to `rewriter error: DOMAIN x where type(x) = CellTFrom({ id: Str, round: Int, src: Str, type: Str }) is not implemented` in a spec I'm experimenting with. I did a quick check on the other array rules and this is the only one that have records on `apply`, so I think this problem is limited to this rule.

PR #2134, opened 2022-08-31
Add de/serialization of `SourceOption` values

### LINKED ISSUE #2129: Add a pureconfig converter that can deserialize a `SourceOption` from configuration input

shonfeder:
Closes #2129

When receiving data from RPC calls, we plan to use our configuration loading
system to read out the application configuration from a JSON payload. This will
need to include the source of the specification, which we represent as a value
of type `SourceOption`. The changes introduced here improve the serialization of
values of this type.

This PR also adds an option `--dump-config=<FILE>` that will write the derived
config of a command into the specified file. This has proved useful for testing,
I think could be useful for other purposes. However, I'm holding off on writing
documentation for it, in case reviewers think the CLI flag shouldn't introduced.

### DIFF
--- a/.unreleased/features/2134-cfg-dump-on-debug.md
+++ b/.unreleased/features/2134-cfg-dump-on-debug.md
@@ -1,0 +1,2 @@
+The application configuration is now dumped into the `run-dir` when the
+`--debug` flag is supplied (see #2134).
--- a/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
+++ b/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
@@ -106,8 +106,6 @@ object Config {
    *   Whether or not to write general profiling data into the `runDir`
    * @param features
    *   Control experimental features
-   * @param dumpConfig
-   *   A file into which the loaded configuration can be written
    */
   case class Common(
       command: Option[String] = None,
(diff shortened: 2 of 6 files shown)


thpani (commented):
Code looks good to me!

I expect this feature to be useful mostly for debugging, so not sure if it's useful to hide it behind a flag.
I'd rather log it at DEBUG level or write to a separate file in the output manager directory by default. Curious what @konnov @Kukovec think.

konnov (approved):
The dumped config looks really nice. Thanks for taking care of it! This work should definitely ease interoperability.

konnov:
> I expect this feature to be useful mostly for debugging, so not sure if it's helpful to hide it behind a flag.
> I'd rather log it at DEBUG level or write to a separate file in the output manager directory by default. Curious what @konnov @Kukovec think.

Yeah, I would enable it in case `--debug` is passed. Also, it would be useful to have it in the automatic bug report.

shonfeder:
Makes sense. I think it might make more sense to treat this just as intermediate file, enabled with `writeIntermediate`? That seems to fit how we are currently handling generation of this kind of artifact.

thpani:
Seems like a sibling of `run.txt` to me, maybe this can go here?

https://github.com/informalsystems/apalache/blob/5e198515d0cb15786ea7aec3e83d828e7bfba989/mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala#L52-L60

shonfeder:
Make sense. I did that! Have another look when you have a chance @thpani ?

shonfeder:
I will merge this in actually, having taken the advice. I can always makes additional corrections in followups. Thanks for the reviews!

Issue #2014, opened 2022-07-26
Shai: Expose `typecheck` functionality via an RPC


shonfeder:
Just like #2013, with the same rationale, but for typechecking.

PR #2138, opened 2022-09-01
Add CmdExecutor RPCs

### LINKED ISSUE #2021: Shai: Expose `parse` functionality as an RPC call
Like #2013 

This functionality will overlap with the `loadModel` rpc added for the `TransExplorer` service, and reuse most of the same code. The main difference is that it will not store the loaded model for stateful exploration, but will function as the `parse` subcommand, and help fill out the RPCs provided by #2020.

shonfeder:
Closes #2021, #2014, and #2013

This plumbs through gRPC calls to dispatch the CLI commands for `parse`, `check`, and `typecheck`.

Note that the returned output is not currently well structured. This is to be added in followup work that will require working out a solution for returning structured counterexample and error data from the `PassChainExecutor`.


shai/src/main/protobuf/cmdExecutor.proto
```diff
@@ -10,4 +10,26 @@ option objc_class_prefix = "TEP";
 package shai;
 
 service CmdExecutor {
+  rpc run(CmdRequest) returns (CmdResponse);
+}
+
+message CmdRequest {
+  Cmd cmd = 1;
+
+  string config = 2;
```
thpani, line 19: Does this also take serialized JSON, like `result` and `failure` below?
Maybe we should add a comment.

shai/src/main/scala/at/forsyte/apalache/shai/v1/CmdExecutorService.scala
```diff
@@ -18,6 +35,63 @@ import com.typesafe.scalalogging.Logger
...
+    for {
+      cfg <- parseConfig(cfgStr)
+
+      toolModule <- {
+        import OptionGroup._
+        cmd match {
+          case Cmd.PARSE           => convErr(WithIO(cfg)).map(new ParserModule(_))
+          case Cmd.CHECK           => convErr(WithCheckerPreds(cfg)).map(new CheckerModule(_))
+          case Cmd.TYPECHECK       => convErr(WithTypechecker(cfg)).map(new TypeCheckerModule(_))
+          case Cmd.Unrecognized(_) => throw new Exception("invalid: toolModuleOfCmd applied before validateCmd")
```
thpani, line 64: What does `toolModuleOfCmd` refer to? Should it be `executeCmd`?

thpani: Since this will be propagated to the client-side, it's probably better to just report that an unsupported command has been passed?

thpani (approved):
Looks great, I only have 2 small comments inline:

shai/src/main/scala/at/forsyte/apalache/shai/v1/CmdExecutorService.scala
shonfeder: The user facing error gets handled during validation here: https://github.com/informalsystems/apalache/blob/9440e380656cbc7580ee48a8527b787a9400debf/shai/src/main/scala/at/forsyte/apalache/shai/v1/CmdExecutorService.scala#L78-L85

This case should be impossible. The error message is out of date! Thanks :)

shonfeder:
Comments addressed. Thanks for the reviw!

shai/src/main/scala/at/forsyte/apalache/shai/v1/CmdExecutorService.scala
thpani: Already merged, but can I suggest to `s/Exception/IllegalArgumentException` in this case?

shonfeder: Oh good idea. Can add in a followup.

Issue #2143, opened 2022-09-02
Add an additional search parameter to `simulate`


konnov:
As raised in #2139, it would be good to extend `simulate` with an additional option `--max-error-per-run=k` that changes the behavior of the search in conjunction with `--max-error=m` as follows:

 - For a randomly generated prefix of actions, the model checker keeps finding errors until `k` errors are found for the prefix, or `m` errors were generated in total (for all prefixes visited so far).
 - Once `k` errors were found for a prefix, switch to another prefix.
 - If `m` errors were found in total, terminate the search.

@andrey-kuprianov, is it what you had in mind in #2139?

@andrey-kuprianov, @tesnimab, what is the priority of this issue? We planned to work on the symbolic simulator in Q4.

thpani:
Given the outgrowth of CLI flags, maybe there is a default we can implement first?
We can always expose more configurable behavior if it's requested by the users.

andrey-kuprianov:
@konnov yes, I think this is very reasonable. It is not really urgent, rather a nice-to-have from my point of view. My guess is that it's probably also better to wait for the full-fledged server mode before diving in here, as the server mode should provide better means for exploration? Only guessing here.

konnov:
With the server mode, you would need us to implement anything in the addition. You would just implement your exploration strategy on top of the client library.

PR #2144, opened 2022-09-02
report an error when --max-error > 1 and no --view is provided

### LINKED ISSUE #1004: [BUG] When using `--max-error=N` where `--view` is not provided, Apalache reports the same counterexample N times
## Description

See title.

## Input specification

Any, where `Inv` has multiple counterexamples.

`apalache-mc check --inv=Inv --max-error=10 MySpec.tla `

## Expected behavior

N(+1) unique counterexamples.

Suggestion: force `--view` whenever `--max-error` is given

### LINKED ISSUE #2139: Employing Apalache for test case generation
## Problem description

I am trying to use Apalache for test case generation, and have found that it behaves in an interesting way; would appreciate any help to improve my understanding.

Suppose I have the following (toy) spec, for which I want to generate test cases:

```tla
---- MODULE Test5 ----
EXTENDS Integers
VARIABLE
    \* @type: Int;
    x,
    \* @type: Int;
    y

Init == x = 0 /\ y = 0

Next == \/ x' = x + 1 /\ UNCHANGED <<y>> 
        \/ y' = y + 1 /\ UNCHANGED <<x>>

Sum5And2x == x + y >= 5 /\ x >=2
NotSum5And2x == ~Sum5And2x

Sum5And2y == x + y >= 5 /\ y >=2
NotSum5And2y == ~Sum5And2y

======================
```

So either `x` or `y` is incremented at each step, and I am interested in traces where the sum reaches 5, and either `x` or `y` is incremented at least twice.

I do it by running Apalache like this: `apalache check --inv=NotSum5And2x --max-error=10 Test5.tla`. I expect to get 10 different traces where `x` is implemented at least twice. Instead, I get 10 identical traces like this:

```tla

EXTENDS Test5

ConstInit == TRUE

(* Initial state *)
State0 == x = 0 /\ y = 0

(* Transition 0 to State1 *)
State1 == x = 1 /\ y = 0

(* Transition 0 to State2 *)
State2 == x = 2 /\ y = 0

(* Transition 0 to State3 *)
State3 == x = 3 /\ y = 0

(* Transition 0 to State4 *)
State4 == x = 4 /\ y = 0

(* Transition 0 to State5 *)
State5 == x = 5 /\ y = 0

InvariantViolation == x + y >= 5 /\ x >= 2

```

From `apalache check --inv=NotSum5And2y --max-error=10 Test5.tla` I get the traces that increment `y` twice, but also 10 identical ones:

```tla

EXTENDS Test5

ConstInit == TRUE

(* Initial state *)
State0 == x = 0 /\ y = 0

(* Transition 0 to State1 *)
State1 == x = 1 /\ y = 0

(* Transition 1 to State2 *)
State2 == x = 1 /\ y = 1

(* Transition 1 to State3 *)
State3 == x = 1 /\ y = 2

(* Transition 1 to State4 *)
State4 == x = 1 /\ y = 3

(* Transition 0 to State5 *)
State5 == x = 2 /\ y = 3

InvariantViolation == x + y >= 5 /\ y >= 2

```

I can try to use Apalache in simulation mode, like this `apalache simulate --length=5 --max-run=10 --output-traces Test5.tla`. This way I do get 10 different traces, but they are not constrained by the invariant, so I may get traces that do not increment `x` twice, for example; in other words I can't focus the search.

I can try to do invariant checking in simulation mode, like this: `apalache simulate --inv=NotSum5And2x --max-error=10 --length=5 --max-run=10 --output-traces Test5.tla`; but then after the first or second of generated example traces violate the invariant, simulation stops, and again 10 equivalent traces are produced, like this one:

```tla

EXTENDS Test5

ConstInit == TRUE

(* Initial state *)
State0 == x = 0 /\ y = 0

(* Transition 1 to State1 *)
State1 == x = 0 /\ y = 1

(* Transition 1 to State2 *)
State2 == x = 0 /\ y = 2

(* Transition 0 to State3 *)
State3 == x = 1 /\ y = 2

(* Transition 1 to State4 *)
State4 == x = 1 /\ y = 3

(* Transition 0 to State5 *)
State5 == x = 2 /\ y = 3

InvariantViolation == TRUE

```

Note that this trace is different than the one obtained via `apalache check`, because it was obtained via simulation first; but then it is again multiplied 10 times.

To summarize: I don't see the way how to currently employ Apalache for test case generation. Probably I am missing some options or configuration parameters?

## Desired solution

I would like to be able to do the following:
- When used in invariant checking mode, `apalache check --max-error=N` should produce `N` _different_ traces violating the invariant
- When used in simulation mode, `apalache simulate --inv=INV --max-error=N --max-run=M` should do `M` simulation runs even if invariant is violated. If it is violated for one of the simulated traces, then the model checker should try to produce `N` _different_ violating traces that are similar to the current violating trace. The important point is that `M` _different_ simulation traces should be produced in any case, violating the invariant or not.

I don't know whether the desired solution is feasible, only describing my ideal scenario. I would really appreciate your help here in either improving my understanding, or in bringing some enhancements to Apalache.

konnov:
Closes #1004. Closes #2139. This is the simplest fix to the problem that occurred several times. Since a default view (equal to the whole state) would slow down the search beyond reasonable, we simply tell the user to find a good view, when `--max-error > 1`.


mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -580,25 +573,34 @@ object OptionGroup extends LazyLogging {
...
+      )
+      // do some validation
+      configured.foreach(c =>
+        if (checker.view.isEmpty && c.maxError > 1) {
+          throw new PassOptionException(s"Option maxError = ${c.maxError} requires a view. None specified.")
+        })
+      // produce the configured checker
+      configured
+    }
   }
```
shonfeder, line 604: I think the way the validation logic has been added here is more invasive than necessary and makes it harder to reason about the validation logic. 

I've two main points in mind:

1. Each binding in the `for` comprehension is already a validation step, and since we are here validating the value for `maxError`, our validation can be focused on that one binding.
2. The `apply` method is already managing errors via the `Try`, so throwing an exception makes the error handling more complex. Instead, we can use the try to put the throwable in a `Failure`.

Those lead me to suggest this alternative:

```suggestion
    def apply(checker: Config.Checker): Try[Checker] = for {
      algo <- checker.algo.toTry("checker.algo")
      discardDisabled <- checker.discardDisabled.toTry("checker.discardDisabled")
      length <- checker.length.toTry("checker.length")
      noDeadlocks <- checker.noDeadlocks.toTry("checker.noDeadlocks")
      nworkers <- checker.nworkers.toTry("checker.nworkers")
      smtEncoding <- checker.smtEncoding.toTry("checker.smtEncoding")
      tuning <- checker.tuning.toTry("checker.tuning")
      maxError <- checker.maxError.toTry("checker.maxError").flatMap { n =>
        if (n > 1 && checker.view.isEmpty) {
          val msg = s"Option maxError = ${n} requires a view. None specified."
          Failure(new PassOptionException(msg))
        } else {
          Success(n)
        }
      }
    } yield Checker(
        algo = algo,
        discardDisabled = discardDisabled,
        length = length,
        maxError = maxError,
        noDeadlocks = noDeadlocks,
        nworkers = nworkers,
        smtEncoding = smtEncoding,
        tuning = tuning,
    )
```

I think this is fairly readable, but it could perhaps be improved further with a helper function.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
```diff
@@ -51,13 +48,13 @@ class CheckCmd(name: String = "check", description: String = "Check a TLA+ speci
...
-        default = None)
+      description =
+        "do not stop on first error, but produce at most the given number of errors (requires --view when greater than 1), default: 1",
+      default = None)
 
   var view: Option[String] =
-    opt[Option[String]](name = "view",
-        description = "the state view to use with --max-error=n, default: transition index", default = None)
+    opt[Option[String]](name = "view", description = "the state view to use with --max-error=n, default: none",
+      default = None)
```
shonfeder, line 57: iiuc, the logic around this indicates that `view` is only useful when `max-error` is used and `max-error` is only useful when `view` is used. Doesn't this indicate that these should be combined into a single flag?

Could it be that both of these flags are really named after parts of the implication details, and what we actually want is some sort of flag like: `--view-error-traces=view:n`?

Just a thought. Would be curious to know what @thpani thought on this point.

shonfeder (changes requested):

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
thpani: @konnov do we still plan to have better default behavior when `--view` is not provided?
Otherwise I agree with Shon, we should either put them into a single flag or at least have two flags with similar names like `--max-error.num=N` and `--max-error.view=View`.

thpani:
@konnov I think you pointed out here https://github.com/informalsystems/apalache/issues/1004#issuecomment-1235252123 that we also need to update documentation?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/CheckCmd.scala
konnov: There is #2143, which should work without view.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
konnov: True. I had no idea how to do that properly. @shonfeder could you maybe take over this PR. I feel like I am losing the context way too fast :)

shonfeder: Yep, sure! Happy to make the changes here.

shonfeder (approved):

PR #2145, opened 2022-09-02
Expose the JSON representation of TlaModules

shonfeder:
Required for #2013 etc.

In order to incorporate the JSON representation of a TLA module into
replies sent to RPC requests, we need to embed the JSON represention
in other JSON data structures. The current mechanism for generating JSON
of a `TlaModule` only allowed serializing to a string, but didn't
provide a mechanism for obtaining the JSON representation.  This change
exposes the underlying representation.

-------


tla-io/src/main/scala/at/forsyte/apalache/io/json/JsonDecoder.scala
```diff
@@ -5,10 +5,13 @@ import at.forsyte.apalache.tla.lir.{TlaDecl, TlaEx, TlaModule}
...
  * A JsonDecoder defines a conversion from a json (as represented by T) to a TLA+ expression/declaration/module
  *
+ * @tparam R
+ *   The class of the value which the JsonRepresentation uses to reprsent JSON
+ *
  * @tparam T
  *   Any class extending JsonRepresentation
  */
-trait JsonDecoder[T <: JsonRepresentation] {
+trait JsonDecoder[R, T <: JsonRepresentation[R]] {
```
Kukovec, line 14: If you're going to parameterize `JsonRepresentation`, you could just have ` JsonDecoder[R]`, and then have the methods be e.g. `asTlaModule(moduleJson: JsonRepresentation[R]): TlaModule`. Unless you're thinking of having different subclasses of `JsonRepresentation[R]` for the same `R` (which I find unlikely).

shonfeder: I fixed this a different way, using an abstract type member, but it much cleaner and less invasive than what I'd first proposed, so thanks for pushing back.

tla-io/src/main/scala/at/forsyte/apalache/io/json/JsonRepresentation.scala
```diff
@@ -3,8 +3,18 @@ package at.forsyte.apalache.io.json
...
  * A representation of json. The concrete implementation may depend on external json libraries. Defines a toString
  * method, to be used when performing IO.
+ *
+ * @tparam T
+ *   The class of the value which the JsonRepresentation uses to reprsent JSON
  */
 trait JsonRepresentation {
+
+  /** The type of tused to represent JSON */
+  type Value
```
Kukovec, line 13: `T` seems to be unused here, why don't you just make `JsonRepresentation[Value]`?

shonfeder: Oh that documentation comment should be removed.

I don’t want to make JasonRepresentation parametric now because it’s more work and not needed to achieve the end I need.

When trying to parameterize it the way you suggested in our meeting I hit problems due to the sub typing currently used in the interfaces. This way seems clean and easy: a 2 line change :)

shonfeder: Are you worried that this will cause problems some how?

Kukovec: No, it just seemed odd to have both `tparam` and `type`. If you want to just keep value I won't block.

tla-io/src/main/scala/at/forsyte/apalache/io/json/JsonRepresentation.scala
```diff
@@ -3,8 +3,18 @@ package at.forsyte.apalache.io.json
 /**
  * A representation of json. The concrete implementation may depend on external json libraries. Defines a toString
  * method, to be used when performing IO.
+ *
+ * @tparam T
+ *   The class of the value which the JsonRepresentation uses to reprsent JSON
```
shonfeder, line 8:
```suggestion
```

tla-io/src/main/scala/at/forsyte/apalache/io/json/JsonRepresentation.scala
shonfeder: Yeah, the tparam docs where left over.  Sorry I missed that.

Kukovec (approved):

shonfeder:
Thanks for the review!

PR #2148, opened 2022-09-05
Refactor Z3 timeout

thpani:
Check Z3 timeout bounds and safely convert timeout parameter to int.

Z3 takes the timeout in milliseconds as unsigned int. So we can safely store timeout in seconds as (signed) int.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -481,20 +481,24 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext {
...
+      def setTimeout(tm: Int): Unit = {
         val params = z3context.mkParams()
-        params.add("timeout", BigInt(tm).toInt)
+        params.add("timeout", tm)
         z3solver.setParameters(params)
       }
+      // Z3 expects `timeout` to be in milliseconds passed as unsigned int, i.e., with a maximum value of 2^32 - 1.
+      // The Z3 Java API only allows passing it as signed int, i.e., the maximum allowed millisecond value is 2^31 - 1 (`Int.MaxValue`).
+      // Check that `timeoutSec` converted to milliseconds fits without truncation.
+      require(timeoutSec < Int.MaxValue / 1000, s"Expected a timeout of <=${Int.MaxValue / 1000} seconds.")
```
konnov, line 496: Since we know this constraint when parsing the configuration parameters, shall we validate the input there too?

konnov (approved):
Looks good! Thanks for taking care of it!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
thpani: Yes, that's a good point!

thpani: Oh wait, since this is a tuning parameter, we're parsing it as `Map[String, String]`.
We only convert this to `Int` in `ModelCheckerParams`, do you think we should validate it there?

konnov: We could also validate that it is an integer there.

thpani: If we want to validate tuning options, we should probably come up with a solution for all of them – created #2168 to track.

PR #2154, opened 2022-09-07
Update logging libraries

shonfeder:
- Updates slf4j and lockback
- Required one update to an extension of the logback configuration class


thpani (approved):
👍

Issue #1761, opened 2022-05-16
Parallelize unit test runs


thpani:
We currently [disable running some unit tests in parallel](https://github.com/informalsystems/apalache/blob/de1e90b0f01ac84aeb20c64efc74838623c5cb5a/build.sbt#L99) to work around the SANY bug https://github.com/tlaplus/tlaplus/issues/688.

Once that bug is resolved, we should remove `Test / parallelExecution := false` from `build.sbt` to lift the restriction.

shonfeder:
I don't think we'll be able to parallelize the unit tests even once https://github.com/tlaplus/tlaplus/issues/688 is fixed, because Sany seems to rely on a global state, and trying to start two Sany parser sessions concurrently causes the failure noted here: https://github.com/informalsystems/apalache/issues/1114#issuecomment-1180534894

thpani:
To summarize my understanding on this, we'd have to introduce global state in Apalache (like a lock or queue) to sequentialize concurrent calls to SANY. Which should work for the CLI, and server, not sure how the JUnit runner maintains state.

shonfeder:
I prepped the fix for this before seeing that you'd planned to take up on work it, @thpani. Sorry for the race condition on my part!

thpani:
> I prepped the fix for this before seeing that you'd planned to take up on work it, @thpani. Sorry for the race condition on my part!

No worries, I didn't have any cycles for this anyways!
Thanks for picking it up!

PR #2156, opened 2022-09-07
Add mutex around invocation of SANY

### LINKED ISSUE #1963: Protect against Sany's lack of thread safety by ensuring any access to Sany is synchronized
See @thpani's suggestion at https://github.com/informalsystems/apalache/pull/1855#discussion_r918669410

This is to address the problem identified at https://github.com/informalsystems/apalache/issues/1114#issuecomment-1180534894

Note that this probably is not the only thread-unsafe chunk of code in Apalache (the singleton object used for the output manager comes to mind, and we also have #1872), so this could open a bigger can of worms.

shonfeder:
Fixes #1963

This adds a mutex around our invocation of SANY. It is very simple: 

- We add a singleton object to synchronize on
- We call SANY through a method on that object that protects the call with `this.synchronized`

Some pointers to instruction guiding my choice are included in comments.

The tests in #2157 demonstrate the synchronization in effect: for without it we would see test failures from SANY due to concurrent invocations.

:warning: Auto merge enabled.

-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
@@ -13,6 +13,29 @@ import java.nio.file.Files
 import scala.io.Source
 import scala.util.{Failure, Success, Try}
 
+// The SANY parser is not thread safe. This singleton object is used to create a
+// mutex around its execution.
+//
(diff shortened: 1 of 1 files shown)


thpani:
> The tests in #2157 demonstrate the synchronization in effect: for without it we would see test failures from SANY due to concurrent invocations.

This is amazing! 😻
Should also unblock #1761

Kukovec (approved):

PR #2157, opened 2022-09-07
Shai: Remove semaphore

shonfeder:
With #2156 in place, we don't need to put a semaphore around invocations
of our pass executions any more, since the synchronization is assured via
the much more focused scope of invocations of SANY.


shonfeder:
This will need to be rebased on main after #2156 is merged in. Still, review should be straight forward here if the `SanyImporter` is just ignored.

thpani (approved):

PR #2160, opened 2022-09-09
Remove the Executor class

shonfeder:
Grew out of #2138, as I realized that we are ready to abstract the
configuration and execution of pass chains.

Now that we've cleaned up the pass chain configuration and command dispatch,
the intermediate `Executor` class has become a needless indirection. It
was repeating some of the same logic we were doing in the
`PassChainExecutor` and obfuscating the sequence of operations needed to
execute a chain of passes.

In the process of refactoring, it also became clear that the
PassChainExecutor is really just defining a single function, and needn't
be a class, so it's been objectified (does this restore its humanity?).

-------


konnov (approved):
Good refactoring!

shonfeder:
Thanks for the review!

PR #2166, opened 2022-09-13
Fix support for one-line comments in type annotations

### LINKED ISSUE #2162: One line comments inside type annotations aren't supported
We are meant to support comments inside type annotations (inside of comments), as per https://apalache.informal.systems/docs/adr/002adr-types.html?highlight=Comments%20inside%20types#14-comments-inside-types.

However, it looks like the SANY importer annotation parser is stripping new lines out of the annotations, so one line comments become syntactically invalid.

For example, if we run

```tla
------------------- MODULE Test -----------------------
CONSTANT
\* @type: {
\*   // some comment
\*   f : Int
\*};
  test

============================================================
```

thru `SanyImporter`'s `loadFromSource`, the annotation we pull out is:

```
@type(" { // some comment f : Int}")
```

While we have unit tests for the comments (e.g., https://github.com/informalsystems/apalache/blob/0cab80df2a9f6d2d54459e0f60fecdd0d5586b13/tla-types/src/test/scala/at/forsyte/apalache/tla/typecheck/TestDefaultType1Parser.scala#L190-L202), afaict we don't have integration tests to ensure that the comments can actually be parsed out of the annotations we can extract from the annotations.

shonfeder:
Fixes #2162

We heard you like comments in your type annotations inside of comments, so we added support for single-line comments, beginning with `//`, that can be used inside of type annotations inside of multi- and single-line TLA+ comments, so you can comment your annotations while annotating inside your comments.

This required tweaking the lexer and preprocessor so that we preserve new lines (and avoid inserting duplicate new lines, added by SANY), and then updating a bunch of finicky tests to accommodate quirks to intermediate strings.

### DIFF
--- a/.unreleased/bug-fixes/2169.md
+++ b/.unreleased/bug-fixes/2169.md
@@ -1,0 +1,2 @@
+Fix missing support for single-line comments inside of type annotations (see
+https://github.com/informalsystems/apalache/issues/2162) 
--- a/test/tla/CommentedTypeAnnotation.tla
+++ b/test/tla/CommentedTypeAnnotation.tla
@@ -1,0 +1,19 @@
+------------------- MODULE CommentedTypeAnnotation -----------------------
+
+(* Used as regression test for https://github.com/informalsystems/apalache/issues/2163 *)
+
+CONSTANT
+  \* @type: Set({
+  \*   // a comment
+  \*   x : Int
+  \*});
+  v,
+  \* @type: {
+  \*   // another comment
+  \*   x : Int,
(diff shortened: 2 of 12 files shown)


konnov (approved):
This looks like it should work! I am actually surprised by how small the change set was. I was expecting something more complex in the comment preprocessor.

shonfeder:
> I was expecting something more complex in the comment preprocessor.

Yeah, really just a few tweaks! Most of the time was just spent on figuring out the source of the problem and fixing the tests :D

PR #2172, opened 2022-09-20
Use IllegalArgumentException for Unrecognized command in CmdExecutorService

shonfeder:
Followup to #2138, suggested by @thpani in a comment
https://github.com/informalsystems/apalache/pull/2138#discussion_r975077165

-------


thpani (approved):
Thanks for the follow-up! 🙌

PR #2183, opened 2022-09-26
Use case class to represent algorithm option

shonfeder:
Reduce the surface area of our runtime errors a smidge :).

:warning: Automerge enabled.


thpani (approved):
Yippee! 🎉

PR #2189, opened 2022-09-28
ADR23 - Trace evaluation

### LINKED ISSUE #2188: Write ADR for trace evaluation
Title

Kukovec:
- ~[ ] Tests added for any new code~
- ~[ ] Ran `make fmt-fix` (or had formatting run automatically on all files edited)~
- ~[ ] Documentation added for any new functionality~
- ~[ ] [Entries added to `./unreleased/`][changelog format] for any new functionality~

closes #2188

### DIFF
--- a/docs/src/adr/023adr-trace-evaluation.md
+++ b/docs/src/adr/023adr-trace-evaluation.md
@@ -28,7 +28,7 @@ accepting the development costs.\
      This is the story explaining the problem we are looking to resolve.
 -->
 As explained in [#1845](https://github.com/informalsystems/apalache/issues/1845), one often runs into the problem of unreadable counterexamples; 
-for complex specifications, it is often the case that either the sate contains many variables, or the invariant contains many conjunctive components (or both).
+for complex specifications, it is often the case that either the state contains many variables, or the invariant contains many conjunctive components (or both).
 In that case, trying to determine exactly why e.g. `A1 /\ ... /\ An` was violated basically boils down to manually evaluating each formula `Ai` with whatever variables are present in the state.
 This is laborious and error-prone, and should be automated.
 
--- a/docs/src/adr/023adr-trace-evaluation.md
+++ b/docs/src/adr/023adr-trace-evaluation.md
@@ -1,0 +1,75 @@
+# ADR-023: Trace evaluation
+
+| authors                                | last revised    |
+| -------------------------------------- | --------------: |
+| Jure Kukovec                           | 2022-09-28      |
+
+**Table of Contents**
+
+- [Summary (Overview)](#summary)
+- [Context (Problem)](#context)
+- [Options (Alternatives)](#options)
+- [Solution (Decision)](#solution)
+- [Consequences (Retrospection)](#consequences)
+
+## Summary
+
+<!-- Statement to summarize, following the following formula: -->
+
+In the context of improving usability\
+facing difficulties understanding counterexample traces\
+we decided to implement trace evaluation\
+to achieve a better user experience\
+accepting the development costs.\
+
+## Context
+
+<!-- Communicates the forces at play (technical, political, social, project).
+     This is the story explaining the problem we are looking to resolve.
+-->
+As explained in [#1845](https://github.com/informalsystems/apalache/issues/1845), one often runs into the problem of unreadable counterexamples; 
+for complex specifications, it is often the case that either the sate contains many variables, or the invariant contains many conjunctive components (or both).
+In that case, trying to determine exactly why e.g. `A1 /\ ... /\ An` was violated basically boils down to manually evaluating each formula `Ai` with whatever variables are present in the state.
+This is laborious and error-prone, and should be automated.
+
+
+## Options
+
+<!-- Communicate the options considered.
+     This records evidence of our circumspection and documents the various alternatives
+     considered but not adopted.
+-->
+1. Call REPL in each state:
+    - No convenient REPL integration at the moment
+    - No clear way of saving outputs
+1. Encode trace traversal as an Apalache problem and use the solver
+    - No additional rules or IO needed
+
+## Solution
+
+<!-- Communicates what solution was decided, and it is expected to solve the
+     problem. -->
+
+We choose option (2). 
+
+Suppose we are given a trace `s0 -> s1 -> ... -> sn` over variables `x1,...,xk` as well as expressions `E1,...,Em`, such that all free variables of `E1,...,Em` are among `x1,...,xk`. W.l.o.g. assume all constants are inlined.
+
+The above defines a trace `t0 -> t1 -> ... -> tn` over variables `v1,...,vm`, such that 
+```
+ti.vj = Ej[si.x1/x1,...,si.xk/xk]
+```
+
+for all `i \in 0..n, j \in 1..m`. In other words, `vj` in state `ti` is the evaluation of the expression `Ej` in state `si`.
+
+By using TransitionExecutor instead of a generic Next-decomposition, this can be encoded as a specification free of control-nondeterminism, in-state computation, or invariants, and is thus incredibly efficient to represent in SMT.
+
+Then, the solver will naturally return an ITF trace containing the evaluations of `E1,...,Em` in each state `s0,...,sn` (the values of `v1,...,vm`).
+
+
+## Consequences
+
+<!-- Records the results of the decision over the long term.
+     Did it work, not work, was changed, upgraded, etc.
+-->
+
+TBD
--- a/docs/src/adr/023adr-trace-evaluation.md
+++ b/docs/src/adr/023adr-trace-evaluation.md
@@ -20,7 +20,7 @@ In the context of improving usability\
 facing difficulties understanding counterexample traces\
 we decided to implement trace evaluation\
 to achieve a better user experience\
-accepting the development costs.\
+accepting the development costs.
 
 ## Context
 
@@ -85,9 +85,19 @@ The above command should produce an Apalache trace (in all available formats), w
   - If `--expressions=E_1,...,E_m` is used, the variables of the output trace are `E_1,...,E_m`.
   - For all `i,j`, the value of `E_i` in state `j` of the output trace is equal to the evaluation of `E_i`, as defined in `<source>`, using the values the variables of the input trace hold in state `j` of the input trace.
 
+Recall that the output trace will only display the expressions `E_1,...E_m` as the output state variables. Should you wish to view the original trace variables, you need to add an expression, like one of the ones below for example:
+```tla
+E_single == x_1
+E_state == [ x1 |-> x_1, ..., xk |-> x_k ]
+E_vars == <<x_1, ..., x_k>>
+```
+
+
 Optionally, we could investigate one of the following two alternatives to the output format:
   1. The output trace variables are `x_1,...,x_k,E_1,...,E_m` instead, where `x_1,...,x_k` are the variables of the original trace. The value of each variable from the input trace has the same value in every state of the output trace, as it does in the corresponding state of the input trace.
+  This is perhaps preferable to use with the [ITF trace viewer](https://github.com/informalsystems/vscode-itf-trace-viewer).
   2. The output contains _both_ the input trace and the output trace (as it would have been produced in the original suggestion) in the same file, but separately.
+
 
 ## Example
 
--- a/docs/src/adr/023adr-trace-evaluation.md
+++ b/docs/src/adr/023adr-trace-evaluation.md
@@ -52,19 +52,355 @@ This is laborious and error-prone, and should be automated.
 
 We choose option (2). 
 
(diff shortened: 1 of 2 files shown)


docs/src/adr/023adr-trace-evaluation.md
```diff
@@ -0,0 +1,75 @@
...
+to achieve a better user experience\
+accepting the development costs.\
+
+## Context
+
+<!-- Communicates the forces at play (technical, political, social, project).
+     This is the story explaining the problem we are looking to resolve.
+-->
+As explained in [#1845](https://github.com/informalsystems/apalache/issues/1845), one often runs into the problem of unreadable counterexamples; 
+for complex specifications, it is often the case that either the sate contains many variables, or the invariant contains many conjunctive components (or both).
```
konnov, line 31:
```suggestion
for complex specifications, it is often the case that either the state contains many variables, or the invariant contains many conjunctive components (or both).
```

docs/src/adr/023adr-trace-evaluation.md
```diff
@@ -0,0 +1,75 @@
...
+     problem. -->
+
+We choose option (2). 
+
+Suppose we are given a trace `s0 -> s1 -> ... -> sn` over variables `x1,...,xk` as well as expressions `E1,...,Em`, such that all free variables of `E1,...,Em` are among `x1,...,xk`. W.l.o.g. assume all constants are inlined.
+
+The above defines a trace `t0 -> t1 -> ... -> tn` over variables `v1,...,vm`, such that 
+```
+ti.vj = Ej[si.x1/x1,...,si.xk/xk]
+```
```
konnov, line 60: Could you use underscores in the variable names, e.g., `t_i`, `v_j`. It's hard to parse the current write up.

docs/src/adr/023adr-trace-evaluation.md
```diff
@@ -0,0 +1,75 @@
...
+
+for all `i \in 0..n, j \in 1..m`. In other words, `vj` in state `ti` is the evaluation of the expression `Ej` in state `si`.
+
+By using TransitionExecutor instead of a generic Next-decomposition, this can be encoded as a specification free of control-nondeterminism, in-state computation, or invariants, and is thus incredibly efficient to represent in SMT.
```
konnov, line 64: The transition executor is very much in the guts of the model checker. While it is possible to use it, I do not think it is the simplest way of solving the problem.

Why cannot we do the following. Assume we have the following trace:

```tla
State0 ==
  /\ x = 2
  /\ y = 3

State1 ==
  /\ x = 4
  /\ y = 5

State2 ==
  /\ x = 6
  /\ y = 4
```

We could encode the above trace as a sequence of records:

```tla
Trace == <<
  [ x |-> 2, y |-> 3],
  [ x |-> 4, y |-> 5],
  [ x |-> 6, y |-> 4]
>>
```

Next, suppose we want to compute the value of `x + y` in every state of the trace. This could be done by producing the following spec:

```tla
\* the trace as defined above
Trace == << ... >>

VARIABLE
  \* the result of evaluating x + y
  \* @type: Int;
  eval1,
  \* The state number in the trace
  \* @type: Int;
  index

Init ==
  /\ eval1 = Trace[1].x + Trace[1].y
  /\ stateNo = 1

Next ==
  LET nextIndex == index + 1 IN
  /\ index' = nextIndex
  /\ eval1 = Trace[nextIndex].x + Trace[nextIndex].y
```

Note that the above specification can be produced by a preprocessing step. There is no need for direct interaction with the transition executor.

konnov (commented):
It is clear to me that we can solve the problem in many ways. I would prefer to solve it by pure preprocessing (see my comment). Also, this ADR is written in a rush. Could you add an example that illustrates the input and the output. In this case, it should be really easy to do that.

docs/src/adr/023adr-trace-evaluation.md
Kukovec: You could do that, but then you're actually solving additional sequence and record applications in SMT. The fact that the executor is "in the guts of the model checker" is irrelevant; the user doesn't need to know or care. 
If your claim is that the executor is too difficult for Apalache developers to use, however, that is an entirely separate discussion.

docs/src/adr/023adr-trace-evaluation.md
Kukovec: I can, but I don't think markdown prettyprints those.

docs/src/adr/023adr-trace-evaluation.md
konnov: > If your claim is that the executor is too difficult for Apalache developers to use, however, that is an entirely separate discussion.

Why? This is an ADR. We can discuss the implications of development costs and long-term support :)

Kukovec: > Why? This is an ADR. We can discuss the implications of development costs and long-term support :)

I think you have to write your ADRs either assuming the rest of the code is what you've got, OR, write an ADR on how to refactor/remove TEX _first_. Otherwise you get scope creep and your ADRs inevitably become a series of cascading changes and refactoring.

Kukovec: I've explored it a bit, and I suppose we can keep this high level by using transition filtering instead:
```
0->0|1->1|...|n->n
```
should give us the i-th Next in the i-th step while allowing us to just call the model checker pass w/o disassembling it.

docs/src/adr/023adr-trace-evaluation.md
shonfeder: I think you can use latex, in case that would help: https://github.blog/2022-05-19-math-support-in-markdown/

shonfeder (commented):
This proposal looks pretty sensible to me and I think I can understand the motivation.

It's not clear to me exactly what the tradeoffs are between Jure's proposed approach and what Igor is proposing, but it seems like it would make sense to incorporate the alternative Igor is suggesting and add some justification for why we do or do not take that proposed route.

docs/src/adr/023adr-trace-evaluation.md
Kukovec: If you're using trace exploration, user-supplied transition filtering makes no sense; there is exactly one transition between each two given states on a trace.

thpani: > user-supplied transition filtering makes no sense

Any other tuning option then. Currently, tuning options are just a serialized String->String map.

Kukovec: Trace evaluation isn't really typical BMC (it just uses it as a backend). The user should not be providing any additional parameters at all, besides the trace file and a list of expressions.

thpani: Maybe you can explain what "_By using TransitionExecutor instead of a generic Next-decomposition, this can be encoded as a specification_" actually means.
It would be good to include such a specification, as an example. It's hard to argue about things that aren't explained in the ADR.

Kukovec: I actually have a prototype, and I'll demo it tomorrow. Afterwards, we can update the ADR.

Kukovec: Also, this design has been discussed offline for at least 2-3 months before the ADR.

Kukovec: I think a lot of the discussion happening here has nothing to do with _design_, but is nitpicking implementation details.

thpani: > the thing being discussed in a very specific implementation detail, and the prototype has uncontroversial stuff

Then demoing the prototype won't change the discussion here.

shonfeder: I also think it's worth keeping in mind that our use of ADRs are kind of overloaded. iiuc, they were principally intended to be architectural **decision** records: and in that function it makes sense to me we may want a light-weight prototype to help inform our decisions. But we have also often used them as design documents, and perhaps this dual use can cause some confusion w/r/t expectations?

Kukovec: > > Also, this design has been discussed offline for at least 2-3 months before the ADR.
> 
> Apparently not to a point where there's consensus on how to steer the transition exploration. Documenting all options on the table and weighing their benefits is exactly the point of an ADR.

But there _is_ consensus on how to steer transition exploration. Take `Next0` in the initial state, then `Next1`, etc.

The different options are about facilitating this choice of transitions (either via TRex calls or via tuning parameters, which get internally propagated into TRex calls).

docs/src/adr/023adr-trace-evaluation.md
konnov: > I think you can use latex, in case that would help: https://github.blog/2022-05-19-math-support-in-markdown/

Yeah, that's a very cool feature. But the ADRs are going via mdbook, so this would not work I suppose? We had some other special support for math environment.

Kukovec: I used some mdbook-latex library in https://apalache.informal.systems/docs/tutorials/symbmc.html.
In hindsight, probably a massive waste of time, since I reckon like 3 people ever read that.

docs/src/adr/023adr-trace-evaluation.md
konnov: > The different options are about facilitating this choice of transitions (either via TRex calls or via tuning parameters, which get internally propagated into TRex calls).

Or via preprocessing without using TRex ;-)

Kukovec: > Or via preprocessing without using TRex ;-)

Name one advantage this has over the solution with `tuning`.

Kukovec: Plus, `Trace[nextIndex]` dodges all preprocessing and, most importantly, const simplification.

thpani: > they were principally intended to be architectural **decision** records

Sure, but where do we discuss a proposed decision, if not on the PR merging the decision record?

> in that function it makes sense to me we may want a light-weight prototype to help inform our decisions

Sure, as long as such a prototype informs decisions and does not preempt them :)

Kukovec: > Of course, as long as such a prototype informs decisions and does not preempt them :)

Pretty much the only tradeoff/downside of my proposed approach is _potential_ dev/maintenance cost. I find it pointless to argue how complicated an abstract implementation would be to maintain/extend. If you have a prototype in front of you, you can start making claims about which parts are maintainable and which aren't.

thpani:
@Kukovec I think you might've pushed the wrong branch, or some merge commit crept in here?

Kukovec:
> @Kukovec I think you might've pushed the wrong branch, or some merge commit crept in here?

Huh, I only did `rebase origin/main`, but let me look at that again.

Kukovec:
Yeah, guess it was just that. Thanks for noticing.

docs/src/adr/023adr-trace-evaluation.md
```diff
@@ -0,0 +1,411 @@
...
+(* Transition 3 to State4 *)
+State4 == A = 25/\ B = 1/\ C = {}/\ D = FALSE
+
+(* The following formula holds true in the last state and violates the invariant *)
+InvariantViolation == TRUE
+
+================================================================================
+(* Created by Apalache on Mon Oct 17 16:06:59 CEST 2022 *)
+(* https://github.com/informalsystems/apalache *)
+```
```
shonfeder, line 401: Thanks! do find having the example here really helps.

After our discussion today, I think it would be worth including a note about how operators can be easily added to get views into regions of the state one may be interested in.

Also, I do think it's worth considering the interaction with the trace viewer, and I do think @thpani makes a strong case here.

IMO, it's fine if we decide to not pursue the course he advises (to extend the state with these  evaluations, and leave filtering to the viewer), but the rationale for that choice is probably good to record here (perhaps it's just because it takes less work! :) .

With those additions, for my part, I'm quite happy to have this merged in.

shonfeder (approved):

docs/src/adr/023adr-trace-evaluation.md
```diff
@@ -0,0 +1,411 @@
...
+
+## Summary
+
+<!-- Statement to summarize, following the following formula: -->
+
+In the context of improving usability\
+facing difficulties understanding counterexample traces\
+we decided to implement trace evaluation\
+to achieve a better user experience\
+accepting the development costs.\
```
thpani, line 23: Somehow github renders this as literal `\`:

```suggestion
accepting the development costs.
```

thpani (approved):
LGTM, thanks for extending the document!

I've noted my preference for output option (1), because I think it's a more universally usable output format. That said, I'm fine going either way and if necessary iterating upon UX feedback.

docs/src/adr/023adr-trace-evaluation.md
thpani: That's a good point, maybe we can include a link to https://github.com/informalsystems/vscode-itf-trace-viewer and note that it supports filtering by variable?

PR #2186, opened 2022-09-28
Return failure data from pass chain executions

shonfeder:
Closes #2185, #2170

This PR enriches the `PassResult` so that it can it can include data about the reasons for pass failures. The approach here is currently going for the shortest path to getting the needed JSON-encodable data for RPC calls. 

Further work could be useful to impose a more rigorous interface. In particular, it could be helpful to reorganize the package dependencies so that the failure data returned by different passes could be enumerated in a case class, allowing static specification of the supported data structures. This is apposed to he approach taken here, which just converts the pass data into ujson. The latter approach is expedient given the current motivating need, but won't be nice to work with if we decide we want to use this data  for purposes other than returning RPC results.

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] Documentation added for any new functionality
- [x] [Entries added to `./unreleased/`][changelog format] for any new functionality

[changelog format]: https://github.com/informalsystems/apalache/blob/main/CONTRIBUTING.md#how-to-record-a-change

PR #2193, opened 2022-10-05
Shai: Add ping RPC to services

### LINKED ISSUE #2192: Shai: Add Ping RPC to serivces
While writing the integration tests for the client, it's clear that one of the pain points in getting the client set up is to be able to wait for a server to be ready in the likely case that an application is starting an apalache service for its use.

To make this easier, we should expose a `ping` rpc from each service that can be used to check if the service is up and running.

shonfeder:
Closes #2192

Adds a no-op RPC call that can be used to ping a service to see if it is available. This will be used on the client side to make it easy to wait until the server is up an running in the likely case that the users of the library is starting an apalache server alongside their application.

NOTE: Will need to rebase after #2186 is in, only https://github.com/informalsystems/apalache/pull/2193/commits/59b4f278e9b0547673746f88c26e7532252e7e6d is added here, and review only needs to consider that commit.


shonfeder:
I will need to rework this likely because of the gnarliness of python's protbuf support: https://github.com/grpc/grpc/issues/9575

shai/src/main/protobuf/common.proto
```diff
@@ -0,0 +1,16 @@
+// Protobuf specification for messages shared between services
+
+syntax = "proto3";
+
+package shai;
+
+option java_multiple_files = true;
+option java_package = "at.forsyte.apalache.shai.v1";
+option java_outer_classname = "Common";
+option objc_class_prefix = "TEP";
```
thpani, line 10: This looks like an ObjectiveC-specific option? Why do we need it here?

thpani (approved):
Code looks good!

For my own understanding, could you explain a bit more how this will be used?

Since the underlying protocol is handshaking via TCP, doesn't the client establishing a connection already imply that the server is up?

shonfeder:
> Since the underlying protocol is handshaking via TCP, doesn't the client establishing a connection already imply that the server is up?

Even gooder question: in theory we should be able to use to the [channel state API](https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md#channel-state-api) to check whether the server is up and healthy. However, in practice -- as I discovered through hours of trial and error a couple months back -- if we try to move forward with RPC calls as soon as the channel state reports "READY", the request can actually be missed, leaving the client waiting indefinitely. I haven't looked at the implementation-level source of this problem, but presumably there transition between having an TCP connection up and ready and the gRPC plumbing to actually handle requests is not atomic. 

I guess the potential gap between having a ready connection and the gRPC system actually up and ready to handle requests is also indicated by https://stackoverflow.com/a/51250184/1187277, which notes:

> Note that the API does not actively monitor the service's health; it just informs whether the server is currently known to be running and reachable.

I'd be quite happy to switch to some built in mechanism if we can find one!

thpani:
> presumably there transition between having an TCP connection up and ready and the gRPC plumbing to actually handle requests is not atomic.

Interesting, thanks for explaining!

Issue #1845, opened 2022-06-01
Make it possible to evaluate expressions over (ITF) traces


Kukovec:
Often, it is difficult for humans to _evaluate_ or _inspect_ a counterexample and mine it for information. 
This is mainly due to the fact that invariants, for which counterexamples are found, are often sizeable and made up of numerous sub-components, or the states themselves contain a large number of variables.

We propose the following feature: Given an ITF trace and a collection of labeled (named) state-level TLA expressions , have Apalache output a mapping over the trace, where the expressions are evaluated at each state.

Example:
Assume a specification over two `Int` variables, `x,y` and an invariant `x < 10 /\ y < 10` produced a trace 
```
(x=1,y=2) -> (2,2) -> (4,2) -> (4,6) -> (10, 6)
Invariant violation: x >= 10 \/ y >= 10 
```
If we want to evaluate the expressions `A: x + y`, `B: x > y` and `C: y < 10` over the trace, we should see:
```
(A: 3, B: FALSE, C: TRUE) -> (4, FALSE, TRUE) -> (6, TRUE, TRUE) -> (10, FALSE, TRUE) -> (16, TRUE, TRUE)
```
or 
```
A: 3 -> 4 -> 6 -> 10 -> 16
B: ...
C: ...
```

Kukovec:
@konnov @shonfeder @thpani @bugarela thoughts?

konnov:
As an idea: This also has flavor of runtime monitoring.

konnov:
adding @tesnimab to follow up on the discussion of input/output

konnov:
Currently, this issue in the bugfixing milestone. If it's more than a week of work, we should probably make a separate milestone for it. @Kukovec what do you think?

shonfeder:
Done! :tada:

PR #2202, opened 2022-10-10
TraceE prototype

Kukovec:


Kukovec:
I'll open this as a real PR now, despite it not being 100% production-ready.
Despite the seemingly high LOC count, a lot of it is boilerplate. The files to focus on are in the [`tracee`](https://github.com/informalsystems/apalache/tree/cfff69229128a27dac6b9e5868752437f9bf77f0/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee) folder, as well as `TraceeCmd`.
A short summary:
1. User calls `tracee --trace=trace.json --expressions=e1,...,en file`. The behavior is defined in `TraceeCmd`
2. Internally, the `TraceePass` will turn the original module into one where the transitions follow the states in the trace, and the new variables are the expressions given by `--expressions`. This is defined in `TraceePassImpl`.
3. The checker, using the tuning options derived in `TraceeCmd`, checks the (simple) spec, finds a witness (as `--output-traces` is force-set to true). That witness is the final output.

TODOs, not addressed by this PR: 
- #2222 (and a lot of ITF in general, needs discussion w/ the team)
- #2209, and some requirements handling.
- "nicer" ouputs: This MWE hijacks the `CoutnerexampleWriter`, and the associated output strings, to an extent.
- tests: Related to #2223, generating well-typed JSON traces, which are a prerequisite for a test suite, is presently an issue.

shonfeder:
Could you include some tests or examples showing how to specify the trace evaluation and then what the resulting trace looks like? I think we should probably start the review by considering those factors... WDYT?

Kukovec:
> Could you include some tests or examples showing how to specify the trace evaluation and then what the resulting trace looks like? I think we should probably start the review by considering those factors... WDYT?

Sure. Suppose you start with
```tla
-------------------------- MODULE test -----------------------------

EXTENDS Integers

VARIABLE
  \* @type: Int;
  x

A == x * x
B == IF x < 3 THEN 0 ELSE 1
C == [y \in {1,2,4} |-> {y} ][x]
D == x % 2 = 0

Init == x = 1
Next == x' = x + 1

Inv == TRUE
  

=========================================================================
```
and Apalache generates the trace (witness) `testTrace.json`
```json
{
  "name": "ApalacheIR",
  "version": "1.0",
  "description": "https://apalache.informal.systems/docs/adr/005adr-json.html",
  "modules": [
    {
      "kind": "TlaModule",
      "name": "counterexample",
      "declarations": [
        {
          "type": "() => Bool",
          "kind": "TlaOperDecl",
          "name": "ConstInit",
          "formalParams": [
            
          ],
          "isRecursive": false,
          "body": {
            "type": "Untyped",
            "kind": "ValEx",
            "value": {
              "kind": "TlaBool",
              "value": true
            }
          }
        },
        {
          "type": "() => Bool",
          "kind": "TlaOperDecl",
          "name": "State0",
          "formalParams": [
            
          ],
          "isRecursive": false,
          "body": {
            "type": "Bool",
            "kind": "OperEx",
            "oper": "AND",
            "args": [
              {
                "type": "Bool",
                "kind": "OperEx",
                "oper": "EQ",
                "args": [
                  {
                    "type": "Int",
                    "kind": "NameEx",
                    "name": "x"
                  },
                  {
                    "type": "Int",
                    "kind": "ValEx",
                    "value": {
                      "kind": "TlaInt",
                      "value": 1
                    }
                  }
                ]
              }
            ]
          }
        },
        {
          "type": "() => Bool",
          "kind": "TlaOperDecl",
          "name": "State1",
          "formalParams": [
            
          ],
          "isRecursive": false,
          "body": {
            "type": "Bool",
            "kind": "OperEx",
            "oper": "AND",
            "args": [
              {
                "type": "Bool",
                "kind": "OperEx",
                "oper": "EQ",
                "args": [
                  {
                    "type": "Int",
                    "kind": "NameEx",
                    "name": "x"
                  },
                  {
                    "type": "Int",
                    "kind": "ValEx",
                    "value": {
                      "kind": "TlaInt",
                      "value": 2
                    }
                  }
                ]
              }
            ]
          }
        },
        {
          "type": "() => Bool",
          "kind": "TlaOperDecl",
          "name": "State2",
          "formalParams": [
            
          ],
          "isRecursive": false,
          "body": {
            "type": "Bool",
            "kind": "OperEx",
            "oper": "AND",
            "args": [
              {
                "type": "Bool",
                "kind": "OperEx",
                "oper": "EQ",
                "args": [
                  {
                    "type": "Int",
                    "kind": "NameEx",
                    "name": "x"
                  },
                  {
                    "type": "Int",
                    "kind": "ValEx",
                    "value": {
                      "kind": "TlaInt",
                      "value": 3
                    }
                  }
                ]
              }
            ]
          }
        },
        {
          "type": "() => Bool",
          "kind": "TlaOperDecl",
          "name": "State3",
          "formalParams": [
            
          ],
          "isRecursive": false,
          "body": {
            "type": "Bool",
            "kind": "OperEx",
            "oper": "AND",
            "args": [
              {
                "type": "Bool",
                "kind": "OperEx",
                "oper": "EQ",
                "args": [
                  {
                    "type": "Int",
                    "kind": "NameEx",
                    "name": "x"
                  },
                  {
                    "type": "Int",
                    "kind": "ValEx",
                    "value": {
                      "kind": "TlaInt",
                      "value": 4
                    }
                  }
                ]
              }
            ]
          }
        },
        {
          "type": "() => Bool",
          "kind": "TlaOperDecl",
          "name": "State4",
          "formalParams": [
            
          ],
          "isRecursive": false,
          "body": {
            "type": "Bool",
            "kind": "OperEx",
            "oper": "AND",
            "args": [
              {
                "type": "Bool",
                "kind": "OperEx",
                "oper": "EQ",
                "args": [
                  {
                    "type": "Int",
                    "kind": "NameEx",
                    "name": "x"
                  },
                  {
                    "type": "Int",
                    "kind": "ValEx",
                    "value": {
                      "kind": "TlaInt",
                      "value": 5
                    }
                  }
                ]
              }
            ]
          }
        },
        {
          "type": "() => Bool",
          "kind": "TlaOperDecl",
          "name": "InvariantViolation",
          "formalParams": [
            
          ],
          "isRecursive": false,
          "body": {
            "type": "Bool",
            "kind": "ValEx",
            "value": {
              "kind": "TlaBool",
              "value": true
            }
          }
        }
      ]
    }
  ]
}
```
(note, Apalache will actually output this trace with the "type" fields set to `Untyped`, this was manually edited by me, see #2223)
After running `tracee --trace=testTrace.json --expressions=A,B,C,D test.tla`, you should see
```
Constructing an example run                                       I@16:06:59.450
Check the trace in: <PATH>/example0.tla, ... I@16:06:59.563
The outcome is: NoError                                           I@16:06:59.571
Trace successfully generated.   
```
where `example0.tla` looks like
```tla

EXTENDS test

ConstInit == TRUE

(* Initial state *)
State0 == A = 1/\ B = 0/\ C = {1}/\ D = FALSE

(* Transition 0 to State1 *)
State1 == A = 4/\ B = 0/\ C = {2}/\ D = TRUE

(* Transition 1 to State2 *)
State2 == A = 9/\ B = 1/\ C = {}/\ D = FALSE

(* Transition 2 to State3 *)
State3 == A = 16/\ B = 1/\ C = {4}/\ D = TRUE

(* Transition 3 to State4 *)
State4 == A = 25/\ B = 1/\ C = {}/\ D = FALSE

InvariantViolation == TRUE

(* Created by Apalache on Mon Oct 17 16:06:59 CEST 2022 *)
(* https://github.com/informalsystems/apalache *)
```

Kukovec:
> prima facie, splitting operators doesn't seem to me like the best way of doing this in terms of readability or machine processing to me.

Here's an actual CE example from the PBTS experiments:
```tla
State0 ==
  Proposer = SetAsFun({ <<0, "f4">>, <<1, "f4">>, <<2, "f4">>, <<3, "f4">> })
    /\ action = "Init"
    /\ beginRound
      = SetAsFun({ <<<<0, "c1">>, 3>>,
        <<<<2, "c1">>, 7>>,
        <<<<1, "c1">>, 7>>,
        <<<<2, "c2">>, 7>>,
        <<<<1, "c2">>, 7>>,
        <<<<3, "c2">>, 7>>,
        <<<<0, "c2">>, 2>>,
        <<<<3, "c1">>, 7>> })
    /\ decision
      = SetAsFun({ <<"c1", <<<<"None", -1, -1>>, -1>>>>,
        <<"c2", <<<<"None", -1, -1>>, -1>>>> })
    /\ evidence = {}
    /\ localClock = SetAsFun({ <<"c1", 3>>, <<"c2", 2>> })
    /\ lockedRound = SetAsFun({ <<"c1", -1>>, <<"c2", -1>> })
    /\ lockedValue = SetAsFun({ <<"c1", "None">>, <<"c2", "None">> })
    /\ msgsPrecommit
      = SetAsFun({ <<
          0, { [id |-> <<"v0", 3, 0>>,
              round |-> 0,
              src |-> "f3",
              type |-> "PRECOMMIT"],
            [id |-> <<"v0", 3, 0>>,
              round |-> 0,
              src |-> "f4",
              type |-> "PRECOMMIT"],
            [id |-> <<"v1", 2, 0>>,
              round |-> 0,
              src |-> "f3",
              type |-> "PRECOMMIT"],
            [id |-> <<"v1", 2, 0>>,
              round |-> 0,
              src |-> "f4",
              type |-> "PRECOMMIT"] }
        >>,
        <<1, {}>>,
        <<
          2, {[id |-> <<"v2", 3, 2>>,
            round |-> 2,
            src |-> "f3",
            type |-> "PRECOMMIT"]}
        >>,
        <<
          3, {[id |-> <<"v2", 7, 3>>,
            round |-> 3,
            src |-> "f4",
            type |-> "PRECOMMIT"]}
        >> })
    /\ msgsPrevote
      = SetAsFun({ <<
          0, { [id |-> <<"v0", 3, 0>>,
              round |-> 0,
              src |-> "f3",
              type |-> "PREVOTE"],
            [id |-> <<"v0", 3, 0>>,
              round |-> 0,
              src |-> "f4",
              type |-> "PREVOTE"],
            [id |-> <<"v1", 2, 0>>,
              round |-> 0,
              src |-> "f3",
              type |-> "PREVOTE"],
            [id |-> <<"v1", 2, 0>>,
              round |-> 0,
              src |-> "f4",
              type |-> "PREVOTE"] }
        >>,
        <<1, {}>>,
        <<2, {}>>,
        <<3, {}>> })
    /\ msgsPropose
      = SetAsFun({ <<
          0, { [proposal |-> <<"v0", 3, 0>>,
              round |-> 0,
              src |-> "f4",
              type |-> "PROPOSAL",
              validRound |-> 2],
            [proposal |-> <<"v1", 2, 0>>,
              round |-> 0,
              src |-> "f4",
              type |-> "PROPOSAL",
              validRound |-> -1] }
        >>,
        <<1, {}>>,
        <<2, {}>>,
        <<3, {}>> })
    /\ proposalReceptionTime
      = SetAsFun({ <<<<0, "c1">>, -1>>,
        <<<<2, "c1">>, -1>>,
        <<<<1, "c1">>, -1>>,
        <<<<2, "c2">>, -1>>,
        <<<<1, "c2">>, -1>>,
        <<<<3, "c2">>, -1>>,
        <<<<0, "c2">>, -1>>,
        <<<<3, "c1">>, -1>> })
    /\ realTime = 0
    /\ round = SetAsFun({ <<"c1", 0>>, <<"c2", 0>> })
    /\ step = SetAsFun({ <<"c1", "PROPOSE">>, <<"c2", "PROPOSE">> })
    /\ validRound = SetAsFun({ <<"c1", -1>>, <<"c2", -1>> })
    /\ validValue
      = SetAsFun({ <<"c1", <<"None", -1, -1>>>>, <<"c2", <<"None", -1, -1>>>> })
```
Mind you, this is a single state. It's unreadable. I view trace evaluation as a sort of view, which lets you shrink this mess into e.g. 
```tla
State7 ==
    /\ AgreementOnValue = TRUE
    /\ ConsensusValidValue = FALSE
    /\ ConsensusTimeValid = TRUE
    /\ TimeLiveness = TRUE
```

Kukovec:
Alternatively, what you might actually be asking for is some visualization support, where you can see the two traces side-by-side, e.g. in your browser.

shonfeder:
Of course with a large state, each state of a trace will be large. But it seems to me you are combining the functionality of adding trace evaluation with an orthogonal concern of filtering out or limiting data presented in a trace. 

It seem cleaner to me that filtering the views of states  can be left to tooling like MBT's trace visualizer, and that the functionality of adding trace evaluation should be narrowly focused on this expansion of states, not also introducing a novel and alternative view of traces.

shonfeder:
> this entire idea started when I, in the role of a user and spec author, felt like I needed to be able to project away irrelevant parts of a counterexample trace, to understand what was going on.

Ah I see! I had the impression this connected to, and was the crystallization of some long-running requests and needs of multiple users (including you). Sorry for missing the relevant context here.

> While I do see your point, my question is whether this is a discussion is to be held here (i.e., is it blocking for this PR?) or as part of https://github.com/informalsystems/apalache/issues/1845.

I may be misunderstanding the intention of this PR (some explanation of what this PR is aiming at and related to and what kind of review you'd like in the description would help with that in the future): I though this prototype PR was meant as a way of making the discussion from #2189 more concrete, allow us to have a more informed and grounded discussion over the proposed feature and it's implementation.

Is the point of the PR actually to introduce prerequisite plumbing rather than to propose a prototype for consideration?

Kukovec:
> I may be misunderstanding the intention of this PR (some explanation of what this PR is aiming at and related to and what kind of review you'd like in the description would help with that in the future): I thought this prototype PR was meant as a way of making the discussion from #2189 more concrete, allow us to have a more informed and grounded discussion over the proposed feature and it's implementation.
> 
> Is the point of the PR actually to introduce prerequisite plumbing rather than to propose a prototype for consideration?

This goes for all PRs, but the way I see it, there's two kinds of "changes" people can request; refinement changes ("tweak this", "improve that") and blocking changes ("this cannot be the way it is now"). With refinement changes, you can merge the current PR and postpone the requests (make an issue, new PR later), whereas blocking changes prevent a PR from merging at all until addressed. 
I typically prefer to have discussions regarding refinement changes on issues (or just open new issues as needed), whereas blocking changes need to be addressed at the PR level. The reason is that all (non-experimental) PRs aim to eventually get merged.

Concretely, you mention alternative ways the output could be formatted. It was unclear to me whether I was expected to modify the actual code _in this PR_ to accommodate that, or not.

shonfeder:
Ah, so I understood we were having a high-level discussion about the feature development based on your prototype. I have not been deeply involved in all the discussion prior, but my understanding was that you had proposed a design in an ADR, got some pushback, and decided to offer a prototype to give a more concrete basis for the design discussion.

I haven't yet made any change requests, but was using the prototype PR for the design discussion. Happy to move that discussion elsewhere if you prefer.

> Concretely, you mention alternative ways the output could be formatted. It was unclear to me whether I was expected to modify the actual code in this PR to accommodate that, or not.

What do you think about the suggestion made by @thpani and I re: the format? Do you agree it would be helpful to include the full state?

This PR is already relatively large as is, so changes may be be best to add as a followup. Alternatively, you could factor out just the plumbing work that adds the CLI command and options, and break out the generation of the trace itself into a separate PR, where it would perhaps make sense to adjust the format.

I haven't reviewed the implementation yet, as I wanted to be clear on the high-level intent and design choices first :D.

thpani:
I think we need some kind of design document before we review implementation.
@Kukovec Do you plan to update #2189 with design implications from writing this prototype?

If I currently wanted to do a thorough review of this, I'd have to:
* Read the code to understand the proposed design
* Think of and weigh it against other possible designs (since we deferred the design discussion to this point)
* Depending on severity, ask for design changes to be implemented before another round of review
* Do a second round of review, for style and quality

That's a lot of effort, and if multiple people review, a lot of duplicate effort.

> I haven't reviewed the implementation yet, as I wanted to be clear on the high-level intent and design choices first :D.

Same here, 600 LOC is a laborious review. Before I start, I would like to understand which parts of this are necessary to merge with the main codebase, and which parts are prototypey / design exploration.

@shonfeder's recommendation to factor out obvious implementation parts lessens that burden a bit, but I'm still cautious about the design-review portion.

> This PR is already relatively large as is, so changes may be be best to add as a followup.

I agree on the code quality side, but since we're also discussing design here, there's a chance that we lack understanding of what the coherent whole will look like until it's fully implemented.

konnov:
It would be good to think about CLI UX before going into a full-scale implementation. For instance, [Working backwards](https://www.allthingsdistributed.com/2006/11/working_backwards.html) suggests writing user manuals before implementing things.

Kukovec:
What exactly do you hope to see from the ADR that hasn't reached consensus at this point? That's such an unrecoverable design decision that needs to block any implementation? I went through the ADR PR, and here's a list of all the claims made therein, as well as their evaluations:
"I would prefer to solve it by pure preprocssing" (+ similar) - Igor admitted, during the demo, that what I'm doing is what he meant by preprocessing
"The transition executor is very much in the guts of the model checker. While it is possible to use it, I do not think it is the simplest way of solving the problem." - we all agree it's not and the implementation isn't using it
"We could encode the above trace as a sequence of records" -> "You could do that, but then you're actually solving additional sequence and record applications in SMT, as well as integer arithmetic (index + 1)" + "Plus, Trace[nextIndex] dodges all preprocessing and, most importantly, const simplification." - stands unrefuted
"The transition filter tuning option is a bit fiddly. It's probably enough for this use case, but in general, constructing regexes is not the best abstraction for this." - statement without justification (and even then, "enough for this use case")
"The user should not be providing any additional parameters at all, besides the trace file and a list of expressions." - stands unrefuted

Even if we, hypothetically, wanted to pivot to the "trace as a sequence of records" approach, the changes are trivial and isolated to the `TraceePassImpl`. And that is, to date, just about the only alternative anyone has ever proposed, and the proposer himself agreed that the downside of sequence/record access dodging preprocessing is severe.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,92 @@
...
+  // Can't pass tuning via flags
+  override def collectTuningOptions(): Map[String, String] = {
+
+    val len = getLenFromFile(trace)
+    persistentLen = len - 1 // 1st state is init, len-1 transitions
+
+    val txFilter = ("0->0" +: (1 until len)
+      .map { i =>
+        s"$i->${i - 1}" // -1, since the 1st state is init
+      }).mkString("|")
```
thpani, line 63: The offsetting here is really hard to follow, I think some textual description would help.

We should add unit tests for this function.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/DrivenTransitionConstructor.scala
```diff
@@ -0,0 +1,46 @@
...
+import at.forsyte.apalache.tla.lir.transformations.TransformationTracker
+import at.forsyte.apalache.tla.lir.transformations.standard.ReplaceFixed
+
+/**
+ * @author
+ *   Jure Kukovec
+ */
+class DrivenTransitionConstructor(
+    tracker: TransformationTracker,
+    exs: Map[String, TlaEx]) {
```
thpani, line 15: Could you add a doccomment on this parameter?

thpani, line 15:
```suggestion
    expressions: Map[String, TlaEx]) {
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPass.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+   * The exitcode that is used to stop the system when one thread has failed with an exception.
+   */
+  val EXITCODE_ON_EXCEPTION = 30
+
+  /**
+   * The exitcode that is used to stop the system when the system did not stop gracefully.
+   */
+  val EXITCODE_ON_SHUTDOWN = 40
```
thpani, line 27: Where are these used?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePass.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * The exitcode that is used to stop the system when one thread has failed with an exception.
+   */
+  val EXITCODE_ON_EXCEPTION = 30
+
+  /**
+   * The exitcode that is used to stop the system when the system did not stop gracefully.
+   */
+  val EXITCODE_ON_SHUTDOWN = 40
```
thpani, line 26: Where are these used?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPass.scala
```diff
@@ -0,0 +1,28 @@
...
+import at.forsyte.apalache.infra.passes.Pass
+
+/**
+ * Bridging pass before BMC, to tie up any configuration that would have been set by passes not utilized in trace
+ * evaluation.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait BridgingPass extends Pass {
```
thpani, line 12: Should we call this

```suggestion
trait TraceeBridgingPass extends Pass {
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPassImpl.scala
```diff
@@ -0,0 +1,24 @@
+package at.forsyte.apalache.tla.tracee.pass
+
+import at.forsyte.apalache.infra.passes.Pass.PassResult
+import at.forsyte.apalache.tla.lir._
+import com.google.inject.Inject
+import com.typesafe.scalalogging.LazyLogging
+
+/**
+ * The implementation of a bounded model checker with SMT.
```
thpani, line 9:
```suggestion
 * Bridging pass before BMC, to tie up any configuration that would have been set by passes not utilized in trace
 * evaluation.
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPassImpl.scala
```diff
@@ -0,0 +1,24 @@
...
+class BridgingPassImpl @Inject() extends BridgingPass with LazyLogging {
+
+  override def name: String = "BridgingPass"
+
+  override def execute(module: TlaModule): PassResult = Right(module)
+
+  override def dependencies = Set()
+
+  // Required by BMC, set by a pass that isn't needed in TraceE mode.
+  override def transformations = Set(ModuleProperty.Analyzed)
```
thpani, line 23: What pass is setting this and what is it needed for?
Why can't we refactor BMC instead of hacking this in?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  // If we set the --length inside collectTuningOptions it gets overwritten later.
+  // therefore, we store it locally, and only update the options right before calling run()
+  private var persistentLen: Int = 0
+
+  def getLenFromFile(file: File): Int = Try(UJsonRep(ujson.read(file))) match {
+    case Success(json) =>
+      // ITF case
+      val statesOpt = json.getFieldOpt("states").map(UJsonScalaFactory.asSeq)
+      statesOpt match {
```
thpani, line 36: You already have a trace reader in `UJsonTraceReader`, in particular the dropping of `Cinit` and `Inv` is duplicated.

thpani (commented):
Preliminary comments below.

In general, there is _a lot_ of plumbing here (Trace reader, CLI cmd, ...) that could've easily been factored into separate PRs.
It's hard to review core pass logic if it's interleaved by arbitrary bits of plumbing. Also, we could've reviewed low risk pieces of code in parallel.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,92 @@
...
+import scala.util.{Failure, Success, Try}
+
+/**
+ * This command initiates the 'tracee' command line.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+class TraceeCmd(name: String = "tracee", description: String = "Evaluate expressions over a trace.")
+    extends CheckCmd(name, description) {
```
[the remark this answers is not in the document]
Kukovec: Ultimately, we'd want to extend not `CheckCmd` but some intermediate command, which defines `init, next, length` and `tuning-options`, which `CheckCmd` _also_ extends, but this felt like too much refactoring for the prototype.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  var trace: File = arg[File](description = "a file containing an ITF trace. Must also define --expressions.")
+
+  var expressions: List[String] =
+    arg[List[String]](name = "expressions",
+        description = "TLA+ expressions to be evaluated over a given trace. Must also define --trace.")
+
+  // If we set the --length inside collectTuningOptions it gets overwritten later.
```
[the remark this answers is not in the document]
Kukovec: Yes, Shon and I already had an offline discussion RE options being modifiable mid-run (which they currently aren't). We're probably going to eventually have the passes propagate modules + options.
The use of the persistent length is in `run`:
https://github.com/informalsystems/apalache/blob/4145b809a1ea652d6f2113bf0a76a8ea5c67a09c/mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala#L75-L76

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,92 @@
...
+          declOpt.map(_.length - 2).getOrElse {
+            throw new JsonDeserializationError(s"${file.getCanonicalPath} is malformed.")
+          }
+      }
+    case Failure(exception) =>
+      throw new JsonDeserializationError(s"Could not read ${file.getCanonicalPath}.", exception)
+  }
+
+  // Can't pass tuning via flags
+  override def collectTuningOptions(): Map[String, String] = {
```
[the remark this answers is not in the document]
Kukovec: Yes, but this first pass is a bit light on UX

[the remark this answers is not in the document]
Kukovec: It should warn on any flag that isn't exactly `trace` or `expressions` in the future.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,92 @@
...
+  private var persistentLen: Int = 0
+
+  def getLenFromFile(file: File): Int = Try(UJsonRep(ujson.read(file))) match {
+    case Success(json) =>
+      // ITF case
+      val statesOpt = json.getFieldOpt("states").map(UJsonScalaFactory.asSeq)
+      statesOpt match {
+        case Some(seq) => seq.length
+        case None      =>
+          // generic case
```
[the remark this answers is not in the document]
Kukovec: Sure, if that's more readable for you.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,92 @@
...
+          val declOpt = for {
+            modules <- json.getFieldOpt("modules")
+            decls <- UJsonScalaFactory.asSeq(modules).head.getFieldOpt("declarations")
+          } yield UJsonScalaFactory.asSeq(decls)
+
+          // we need len-2 for CInit (head) and Inv (last)
```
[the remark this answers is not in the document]
Kukovec: Yes, due to the fact that they're written using CEWriter. That's one of the problems I see, but not really something this PR is going to fix (we had similar issues when Igor first implemented the "print witness" command, people found the output referencing CE and invariants a bit unintuitive).

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
Kukovec: "this function" = `collectTuningOptions()` or "this function" = `txFilter(len)`?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/DrivenTransitionConstructor.scala
Kukovec: Sure

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPass.scala
Kukovec: I thought all passes had to define those, IDK, this I just copied and modified from the BMC pass.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPassImpl.scala
Kukovec: This is set by `AnalysisPass`, indicating things like e.g. skolemization being properly applied IIRC.
In a normal run of BMC, with generic TLA input, we want (need) the analysis pass to run. In the tracee case, the input module is modified in such a way that  there's no expansion or skolemization anyway.
If we refactor BMC, we lose this safeguard with the normal `check`. The alternative is plugging `AnalysisPass` back in, even though it does nothing for the tracee case.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
Kukovec: That's true for reading, and it should be used thhere, but running the entire `convert`, just to read the length of a json array (the contents of which do not need to be converted at this point, and cannot be made persistent) seems wasteful. The length of the unconverted array should be the same as that of the converted one.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
thpani: Can we add an issue or TODO comment to track this future work?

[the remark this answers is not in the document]
Kukovec: Sur

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
thpani: Why do we need a stateful solution that assign `persistentLen` in `collectTuningOptions` tho?

Without knowing the pass executor, it's not obvious that `collectTuningOptions` is called before `run`.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
thpani: I think "generic case" here means IR JSON, right?
The way it's written, the `None` branch is a generic case of ITF.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
thpani: I meant, if we feed an `--ouput-traces` trace into tracee, is it sound to drop `last`?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
thpani: I guess you'd want to factor out a method that takes a trace and constructs `txFilter`, then test that?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
Kukovec: So I first tried modifying the `var length` (inherited from `CheckerCmd`) inside `collectTuningOptions`. The problem with that was that one of the option instantiation passes overwrote it again with the default 10. Then I made a class variable `persistentLen`, which remembers the trace length from `collectTuningOptions` and overwrites the `length` in run, called later

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPassImpl.scala
thpani: So BMC has a safeguard checking what ran before?
I'm fine with keeping this as-is, if we can improve the comment to point out where that happens, and that we fake a superfluous `AnalysisPass` for Tracee.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
Kukovec: This is a deficiency of our options architecture, it wasn't really designed to change options mid-run.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
Kukovec: Yes, I suppose `JSON case` is better.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
Kukovec: Yes, `--output-traces` adds an `Inv = TRUE` (though there is still #2223).

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
thpani: Looking up JSON fields in a `Cmd` feels wrong, why not move this to `UJsonTraceReader`?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
thpani: Wanna track that somewhere?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
Kukovec: That's probably not a bad idea.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
Kukovec: I'll make an issue.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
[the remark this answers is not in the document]
shonfeder: I like Thomas' reasoning here.

Another option to avoid the mutable state is to call the options pass executor on its own with options set, then change the options as you need for your tracee passchain.

You can see the Shai code for an example of how to call a pass within a command: https://github.com/informalsystems/apalache/blob/e03bc42258aa844869794bf5efc7bb440628db93/shai/src/main/scala/at/forsyte/apalache/shai/v1/CmdExecutorService.scala#L72-L93

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,95 @@
...
+import at.forsyte.apalache.tla.imp.src.SourceStore
+import at.forsyte.apalache.tla.lir.oper.{TlaBoolOper, TlaOper}
+import at.forsyte.apalache.tla.lir.{NameEx, OperEx, TlaEx, TlaOperDecl}
+import at.forsyte.apalache.tla.typecomp.json.UJsonToTlaViaBuilder
+
+import scala.annotation.unused
+import scala.util.{Failure, Success, Try}
+
+/**
+ * Reads an ITF-format or a TlaJSON-format trace into a TLAIR trace, with an intermediate UJson representation.
```
shonfeder, line 17: Could you note here what the intermediate ujson representation is for?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @author
+ *   Jure Kukovec
+ */
+class UJsonTraceReader(sourceStoreOpt: Option[SourceStore], tagReader: TypeTagReader) extends TraceReader[UJsonRep] {
+  private val builder = new UJsonToTlaViaBuilder(sourceStoreOpt)(tagReader)
+  override def read(source: SourceOption): UJsonRep = {
+    val readable: ujson.Readable = source match {
+      case FileSource(f, Format.Json) => f
+      case _                          => throw new IllegalArgumentException("File provided with --trace is not a JSON.")
```
shonfeder, line 27: This should also support `StringSource` I think, no?

Also, imo, probably not a great idea to mention the CLI flag in this error. If the flag name changes this will be out of date, and users could load this through a config or RPC call, or this class could be used for other purposes, no through your new CLI endpoint. So probably better to have something like:

```scala
      case _                          => throw new IllegalArgumentException("Tried to load trace from a non-JSON source.")
```

If you need give more specific cli input based input, you can catch this exception (or, better, maybe this should be `Try`  ;P) and extend the error.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,95 @@
...
+    // Rethrow as JsonDeserializationError if unable to read
+    Try(ujson.read(readable)) match {
+      case Success(ujsonVal) => UJsonRep(ujsonVal)
+      case Failure(exception) =>
+        throw new JsonDeserializationError("Unable to read --trace as JSON.", exception)
+    }
+  }
+
+  // TODO
+  private def convertITF(@unused json: UJsonRep): Trace = IndexedSeq.empty
```
shonfeder, line 39: Could this raise a not-implemented instead of an empty sequence? seems better protection against mistaken use.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  private def kvFromAsgn(ex: TlaEx): (String, TlaEx) = ex match {
```
shonfeder, line 41: I didn't know what his mean until I looked at error, so I think the existing name won't read very well in its use site:

```suggestion
  private def kvFromAssignment(ex: TlaEx): (String, TlaEx) = ex match {
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,94 @@
...
+    case OperEx(TlaOper.eq, NameEx(name), rhs) => name -> rhs
+    case _ => throw new JsonDeserializationError(s"Cannot read variable assignment from $ex.")
+  }
+
+  private def convertGenericJSON(json: UJsonRep): Trace = {
```
shonfeder, line 46: I'm not sure what "generic" means in this context.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }
+    // From an indexed sequence of declarations, create a sequence of State-maps.
+    operDecls
+      .map { decls =>
+        decls.map { decl =>
+          decl.body match {
+            // Edge case: single-var specs
+            case OperEx(TlaOper.eq, NameEx(name), rhs) => Map(name -> rhs)
+            // general conjunctive case
```
shonfeder, line 63: This looks like `kvFromAsgn`. A bit strange to me that we have the same logic here and in that abstraction... perhaps you should replace `kvFromAsgn` with a function performing this transformation into maps (lines 62-68)?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,94 @@
...
+              throw new JsonDeserializationError(s"Cannot read state from $body.")
+          }
+        }
+      }
+      .getOrElse {
+        throw new JsonDeserializationError(s"Trace JSON is incorrectly formatted.")
+      }
+  }
+
+  // Assume here that the json is in ITF format
```
shonfeder, line 77: I'm confused about this comment. How does it relate to the unimplemented `convertITF` method? If I am reading line 80 right, this doesn't actually assume an ITF format but supports both ITF and IR?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,94 @@
...
+    val nameOpt = json.getFieldOpt("name").map(UJsonScalaFactory.asStr)
+    if (nameOpt.contains("ApalacheIR")) {
+      convertGenericJSON(json)
+    } else {
+      val formatOpt = for {
+        meta <- json.getFieldOpt("#meta")
+        format <- meta.getFieldOpt("format")
+      } yield UJsonScalaFactory.asStr(format)
+      if (formatOpt.contains("ITF")) {
+        convertITF(json)
```
shonfeder, line 88: Why are we calling this function if it's not implemented yet?

shonfeder, line 88: This branch should instead be delegated to when reading from the `OptionSource` because we already know at that point whether we have an `ITF` format. Otherwise, there'd have been no reason to add support for the `ITF` format :D

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/BridgingPass.scala
shonfeder: I'd check whether these are actually needed and remove them if not.

shonfeder (changes requested):

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -203,6 +203,22 @@ object Config {
...
+  /**
+   * Configuration of trace evaluation
+   *
+   * @param trace
+   *   an ITF trace file
+   * @param expressions
+   *   a list of expressions, to be evaluated at every state in the trace
+   */
+  case class Tracee(
+      trace: Option[File] = None,
```
shonfeder, line 215: I think this should take a `SourceOption`, just as we do for the `Input` configuration:

https://github.com/informalsystems/apalache/blob/e03bc42258aa844869794bf5efc7bb440628db93/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala#L118-L127

This will allow us to support RPCs and other modes of invocation.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -585,6 +602,29 @@ object OptionGroup extends LazyLogging {
     }
   }
 
+  /** Options used to configure trace evaluation. */
+  case class Tracee(
+      trace: File,
```
shonfeder, line 607: This should also be a `SourceOption`.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -585,6 +602,29 @@ object OptionGroup extends LazyLogging {
...
+
+  object Tracee extends Configurable[Config.Tracee, Tracee] with LazyLogging {
+    def apply(tracee: Config.Tracee): Try[Tracee] = {
+
+      def validateExprs(lst: List[String]): Try[List[String]] =
+        if (lst.nonEmpty) Success(lst)
+        else Failure(new PassOptionException("Trace evaluation requires a nonempty list of expressions."))
+
+      for {
+        trace <- tracee.trace.toTry("trace.trace")
```
shonfeder, line 619: Once this is a `SourceOption` you can validate at this point that you only hava JSON or ITF file.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
```diff
@@ -0,0 +1,116 @@
...
+    sourceStore: SourceStore)
+    extends TraceePass with LazyLogging {
+
+  override def name: String = "TraceePass"
+
+  // TODO: turn requires into proper error handling?
+  override def execute(module: TlaModule): PassResult = {
+
+    val traceJsonFile = options.tracee.trace
+    val traceSource = SourceOption.FileSource(traceJsonFile)
```
shonfeder, line 39: This can be removed once `HasTrace` is changed to already load a `SourceOption`, and it should support both a `FileSource` and a `StringSource`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
```diff
@@ -0,0 +1,116 @@
...
+    val traceReader = new UJsonTraceReader(Some(sourceStore), DefaultTagReader)
+
+    val trace = traceReader.convert(traceReader.read(traceSource))
+    require(trace.nonEmpty)
```
shonfeder, line 43: I agree with your TODO, this should return an error.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
```diff
@@ -0,0 +1,116 @@
...
+        module.operDeclarations.filterNot(d => droppedOperNames.contains(d.name)) ++ (initOper +: nextOpers) ++ cinitOpt
+
+    val outModule = TlaModule(module.name, newDecls.toSeq)
+
+    // Mark introduced transition operators as persistent, so they don't get wiped by the inlinig pass.
+    derivedPreds.addPersistent(initOper.name +: nextOpers.map(_.name): _*)
+
+    writeOut(writerFactory, outModule)
+
+    Right(outModule)
```
shonfeder, line 110: Instead of relying on exception raising for the errors for this, it would help with integration with the RPC system if the error from this is returned as a `Left`.

shonfeder (commented):
Finishing up my first review pass finally!

Great work on the MVP here.

I think some passes to clean up TODOs and improve the input validation logic and error handling are in order, but otherwise looking good! :D

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
Kukovec: Both `.itf.json` and `.json` are JSONs. Thee former is an ITF JSON, and the latter needs some kind of adjective. Would you prefer `ApalacheJSON`?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
Kukovec: The reason I don't call `kvFromAsgn` here again is that you already do the pattern matching that happens inside `kvFromAsgn` beforehand, whereas in the following lines, you map `kvFromAsgn` over the contents of a sequence of `TlaEx`, the shapes of which are unknown.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
Kukovec: No, that seems to be a leftover comment from an old iteration, will remove.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
Kukovec: But `TraceeCmd.trace` still has to be a `File`, no?

shonfeder: I don't see why!

shonfeder: Oh sorry, yes, the CLI input itself is a file.

shonfeder: But allowing the config to take an `OptionSource` allows us to pass in strings via JSON config input, either loading it from a config source on disk or via RPC data.

shonfeder: So the `SourceOption`s come in at the green level here: https://github.com/informalsystems/apalache/blob/main/docs/src/adr/022adr-unification-of-configs-and-options.md#solution

Kukovec: I think I got it, commit incoming soon

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
Kukovec: The name got changed in the latest commit

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
shonfeder: Why should this support reading from files but not reading from strings if it is going to suppoort `SourceOption` at all? If you change the method, so it only takes `ujson` that seems consistent, but if you are accepting `SourceOption` here, I see no reason it shouldn't also support `StringSource`. Adding support should also be a one line addition iiuc.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -585,6 +603,34 @@ object OptionGroup extends LazyLogging {
...
+  object Tracee extends Configurable[Config.Tracee, Tracee] with LazyLogging {
+    def apply(tracee: Config.Tracee): Try[Tracee] = {
+
+      def validateExprs(lst: List[String]): Try[List[String]] =
+        if (lst.nonEmpty) Success(lst)
+        else Failure(new PassOptionException("Trace evaluation requires a nonempty list of expressions."))
+
+      def validateSource(sourceOption: SourceOption): Try[SourceOption] = sourceOption match {
+        case FileSource(_, SourceOption.Format.Itf | SourceOption.Format.Json) => Success(sourceOption)
+        case _ => Failure(new PassOptionException("Trace evaluation requires an ITF or JSON trace."))
```
shonfeder, line 621: I think we've also discussed this elsewhere, but why not also support string inputs?

```suggestion
      def validateSource(sourceOption: SourceOption): Try[SourceOption] = sourceOption.format match {
        case SourceOption.Format.Itf | SourceOption.Format.Json => Success(sourceOption)
        case _ => Failure(new PassOptionException("Trace evaluation requires an ITF or JSON trace."))
```

Kukovec: String inputs don't tell you if the insides should follow ITF or ApalacheJSON formatting.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
```diff
@@ -0,0 +1,85 @@
...
+  override def run() = {
+    val cfg = configuration.get
+    val options = OptionGroup.WithTracee(cfg).get
+
+    // We cannot store the trace length persistently in `var length`, because it gets overwritten by options management
+    // defaults by the time run is called.
+    // Therefore, we just read it again. The execution length is 1 shorter than the trace length,
+    // because the trace contains the initial state.
+
+    val executionLength = getLenFromFile(trace) - 1
```
shonfeder, line 66: I missed this before, but what does `var length` refer to here?

Kukovec: `CheckCmd` declares a `length` field.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,113 @@
...
+  private def tryRead(file: File): UJsonRep = Try(ujson.read(file)) match {
+    case Success(ujsonVal) => UJsonRep(ujsonVal)
+    case Failure(exception) =>
+      throw new JsonDeserializationError(s"Unable to read $file as JSON.", exception)
+  }
+
+  override def read(source: SourceOption): TraceUJson = source match {
+    case FileSource(f, Format.Json) => ApalacheJson(tryRead(f))
+    case FileSource(f, Format.Itf)  => ITFJson(tryRead(f))
+    case src => throw new IllegalArgumentException(s"Tried to load state from an invalid source: $src.")
```
shonfeder, line 45: I think something like this should suffice to support string inputs, if you just change `tryRead` to take a `ujson.Readable`.

```suggestion
    case FileSource(f, Format.Json)     => ApalacheJson(tryRead(f))
    case FileSource(f, Format.Itf)      => ITFJson(tryRead(f))
    case StringSource(s, _, Format.Json => ApalacheJson(tryRead(s))
    case StringSource(s, _, Format.Itf  => ITFJson(tryRead(f))
    case src => throw new IllegalArgumentException(s"Tried to load state from an invalid source: $src.")
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  override def convert(traceJson: TraceUJson): Trace = traceJson match {
+    case ITFJson(json)      => convertITF(json)
+    case ApalacheJson(json) => convertApalacheJson(json)
+  }
+
+  override def getTraceLength(traceJson: TraceUJson): Int = traceJson match {
+    case ITFJson(json)      => getLengthITF(json)
+    case ApalacheJson(json) => getLengthApalacheJson(json)
+  }
```
shonfeder, line 56: Nice! Seem squite a bit cleaner to me. Thanks! :D

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
```diff
@@ -0,0 +1,123 @@
...
+  override def execute(module: TlaModule): PassResult = {
+
+    val traceSource = options.tracee.trace
+    val traceReader = new UJsonTraceReader(Some(sourceStore), DefaultTagReader)
+
+    val trace = traceReader.convert(traceReader.read(traceSource))
+    require(trace.nonEmpty)
+    if (trace.isEmpty) {
+      passFailure("The provided trace is empty.", EXITCODE_ON_EXCEPTION)
+    } else {
```
shonfeder, line 45: iiuc, the error handling here should replace the `require`?

```suggestion
    if (trace.isEmpty) {
      passFailure("The provided trace is empty.", EXITCODE_ON_EXCEPTION)
    } else {
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
```diff
@@ -0,0 +1,123 @@
...
+      // TODO: make findBodyOf return Option? (unrelated to MVP)
+      val expressions = options.tracee.expressions.map { exName =>
+        findBodyOf(exName, module.operDeclarations: _*) match {
+          case NullEx => throw new ConfigurationError(s"Operator $exName undefined in module ${module.name}.")
+          case ex     => exName -> ex
+        }
+      }.toMap
+      if (expressions.isEmpty)
+        passFailure("The provided list of expressions is empty.", EXITCODE_ON_EXCEPTION)
+      else {
```
shonfeder, line 56: Don't we already validate this when the `Option` group is constructed?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
```diff
@@ -0,0 +1,123 @@
...
+
+      // TODO: make findBodyOf return Option? (unrelated to MVP)
+      val expressions = options.tracee.expressions.map { exName =>
+        findBodyOf(exName, module.operDeclarations: _*) match {
+          case NullEx => throw new ConfigurationError(s"Operator $exName undefined in module ${module.name}.")
+          case ex     => exName -> ex
+        }
+      }.toMap
+      if (expressions.isEmpty)
+        passFailure("The provided list of expressions is empty.", EXITCODE_ON_EXCEPTION)
```
shonfeder, line 55: The error code name doesn't seem to fit, as this isn't stopping with an exception, unless i misunderstand. Should this not just be `ExitCodes.ERROR`?

I think i don't understand the intent of these `EXITCODE_*`  constants when we have the `ExitCodes` class already...

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
```diff
@@ -0,0 +1,123 @@
...
+      else {
+
+        /*
+        In this mode, we will inject our own
```
shonfeder, line 59: This is all very nicely commented. Thanks for that!

shonfeder (commented):
I'm still pressing on this support for `StringSource` because I don't understand why we'd want to exclude that. Otherwise, just a few clarifying questions and a superficial, non-blocking suggestions.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
shonfeder: I think they do, unless i misunderstand. See https://github.com/informalsystems/apalache/blob/e03bc42258aa844869794bf5efc7bb440628db93/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala#L302

It has exactly the same format info as `FileSource`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
Kukovec: Ah, I didn't realize `StringSource` also carried format info. If that's the case then you're right, we can support it.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/pass/TraceePassImpl.scala
Kukovec: Yes, but you can technically create a `TraceCmd` w/o adhering to this pipeline (e.g. in a test somewhere).

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TraceeCmd.scala
shonfeder: So is the point here that you want to set a configured length, not depend on the length attribute that is used for CLI parsing?

Kukovec: Yes, I want to programmatically override the checker's `length` parameter to be the length of the provided trace (-1).

shonfeder: I see! I'd not mention `var length` because it doesn't appear in this class, and won't even be relevant once the TODO on line 16 is done, as we'll no longer be inheriting from `CheckCmd`. Perhaps just note that the length is computed from the input source.

shonfeder (approved):
LGTM!

Thanks for your patience and communicativity as we worked out our understanding and converged here!

thpani (approved):
Looks good, thanks for the cleanup work!

Can't wait to try Tracee in practice :)

PR #2231, opened 2022-10-18
Small cleanup on Shai RPCs

shonfeder:
Two tiny fixes here:

- Use lower-case consistently for gRPC service methods (this doesn't make a difference
for Scala code generation but does for Python)
- Use the same field name and representation for returning counterexamples
coming from deadlock and violation errors 


konnov (approved):
LGTM

PR #2233, opened 2022-10-18
Extend SourceOption.FileSource for itf.json

### LINKED ISSUE #2222: Extend `SourceOption` with ITF format support
Currently, `SourceOption.FileSource.apply` uses `FilenameUtils.getExtension` to determine the file extension. Unfortunately, ITF file names have the shape `<name>.itf.json`. As `getExtension("a.b.c")` will return `c` we should either:

- Change the ITF file names to just `<name>.itf`, or
- Implement a custom `getExtension`

shonfeder:
Closes #2222

After noodling around to confirm that detecting ".itf.json" would be reasy
enough, it seemed to make sense to offer the solution as a PR.

-------

### DIFF
--- a/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
+++ b/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
@@ -261,6 +261,7 @@ object SourceOption {
   object Format {
     case object Tla extends Format
     case object Json extends Format
+    case object Itf extends Format
   }
 
   /** Data to be loaded from a file */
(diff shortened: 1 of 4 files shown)


konnov (approved):
Look great!

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -274,12 +275,17 @@ object SourceOption {
...
     def apply(file: java.io.File): FileSource = {
-      val format = FilenameUtils.getExtension(file.getName()) match {
-        case "json"  => Format.Json
-        case "tla"   => Format.Tla
-        case unknown => throw new PassOptionException(s"Unsupported file format ${unknown}")
+      val fname = file.getName()
+      val format = FilenameUtils.getExtension(fname) match {
+        case "tla"                               => Format.Tla
+        case "json" if hasItfSubExtension(fname) => Format.Itf
+        case "json"                              => Format.Json
```
Kukovec, line 287: How about
```suggestion
        case "json" => if (hasItfSubExtension(fname)) Format.Itf else Format.Json
```
?

Kukovec (approved):

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
Kukovec: I'm not going to push back on that, I defer to your preference.

Issue #2199, opened 2022-10-08
Shai: Return proper data from parse errors 


shonfeder:
Some followup is needed to #2186, I missed picking up the parser errors, since those are all currently handled by exceptions. I should be able to pick them up by wrapping the outerlayer in the Sany pass with a `Try`

PR #2234, opened 2022-10-18
Allow SanyImporter.loadFromSource to return a try

shonfeder:
Contributes to #2199

## Context

Our current practice in the parsing pass is to signal all errors via exceptions
that bubble up to the top, but the rest of the passes are able to return their
errors via the `PassResult`. Much (if not most) of the SANY code is already
hadnling errors through `Try`s or `Either`s, but we've thus far been casting
those into exceptions at the final mile.

This PR starts the process of converting the exception-based error handling into
ADT-based error handling in the parser pass ([as recommended in the Scala
docs](https://docs.scala-lang.org/overviews/scala-book/functional-error-handling.html)).

## Reviewing

Reviewing by commit should be helpful.

This initial, small PR is meant to illustrate how these conversions from
exceptional errors to `Try`/`Either` errors will look, and followups towards
#2199 will likely proceed in the same way but in larger batches. So please
consider the general shape of the change proposed here to see whether it make
sense, or whether you think an alternative approach would be preferable.

### DIFF
--- a/shai/src/main/scala/at/forsyte/apalache/shai/v1/TransExplorerService.scala
+++ b/shai/src/main/scala/at/forsyte/apalache/shai/v1/TransExplorerService.scala
@@ -26,6 +26,7 @@ import zio.{Ref, ZEnv, ZIO}
 import com.typesafe.scalalogging.Logger
 import at.forsyte.apalache.infra.passes.options.OptionGroup
 import at.forsyte.apalache.infra.passes.PassChainExecutor
+import ujson.Obj
 
 // TODO The connection type will become enriched with more structure
 // as we build out the server
@@ -178,7 +179,11 @@ class TransExplorerService(connections: Ref[Map[UUID, Conn]], logger: Logger)
             output = Output(Some(new java.io.File("."))),
         )
       }
-      PassChainExecutor.run(new ParserModule(options)).left.map(code => s"Parsing failed with error code: ${code}")
+      PassChainExecutor
+        .run(new ParserModule(options))
+        .left
+        // TODO: Consolidate with pass error json formed in CmdExecutorService
+        .map(err => Obj("error_type" -> "pass_failure", "data" -> err).toString())
     } catch {
       case e: Throwable => Left(s"Parsing failed with exception: ${e.getMessage()}")
     }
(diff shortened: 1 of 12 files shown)


shonfeder:
Thinking on this a bit more: a shorter path to just getting the data we need out for Shai RPCs is to wrap the outer-level invocation in a catch that will let us obtain the error data raised.

I think given the amount of changes required here could signal that this path would embark on a refactor that would delay the planned delivery for MBT stuff.

thpani:
> That said, I think this change is still for the better, as each step away from using exception for errors is a step in the right direction IMO.

While functionally equivalent, exceptions are the more powerful construct, precisely because they can jump the call stack. In a multi-purpose language, I'm rather skeptical of dragging error behavior information around in all function signatures.

That said, I think Scala botched it by removing Java's checked exceptions, leaving monadic error handling as the only safe option. Looks like this is going to change though with [CanThrow Capabilities](https://docs.scala-lang.org/scala3/reference/experimental/canthrow.html). (Also see [this blog post](https://medium.com/scala-3/safer-exceptions-in-scala-3-12c47100f050).)

I'm curious what others think.

> I think given the amount of changes required here could signal that this path would embark on a refactor that would delay the planned delivery for MBT stuff.

Very thoughtful observation! 👍

konnov:
I think that replacing exceptions with `Either` is a good idea for passes, as they are our high-level access points. However, I am with @thpani on that replacing all exceptions with `Either` will only complicate matters. Especially, in the parser, there is no meaningful path one can take on parse errors, but bubble up the error. So I really like the idea of catching up exceptions in passes and wrapping them with `Either`.

shonfeder:
Good points, and thanks also for the discussion during today's standup.

Happy to concede the point on this particular changeset and close the PR. For the longer-term design practice, I'll keep advocating for tracking recoverable errors into the type system when I think there's benefit :D 

> In a multi-purpose language, I'm rather skeptical of dragging error behavior information around in all function signatures.

In my experience, it is generally quite light weight and really improves ability to reason about the code once you're using it consistently. It really feels quite freeing to call a function `f : a -> b` and know that you're just going to get a `b`, unless something truly unrecoverable happens.

> there is no meaningful path one can take on parse errors, but bubble up the error

I think this is maybe true for errors arising from SANY in our current set up, but in the  general case, you might collect multiple different parser errors and report them all at once, or try alternative parsing strategies, or run some heuristics on the error and the input to provide really clear feedback, etc..

shonfeder:
Thanks both for the thoughtful feedback and review!

Issue #2204, opened 2022-10-11
Bogus error with mutual recursive operators


lemmy:
Apalache claims that TLA+ parser rejects (`Error by TLA+ parser: Found a cyclic dependency among operators: IsEven, IsOdd`) a spec that it doesn't:

```tla
---- MODULE F ----
EXTENDS Naturals

RECURSIVE IsEven(_)

RECURSIVE IsOdd(_)

IsEven(n) ==
    IF n = 0
    THEN TRUE
    ELSE IsOdd(n-1)

IsOdd(n) ==
    IF n = 0
    THEN FALSE
    ELSE IsEven(n-1)

==================
```

```zsh
java -cp ~/TLC/tla2tools_1-7-0.jar tla2sany.SANY F.tla

****** SANY2 Version 2.1 created 24 February 2014

Parsing file /Users/markus/src/TLA/_specs/ewd998/F.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/Naturals.tla
Semantic processing of module Naturals
Semantic processing of module F
```

```zsh
apalache-mc check F.tla                               
16:46:05.965 [main] INFO at.forsyte.apalache.tla.tooling.opt.CheckCmd - Loading configuration
16:46:05.984 [main] INFO at.forsyte.apalache.tla.tooling.opt.CheckCmd -   > TLC config file found in specification directory. To enable it, pass --config=F.cfg.
# Usage statistics is ON. Thank you!
# If you have changed your mind, disable the statistics with config --enable-stats=false.

Output directory: /Users/markus/src/TLA/_specs/ewd998/_apalache-out/F.tla/2022-10-11T16-46-06_16504550803085844238
# APALACHE version: 0.29.2 | build: 8a342d2                       I@16:46:06.168
Tuning: search.outputTraces=false                                 I@16:46:06.186
PASS #0: SanyParser                                               I@16:46:06.323
Parsing file /Users/markus/src/TLA/_specs/ewd998/F.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/Naturals.tla
Check the dependency graph in dependencies.dot. You can view it with graphviz. E@16:46:06.453
Error by TLA+ parser: Found a cyclic dependency among operators: IsEven, IsOdd E@16:46:06.455
It took me 0 days  0 hours  0 min  0 sec                          I@16:46:06.455
[...]

PR #2242, opened 2022-10-19
Shai: Return parsing pass error data in RPCs

### LINKED ISSUE #2199: Shai: Return proper data from parse errors

shonfeder:
Closes #2199, #2204

(See #2234 for additional notes on context and the motivation for this approach).

- Catches `SanyException`s from the highest level of the SANY parser
pass and makes these available in the `PassResult`.
- Promotes the general RPC error data into protbuf specs, allowing more
consistentcy and reducing the amount of dynamic JSON.
- Includes some light refactoring as incidental cleanup.
- Updates tests.

-------

### DIFF
--- a/passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/SanyParserPassImpl.scala
+++ b/passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/SanyParserPassImpl.scala
@@ -19,6 +19,7 @@ import at.forsyte.apalache.infra.passes.options.OptionGroup
 import scala.util.Try
 import scala.util.Failure
 import scala.util.Success
+import at.forsyte.apalache.tla.imp.SanyException
 
 /**
  * Parsing TLA+ code with SANY.
@@ -71,6 +72,11 @@ class SanyParserPassImpl @Inject() (
     Right(modules.get(rootName).get)
   }
 
+  private val loadFromTlaSource: SourceOption => PassResult = {
+    case SourceOption.StringSource(content, aux, _) => loadFromTlaString(content, aux)
+    case SourceOption.FileSource(file, _)           => loadFromTlaFile(file)
+  }
+
   private def saveLoadedModule(module: TlaModule): Either[PassFailure, Unit] = {
     // save the output
     writeOut(writerFactory, module)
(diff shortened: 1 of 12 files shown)


passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/SanyParserPassImpl.scala
```diff
@@ -89,25 +95,29 @@ class SanyParserPassImpl @Inject() (
...
-      sortedModule <- sortDeclarations(rootModule)
-      _ <- saveLoadedModule(sortedModule)
-    } yield sortedModule
+    try {
+      parseSource(options.input.source)
+    } catch {
+      case err: SanyException =>
+        val msg = err.getMessage()
+        logger.error(s"Parsing error: ${msg}")
+        passFailure(List(msg), ExitCodes.ERROR)
```
thpani, line 109: It think this obsoletes one path in some of the exception adapters, e.g.

https://github.com/informalsystems/apalache/blob/7477b1be3dae41b187298427d2559da3d4bf1f0d/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/ParserExceptionAdapter.scala#L18-L19

passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/SanyParserPassImpl.scala
```diff
@@ -89,25 +95,29 @@ class SanyParserPassImpl @Inject() (
...
-            case FileSource(file, _)           => loadFromTlaFile(file)
-          }
-      }
-      sortedModule <- sortDeclarations(rootModule)
-      _ <- saveLoadedModule(sortedModule)
-    } yield sortedModule
+    try {
+      parseSource(options.input.source)
+    } catch {
+      case err: SanyException =>
```
thpani, line 106: Could you explain why we catch `SanyException` here, but not `AnnotationParserError` (like in `ParserExceptionAdapter`)?

passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/SanyParserPassImpl.scala
```diff
@@ -89,25 +95,29 @@ class SanyParserPassImpl @Inject() (
...
+        val msg = err.getMessage()
+        logger.error(s"Parsing error: ${msg}")
```
thpani, line 108: I think this closes #2204

thpani (approved):
Looks good, also nice work on replacing string/JSON with more structured protobuf data!

Just a question and a comment on how this relates to our current exception adapters.

shonfeder:
Very good question about the relation to the exception adapters. I'll need to do a grep of the codebase to ensure that:

- All places where `SanyException` are raised are now caught by this `Either`
- That there are not additional exceptions we use to signal parse errors, which should also be caught.

Once those are addressed, I should remove those cases from the exception adapters.

passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/SanyParserPassImpl.scala
shonfeder: I just didn't realize that this would be needed, and no tests or static analysis was able to warn me. Thank you for catching this!

thpani: 😬

PR #2244, opened 2022-10-20
ERC20 example

bugarela:
Hello :octocat: 

This adds the ERC20 example, written by @konnov and adapted to use variants by me.


test/tla/ERC20_typedefs.tla
```diff
@@ -0,0 +1,26 @@
...
+  Type definitions for the module ERC20.
+
+  An account address, in our case, simply an uninterpreted type ADDR.
+
+  @typeAlias: transfer = { id: Int, fail: Bool, sender: ADDR, toAddr: ADDR, value: Int };
+  @typeAlias: approve = { id: Int, fail: Bool, sender: ADDR, spender: ADDR, value: Int };
+  @typeAlias: transferFrom = { id: Int, fail: Bool, sender: ADDR, fromAddr: ADDR, toAddr: ADDR, value: Int };
+
+  A transaction is a discriminated union:
+  @typeAlias: TX = None({ id: Int, fail: Bool }) | Transfer($transfer) | Approve($approve) | TransferFrom($transferFrom) ;
```
konnov, line 12: How could `None` fail?

test/tla/ERC20.tla
```diff
@@ -0,0 +1,314 @@
...
+Approve(fields) ==
+    Variant("Approve", fields)
+
+\* @type: $transferFrom => TX;
+TransferFrom(fields) ==
+    Variant("TransferFrom", fields)
+
+\* Initialize an ERC20 token.
+Init ==
+    /\ balanceOf = [a \in ADDR |-> 100]
```
konnov, line 79: Shall we introduce a constant for the initial balance?

konnov (approved):
Thanks for refactoring!

test/tla/ERC20_typedefs.tla
bugarela: The value is unused, as is the `id`. I removed them.

test/tla/ERC20.tla
bugarela: I don't think so. As we were discussing yesterday, `Init` is also a form of parametrization, so that would be just a transfer of parameters IMO.

konnov: I mean, `ERC20.tla` is parameterized with `ADDR` and `AMOUNTS`. This parameters are instantiated in `MC_ERC20.tla`. In this sense, we already have parameters in this spec, so why not introducing another one?

bugarela: Ok, I think it makes sense to prioritize parametrization by constants when possible (as oppose to defining multiple `Init`s). I'll add the parameter.

bugarela: I've added a constant of type `ADDR -> Int` to hold the initial balances for all accounts, so we can define accounts with different initial balances.

shonfeder:
The failing unit test is unrelated, and is a nondet failure due to some PBT generation.

konnov:
Could you maybe update the same spec in https://github.com/informalsystems/tla-apalache-workshop/

Issue #2258, opened 2022-10-28
Extend ITF format with variable types


Kukovec:
At present, only the variable names and values are recorded. This can be a problem in reconstruction via the typed builder in the presence of empty sets/sequences, because empty collections need type tag annotations in the builder.
Concretely, we should either
1. Change the "vars" field to an object, the keys of which are the variables, and the values of which are string representations of TT1 types, or
2. Add a new field "varTypes" (or similar), with that same info, preserving "vars" as a list, as not to break compatibility.

shonfeder:
I think (1) is a nicer option.

I think this also opens up an opportunity to make the trace much more concise and readable!

At the time ITF was designed I argued against the current data representation, because it's treatment of scalars and compound values is so different: https://github.com/informalsystems/apalache/blob/main/docs/src/adr/015adr-trace.md#discussion

At that time, Igore argued that

> The more regular approach is less concise. In the future, we might want to add a flag that lets the user choose between the regular output and the output proposed in this ADR, which is more ad hoc.

However, if we put the variable types in this header data, then we can get rid of all the `#map` and `#set` tags, cause all that data would be readily available in the variable's types. WDYT?

Kukovec:
Yes, I'm more pro (1) myself, ITF isn't really mature enough where we'd need to care about compatibility IMO.
RE compound types, that's more of an orthogonal discussion, but the `#` values did strike me as odd at the time, I wouldn't mind changing them in the future.

shonfeder:
This will be much easier to communicate and gather feedback on with an example.

Here's a fragment of the current ITF format, as per ADR015:

```json
{
  "#meta": {
    "source": "MC_MissionariesAndCannibalsTyped.tla"
  },
  "vars": [ "bank_of_boat", "who_is_on_bank" ],
  "states": [
    {
      "#meta": { "index": 0 },
      "bank_of_boat": "E",
      "who_is_on_bank": {
        "#map": [
          [ "E", { "#set": [ "c1_OF_PERSON", "c2_OF_PERSON",
                             "m1_OF_PERSON", "m2_OF_PERSON" ] } ],
          [ "W", { "#set": [] } ]
        ]  
      }
    },
    {
      "#meta": { "index": 1 },
      "bank_of_boat": "W",
      "who_is_on_bank": {
        "#map": [
          [ "E", { "#set": [ "c1_OF_PERSON", "m1_OF_PERSON" ] } ],
[...]

PR #2267, opened 2022-11-03
ITF ADR update

Kukovec:

First half of #2258 


docs/src/adr/015adr-trace.md
```diff
@@ -230,11 +230,11 @@ as `CONSTANTS` in TLA+. For example, the field may look like:
   "params": [ "Missionaries", "Cannibals" ]
 ```
 
-The field `vars` is an array of names that must be set in every state.
+The field `vars` is an object, the keys of which are names that must be set in every state, and the values are string representations of their types (as defined in [ADR002]).
 For example, the field may look like:
 
 ```js
-  "vars": [ "bank_of_boat", "who_is_on_bank" ]
+  "vars": { "bank_of_boat": "Str", "who_is_on_bank": "Str -> Set(PERSON)" }
```
thpani, line 237: Could "vars" carry more information in the future?
A more future-proof / backwards-compatible encoding would be

```suggestion
  "vars": [ { "name": "bank_of_boat", "type": "Str" },
            { "name": "who_is_on_bank", "type": "Str -> Set(PERSON)" } ]
```

thpani (commented):
I haven't really followed the discussion on this, so I'll leave the review to @shonfeder.

One question on the chosen encoding though:

thpani:
afaict @konnov authored this ADR, should we wait for his review next week?

Kukovec:
> afaict @konnov authored this ADR, should we wait for his review next week?

We can, but then I'll be blocked until he comes back + we have consensus.

shonfeder:
I think Igor's point about using the meta to maintain backwards compat makes sense.

Or, if we are going to break it, given we have known users, we should be planning a versioned format release, so that might call for a new ADR (or just an issue) explaining the versioning scheme, and then implementing the breaking changes on top of that.

But imo, just putting the variable type data in the meta seems most expedient.

shonfeder:
Oh, I see. Maybe your suggestion for a new field for this is the good middle path then?

konnov (commented):
As we briefly discussed online, I would prefer to have the type data in the `#meta` object. The reasons are:

 - The types syntax and the type system may change in the future. Moreover, with the current approach there is no issue of having traces that use different type systems.
 - The types are optional, e.g., if we produce an ITF trace with TLC.
 - In principle, the type information is redundant. Most of the time, it could be inferred from the right-hand sides of the assignments. However, types cannot be uniquely inferred for empty sets and empty functions.

Additionally, the current format is as minimal as possible. All additional info goes into the meta object. This gives us a bit of resiliency w.r.t. the format changes. All tools should support the absolute minimum. If they need additional data, it is optionally provided in the meta object.

Issue #2265, opened 2022-11-02
Rework all unit tests, which depend on `APALACHE_HOME`


Kukovec:
Following a Slack troubleshooting session, it seems to be a QoL barrier for first-time users, since Apalache reports test failures if one simply runs `make` on a freshly cloned repository, without first configuring the home variable.
To avoid this, we should either turn affected unit tests into integration tests, or embed the files, which are now being read from the folder.

thpani:
>  To avoid this, we should either turn affected unit tests into integration tests, or embed the files, which are now being read from the folder.

The affected unit tests are for `SanyImporter`, not sure how we would get equivalent tests for this.

IMO it would be better to copy and adept the provisions in `TestTypeCheckerTool` to print a helpful error message:

https://github.com/informalsystems/apalache/blob/20b0c4fdaff09686bbedba479a1866b46012394a/passes/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala#L45-L52

thpani:
Also, FWIW, our install-from-source instructions cover this:

https://github.com/informalsystems/apalache/blob/20b0c4fdaff09686bbedba479a1866b46012394a/docs/src/apalache/installation/source.md?plain=1#L13

Kukovec:
> The affected unit tests are for `SanyImporter`, not sure how we would get equivalent tests for this.

You can import from an embedded string, that avoids looking up anything in the filesystem and the need for the home variable.

> Also, FWIW, our install-from-source instructions cover this:

You shouldn't have to use direnv when first installing IMO. It should be made possible to just run `make` and then `<dir>/bin/apalache-mc ...`. Advanced/efficient invocations should be the next step, once you realize how tedious the alternative is.

thpani:
> You can import from an embedded string, that avoids looking up anything in the filesystem and the need for the home variable.

afaict these instance/extend standard modules, how do you load those from a variable?

> You shouldn't have to use direnv when first installing IMO

you don't have to use direnv, one can simply source the files as it says in the installation docs. this is not about advanced/efficient invocation, simply setting up the build env.

Kukovec:
It's less about what is in the docs (after all, the docs _do_ mention setting `APALACHE_HOME`), and more about what the most streamlined new-user experience is.

PR #2270, opened 2022-11-04
Don't run tests on the default make target

shonfeder:
Aims to address the problem motivating #2265, which was a build failure
by a new user due to tests failing due to unset `APALACHE_HOME`.

Thanks to @lasarojc for initially reporting the build problem.

This change was discussed and decided upon with @Kukovec and @thpani during a standup.


thpani (approved):
LGTM, thanks for picking this up!

PR #2273, opened 2022-11-08
ITF trace reader

### LINKED ISSUE #2272: Implement a reader for ITF -> TlaIR
We need to be able to read ITF files as a sequence of `State`s in the IR (e.g. for trace evaluation).

Kukovec:

Closes #2272

Adds support for reading in ITF-format traces (as defined by the latest update to ADR15: #2267). 
Followup: Use the reader to support ITF input to `tracee --trace`.

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
@@ -218,7 +218,7 @@ class ItfToTla[T <: JsonRepresentation](
       stateJson.allFieldsOpt match {
         case None => throw new JsonDeserializationError(s"Malformed JSON: Value at $STATES_FIELD is not an object.")
         case Some(varsInState) =>
-          val undeclared = varsInState -- varSet
+          val undeclared = (varsInState -- varSet) - META_FIELD // #meta is the only extra field allowed in states
           if (undeclared.nonEmpty)
             throw new JsonDeserializationError(
                 s"Undeclared variable(s) present in state: ${undeclared.mkString(", ")}.")
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
@@ -1,0 +1,249 @@
+package at.forsyte.apalache.io.itf
+
+import at.forsyte.apalache.io.itf.ItfToTla._
+import at.forsyte.apalache.io.json.{JsonDeserializationError, JsonRepresentation, ScalaFactory}
+import at.forsyte.apalache.io.lir.Trace
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.typecomp.TBuilderInstruction
+import at.forsyte.apalache.tla.types.parser.DefaultType1Parser
+import at.forsyte.apalache.tla.types.{tla, ModelValueHandler}
+
+/**
+ * Reads a sequence of States, i.e. mappings from variable names to TlaEx values, from an ITF json.
+ *
+ * @tparam T
+ *   Any class extending JsonRepresentation
+ * @author
+ *   Jure Kukovec
+ */
+class ItfToTla[T <: JsonRepresentation](
+    scalaFactory: ScalaFactory[T]) {
+
+  /**
+   * Checks that the Apalache-specific ITF requirements are met, as defined in ADR15. If yes, returns a mapping from
+   * variables to their types.
+   */
+  private[itf] def validateShapeAndGetTypes(json: T): Map[String, TlaType1] = {
+    val varTypesOpt = for {
+      meta <- json.getFieldOpt(META_FIELD)
+      varTypes <- meta.getFieldOpt(VAR_TYPES_FIELD)
+    } yield varTypes
+    val varTypesJson = varTypesOpt.getOrElse {
+      throw new JsonDeserializationError("ITF JSON does not satisfy Apalache requirements. " +
+        "Toplevel \"#meta\" field must exist and contain \"varTypes\". " +
+        "See ADR15: https://apalache.informal.systems/docs/adr/015adr-trace.html#itf-as-input-to-apalache. ")
+    }
+
+    val diffOpt = for {
+      vars <- json.getFieldOpt(VARS_FIELD)
+      varTypesKeys <- varTypesJson.allFieldsOpt
+    } yield {
+      val varsSet = scalaFactory.asSeq(vars).map { scalaFactory.asStr }.toSet
+      // Symmetric difference
+      (varsSet -- varTypesKeys) ++ (varTypesKeys -- varsSet)
+    }
+
+    val diff = diffOpt.getOrElse {
+      throw new JsonDeserializationError(s"\"$VARS_FIELD\" or \"$VAR_TYPES_FIELD\" contents are malformed.")
+    }
+    if (diff.nonEmpty)
+      throw new JsonDeserializationError(s"\"$VARS_FIELD\" and \"$VAR_TYPES_FIELD\" define different states. " +
+        s"Variable(s) ${diff.toSeq.mkString(", ")} present in exactly one of the two fields.")
+
+    varTypesJson.allFieldsOpt.get.map { k =>
+      k -> DefaultType1Parser.parseType(scalaFactory.asStr(varTypesJson.getFieldOpt(k).get))
+    }.toMap
+  }
+
+  /**
+   * Some values, like the predefined sets (Int, Nat, etc.) are encoded as "UNSERIALIZABLE_FIELD": `<value>` Currently,
+   * per our discussions, we disallow the use of this tag, but the function can be extended in the future.
+   *
+   * E.g. we could return Some(tla.intSet()) on str = "Int".
+   */
+  private[itf] def attemptUnserializable(json: T): Option[TBuilderInstruction] =
+    if (json.allFieldsOpt.contains(Set(UNSERIALIZABLE_FIELD))) {
+      val str = scalaFactory.asStr(json.getFieldOpt(UNSERIALIZABLE_FIELD).get)
+      throw new JsonDeserializationError(s"Unserializable value $str is disallowed.")
+    } else None
+
+  // Note: attemptUnserializable + getOrElse is future-proof, and allows us to eventually case-match
+  // unserializable values.
+
+  private[itf] def typeDrivenBuild(json: T, tt1: TlaType1): TBuilderInstruction =
+    attemptUnserializable(json).getOrElse {
+      tt1 match {
+        case BoolT1 =>
+          tla.bool(scalaFactory.asBool(json))
+        case StrT1 =>
+          val s = scalaFactory.asStr(json)
+          if (ModelValueHandler.isModelValue(s))
+            throw new JsonDeserializationError(
+                s"$s is annotated as a string, but has the value of an uninterpreted literal.")
+          tla.str(s)
+        case ct: ConstT1 =>
+          val raw = scalaFactory.asStr(json)
+          // ConstT1-type literals must adhere to the pattern regex
+          val typeIndexOpt = ModelValueHandler.typeAndIndex(raw)
+          val (constT, idx) = typeIndexOpt.getOrElse {
+            throw new JsonDeserializationError(s"$raw is annotated as an uninterpreted literal, but has string form.")
+          }
+          // The pattern-suffix must also match the declared type
+          if (ct != constT) {
+            throw new JsonDeserializationError(s"$raw is annotated as $ct but has a $constT value.")
+          }
+          tla.const(idx, constT)
+        case IntT1 =>
+          // Big integers are written as strings, not JS integers
+          if (json.allFieldsOpt.contains(Set(BIG_INT_FIELD))) {
+            val biString = scalaFactory.asStr(json.getFieldOpt(BIG_INT_FIELD).get)
+            tla.int(BigInt(biString))
+          } else tla.int(scalaFactory.asInt(json))
+
+        case SeqT1(elemT) =>
+          val elems = scalaFactory.asSeq(json)
+          val args = elems.map { typeDrivenBuild(_, elemT) }
+          // Empty collections have special builder ctors
+          if (args.isEmpty) tla.emptySeq(elemT)
+          else tla.seq(args: _*)
+
+        case RecT1(fieldTypes) =>
+          val keys = fieldTypes.keySet
+          val diffOpt = for {
+            fields <- json.allFieldsOpt
+          } yield (keys -- fields) ++ (fields -- keys) // symmetric difference
+
+          // The above is None, iff json is not an Object with fields
+          val diff = diffOpt.getOrElse {
+            throw new JsonDeserializationError(s"Record object contents are malformed.")
+          }
+
+          // The record type and value must declare the _same_ fields
+          if (diff.nonEmpty)
+            throw new JsonDeserializationError(
+                s"Record object and its type annotation $tt1 define different key sets. " +
+                  s"The keys ${diff.toSeq.mkString(", ")} are present in exactly one of the two.")
+
+          // TODO: We currently don't check the key shape against any sort of pattern.
+          // Technically, "_-?*+" is a valid record field, as far as this code is concerned.
+          // Do we want to restrict the permissible record field names, and if so, to what?
+
+          val recElems = json.allFieldsOpt.get.toSeq.map { k =>
+            k -> typeDrivenBuild(json.getFieldOpt(k).get, fieldTypes(k))
+          }
+
+          // Records must contain at least 1 kv pair
+          if (recElems.isEmpty) throw new JsonDeserializationError(s"Records may not be empty.")
+
+          tla.rec(recElems: _*)
+
+        case TupT1(elems @ _*) =>
+          if (!json.allFieldsOpt.contains(Set(TUP_FIELD)))
+            throw new JsonDeserializationError(s"Tuple-type annotated values must declare a \"$TUP_FIELD\" field.")
+          val tupSeq = scalaFactory.asSeq(json.getFieldOpt(TUP_FIELD).get)
+
+          // The declared tuple length must match the # of elements
+          val valLen = tupSeq.length
+          val typeLen = elems.length
+          if (valLen != typeLen)
+            throw new JsonDeserializationError(
+                s"Tuple value and type have $valLen and $typeLen elements respectively."
+            )
+
+          val args = tupSeq.zip(elems).map { case (elemJson, elemT) =>
+            typeDrivenBuild(elemJson, elemT)
+          }
+          tla.tuple(args: _*)
+
+        case SetT1(elemT) =>
+          if (!json.allFieldsOpt.contains(Set(SET_FIELD)))
+            throw new JsonDeserializationError(s"Set-type annotated values must declare a \"$SET_FIELD\" field.")
+          val setSeq = scalaFactory.asSeq(json.getFieldOpt(SET_FIELD).get)
+          val args = setSeq.map { elemJson =>
+            typeDrivenBuild(elemJson, elemT)
+          }
+          // Empty collections have special builder ctors
+          args match {
+            case Seq() => tla.emptySet(elemT)
+            case seq   => tla.enumSet(seq: _*)
+          }
+
+        case FunT1(argT, resT) =>
+          if (!json.allFieldsOpt.contains(Set(MAP_FIELD)))
+            throw new JsonDeserializationError(s"Function-type annotated values must declare a \"$MAP_FIELD\" field.")
+
+          val funSeq = scalaFactory.asSeq(json.getFieldOpt(MAP_FIELD).get)
+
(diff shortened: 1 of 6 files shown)


tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+ *   Jure Kukovec
+ */
+class ItfToTla[T <: JsonRepresentation](
+    scalaFactory: ScalaFactory[T]) {
+
+  /**
+   * Checks that the Apalache-specific ITF requirements are met, as defined in ADR15. If yes, returns a mapping from
+   * variables to their types.
+   */
+  private[itf] def validateShapeAndGetTypes(json: T): Map[String, TlaType1] = {
```
shonfeder, line 26: What about making this and similar methods in here return an `Either` that tracked the error rather than throwing exceptions? It would make it easier to reason about the possible error cases, and you could drop the `validate` or `attempt` prefixes in the methods, since it'd be clear from the types that they could fail, and you can sequence with for comprehensions.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+      vars <- json.getFieldOpt(VARS_FIELD)
+      varTypesKeys <- varTypesJson.allFieldsOpt
+    } yield {
+      val varsSet = scalaFactory.asSeq(vars).map { scalaFactory.asStr }.toSet
+      // Symmetric difference
+      (varsSet -- varTypesKeys) ++ (varTypesKeys -- varsSet)
+    }
+
+    val diff = diffOpt.getOrElse {
+      throw new JsonDeserializationError(s"\"$VARS_FIELD\" or \"$VAR_TYPES_FIELD\" contents are malformed.")
```
shonfeder, line 47: Why do we report `VAR_TYPES_FIELD` malformed here, rather than when the types are extracted on line 27? Also, what kind of malformed could this be?

shonfeder, line 47: Seems like the logic would be clearer if you first validate whether the fields are present and in the expected form, and handle reporting errors on that, then do the validation of whether there is a type for each variable, rather than mixing up the two different validation checks here.

shonfeder (changes requested):
I left just a few initial comments, but I wanted to propose this alternative approach:

I believe that if we define a representation of the ITF format in Scala classes, we can serialization and deserialization more or less for free via ujson and upickle.

This could give us an approach that covered this new use case but resulted in overall reduction of code.

We'd end up with two mapping `TlaModule -> ITF` and `ITF -> TlaModule`, and then the de/serialization of the `ITF` type itself should be automated.

Any thoughts on this alternative?

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
Kukovec: This feels reminiscent of https://github.com/informalsystems/apalache/pull/2234

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
Kukovec: l28 checks if `#meta` exists on the toplevel object.
l29 checks if `varTypes` exists in the meta.

l38 checks `vars` exists on the toplevel object
l39 checks `varTypes` (which we know exists) is a JSON Object (something that, itself, has fields)

Malformed-ness here is when `vars` either doesn't exist, or `varTypes` is e.g. an int. You can see a malformed example here

https://github.com/informalsystems/apalache/blob/1707cf226fa41065d831aee82867ca74671eae60/tla-io/src/test/scala/at/forsyte/apalache/io/itf/TestItfToTla.scala#L30-L46

But good catch, I didn't add a test for when `varTypes` is not an Object, I should do that.

Kukovec: My reasoning is a bit different, I try to catch every form of exception the moment there is sufficient evidence to record a violation. Also, to me, whether there is a type for every variable and whether fields are missing are not really different things; in both cases the input is equally malformed.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+  private[itf] def validateShapeAndGetTypes(json: T): Map[String, TlaType1] = {
+    val varTypesOpt = for {
+      meta <- json.getFieldOpt(META_FIELD)
+      varTypes <- meta.getFieldOpt(VAR_TYPES_FIELD)
+    } yield varTypes
+    val varTypesJson = varTypesOpt.getOrElse {
+      throw new JsonDeserializationError("ITF JSON does not satisfy Apalache requirements. " +
+        "Toplevel \"#meta\" field must exist and contain \"varTypes\". " +
+        "See ADR15: https://apalache.informal.systems/docs/adr/015adr-trace.html#itf-as-input-to-apalache. ")
+    }
```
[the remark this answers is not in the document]
Kukovec: > I left just a few initial comments, but I wanted to propose this alternative approach:
> 
> I believe that if we define a representation of the ITF format in Scala classes, we can serialization and deserialization more or less for free via ujson and upickle.
> 
> This could give us an approach that covered this new use case but resulted in overall reduction of code.
> 
> We'd end up with two mapping `TlaModule -> ITF` and `ITF -> TlaModule`, and then the de/serialization of the `ITF` type itself should be automated.
> 
> Any thoughts on this alternative?

Two, mainly. Firstly, we already have an in-Scala representation: `IndexedSeq[Map[String, TlaEx]]` (`lir.Trace`). Is is the only real programmatic representation of a trace you need IMO.
Secondly, the problem I have with outsourcing (de)serialization, is that you use any sort of useful error message feedback; Assume you have a more complicated Scala `ITF` class and you want to use a library to automatically deserialize it from JSON. Because you're outsourcing, the only error message you can get is `SomeException("Input file malformed.")`. It won't tell you, for example, that your JSON declares 3 values in `vars`, but the states only have 2 variables, because that's not a JSON formatting issue, that's a logical constraint not being satisfied.

Kukovec:
> I left just a few initial comments, but I wanted to propose this alternative approach:
> 
> I believe that if we define a representation of the ITF format in Scala classes, we can serialization and deserialization more or less for free via ujson and upickle.
> 
> This could give us an approach that covered this new use case but resulted in overall reduction of code.
> 
> We'd end up with two mapping `TlaModule -> ITF` and `ITF -> TlaModule`, and then the de/serialization of the `ITF` type itself should be automated.
> 
> Any thoughts on this alternative?

Two, mainly. Firstly, we already have an in-Scala representation: `IndexedSeq[Map[String, TlaEx]]` (`lir.Trace`). It's the only real programmatic representation of a trace you need IMO.
Secondly, the problem I have with outsourcing (de)serialization, is that you use any sort of useful error message feedback; Assume you have a more complicated Scala `ITF` class and you want to use a library to automatically deserialize it from JSON. Because you're outsourcing, the only error message you can get is `SomeException("Input file malformed.")`. It won't tell you, for example, that your JSON declares 3 values in `vars`, but the states only have 2 variables, because that's not a JSON formatting issue, that's a logical constraint not being satisfied.

shonfeder:
> we already have an in-Scala representation: IndexedSeq[Map[String, TlaEx]]

This is a representation of a counter example, not of the ITF format.

Kukovec:
> > we already have an in-Scala representation: IndexedSeq[Map[String, TlaEx]]
> 
> This is a representation of a counter example, not of the ITF format.

ITF is just a particular presentation (output format) of the logical concept of a trace, at least in my view.

shonfeder:
> ITF is just a particular presentation (output format) of the logical concept of a trace, at least in my view.

This perspective certainly makes sense in the abstract. But it doesn't change the fact that the format has a particular structure which is significantly different from a sequence of maps of TLA expressions (which isn't, btw, the platonic form of a trace, but also just a particular format ;) ).

From my perspective, you seem to be conflating deserialization and data validation and this seems to be a persistent difference in our points of view. Because we have not taken a very disciplined approach to serialization and deserialization in most of this code, we are doing a lot of manipulation with representations of JSON. But JSON is meant as a data exchange format, not as a data model, so the guarantees it can give and the kinds of representation it supports are very limited.

We can fix this latter problem by using automated serialization and deserialization to and from objects that are more faithful representations of our data model. This usually involves two steps:

1. Automatic deserialization.
2. Validation to ensure any needed contracts are enforced, if they exceed what can be captured in our data model.

This difference is discussed in the [upickle documentation](http://www.lihaoyi.com/post/HowtoworkwithJSONinScala.html#converting-tofrom-scala-data-types):

> Often you do not just want dynamically-typed JSON trees: you expect the data to conform to a particular schema, have your code use that schema in a safe way, and fail early (and loudly!) if the incoming data doesn't conform.

So what I'm suggesting is using the existing, out of the box functionality of the libraries we already depend on to perform this task, rather than rolling our own implementations of this pluming work.

>  Secondly, the problem I have with outsourcing (de)serialization, is that you use any sort of useful error message feedback; Assume you have a more complicated Scala ITF class and you want to use a library to automatically deserialize it from JSON. Because you're outsourcing, the only error message you can get is SomeException("Input file malformed."). It won't tell you, for example, that your JSON declares 3 values in vars, but the states only have 2 variables, because that's not a JSON formatting issue, that's a logical constraint not being satisfied.

I can run a test to verify, but my experience with ujson/upickle so far is that it gives quit clear errors regarding any parts of the JSON input that fail to deserialize (missing keys, invalid values, etc.). The validation errors you mention would need to be handled in a subsequent validation pass once the data is deserialized into our data model. 

It's possible that we've design the format in such a way that automated serialization and deserialization into types representing the data model isn't really feasible. But IMO we won't know until we try. If it does work out, we can likely save several hundred lines of code and reduce our maintenance burden.

I'd be willing to give a shot, if that would be helpful.

shonfeder:
Looking at the implementation of ITF currently, given the form that was decided on etc., I think there's not a great gain to trying to switch to auto deserialization at this point.

I'll retract that suggestion.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
shonfeder: > I try to catch every form of exception the moment there is sufficient evidence to record a violation

I guess you mean "catch every form of error". This is basically what I'm encouraging: you are currently performing inspection of the sets in the code "prior" to handling errors from invalid data. And, more generally, the code does not handle report the error at the earliest point it is detectable, instead requiring followup computations after other things have been processed, and wrapping several different kinds of errors into one.

Here's an example of how using `Either`s as I'm suggesting could help clean up the error handling here (imo). iiuc, this snippet accounts for all the entire body of this method:

```scala
  private[itf] def validateShapeAndGetTypes(json: T): Either[String, Map[String, TlaType1]] = for {
    meta <- json.getFieldOpt(META_FIELD).toRight("'#meta' field must exist")
    varTypes <- meta.getFieldOpt(VAR_TYPES_FIELD).toRight("'#meta' field must contain a `varTypes` field")
    varTypesKeys <- varTypes.allFieldsOpt.toRight("'#meta.varTypes' must be a map")
    vars <- json.getFieldOpt(VARS_FIELD).toRight("'vars' field must exist")
    varsSet = scalaFactory.asSeq(vars).map { scalaFactory.asStr }.toSet
    _ <- {
      val varsWithoutTypes = (varsSet -- varTypesKeys)
      Either.cond(varsWithoutTypes.isEmpty, (),
          s"$VARS_FIELD contains variables missing types in $VAR_TYPES_FIELD: ${varsWithoutTypes}")
    }
    _ <- {
      val typesWithoutVars = (varTypesKeys -- varsSet)
      Either.cond(typesWithoutVars.isEmpty, (),
          s"$VAR_TYPES_FIELD contains variables missing types in $VARS_FIELD: ${typesWithoutVars}")
    }
  } yield varTypes.allFieldsOpt.get.map { k =>
    k -> DefaultType1Parser.parseType(scalaFactory.asStr(varTypes.getFieldOpt(k).get))
  }.toMap

```

IMO, this has two benefits:

1. Users get more precise information about the source of each error.
2. Readers of the code get one logical line for each error-prone case, they can see at a glance what the error case is, rather than having to look several lines to a subsequent test to sees what exception is raised.

Once you're finished sequencing your error-prone computations, you can always drop out of the `Either` monad by throwing an exception with the error, but this approach lets us clearly isolate and declare the sequence of validation steps in a safe and, imo, quite perspicuous way.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+      val varsSet = scalaFactory.asSeq(vars).map { scalaFactory.asStr }.toSet
+      // Symmetric difference
+      (varsSet -- varTypesKeys) ++ (varTypesKeys -- varsSet)
+    }
+
+    val diff = diffOpt.getOrElse {
+      throw new JsonDeserializationError(s"\"$VARS_FIELD\" or \"$VAR_TYPES_FIELD\" contents are malformed.")
+    }
+    if (diff.nonEmpty)
+      throw new JsonDeserializationError(s"\"$VARS_FIELD\" and \"$VAR_TYPES_FIELD\" define different states. " +
```
shonfeder, line 50: Wouldn't a JSON deserialization error would occur prior to obtaining a `ujson.Value`?. This is rather a data-validation error.

Not sure how you want to do error handling here, but I wouldn't wrap it up with JSON deserialization.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+  private[itf] def typeDrivenBuild(json: T, tt1: TlaType1): TBuilderInstruction =
+    attemptUnserializable(json).getOrElse {
+      tt1 match {
+        case BoolT1 =>
+          tla.bool(scalaFactory.asBool(json))
+        case StrT1 =>
+          val s = scalaFactory.asStr(json)
+          if (ModelValueHandler.isModelValue(s))
+            throw new JsonDeserializationError(
+                s"$s is annotated as a string, but has the value of an uninterpreted literal.")
```
shonfeder, line 82: This is definitely not a JSON deserialization error. This is a type checking error :D

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+          if (ct != constT) {
+            throw new JsonDeserializationError(s"$raw is annotated as $ct but has a $constT value.")
+          }
+          tla.const(idx, constT)
+        case IntT1 =>
+          // Big integers are written as strings, not JS integers
+          if (json.allFieldsOpt.contains(Set(BIG_INT_FIELD))) {
+            val biString = scalaFactory.asStr(json.getFieldOpt(BIG_INT_FIELD).get)
+            tla.int(BigInt(biString))
+          } else tla.int(scalaFactory.asInt(json))
```
shonfeder, line 101: How will this work if the user provided an expression here like:

```json
{"#bigint": "12312312", "foo": "some junk"}
```

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+          // The above is None, iff json is not an Object with fields
+          val diff = diffOpt.getOrElse {
+            throw new JsonDeserializationError(s"Record object contents are malformed.")
+          }
+
+          // The record type and value must declare the _same_ fields
+          if (diff.nonEmpty)
+            throw new JsonDeserializationError(
+                s"Record object and its type annotation $tt1 define different key sets. " +
+                  s"The keys ${diff.toSeq.mkString(", ")} are present in exactly one of the two.")
```
shonfeder, line 125: Perhaps handle the key differences separately here, so you can report which keys are missing in which place.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+
+          // TODO: We currently don't check the key shape against any sort of pattern.
+          // Technically, "_-?*+" is a valid record field, as far as this code is concerned.
+          // Do we want to restrict the permissible record field names, and if so, to what?
```
shonfeder, line 129: I think it's ok to omit this.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+
+          val recElems = json.allFieldsOpt.get.toSeq.map { k =>
+            k -> typeDrivenBuild(json.getFieldOpt(k).get, fieldTypes(k))
```
shonfeder, line 132: Doesn't ujson provide an easy way of mapping over key/value pairs of an object? Why the need for the `option -> get -> toSeq` roundabout?

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,249 @@
...
+            val v = typeDrivenBuild(vJson, resT)
+            tla.tuple(k, v) // for funAsSet, we need a set of paris
+          }
+          // Empty collections have special builder ctors
+          val set =
+            if (args.isEmpty) tla.emptySet(TupT1(argT, resT))
+            else tla.enumSet(args: _*)
+
+          // TODO: Check if keys are duplicated or let SetAsFun rule catch that? Aternatively,
+          // treat it as double-except semantics
```
shonfeder, line 195: Seems this might be worth tracking in an issue?

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
Kukovec: Well. that's a matter of perspective. For me, the step "read a trace from a json file" is, in a sense, atomic, so anything that happens inside the code that does that "failed to deserialize" the json. It's also the most useful for the user, the end message should say that something is wrong with the json they've provided.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
Kukovec: Yes, in the provided json, which is why the json can't be read into a trace.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
Kukovec: Well, the problem is, we've set up reserved identifiers, which use tokens that are illegal in valid TLA names (e.g. `$`). If we allow arbitrary identifiers here, I can foresee a bug, several months down the line, when someone just happens to use `v$1` or something similar, and we get a cell-name collision.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
Kukovec: Because this is not written for `ujson`, but for the `JsonRepresentation` abstraction.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
Kukovec: The thing is, I'm not 100% sure about  `SetAsFun` semantics, I was hoping for @konnov's input

konnov: The semantics is here: https://github.com/informalsystems/apalache/pull/2273#discussion_r1018967098

Kukovec: This just links to my comment btw.

konnov: In short, `SetAsFun` would not do any error reporting on `SetAsFun({ <<1, 2>>, <<1, 3>>) })`. It would just interpret the set somehow. Looking at the code, it may even produce unsatisfiable constraints.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
shonfeder: To clarify: I think it's worth considering this approach, but I won't block on the use  of `Either`.

I do think it may be worth being more precise about which errors arising when, but if you think there are good reasons not to, I'm open to learn.

Kukovec:
Closed in favor of #2283

PR #2283, opened 2022-11-11
ITF trace reader - `Either`-based alternative

### LINKED ISSUE #2272: Implement a reader for ITF -> TlaIR

Kukovec:

closes #2272

Alternative to #2273, uses `Either`.

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
@@ -218,7 +218,7 @@ class ItfToTla[T <: JsonRepresentation](
       stateJson.allFieldsOpt match {
         case None => throw new JsonDeserializationError(s"Malformed JSON: Value at $STATES_FIELD is not an object.")
         case Some(varsInState) =>
-          val undeclared = varsInState -- varSet
+          val undeclared = (varsInState -- varSet) - META_FIELD // #meta is the only extra field allowed in states
           if (undeclared.nonEmpty)
             throw new JsonDeserializationError(
                 s"Undeclared variable(s) present in state: ${undeclared.mkString(", ")}.")
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
@@ -19,219 +19,285 @@ import at.forsyte.apalache.tla.types.{tla, ModelValueHandler}
 class ItfToTla[T <: JsonRepresentation](
     scalaFactory: ScalaFactory[T]) {
 
+  // Since we want to use Either, but scalaFactory methods throw, we try-catch and wrap in Left/Right
+  private def thrownToEither[K, V, E](eBuilder: String => E, method: K => V)(arg: K): Either[E, V] = try {
+    Right(method(arg))
+  } catch {
+    case e: JsonDeserializationError => Left(eBuilder(e.getMessage))
+  }
+
+  private def asStr(json: T): Either[ItfFormatError, String] = thrownToEither(ItfFormatError, scalaFactory.asStr)(json)
+  private def asInt(json: T): Either[ItfFormatError, Int] = thrownToEither(ItfFormatError, scalaFactory.asInt)(json)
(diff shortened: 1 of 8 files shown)


tla-io/src/test/scala/at/forsyte/apalache/io/itf/TestItfToTla.scala
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    val nat = singleUS("Nat")
+
+    assert {
+      itfToTla.attemptUnserializable(nat).exists(_.isLeft)
+    }
+  }
+
+  test("typeDrivenBuild") {
```
shonfeder, line 133: This test case is really long. We get much better feedback from tests if we can break them into smaller units, so that when we get "test case 'such and such property should hold' failed" we know from the name of the test what has failed.

It's not essential, but imo these tests will be much more helpful in case of breakage if they are broken into units that note what they are testing. Seems like the natural categorization based on your current tests is to break each type into its own test case?

This basically just amounts to replacing the (tbh, slightly cryptic ;) ) comments with test descriptions.

tla-io/src/test/scala/at/forsyte/apalache/io/itf/TestItfToTla.scala
```diff
@@ -0,0 +1,480 @@
...
+          .map(_.build)
+          .contains(tla
+                .setAsFun(tla.enumSet(
+                        Seq(1, 2).map { i => tla.tuple(tla.int(i), tla.int(i)) }: _*
+                    ))
+                .build))
+
+  }
+
+  test("getTrace") {
```
shonfeder, line 361: I don't see any tests here to check that the correct behavior occurs in the presence of junk fields, like `{"#set": [], "foo": "bar"}`.

Can you add a named test case for this (and similar edge cases, if you think of any?).

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,315 @@
...
+                      )
+                      bi <- typeDrivenBuild(json.getFieldOpt(k).get, fieldTypes(k))
+                    } yield k -> bi
+                  }
+              )
+            // Records must contain at least 1 kv pair
+            _ <- requirement(
+                recElems.nonEmpty,
+                ItfFormatError(s"Records may not be empty."),
+            )
```
shonfeder, line 192: This `requirement` idiom reads really nice to me!

shonfeder (approved):
Code LGTM!

One thing I'd still like to see here before merging: put the ITF field constants in place where they can be shared with the writer code. And we can replace the magic strings in the writer with reference to the constants in a followup.

I also find the test cases to be much larger than is preferable, and left some notes indicating what I think would help. But if you disagree or don't think it's worth the effort, I'm fine with merging as is.

Thanks for the discussions!

Kukovec:
> One thing I'd still like to see here before merging: put the ITF field constants in place where they can be shared with the writer code. And we can replace the magic strings in the writer with reference to the constants in a followup.

Yes, I have that planned as a followup. I'll open an issue for that.

tla-io/src/main/scala/at/forsyte/apalache/io/itf/ItfToTla.scala
```diff
@@ -0,0 +1,315 @@
...
+  val VAR_TYPES_FIELD: String = "varTypes"
+  val VARS_FIELD: String = "vars"
+  val STATES_FIELD: String = "states"
+
+  val BIG_INT_FIELD: String = "#bigint"
+  val TUP_FIELD: String = "#tup"
+  val SET_FIELD: String = "#set"
+  val MAP_FIELD: String = "#map"
+  val UNSERIALIZABLE_FIELD: String = "#unserializable"
+}
```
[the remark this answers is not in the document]
Kukovec: I'll move it and import it in the writer, see #2284.

tla-io/src/test/scala/at/forsyte/apalache/io/itf/TestItfToTla.scala
Kukovec: Sure, I can make tighter groups, np.

PR #2287, opened 2022-11-14
ITF field constants used in writer

### LINKED ISSUE #2284: ITF counetrexample writer should not use hardcoded field name strings
In #2283, the field name literals were added in an object. This object should be moved to where it can be shared between both the reader and the writer, and the writer should use the object fields instead of hardcoded strings.

Kukovec:

Closes #2284 + adds ITF error handling to exception adapter


shonfeder (approved):
Looks great!

Thanks for the followup!

PR #2307, opened 2022-11-28
Add TNT AST representation and de/serialization

shonfeder:


tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
```diff
@@ -0,0 +1,346 @@
...
+  val id: Int
+}
+
+/** The representation of TNT expressions */
+sealed trait TntEx extends WithID
+
+object TntEx {
+
+  implicit val rw: RW[TntEx] =
+    RW.merge(Name.rw, Bool.rw, TntInt.rw, Str.rw, App.rw, Lambda.rw, Let.rw)
```
konnov, line 75: I wonder whether we should prefix the names of all the classes with `Tnt`. Otherwise, IDEs would keep auto-suggesting these classes whenever somebody is typing `Name` or `Bool`. Scala itself would not have a problem here, but I would not like to have these auxiliary classes in automatic suggestions.

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
```diff
@@ -0,0 +1,346 @@
...
+  trait WithOptionalTypeAnnotation {
+    val typeAnnotation: Option[TntType]
+  }
+
+  trait WithTypeAnnotation {
+    val typeAnnotation: TntType
+  }
+
+  /**
+   * A user-defined operator that is defined via one of the qualifiers: val, def, pred, action, or temporal. Note that
```
konnov, line 151:
```suggestion
   * A user-defined operator that is defined via one of the qualifiers: val, def, action, or temporal. Note that
```

It was probably copied from an outdated comment.

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
```diff
@@ -0,0 +1,346 @@
...
+      StrT.rw,
+      ConstT.rw,
+      VarT.rw,
+      SetT.rw,
+      SeqT.rw,
+      FunT.rw,
+      OperT.rw,
+      TupleT.rw,
+      RecordT.rw,
+      UnionT.rw,
```
konnov, line 246: Same here. How much collision are we going to cause with other names e.g. `BoolT1`? For the time being, I would prefix all of them with `Tnt`.

konnov (approved):
Looks good to me!

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
bugarela: Also, while we don't use this qualifier information (which can be forever afaiu), I think we should keep that as a string and even get rid of the value options in the comments since that will be out of sync pretty easily.

bugarela (approved):
LGTM

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
shonfeder: Given the possibility of namespacing (`TntEx.Bool`) I dislike the redundancy of needing the Hungarian notation on all constructors. But I didn't realize this posed a meaningful problem form auto-suggest.

I'll add the prefixes.

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
shonfeder: The intended way to use this would be to avoid importing the enclosing object. But as with the value constructors, I'll use Hungarian notation.

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
```diff
@@ -0,0 +1,346 @@
...
+  trait WithTypeAnnotation {
+    val typeAnnotation: TntType
+  }
+
+  /**
+   * A user-defined operator that is defined via one of the qualifiers: val, def, pred, action, or temporal. Note that
+   * TntOpDef does not have any formal parameters. If an operator definition has formal parameters, then `expr` should
+   * be a lambda expression over those parameters.
+   */
+  @key("def") case class Def(
```
shonfeder, line 155: @konnov how do you want this prefixed? I assume `TntDef.TntDef` isn't very desirable :)

konnov: Would it make sense to rename the enclosing `TntDef` into `TntUnit` (as a compilation unit) and then we could rename `Def` into `TntDef`?

shonfeder: just checked in the ts source and it is `TntOpDef`. I'll use that.

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/TntIR.scala
shonfeder: Fixed in aa5a4ea946a75749927d6c6fc88975e530f68f53

PR #2311, opened 2022-12-03
Add type mapping

shonfeder:
-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/tnt/Tnt.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/tnt/Tnt.scala
@@ -111,7 +111,7 @@ object Tnt {
     // @typeAlias: u = Foo({a: Int, b: Str}) | Bar(Int);
     // ```
     //
-    // As a result, our conversion fron TNT has to take a list of records of TNT
+    // As a result, our conversion from TNT has to take a list of records of TNT
     // rows convert them into a single TlaType1 record, for which all the values
     // are themselves records, and the keys are given by the values of the `taga
     // field from TNT rows.
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/tnt/Tnt.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/tnt/Tnt.scala
@@ -26,7 +26,7 @@ object Tnt {
     // This little bundle of state lets us track the conversion of TNT var names to
     // TlaType1 var numbers.
     //
-    // Since the scope of type variables in TNT is always limited to single type-level type expressions,
+    // Since the scope of type variables in TNT is always limited to single top-level type expressions,
     // and since the converter class is constructed fresh for each (top-level) type conversion,
     // we don't need to worry about variable name collisions.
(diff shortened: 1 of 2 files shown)


tla-io/src/main/scala/at/forsyte/apalache/io/tnt/Tnt.scala
```diff
@@ -0,0 +1,149 @@
...
+      // Since we have to concat lists here, which can be expensive, we use an
+      // accumulator to benefit from tail call optimization. This is purely a precaution.
+      def aux(r: Row, acc0: List[TlaType1]): List[TlaType1] = r match {
+        // TODO Update with support for row-based tuples: https://github.com/informalsystems/apalache/issues/1591
+        // Row variables in tuples are not currently supported. But we will proceed assuming that this is an
+        // over generalization, and that we can safely treat the tuple as a closed tuple type.
+        case Row.Var(_) =>
+          logger.debug(
+              s"Found open row variable in TNT tuple with fields $row. Polymorphic tuples are not supported, but we will proceed assuming the tuple can be treated as a closed type.")
+          acc0
```
shonfeder, line 53: @konnov and @gabrielamafra I thought of this approach after failing to figure out how sparse tuples could help :sweat_smile: 

I'm not sure if it will actually go thru :sweat_smile: I figure it's worth a shot, unless someone knows prima facie why this won't work. If so, I can convert this into an unsupported form, as we discussed, or pursue a different approach if someone can suggest one.

shonfeder:
The failing integration tests here are an (upstream?) CI-specific thing to do with our nix environment. I'm debugging that now, but it's unrelated to this PR.

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/Tnt.scala
```diff
@@ -0,0 +1,149 @@
...
+  //
+  // - Encapsulate and store benign state used to track variable numbers (see below)
+  // - Support and encapsulate the mutual recursion needed in the methods
+  private[tnt] class typeToTlaType() extends LazyLogging {
+
+    // TNT type variables are represented with strings, but in TlaType1 they are integers.
+    // This little bundle of state lets us track the conversion of TNT var names to
+    // TlaType1 var numbers.
+    //
+    // Since the scope of type variables in TNT is always limited to single type-level type expressions,
```
bugarela, line 29: Did you mean:
```suggestion
    // Since the scope of type variables in TNT is always limited to single top-level type expressions,
```

bugarela (approved):
LGTM! I left only some typo fixes :smile:

tla-io/src/main/scala/at/forsyte/apalache/io/tnt/Tnt.scala
shonfeder: Yep, it definitely makes sense to me that there are some TNT specs we won't be able to typecheck with this approach unless the operators are annotated.

shonfeder (commented):
Thanks very much for the fixes and the review! Sorry for the many typos :(

PR #2319, opened 2022-12-08
Fix failing nix setup in integration tests

### LINKED ISSUE #2318: CI on integration tests for ubuntu and docker are failing during nix setup
E.g., https://github.com/informalsystems/apalache/actions/runs/3649476797/jobs/6164204564

```
 error: could not set permissions on '/nix/var/nix/profiles/per-user' to 755: Operation not permitted
```

shonfeder:
Our nix installation in the CI was failing, presumably due to updates in the underlying runners that conflicted with the quite outdated version of the github action we were using to install nix. This PR just updates the nix action, which also allows us drop some of the configuration, since flakes is now supported without requiring a special URL.

Closes #2318 

:warning: Auto merge enabled

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -152,12 +152,9 @@ jobs:
             ${{ runner.os }}-nix-
             ${{ runner.os }}-
       - name: Install Nix
-        uses: cachix/install-nix-action@v14.1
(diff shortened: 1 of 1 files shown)


konnov (approved):
Looks good!

PR #2326, opened 2022-12-15
Fix shadowing on `unchecked`

### LINKED ISSUE #2325: `ScopedBuilder` reports shadowing when the variable in a quantified expression is `unchecked`
We should fix `boundVarIntroduction-` methods to check if the set of introduced variables is actually empty (which happens when `unchecked` is used), before checking subsumption, because the empty set is trivially a subset of every set.

Kukovec:

closes #2325


thpani (approved):
LGTM

Issue #1907, opened 2022-07-04
Revisit ADR007


konnov:
I am hitting problems with the dependencies across packages again. Looking at [ADR007](https://apalache.informal.systems/docs/adr/007adr-restructuring.html), it is a solution. But I would prefer understanding the problem first. We should find a bit of time to document the current situation with the packages and, based on that, develop a solution. It may happen to be exactly as in ADR007.

PR #2330, opened 2022-12-19
Fix array rule

### LINKED ISSUE #2329: `FunCtorRuleWithArrays` incorrectly specifies undefined values in arrays
In the line
https://github.com/informalsystems/apalache/blob/86dbc4e2ae371e73013c0beb3a5ac825840f1a09/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/FunCtorRuleWithArrays.scala#L59
the assertion is made, that the range value is not an index, for which the array is defined. This is type-incorrect and unsound, the correct alternative would be to use `domElem` in place of `rangeElem`.

Kukovec:

closes #2329 
The bug shows up in tests when the `ScopedBuilder` is used. Since those tests will come as part of a separate refactoring PR, they are omitted here.


thpani (approved):
LGTM

> Since those tests will come as part of a separate refactoring PR, they are omitted here.

Just in case, could you link an issue# for that refactoring for future traceability?

shonfeder:
Would it be possible to pull those tests into this PR without too much trouble? It would help break up that looming PR, and I guess strictly keep things that belong together togehter :D

Kukovec:
> Just in case, could you link an issue# for that refactoring for future traceability?

#1907 generally

> Would it be possible to pull those tests into this PR without too much trouble? It would help break up that looming PR, and I guess strictly keep things that belong together togehter :D

That would be a bit difficult. For the bug to show up in tests, we'd need to use the typed builder in `FunCtorRuleWithArrays`, which would automatically trigger dozens of tests to report type errors. I'd then have to go and extract all of _those_ tests, but rollback any changes to arenas.

shonfeder:
I see. Sounds too costly. I'm fine with merging this as is!

shonfeder (approved):

PR #2340, opened 2023-01-12
Unsound FunApp optimization disabled

### LINKED ISSUE #2338: `FunAppRule` optimizaiton is unsound for the OOPSLA ecoding
The following optimization is unsound:
https://github.com/informalsystems/apalache/blob/1a185d7ad38944685cb553a2ef6befe2ae925c6d/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/FunAppRule.scala#L164-L169

The reason for this is that it does not consider actual membership when inspecting the relation. Function domains may contain duplicate elements, e.g. `{1,1,2}`. In this case `SetOps.dedup` produces a canonical encoding `{c_1, c_2, c_3}`, where `c_1 = c_2 = 1`, but the `in` relation holds for exactly `c_1`. If such a function is applied to a value that evaluates to exactly `c_2`, the optimization above returns whatever codomain value is paired with `c_2`. Instead, it should always return the value paired with `c_1` (the canonical cell representation of 1). Because the value paired with `c_2` is unconstrained, it is possible for the solver to determine both `f[c_1] = f[c_2]` and `f[c_1] /= f[c_2]` despite the fact that `c_1 = c_2 = 1`.

The reason this is currently not an issue on `main` is that the caches aren't being hit consistently (expressions which should have been rewritten via the cache to produce `c_2` produce some other cell that is asserted to have value 1, but is not exactly `c_2`).

Kukovec:

closes #2338

### DIFF
--- a/test/tla/cli-integration-tests.md
+++ b/test/tla/cli-integration-tests.md
@@ -2524,8 +2524,13 @@ $ apalache-mc check --length=0 --inv=AllTests TestFiniteSetsExt.tla | sed 's/[IE
 EXITCODE: OK
 ```
 
+
 ### check TestFunctions.tla reports no error
 
+Disabled because fixing #2338 causes this to stack-overflow
+Revisit once pointers are implemented.
+
+<!-- $MDX skip -->
 ```sh
 $ apalache-mc check --length=0 --inv=AllTests TestFunctions.tla | sed 's/[IEW]@.*//'
 ...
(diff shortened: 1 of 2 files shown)


konnov (approved):
Yes, it looks like this optimization needs more thought

Kukovec:
> Yes, it looks like this optimization needs more thought

I imagine it's sound exactly for `FixedElemPtr(_, true)`

PR #2386, opened 2023-01-27
Add simple unit tests for quint exp conversion

shonfeder:
followup to #2351

adds simple unit tests, following @gabrielamafra's suggestion to illustrate and validate basic logic


konnov (approved):
Looks great!

PR #2396, opened 2023-02-02
SetMapRule

### LINKED ISSUE #2382: Rework set-generating rules to leverage pointers: SetMap
Mapping preserves the pointer-kind of the original set.

Kukovec:

closes  #2382

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/arena/PtrUtil.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/arena/PtrUtil.scala
@@ -25,4 +25,18 @@ object PtrUtil {
     }
   }
 
+  // When looking at cartesian product sets (e.g. for Map), the following holds true:
+  // If c_S represents S, c_a represents a, c_T represents T, c_b represents b, c_ST represents S x T
+  // and c_tup represents <<a,b>>, then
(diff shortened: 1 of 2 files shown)


thpani:
I have zero context on this, a PR description would help.
What are the concrete changes I'm looking at here?

Kukovec:
> I have zero context on this, a PR description would help. What are the concrete changes I'm looking at here?

Essentially this line:
```scala
    nextState = nextState.updateArena(_.appendHas(targetSetCell, tuplePtrCtor(mapResultCell)))
```

Here's the TLDR:
In a previous PR, `ElemPtr`s were introduced in all the rewriting rules, but they were indiscriminately `SmtConstElemPtr`, regardless of what they should be in the finished state.

So what kind of pointers does `SetMap` use for its elements? (specifically, when does it use `FixedElemPtr`?)
If `a \in S` via a fixed pointer, it clearly must be the case that `f(a) \in { f(x) ; x \in S }` via fixed pointer as well. Even if `f` is horribly non-injective, `FixedElemPtr`s are about guarantees.

This becomes a bit more complicated with map, because the general form is `{f(x1,..., xn) : x1 \in S1, ..., xn \in Sn}`, or rather 
 `{f(<<x1,..., xn>>) : <<x1,...,n>> \in S1\X  ... \X Sn}`. So we have to determine when `<<x1, ..., xn>> \in S1 \X ... \X Sn` via fixed pointer. It turns out that is exactly when all `xi \in Si` via fixed pointers.

thpani (approved):
LGTM

PR #2425, opened 2023-02-18
Add support for parsing qnt.json input

### LINKED ISSUE #2421: Add CLI support for receiving input as JSON encoded quint IR
Required for https://github.com/informalsystems/quint/issues/635

shonfeder:
Closes #2421

Adds CLI support for reading JSON encoded quint IR, produced by `quint typecheck
--out spec.qnt.json spec.qnt`.


konnov (approved):
That's cool!

tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
```diff
@@ -66,7 +66,7 @@ class TestQuintEx extends AnyFunSuite {
   // We construct a converter supplied with the needed type map
   val quint = new Quint(QuintOutput(
           "typechecking",
-          QuintModule(0, "MockedModule", List()),
+          List(QuintModule(0, "MockedModule", List())),
```
thpani, line 69: Nit-picking: do we really need concrete linear sequence types here? (over, say, a trait like `Seq`)

thpani (approved):

tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
shonfeder: I’m not clear on what the practical benefit of using a the less determinate `Seq` would be in this case. I also haven’t been able to find any clear guidelines that seem to be unanimously or even widely accepted on when seq should be preferred over list in case classes.


It might also be worth noting that the only real function of these case classes is to produce the deserializer that reads quint data out of JSON, and there will only ever be a single module in a well formed output.

These classes aren’t really part of a public API, and we are only really using the lists for iteration.  WDYT? Any compelling reasons for use of Seq come to mind?

thpani: If you don't need constant-time `.head`, `.tail`, or `.prepend`, it seems easier to use `Seq`? That way, calling code doesn't have to construct a list (which could be expensive) if you're passing in any other kind of collection.

> These classes aren’t really part of a public API, and we are only really using the lists for iteration. WDYT? Any compelling reasons for use of Seq come to mind?

I just don't see a compelling reason to require `List` :) Am I missing something?

shonfeder: Well, as I’ve noted, this code is only used internally to parse out JSON, and no discussion or documentation I’ve found has given a compelling reason to avoid lists in this kind of case. 

I can check ujson to see if it will deserialize into a Seq just as well, but List is a very common and ubiquitous data type, and without a particular reason to motivate fussing with the case classes here, I don’t feel very motivated to spend cycles on it :)

shonfeder: I’d you can find anything that makes a strong case about why Seq should be preferred in internal case classes like this, I’ll happily read it and revise as needed, but I haven’t found anything.

Calling code doesn’t have to construct a List needlessly when it has a different collection, because this is only used internet ally to produce Aplache IR.

thpani: Many things are only used internally until they aren't :)

I guess the common argument for requiring interfaces/traits over concrete types is abstraction? Not sure if you're making an argument about case classes specifically, though.

Either way, I'm okay merging this, just don't think it's the most general / future-proof.

shonfeder: I’ll keep in mind generalizing to Seq next time I touch these data types.

PR #2427, opened 2023-02-20
Adds pointer caching

Kukovec:

In addition to adding automatic caching to `ElemPtr`s in `PtrUtil`, this PR also changes the way `Z3SolverContext` handles `in`-relation cells. Previously, attempting to assert and expression containing an `in_X_Y` pointer name would throw, unless `declareInPredIfNeeded` was called at some point in the past. This seems completely pointless and unnecessary, so `getInPred` now just calls `declareInPredIfNeeded` if it doesn't already have the expression cached, instead of throwing.

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/arena/PtrUtil.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/arena/PtrUtil.scala
@@ -18,6 +18,9 @@ object PtrUtil {
 
   object Heuristics {
 
+    // for debugging; using this is equivalent to disabling caching
+    def alwaysFalse: CachingHeuristic = _ => false
+
     def notAName: CachingHeuristic = {
       case _: NameEx => false
       case _         => true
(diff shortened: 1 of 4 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/arena/PtrUtil.scala
```diff
@@ -11,6 +14,38 @@ import at.forsyte.apalache.tla.types.tla
...
+    def notAName: CachingHeuristic = {
+      case _: NameEx => false
+      case _         => true
+    }
+
+    def notConjunctionOfNames: CachingHeuristic = {
+      case _: NameEx                          => false
+      case OperEx(TlaBoolOper.and, args @ _*) => args.exists(notConjunctionOfNames)
+      case _                                  => true
+    }
```
konnov, line 33: This does not seem to be a complex expression to me. I was expecting something like a threshold over the expression size.

konnov (approved):
Looks good. I am just not sure about the particular choice of heuristics.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/arena/PtrUtil.scala
Kukovec: I mean, these were really just prototypes meant to be slightly more complex than no-ops, not the final versions. Obviously the actual heuristic will ultimately need some more planning.

Issue #2438, opened 2023-02-27
Convert quint set operators


shonfeder:
`fold` and `flatten` require consideration of rewiring: https://github.com/informalsystems/apalache/issues/2437

PR #2449, opened 2023-03-02
Add conversion of quint Set, exists, and forall

shonfeder:
Contributes to #2438

This includes a non-trivial combinator for converting binary binding
operator from quint to binding operator in ApalacheIR.

-------

### DIFF
--- a/tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
+++ b/tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
@@ -26,7 +26,14 @@ class TestQuintEx extends AnyFunSuite {
       x
     }
 
+    // Operator application
+    def app(name: String, args: QuintEx*): QuintApp = QuintApp(uid, name, args)
+
     val tt = QuintBool(uid, true)
+    val _0 = QuintInt(uid, 0)
+    val _1 = QuintInt(uid, 1)
+    val _2 = QuintInt(uid, 2)
+    val _3 = QuintInt(uid, 3)
     val _42 = QuintInt(uid, 42)
     val s = QuintStr(uid, "s")
     val name = QuintName(uid, "n")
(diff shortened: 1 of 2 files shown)


tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
```diff
@@ -37,23 +44,33 @@ class TestQuintEx extends AnyFunSuite {
...
     val appBar = QuintApp(uid, "bar", List(_42))
     val letBarBeLambdaInAppBar = QuintLet(uid, barDef, appBar)
-
-    // Builtin operator application
-    def app(name: String, args: QuintEx*): QuintApp = QuintApp(uid, name, args)
+    val nIsGreaterThanZero = app("igt", name, _0)
+    // A predicate on ints
+    val intIsGreaterThanZero = QuintLambda(uid, List("n"), "def", nIsGreaterThanZero)
+    val intSet = app("Set", _1, _2, _3)
+    val emptyIntSet = app("Set", _1, _2, _3)
```
thpani, line 51: I couldn't find where this is used.
Would it make sense to include a test for the empty `Set` application?

tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
```diff
@@ -172,4 +189,16 @@ class TestQuintEx extends AnyFunSuite {
   test("can convert builtin iuminus operator application") {
     assert(convert(Q.app("iuminus", Q._42)) == "-42")
   }
+
+  test("can convert builtin Set operator application") {
+    assert(convert(Q.app("Set", Q.s, Q.s, Q.s)) == """{"s", "s", "s"}""")
+  }
+
+  test("can convert builtin exists operator application") {
```
thpani, line 197: Would it make sense to include (negative) tests for the 2 exceptional cases in `binaryBindingApp`?

thpani (approved):
Hooray! 🎉 

I think we lack a test for empty `Set` application, otherwise LGTM

tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
shonfeder: Yes! Absolutely. That war an oversight. Thanks!

tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
shonfeder: I couldn't decide on those. The logic gets a bit more complicated, but there is still only really one way that constructor can go right, and looking at the code it is pretty evident to me that every other way will go wrong and result in the error... 

But maybe it worth adding a test for that.

shonfeder:
Thanks for the review!

konnov (approved):
Looks great to me too!

Issue #2450, opened 2023-03-02
Convert quint's `nondet x = oneOf(...); A` to a nondet existential quantification.


shonfeder:
This will require special treatment in 

https://github.com/informalsystems/apalache/blob/ecdc52ff5cd383ffef03f56a525101608610d679/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala#L90-L99

We'll be mapping

```
nondet x = oneOf(S); A
```

to

```
\E x \in S: A
```

Thanks to @konnov and @gabrielamafra for the pointers here.

konnov:
I realized a few days ago that `nondet x = oneOf(S); A` should be translated to:

```
S /= {} /\ \E x \in S: A
```

Since `oneOf` is a function, it requires non-emptiness of `S`, whereas `\E x \in S: A` simply returns `FALSE` when `S` is empty.

thpani:
> Since `oneOf` is a function, it requires non-emptiness of `S`, whereas `\E x \in S: A` simply returns `FALSE` when `S` is empty.

I don't get it, what does the `S /= {}` do, on top of making the expression evaluate to `FALSE`?

konnov:
> > Since `oneOf` is a function, it requires non-emptiness of `S`, whereas `\E x \in S: A` simply returns `FALSE` when `S` is empty.
> 
> I don't get it, what does the `S /= {}` do, on top of making the expression evaluate to `FALSE`?

Oh, I think you are right. It's actually the other way around: `\E x \in S: A` of TLA+ should be understood as `any { S == Set(), nondet x = oneOf(S); A }`.

thpani:
Still not sure I understand. In the case where `S = {}`, wouldn't

```
any { S == Set(), all { S != Set(), nondet x = oneOf(S); A } }
```

be true, and `\E x \in S: A` be false?

konnov:
> Still not sure I understand. In the case where `S = {}`, wouldn't
> 
> ```
> any { S == Set(), all { S != Set(), nondet x = oneOf(S); A } }
> ```
> 
> be true, and `\E x \in S: A` be false?

Sorry. I am a bit tired today :)

Yes, you are right. We need only the `all` part. Fixed in the comment above.

PR #2451, opened 2023-03-02
Add conversion of remaining quint set operators

### LINKED ISSUE #2438: Convert quint set operators

shonfeder:
Closes #2438

All remaining set operator are added here, with the exception of `oneOf`, which
will be addressed in #2450.

Thanks to @konnov and @bugarela for the assist on reasoning thru the less
straight forward operators, `oneOf` and `chooseSome`.

-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
@@ -44,6 +44,14 @@ class Quint(moduleData: QuintOutput) {
     s"__QUINT_LAMBDA${n}"
   }
 
+  // benign state to generate unique variable names
+  private var uniqueVarNo = 0
+  private def uniqueVarName(): String = {
+    val n = uniqueVarNo
+    uniqueVarNo += 1
+    s"__quint_var${n}"
+  }
+
   // Convert a QuintEx into a TlaEx
   //
   // We implement a small family of mutually recursive conversion functions using this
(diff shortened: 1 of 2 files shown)


tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
```diff
@@ -220,4 +231,73 @@ class TestQuintEx extends AnyFunSuite {
     assert(exn.getMessage.contains(
             "Input was not a valid representation of the QuintIR: too many arguments passed to binary operator forall"))
   }
+
```
bugarela, line 234: I'm missing tests to `forall` and `exists`, specially one that tests them with multiple lambda arguments (I'm not sure if this would support that as is).

bugarela (commented):
It's not clear for me whether the binding we're doing is enough to support lambdas with multiple arguments. In Quint, they are unpacked, while TLA+ supports multiple quantified variables in `forall` and `exists`.

We can postpone this, of course, I'd just like to point out this scenario in case we don't have it mapped out already (either in the code or in an issue).

tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
```diff
@@ -44,6 +44,14 @@ class Quint(moduleData: QuintOutput) {
     s"__QUINT_LAMBDA${n}"
   }
 
+  // benign state to generate unique variable names
```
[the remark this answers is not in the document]
shonfeder: I did mean benign, in the sense of https://www.sciencedirect.com/science/article/abs/pii/0020019088902293

the idea being that from outside of the conversion function we’re building up with this object, the state can’t interact with or track or interfere with anything.

tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
shonfeder: Exists and forall were introduced in the previous PR, and you can see the teats for those on line 200/211 above.

I have totally overlooked the multi argument bindings of forall and exists, and I think of lambdas more generally! Thanks for pointing that out. I will add support for those in a follow up!

bugarela (approved):

shonfeder:
Thanks for the careful review and catching my oversight!

PR #2460, opened 2023-03-06
Support tuppled variables in quantification bindings in builder

### LINKED ISSUE #2292: Extend builder with tupled bound variable methods
The `ScopedBuilder` offers methods with which one can produce IR in normalized form. As such, it cannot be used to construct, e.g.
`\E <<x,y>> \in S: P(x,y)`, since `\E` expects a single variable name.
Instead of modifying `ScopedBuilder`, we should add tupled alternatives to any method, which may introduce bound variables (`\E, \A`, map, filter) in an extension.

Kukovec:

closes #2292


tlair/src/test/scala/at/forsyte/apalache/tla/typecomp/TestTupledBuilder.scala
```diff
@@ -0,0 +1,123 @@
...
+          method,
+          mkWellTyped,
+          mkIllTyped,
+          resultIsExpected(oper),
+      ) _
+
+    checkRun(Generators.typeAndSeqGen)(run(TlaBoolOper.forall, builder.forallAllowTuples))
+    assertThrowsBoundVarIntroductionTernary(builder.forallAllowTuples)
+
+    checkRun(Generators.typeAndSeqGen)(run(TlaBoolOper.exists, builder.existsAllowTuples))
```
shonfeder, line 72: I guess the docs for `forall` and `exists` will need to be updated to indicate they can take arguments that are tupples?gg)

shonfeder (approved):
The code generally looks great to me.

As we discussed, the public API of the builder library would arguably be cleaner if the most general constructor (that supports both tuppled and bare names) where called `forall` and `exists` respectively, and if the more restrictive name were used for the more specific behavior. But this needn't be a blocking point.

shonfeder (approved):
Looks fantastic. Thanks so much!

PR #2474, opened 2023-03-09
Add conversion of quint list operator

shonfeder:
Closes #2440

Implementations for `select`, `range`, and `foldr` are not yet included here,
since they will need more involved conversion rules, as per https://github.com/informalsystems/apalache/issues/2437

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] Documentation added for any new functionality
- [x] [Entries added to `./unreleased/`][changelog format] for any new functionality

[changelog format]: https://github.com/informalsystems/apalache/blob/main/CONTRIBUTING.md#how-to-record-a-change

PR #2495, opened 2023-03-21
Fix index increment logic in Quint index conversion

shonfeder:
We were only handling incrementing integer literals, which is
inadequate, since it doesn't account for expressions such as variables
of compound expressions.

Followup to #2474 

-------


thpani (approved):
Good catch, LGTM!

PR #2480, opened 2023-03-13
Convert quint Map operators (TLA functions)

shonfeder:
Closes #2439 
Closes #2437

Thanks to @konnov for the guidance and @thpani for managing the hairiest part around `put`.

This also includes a fix for misnamed quint addition operator `isum` -> `iadd`.
 
-------

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] Documentation added for any new functionality
- [x] [Entries added to `./unreleased/`][changelog format] for any new functionality

[changelog format]: https://github.com/informalsystems/apalache/blob/main/CONTRIBUTING.md#how-to-record-a-change

PR #2496, opened 2023-03-22
Translate Quint `put`

### LINKED ISSUE #2493: Translate Quint `put`
via `<:` and `@@` function merge: https://github.com/informalsystems/apalache/blob/main/src/tla/__rewire_tlc_in_apalache.tla#L27-L41 :

thpani:
Translate Quint `put`

Merges into the Maps PR (#2480). This is missing a test, that we'll have to add once we have the `Map()` constructor.

- [ ] ~[Entries added to `./unreleased/`][changelog format] for any new functionality~

Closes #2493

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
@@ -341,7 +341,52 @@ class Quint(moduleData: QuintOutput) {
         case "setOfMaps" => binaryApp(opName, tla.funSet)
         case "set"       => ternaryApp(opName, tla.except)
         case "setBy"     => null
-        case "put"       => null
+        case "put"       =>
+          // put(map, key, value) ~~> (key :> value) @@ map
+          quintArgs =>
+            ternaryApp(opName,
+                (map, key, value) => {
+                  // (1) build `key :> value` == `[ _ \in {key} |-> value ]`
+                  // extract types
+                  val mapType = Quint.typeToTlaType(types(quintArgs(0).id).typ)
+                  val keyType = Quint.typeToTlaType(types(quintArgs(1).id).typ)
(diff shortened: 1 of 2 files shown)


tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
```diff
@@ -341,7 +341,52 @@ class Quint(moduleData: QuintOutput) {
...
+          quintArgs =>
+            ternaryApp(opName,
+                (map, key, value) => {
+                  // (1) build `key :> value` == `[ _ \in {key} |-> value ]`
+                  // extract types
+                  val mapType = Quint.typeToTlaType(types(quintArgs(0).id).typ)
+                  val keyType = Quint.typeToTlaType(types(quintArgs(1).id).typ)
+                  val valType = Quint.typeToTlaType(types(quintArgs(2).id).typ)
+                  // f1 == key :> value
+                  val f1 = tla.funDef(value, (tla.name(uniqueVarName(), keyType), tla.enumSet(key)))
```
konnov, line 355: Do we really want to go via this route? Functions are kind of expensive in the OOPSLA encoding, they are sets of pairs (though Jure's improvements may compensate for that).

If we directly specialize the definition of `@@` for the case of adding one pair, this should be more efficient and maybe even simpler.

shonfeder: this looks to use the exact same approach as in our rewiring rule. why would we differ from that? or should that rule be changed?

thpani: Ah, good point! See 435cb627e

shonfeder: Oh i see! Cool :)

konnov: Nice! :)

PR #2499, opened 2023-03-24
Add conversion of quint nondet binders

### LINKED ISSUE #2450: Convert quint's `nondet x = oneOf(...); A` to a nondet existential quantification.

shonfeder:
Closes #2450

Also includes reorg of the long conversion functions into named functions so we
can keep the builtin conversion case analyses easier to scan. This is isolated
in b22cd39.

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
@@ -470,11 +470,28 @@ class Quint(moduleData: QuintOutput) {
       applicationBuilder(quintArgs)
     }
 
+    private val nondetBinding: (QuintDef.QuintOpDef, QuintEx) => TBuilderInstruction = {
+      // val nondet name = oneOf(domain); scope
+      // ~~>
+      // LET wrap = \E name \in domain: scope IN wrap
+      case (QuintDef.QuintOpDef(_, name, "nondet", QuintApp(id, "oneOf", Seq(domain)), _), scope) =>
+        val wrapName = uniqueVarName()
+        val wrap = tla.name(wrapName, OperT1(Seq(), BoolT1))
+        val elemType = Quint.typeToTlaType(types(id).typ)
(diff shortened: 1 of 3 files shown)


tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
```diff
@@ -458,11 +470,33 @@ class Quint(moduleData: QuintOutput) {
...
     }
 
+    // Convert Quint's nondet variable binding
+    //
+    //   val nondet name = oneOf(domain); scope
+    //   ~~>
+    //   LET wrap = \E name \in domain: scope IN wrap
+    //
+    // We must wrap the binding, because this is at the declaration level of the quint
+    // IR, but must be at the expression level in the Apalache IR.
```
konnov, line 480: @Kukovec probably remembers it better, but I think that wrapping an existential in `LET` may prevent the preprocessor from applying Skolemization. I do not see why we could not simply produce an expression. `nondet` declarations cannot appear as top-level declarations in Quint, they only appear inside other operator declarations. Nested `LET-IN` expressions are just expressions.

konnov (approved):
Looks great! Can we get rid of `LET wrap = ...`? I do not fully understand the problem it is supposed to solve.

tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
shonfeder: Yep! Thanks for catching that. I'm not sure why I'd thought it needed wrapping. Fixed.

thpani (approved):

PR #2500, opened 2023-03-24
Fix conversion of quint foldr

shonfeder:
We were constructing an unneeded lambda wrapper, that was also
triggering a type err on the secondary type-checking pass.

```
    <unknown>: internal error in type checking: Bug in ToEtcExpr. Expected an operator name, found: LET __QUINT_LAMBDA1(x, y) ≜ x + y IN __QUINT_LAMBDA1 E@15:46:23.130
    at.forsyte.apalache.infra.AdaptedException: <unknown>: internal error in type checking: Bug in ToEtcExpr. Expected an operator name, found: LET __QUINT_LAMBDA1(x, y) ≜ x + y IN __QUINT_LAMBDA1
        at at.forsyte.apalache.infra.passes.PassChainExecutor$.run(PassChainExecutor.scala:39)
        at at.forsyte.apalache.tla.tooling.opt.CheckCmd.run(CheckCmd.scala:132)
        at at.forsyte.apalache.tla.Tool$.runCommand(Tool.scala:138)
        at at.forsyte.apalache.tla.Tool$.run(Tool.scala:118)
        at at.forsyte.apalache.tla.Tool$.main(Tool.scala:40)
        at at.forsyte.apalache.tla.Tool.main(Tool.scala)
```

This didn't show up until writing integration tests for quint.

Followup to #2488

-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
@@ -278,43 +278,6 @@ class Quint(moduleData: QuintOutput) {
                 tla.foldSeq(testLambda, tla.emptySeq(elemType), seq)
               })(quintArgs)
 
-      def foldr(opName: String): Converter =
-        quintArgs =>
-          ternaryApp(opName,
-              (seq, init, op) => {
-                val seqType = Quint.typeToTlaType(types(quintArgs(0).id).typ)
-                val elemType = seqType match {
-                  case SeqT1(elem) => elem
-                  case invalidType => throw new QuintIRParseError(s"sequence ${seq} has invalid type ${invalidType}")
-                }
-                val accType = Quint.typeToTlaType(types(quintArgs(1).id).typ)
-                // Reverse(__s) ==
-                //  LET __s_len == Len(__s) IN
-                //  LET __get_ith(__i) == __s[__s_len - __i + 1] IN
-                //  SubSeq(__ApalacheMkSeq(__ApalacheSeqCapacity(__s), __get_ith), 1, __s_len)
-                val sParam = tla.param(uniqueVarName(), seqType)
-                val s = tla.name(sParam._1.name, sParam._2)
-                val iParam = tla.param(uniqueVarName(), IntT1)
-                val i = tla.name(iParam._1.name, iParam._2)
-                val s_lenDecl = tla.decl(uniqueLambdaName(), tla.len(seq))
-                val s_len = tla.name(s_lenDecl.name, OperT1(Seq(), IntT1))
-                val get_ith = tla.lambda(uniqueLambdaName(),
-                    tla.app(s, tla.plus(tla.minus(tla.appOp(s_len), i), tla.int(1))), iParam)
-                val reverseDecl = tla.decl(uniqueLambdaName(),
-                    tla.letIn(tla.subseq(tla.mkSeqConst(tla.apalacheSeqCapacity(s), get_ith), tla.int(1),
-                            tla.appOp(s_len)), s_lenDecl), sParam)
-                // FoldRight(__op(_, _), __seq, __base) ==
-                //  LET __map (__y, __x) == __op(__x, __y) IN
-                //  __ApalacheFoldSeq(__map, __base, Reverse(__seq))
-                val xParam = tla.param(uniqueVarName(), elemType)
-                val x = tla.name(xParam._1.name, xParam._2)
-                val yParam = tla.param(uniqueVarName(), accType)
-                val y = tla.name(yParam._1.name, yParam._2)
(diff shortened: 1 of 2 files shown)


shonfeder:
Ah, I missed that! It is working in the integration tests, but because i'm only testing with commutative ops. I better fix that :)

I'll follow up with a fix of this fix :D

shonfeder:
So we cannot define a flip operation by name, and then pass that to fold, because for some reason scoped builder requires the operator given to a fold to be a `LET/IN`

https://github.com/informalsystems/apalache/blob/e7fe0107a2720aa492e46252a00a5eba81b8a4cc/tlair/src/main/scala/at/forsyte/apalache/tla/typecomp/unsafe/UnsafeApalacheBuilder.scala#L131-L139

But when we construct the nested `LET/IN`s required to flip the args inline, we hit the error above due to this rule in the type checker:

https://github.com/informalsystems/apalache/blob/e7fe0107a2720aa492e46252a00a5eba81b8a4cc/tla-typechecker/src/main/scala/at/forsyte/apalache/tla/typecheck/etc/ToEtcExpr.scala#L216-L231

Why can we only apply a name and not an expression (which evaluates to an operator)? Not sure. It's not documented there.

I've not yet traced thru why exactly we hit this, but thought @Kukovec or @konnov might have some idea how we should proceed here.

For reference, here is the TLA we are producing that is triggering the type error:

```
    ApaFoldSeqLeft(LET (*
      @type: ((Set(Int), Int) => Set(Int));
    *)
    __QUINT_LAMBDA5(__quint_var3, __quint_var2) ==
      (LET (*
        @type: ((Int, Set(Int)) => Set(Int));
      *)
      __QUINT_LAMBDA1(x, acc) == acc \union {x}
      IN
      __QUINT_LAMBDA1)(__quint_var2, __quint_var3)
    IN
    __QUINT_LAMBDA5, {}, (__QUINT_LAMBDA4(l))))
      = n + 1
    /\ n = l[(n + 1)]
```

shonfeder:
Writing that out may have helped. If I assign the operator we need to flip to an intermediate name, perhaps we can pass the typecheck. Like:

```tla
    LET opToFlip = LET __QUINT_LAMBDA1(x, acc) == acc \union {x} IN __QUINT_LAMBDA1 IN
    ApaFoldSeqLeft(
    LET
      (* @type: ((Set(Int), Int) => Set(Int)); *)
      __QUINT_LAMBDA5(__quint_var3, __quint_var2) == opToFlip(__quint_var2, __quint_var3)
    IN
      __QUINT_LAMBDA5
    , {}
    , (__QUINT_LAMBDA4(l))
    ))
      = n + 1
    /\ n = l[(n + 1)]
```

That binding seems silly, but if it'll unblock for now :)

shonfeder:
Nah, that doesn't seem to be possible either since `opToFlip` gets typed as `() => (Int, Set(Int)) => Set(Int)`.

thpani:
@shonfeder This came up in synchronous discussion w/ @konnov and @Kukovec this morning.

> * Either we can get the type-checker to consume lambda'd applications `(LET X(a, b) == _ IN X)(y, z)`
> 
> * Or, do a contextual translation of the quint lambda to unmangle it into a LET-IN preceeding the `ApaFoldSeqLeft`

(1) is not legal TLA+.

It seems there's better options for translating (have `foldr` in a future Quint stdlib, support as an Apalache-builtin), our conclusion was to postpone `foldr` and hide it from the Quint docs for now: https://github.com/informalsystems/quint/issues/752

shonfeder:
Thanks for looping me in the sync discussion. This was gonna be my suggestion today as well, so I'm on board. The "fix" here will then become just removal :)

shonfeder:
A point on (2): I created a version of this rewrite to hoist the operation flip into a let above the fold, but it *still* requires application of a let/in:

```
    (LET (* @type: ((Int, Set(Int)) => Set(Int)); *)
    __quint_var4(__quint_var2, __quint_var3) ==
      (LET (*
        @type: ((Int, Set(Int)) => Set(Int));
      *)
      __QUINT_LAMBDA0(x, acc) == acc \union {x}
      IN
      __QUINT_LAMBDA0)(__quint_var2, __quint_var3)
    IN
```

So IIUC, so long as we have this constraint in our type system, and so long as we can only represent lambdas as let/ins, there is no way to write a flip operation on a "lambda".

However, what we *could* probably do is a rewrite, where we flip the params of the operator, rather than effecting that thru TLA's limited powers of functional abstraction. I'll propose that course at the standup.

thpani:
We could do a post-processing (proposed by @Kukovec this morning) of the form

```
(LET X(f1, ..., fn) == body IN X)(p1, ..., pn)  ~~>  LET X(f1, ..., fn) == body IN X(p1, ..., pn)
```

shonfeder:
Simpler fix as per our discussions: now we just remove the operator. Thanks for helping think thru the options and tradeoffs!

thpani (approved):
LGTM

Issue #2532, opened 2023-04-13
Support `LAMBDA` in the Apalache IR


shonfeder:
The Apalache IR has not yet picked up support for TLA `LAMBDA`. Instead we perform a [de-lambdification](https://en.wikipedia.org/wiki/Let_expression#Conversion_from_lambda_to_let_expressions) to convert any lambdas into let expressions. However, let expressions cannot be applied, as noted in #2517.

This has required workarounds in quint conversion:

- #2500 
- #2515 
- #2531 

Adding support for `LAMBDA` in the IR will have at least the following benefits:

- Reduce complexity of quint conversion.
- Improve readability of TLA we generate from parsed specs (e.g., as recorded with `--write-intermediate`)
- Reduce complexity of our pre-processing, since we won't need to de-lambdafy.

PR #2534, opened 2023-04-14
DeLambda transformation

Kukovec:

Since #2532 is likely to require significant planning and refactoring, here's a temporary workaround for unsupported lambda-generated let applications.

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/DeLambdifier.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/DeLambdifier.scala
@@ -1,0 +1,29 @@
+package at.forsyte.apalache.tla.pp
+
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.oper.TlaOper
+import at.forsyte.apalache.tla.lir.transformations._
+
+/**
+ * DeLambdifier turns applications of operator-typed LET-expressions, into LET-expressions with applications pushed into
+ * the body.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+class DeLambdifier(tracker: TransformationTracker) extends TlaExTransformation {
+  override def apply(input: TlaEx): TlaEx = {
+    transform(input)
+  }
+
+  private def transform: TlaExTransformation = tracker.trackEx {
+    case ex @ OperEx(TlaOper.apply, LetInEx(nameEx @ NameEx(operName), decl @ TlaOperDecl(declName, _, _)), args @ _*)
+        if operName == declName =>
+      LetInEx(
+          OperEx(TlaOper.apply, nameEx +: args.map(transform): _*)(ex.typeTag),
+          decl.copy(body = transform(decl.body)),
+      )(ex.typeTag)
+
+    case ex @ _ => ex
+  }
+}
--- a/passes/src/main/scala/at/forsyte/apalache/tla/passes/pp/PreproPassImpl.scala
+++ b/passes/src/main/scala/at/forsyte/apalache/tla/passes/pp/PreproPassImpl.scala
@@ -1,16 +1,16 @@
 package at.forsyte.apalache.tla.passes.pp
 
+import at.forsyte.apalache.infra.passes.DerivedPredicates
 import at.forsyte.apalache.infra.passes.Pass.PassResult
-import at.forsyte.apalache.tla.imp.src.SourceStore
+import at.forsyte.apalache.infra.passes.options.OptionGroup
 import at.forsyte.apalache.io.lir.TlaWriterFactory
+import at.forsyte.apalache.tla.imp.src.SourceStore
 import at.forsyte.apalache.tla.lir.storage.ChangeListener
 import at.forsyte.apalache.tla.lir.transformations.standard._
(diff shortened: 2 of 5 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/DeLambdifier.scala
```diff
@@ -0,0 +1,29 @@
...
+import at.forsyte.apalache.tla.lir.transformations._
+
+/**
+ * DeLambdifier turns applications of operator-typed LET-expressions, into LET-expressions with applications pushed into
+ * the body.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+class DeLambdifier(tracker: TransformationTracker) extends TlaExTransformation {
```
shonfeder, line 14: I think `LetInApplier` is probably a more accurate name? There seems to already be precedent for `de-lambda` as a transformation that converts (proper) lambdas into let/in expressions, which is not what we're ding here :) https://en.wikipedia.org/wiki/Let_expression#Conversion_from_lambda_to_let_expressions

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/DeLambdifier.scala
```diff
@@ -0,0 +1,29 @@
+package at.forsyte.apalache.tla.pp
+
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.oper.TlaOper
+import at.forsyte.apalache.tla.lir.transformations._
+
+/**
+ * DeLambdifier turns applications of operator-typed LET-expressions, into LET-expressions with applications pushed into
+ * the body.
+ *
```
shonfeder, line 10: Maybe note that this is a workaround motivated by lack of support for lambdas on the IR, with a link to the issue?

shonfeder (approved):
Great, thanks! 
Two suggestions, tho they are not essential.

And, just to confirm, this rewriting pass will run on code that is supplied by to the checker via loading from JSON in https://github.com/informalsystems/apalache/blob/5463ad2c6913ddbdbbbb802a344da1d62a0bd708/passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/SanyParserPassImpl.scala#L126

shonfeder:
Or do i rather need to ensure i call this tranformer on every TlaEx constructed in the conversion?

Kukovec:
> Or do i rather need to ensure i call this tranformer on every TlaEx constructed in the conversion?

We can adapt it a bit, and have it just be one of the transformations in `PreprocessingPass`, if that's easier, from a usability POV

shonfeder:
That would definitely be much easier!

shonfeder (approved):
Awesome. This will be a big help. Thanks very much!!!

shonfeder:
Unfortunately this workaround is not addressing the problem in quint conversion as hoped: 

```
    <unknown>: internal error in type checking: Bug in ToEtcExpr. Expected an operator name, found: LET __QUINT_LAMBDA0(x) ≜ x + 1 IN __QUINT_LAMBDA0 E@14:00:51.892
    at.forsyte.apalache.infra.AdaptedException: <unknown>: internal error in type checking: Bug in ToEtcExpr. Expected an operator name, found: LET __QUINT_LAMBDA0(x) ≜ x + 1 IN __QUINT_LAMBDA0
        at at.forsyte.apalache.infra.passes.PassChainExecutor$.run(PassChainExecutor.scala:39)
        at at.forsyte.apalache.tla.tooling.opt.CheckCmd.run(CheckCmd.scala:132)
        at at.forsyte.apalache.tla.Tool$.runCommand(Tool.scala:138)
        at at.forsyte.apalache.tla.Tool$.run(Tool.scala:118)
        at at.forsyte.apalache.tla.Tool$.main(Tool.scala:40)
        at at.forsyte.apalache.tla.Tool.main(Tool.scala)
```

I guess this is because we run typechecking which eats it here, before we do the `PreproPass`?

https://github.com/informalsystems/apalache/blob/aa87a154ea7e7ec59efb7df00fa8e628eddbf062/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/config/CheckerModule.scala#L91-L102

shonfeder:
I can just bypass this check in the typechecker, but then we abandon compat with TLC in our typing, and I agree with @konnov that this edge case doesn't justify it.

If we can't get this working with our current pipeline, probably best to revert.

Kukovec:
> I can just bypass this check in the typechecker, but then we abandon compat with TLC in our typing, and I agree with @konnov that this edge case doesn't justify it.

I don't think you need to "bypass" anything, we just have to add a case for the typechecker to understand that `(LET F == ... IN F)` has an operator type, and continue from there. For the typechecker, `(LET F == ... IN F)(x)` and `F(x)` + in-scope definition should be equivalent, it's just currently not accepting the former on a syntax basis.

shonfeder:
This would make the type checker accept TLA that is not valid with TLC. This is what I mean by “bypass”.

As per our discussion yesterday, it doesn’t seem think there’s any reason to deviate in this way, because we can get what we need from this change with just five lines in the scoped builder, and this is all just a workaround anyhow. :)

Issue #2552, opened 2023-05-02
Quint conversion of applied polymorphic operators is typed too narrowly


shonfeder:
I have been inferring the types of an applied quint operator from the type map recording the types of its arguments and the type of the application. This breaks the coherence checking of the typed builder, since we are ignoring operator polymorphism. We need to instead use the quint lookup table (exposed in https://github.com/informalsystems/quint/pull/852) to type the applied operator with their most generally inferred types at the time of declaration.

PR #2560, opened 2023-05-13
Support application of polymorphic operators in typed builder

### LINKED ISSUE #2559: Typed builder cannot apply construction of polymorphic operators
Blocking #2552 

A (currently failing) test case showing the problem is here: https://github.com/informalsystems/apalache/pull/2560/files#diff-da86be24ba1118825a1ec09e85913af14cd2b28316e1747934f6bd9579e537e3R56-R62

### LINKED ISSUE #2581: Typed builder is not able to construct applications of row-polymorphic operator
This is blocking completion of https://github.com/informalsystems/apalache/milestone/74. When we try to construct the application of a row-polypmorphic operator we get errors originating from the typed builder looking like

```
    error: internal error: Input was not a valid representation of the QuintIR: Conversion failed while building operator definition
    Operator OPER_APP cannot be applied to arguments of types ((({ f1: Int, a }) => { f1: Int, a }), { f1: Int, f2: Str, a })
    Please report an issue: https://github.com/informalsystems/quint/issues/new
```

Of course, this is a valid operator application.

Arguably this is a special case of #2559, but the working fix for that initially proposed in #2560 doesn't suffice for this.

shonfeder:
Closes #2559
Closes #2581

It appears that the typed builder is not currently able to handle constructing application of polymorphic operator.

This adds a regression test for this omission and a fix to unblock it.

The fix is to replace the limited and more ad-hoc type compatibility check used in the builder previously with the same type unifier we use for type checking. The result is

- More code reuse.
- More correct checks on type compatibility in the builder.
- Less code.

This will unblock #2552.

### DIFF
--- a/tlair/src/main/scala/at/forsyte/apalache/tla/typecomp/signatures/FlexibleEquality.scala
+++ b/tlair/src/main/scala/at/forsyte/apalache/tla/typecomp/signatures/FlexibleEquality.scala
@@ -1,53 +1,26 @@
 package at.forsyte.apalache.tla.typecomp.signatures
 
-import at.forsyte.apalache.tla.lir.{FunT1, RecRowT1, RecT1, RowT1, SeqT1, SetT1, TlaType1, TupT1}
+import at.forsyte.apalache.tla.lir.TlaType1
 
-import scala.collection.immutable.SortedMap
-import at.forsyte.apalache.tla.lir.VarT1
+import at.forsyte.apalache.tla.types.TypeUnifier
+import at.forsyte.apalache.tla.types.TypeVarPool
+import at.forsyte.apalache.tla.types.Substitution
 
 /**
  * Determines flexible equality ([[compatible]]), i.e. an equivalence relation over TT1, which is equality for
  * non-record types, but allows the comparison of records types with compatible fields
  *
  * @author
- *   Jure Kukovec
+ *   Jure Kukovec, ShonFeder
  */
 object FlexibleEquality {
 
-  // None if no common supertype, Some(e) if compatible, and common supertype is e.
-  def commonSupertype(lhs: TlaType1, rhs: TlaType1): Option[TlaType1] = (lhs, rhs) match {
-    case (a, b) if a == b     => Some(a)
-    case (VarT1(_), b)        => Some(b)
-    case (a, VarT1(_))        => Some(a)
-    case (SeqT1(l), SeqT1(r)) => commonSupertype(l, r).map(SeqT1)
-    case (SetT1(l), SetT1(r)) => commonSupertype(l, r).map(SetT1)
-    case (FunT1(lD, lC), FunT1(rD, rC)) =>
-      for {
-        domT <- commonSupertype(lD, rD)
-        cdmT <- commonSupertype(lC, rC)
-      } yield FunT1(domT, cdmT)
-    case (TupT1(lElems @ _*), TupT1(rElems @ _*)) if lElems.size == rElems.size =>
-      val commonTup = lElems.zip(rElems).flatMap { case (l, r) => commonSupertype(l, r) }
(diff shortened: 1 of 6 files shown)


tlair/src/main/scala/at/forsyte/apalache/tla/typecomp/signatures/FlexibleEquality.scala
```diff
@@ -16,6 +17,8 @@ object FlexibleEquality {
   // None if no common supertype, Some(e) if compatible, and common supertype is e.
   def commonSupertype(lhs: TlaType1, rhs: TlaType1): Option[TlaType1] = (lhs, rhs) match {
     case (a, b) if a == b     => Some(a)
+    case (VarT1(_), b)        => Some(b)
+    case (a, VarT1(_))        => Some(a)
```
shonfeder, line 21: @Kukovec I'd like to make sure you agree this is is the correct fix. This is overly permissive, as doesn't account for variable binding, but I think we should *not* put that logic in the builder, as we move much closer to reproducing the type checker in the builer then, which smells like tech debt, given that we can just use the type checker if we want :)

If you agree on this fix, then dozens of the builder test cases will need to be updated to account for type variability. 

My proposed fix, as seen below, is to clean up these complicated test cases by replacing them with simpler, PBT-style tests (which have -- or can be made to have -- equivalent or slightly better coverage, iiuc). These replacements use the concrete type builder introduced in #2550, which does not produce type variables.

Checks for logic around type variables can be confined to test of operator application, I believe.

WDYT?

shonfeder:
In conversation last week, @Kukovec rejected this change in preference for avoiding use of `ScopedBuilder` and instead using the unsafe builder methods for converting quint.

I am fine with this approach in principle, however afaict, there is no non-scope-safe builder implementation that has parity with the scoped builder. We don't even have an abstract interface to build against for putting this together. As a result, the scope or work required is much larger than the two line change I've proposed here.

I am asking that we move forward with my proposal here, in light of the following considerations: 

- the scoped builder is broken in this regard: it is impossible to build the application of polymorphic operators
- the scoped builder is meant to be a helper operation that expedites other aims, it is not a goal in itself, whereas completing the integration with quint is high priority
- if it deemed important to enrich the builder with more type guarantees during construction or to add a non-scoped version satisfying the same interface, this can easily be added later, replacing my change.

tlair/src/main/scala/at/forsyte/apalache/tla/types/TypeUnifier.scala
```diff
@@ -223,6 +223,11 @@ class TypeUnifier(varPool: TypeVarPool) {
       case (RecRowT1(RowT1(lfields, lv)), RecRowT1(RowT1(rfields, rv))) =>
         unifyRows(lfields, rfields, lv, rv).map(t => RecRowT1(t))
 
+      case (rec @ RecT1(_), rowRec @ RecRowT1(_)) => compute(rowRec, rec)
+
+      // An old record type can be treated as a monomorphic row-typed record
+      case (rowRec @ RecRowT1(_), RecT1(fields)) => compute(rowRec, RecRowT1(RowT1(fields, None)))
+
```
shonfeder, line 230: The motivation here was that, at some point for a reason I'm not totally clear on, operations in https://github.com/informalsystems/apalache/tree/cbf3870476b035a3280a9cb790edd60803b17a6f/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateDecoder.scala are decoding an Arena that must have been a row-typed record at type checking into an old record, and then passing that on to the builder, which then fails to unify the types.

However, I think this addition is well motivated in any case, since an old record typing should be able to be considered a subtype in this way. This basically picks up a piece of the logic from the `commontSuperType` code that is being removed.

Kukovec (approved):
LGTM

shonfeder:
Thanks for the review!

PR #2567, opened 2023-05-24
Quint conversion: Track nullary operator declarations at top-level

shonfeder:
Closes informalsystems/quint#896

We were not tracking nullary operators defined at the top-level, which was leading to reference to top-level operator hitting the same problem as we saw in the typing of nested definition of nullary operators.

The fix here is to

- Track whether a declaration conversion was for a nullary operator
- If so, add the name of for this operator into the store of nullary operator names used for correcting the types when nullary operator are referenced.
 
Tests for this have been conducted locally and will be added as integration tests in quint: testing this requires construction of a valid quint module, and IMO that would not be an economical addition to our test coverage here given the need to cover it already on the quint side.

- [-] Tests added for any new code


thpani (approved):
Looks great!

PR #2588, opened 2023-06-03
Fix some QuintIR serialization issues

shonfeder:
These have turned up thanks to @konnov's report in
https://github.com/informalsystems/quint/issues/927


Makefile
```diff
@@ -40,14 +40,20 @@ integration: package
...
 	curl https://raw.githubusercontent.com/informalsystems/quint/main/examples/puzzles/tictactoe/tictactoe.qnt > $(TEMP_QNT_TTT_FILE)
 	quint typecheck --out $@ $(TEMP_QNT_TTT_FILE)
 	rm $(TEMP_QNT_TTT_FILE)
 
+TEMP_QNT_CS_FILE := $(shell mktemp)
+tla-io/src/test/resources/clockSync3.json:
+	curl https://raw.githubusercontent.com/informalsystems/quint/main/examples/classic/distributed/ClockSync/clockSync3.qnt > $(TEMP_QNT_TTT_FILE)
+	quint typecheck --out $@ $(TEMP_QNT_CS_FILE)
+	rm $(TEMP_QNT_CS_FILE)
+
```
konnov, line 56: I am a bit worried that we would have a massive Makefile. Is there a way to use pattern rules?

tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintIR.scala
```diff
@@ -5,6 +5,29 @@ package at.forsyte.apalache.io.quint
...
+//    expected <JSON value> got <diff JSON value> at index <N>
+//
+// E.g.,
+//
+//    expected sequence got dictionary at index 525
+//
+// The _index_ is the character count of the JSON data. To debug,
+// first get quint's JSON representation of the spec with
+//
+//    quint typecheck --out spec.qnt.json spec.qnt
```
konnov, line 25: This looks a bit hard to debug. Would not it be easier (for me :)) to save the offending json and print the filename?

konnov (approved):
Looks good to me. Thanks for the fix!

Makefile
shonfeder: I don’t think we should need more of these these, but if we add a third, I’ll work out a pattern :)

tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintIR.scala
shonfeder: We don’t have file names
At this point. But file names are printed via the top level error handler when this is activated h the cli. 

Unfortunately upickle just doesn’t offer very nice error messages when using its generated serialization/deserialization :( this probably wouldn’t be a problem if we were using one of the more popular Json libs? If we were manually deserializing everything we could do our own error handling of course, but I don’t think it would be worth the extra code.

I’m not very happy with upickle so far tho, so maybe we can swap it out to get better debugging. The only issue is that upickle was selected for the Apalache IR, and I’m
Not sure we want two competing serialization formats :)

shonfeder: I’ll think on it!

PR #2608, opened 2023-06-20
Prevent shadowing of "_" Quint lambda parameters

thpani:
Rename quint lambda `_` parameters to `__QUINT_UNDERSCORE_{id}`.

This prevents shadowing of nested underscore parameters.

Closes #926

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
@@ -226,7 +226,13 @@ class Quint(moduleData: QuintOutput) {
                       | ${lambda}""".stripMargin
               )
             case QuintLambda(_, params, _, scope) =>
-              val tlaArgs = params.map(p => tla.name(p.name, Quint.typeToTlaType(types(p.id).typ)))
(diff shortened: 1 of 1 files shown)


Kukovec (approved):

tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
```diff
@@ -226,7 +226,13 @@ class Quint(moduleData: QuintOutput) {
                       | ${lambda}""".stripMargin
               )
             case QuintLambda(_, params, _, scope) =>
-              val tlaArgs = params.map(p => tla.name(p.name, Quint.typeToTlaType(types(p.id).typ)))
+              // uniquely rename parameters of name "_" to prevent shadowing of nested "_"s
```
shonfeder, line 229: This is great! Nice fix :)

we should be sure to add changelog entries for these kinds of bug fixes I think. Do you agree?

thpani: fix in https://github.com/informalsystems/apalache/pull/2615

PR #2628, opened 2023-07-03
Refactor `Quint` class

thpani:
The `Quint` class currently contains all of the translation-related code, including
def/expr translation, type translation, and unique name generation.

Some of this functionality is contained in nested classes + companion objects, both on the `Quint` class and the `Quint` companion object.
With recent issues, it has become hard to debug where these classes / objects are instantiated, what (parent) state they access, and where that is instantiated.

This refactoring PR splits off unique name generation and type conversion into separate classes / files. It also replaces implicit class instantiation in `apply` with explicit constructor calls via `new`, and renames / moves some methods to arrive at a more uniform interface.

The refactor is all mechanical; the only remarkable commit is f1a38d5afc5d00cdd8667ab536f45de40ddd3568, which constructs one translator per module (rather than per definition).

I think its easiest to review the entire code at once, by checking out the branch and diffing sans-whitespace with `git diff -w main`.


konnov (approved):
Makes sense to me. Thank you for splitting the PR into smaller commits. It's quite hard to understand the whole diff, so I could only get the sense of the changes by looking at the individual commits.

PR #2629, opened 2023-07-03
Use single typevar map for entire spec

thpani:
Use a single type converter for an entire quint spec.

We previously used one type converter per sub-expression. This re-initialized the map for translating Quint type variables (named) to Apalache TypeSystem 1 type variables (numbered), causing wrong type annotations. For example the type of the following Quint was translated as `(a,a) => {src: a, ts: a}`.

```bluespec
/// `p.msgFrom(ts)` is a message sent from `p` at time `ts` 
def msgFrom(p, ts) = {src: p, ts : ts}
```

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintTypeConverter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintTypeConverter.scala
@@ -120,7 +120,7 @@ private class QuintTypeConverter extends LazyLogging {
     RowT1(fieldTypes: _*)
   }
 
-  private val convert: QuintType => TlaType1 = {
+  val convert: QuintType => TlaType1 = {
     case QuintBoolT()             => BoolT1
     case QuintIntT()              => IntT1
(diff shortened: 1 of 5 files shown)


tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintTypeConverter.scala
```diff
@@ -135,13 +135,3 @@ private class QuintTypeConverter extends LazyLogging {
...
 }
-
-/**
- * Convert a QuintType to a TlaType1.
- *
- * Constructs a fresh scope for translating Quint type variables to Apalache type variables (see
- * QuintTypeConverter.getVarNo() for details).
- */
-object QuintTypeConverter {
-  def apply(quintType: QuintType): TlaType1 = new QuintTypeConverter().convert(quintType)
```
konnov, line 146: Good catch! I can see how the companion object was creating a new instance every time

konnov (approved):
Nice!

PR #2633, opened 2023-07-05
Extend unification error message in appOp

thpani:
Extend the exception message with expected and actual types, and the affected expression.


konnov (approved):
That's great! I like informative errors!

Issue #2339, opened 2023-01-12
Revisit/refactor caching


Kukovec:
Recent refactoring has uncovered that caches are often not being hit consistently. The reason for this is a type-tag check in the cache lookup, which often fails, as internally TLA expressions are constructed with an untyped builder (as opposed to `ScopedBuilder` used in the refactored files, which type-tags all values generated with `Typed(_)`). Additionally, caches lack a solid test suite.

Hitting caches consistently also highlights correctness bugs (e.g. #2338).

PR #2648, opened 2023-07-13
First attempt at `Cache` rework

Kukovec:

First steps toward #2339.

There's a few notable changes from the current implementation:

- `Recoverable[_]` is not used. While that trait served as a means to transition caches into arbitrary states, originally for multicore, such functionality is only needed in the presence of `SolverContext` as part of the cache data structure, since that class is internally mutable. If no such data is present, the functionality of `recover()` and `snapshot()` can be achieved by just passing around immutable Scala objects.
- `SolverContext` removed from all concrete cache implementations. In this iteration, the plan is to decouple SMT completely from the caches, and delegating the concern of maintaining synchronization obligations between caches/arenas and SMT to some dedicated manager class, if need be (in the future). As such, `SolverContext`s appear only as arguments to public methods, to be invoked on demand.
- Extra methods added to `Stackable`, to facilitate (potentially delayed) constraint generation (see above). 

### DIFF
--- a/tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/IntValueCacheTest.scala
+++ b/tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/IntValueCacheTest.scala
@@ -57,7 +57,6 @@ class IntValueCacheTest extends AnyFunSuite with BeforeAndAfterEach {
     cache.getOrCreate(a1, i2)
     cache.getOrCreate(a1, i2)
 
-    // should be obvious
     assert(fakeCtx.constraints.isEmpty)
 
     cache.addAllConstraints(fakeCtx)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  // reverse mapping
+  private var _reverseCache: Map[TargetT, (SourceT, LevelT)] = HashMap()
+  // inherited classes can read, but not modify without push/pop
+  protected def reverseCache: Map[TargetT, (SourceT, LevelT)] = _reverseCache
+
+  def values(): Iterable[TargetT] = _cache.values.map(_._1)
+
+  /**
+   * Create a target value based on the source value and cache it.
```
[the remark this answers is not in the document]
shonfeder: Maybe the issue is that this interface is not a cache, but an interface for memoized computation of some values that uses a cache?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/IntValueCache.scala
```diff
@@ -0,0 +1,38 @@
...
+ *   Jure Kukovec
+ */
+class IntValueCache extends Cache[PureArena, BigInt, ArenaCell] {
+
+  def create(arena: PureArena, intValue: BigInt): (PureArena, ArenaCell) = {
+    // introduce a new constant
+    val newArena = arena.appendCell(CellT.fromType1(IntT1))
+    val intCell = newArena.topCell
+    (newArena, intCell)
+  }
```
shonfeder, line 22: According to docs on the trait, this should be caching the source, right? But I don't see any cache entry happening here. I'm I missing something?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/Stackable.scala
```diff
@@ -0,0 +1,66 @@
...
+ * An object, the state of which can be saved in a stack-like manner by push and later restored by pop, similar to an
+ * SMT context. This trait is used for: SMT contexts and value caches.
+ *
+ * Some objects maintain a certain contract with the SMT solver, that is, they may add constraints to a
+ * [[at.forsyte.apalache.tla.bmcmt.smt.SolverContext]] on demand.
+ *
+ * @author
+ *   Jure Kukovec
+ */
+trait Stackable {
```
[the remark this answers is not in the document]
shonfeder: I'm not sure `Stackable` is a great name. Stackable suggests to me that a thing operates as a stack, but while this interface has methods `push` and `pop`, they don't have the signatures I'd expect of a stack.

iiuc, 'stack' is only showing up here because of a concept used in designing (but not actually implementing) the implementation? That is, the idea is that we have an interface for an object which can "timestamp" points in its history and then undo changes back to certain points. This is thought of as being like a stack of states, thus borrowing the terms `push` and `pop` from stacks. But the trait itself seems too have too general of a name. Maybe like "Snapshottable" would be better?

I'm also not sure `push` is a great name for the method that takes this snapshot, because it is not pushing anything. Maybe `takeSnapshot` would be clearer? And perhaps `revert` or `undo` for `pop`?

WDYT?

One other question: do you actually expect this trait to be implemented by other classes? It seems very special purpose given the last three methods. Perhaps we can do without the trait for now and just stick with the abstract class?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * An abstract cache that assigns `TargetT`-typed values to `SourceT` typed values, while potentially modifying a
+ * `ContextT`-typed context.
+ *
+ * Example: IntValueCache assigns [[ArenaCell]]s to integers, while modifying a [[PureArena]].
+ *
+ * While it extends [[Stackable]], there's no need to implement an actual stack data structure. By assigning each cache
+ * entry a level, effectively a timestamp for when it was added, we can proxy the state of the stack, by filtering out
+ * entries, based on their levels.
```
shonfeder, line 16: iiuc, this is a comment for the implementation but not needed for documenting the class? Maybe move it into a normal comment near the hashmaps that implement this approach?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,153 @@
...
+import at.forsyte.apalache.tla.bmcmt.stratifiedRules.aux.Stackable
+
+import scala.collection.immutable.HashMap
+
+/**
+ * An abstract cache that assigns `TargetT`-typed values to `SourceT` typed values, while potentially modifying a
+ * `ContextT`-typed context.
+ *
+ * Example: IntValueCache assigns [[ArenaCell]]s to integers, while modifying a [[PureArena]].
+ *
```
shonfeder, line 13: Could you please add a short paragraph explaining *why* this class needs to be able to snapshot and roll back the state? That is the most critical part of the behavior iiuc (otherwise we could just use `lazy` for memoization).

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,153 @@
...
+  private var _level: Int = 0
+
+  // the base map, tracking the level at which each entry was added.
+  private var _cache: Map[SourceT, (TargetT, LevelT)] = HashMap()
+  // inherited classes can read, but not modify without push/pop
+  protected def cache: Map[SourceT, (TargetT, LevelT)] = _cache
+
+  // reverse mapping
+  private var _reverseCache: Map[TargetT, (SourceT, LevelT)] = HashMap()
+  // inherited classes can read, but not modify without push/pop
```
shonfeder, line 37: push/pop are not primarily responsible for the updates, rather `addToCache` seems to be the main method. Perhaps correct that in the comment?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,153 @@
...
+   *
+   * @param source
+   *   the source value
+   * @param target
+   *   the cached target value
+   */
+  protected def addToCache(source: SourceT, target: TargetT): Unit = {
+    _cache += source -> (target, level)
+    _reverseCache += target -> (source, level)
+  }
```
shonfeder, line 84: What is the correct behavior if an equal `source` is added with a different `target` during a "later" snapshot? Should it just overwrite the previously cached value (along with its stored level)?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,153 @@
...
+   */
+  override def pop(n: Int): Unit = {
+    require(level >= n, s"Can't pop $n levels from a cache of level $level.")
+    _level -= n
+
+    def isOld[A, B](mapEntry: (A, (B, Int))): Boolean =
+      mapEntry._2._2 <= level
+
+    _cache = cache.filter(isOld)
+    _reverseCache = reverseCache.filter(isOld)
```
shonfeder, line 138: I'm not sure how big these caches will get, but it seems like this could be quite expensive if it gets large? An alternative that comes to mind is actually using a stack of sets of keys. Rolling back `n` levels would mean removing the keys from the `n` levels of the stack from the hashmap. This would save having to go thru every element of both hashmaps, and may be worh it if the caches are expected to get large. WDYT?

shonfeder (commented):
The general shape of the code looks good to me! I have a few initial questions for now! I figure I'll be able to do a proper review once I get more context thanks to your answers :)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
Kukovec: `addToCache` is protected, the interface entry point is `getOrCreate`, which doesn't allow this to happen.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
Kukovec: Any other alternative comes with a memory overhead, and traversal is `O(n)` anyhow. Since all of the non-SMT solving takes negligible time in practice, I think going for a lower memory footprint is better here.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/IntValueCache.scala
Kukovec: `getOrCreate` should be the only entry point, and `addToCache` is called there, one should never be calling `create` directly. Looking at it, I'm not 100% sure what the public-ness of a method that is declared as `protected` in `Cache` but def here is, but I would assume `protected` too.

shonfeder: Apologies if I am missing something, but `create` is defined as an abstract method on the the parent class, and is documented to say `Create a target value based on the source value and cache it.` Here you are implementing the method, but I don't see it caching. May that comment is stale?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
shonfeder: Yeah, but any child class could call this directly and overwrite a key, right? Maybe this should be `private`? If else maybe the check for collisions should be moved here?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
shonfeder: Makes sesns!

konnov (commented):
I am a bit confused about renaming `push/pop` into `snapshot/revert`. IMO, the new names do not convey the right message. In the old implementation, `snapshot` was producing a snapshot. In the new one, the state is kept internally, so `push/pop` actually make more sense to me.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/Snapshottable.scala
```diff
@@ -0,0 +1,43 @@
...
+   * Importantly, `revert` may be called multiple times and thus it is not sufficient to save only the latest state.
+   */
+  def revert(): Unit
+
+  /**
+   * Revert n times.
+   * @param n
+   *   Revert n times, if n > 0, otherwise, do nothing
+   */
+  def revert(n: Int): Unit
```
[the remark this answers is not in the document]
Kukovec: @shonfeder asked me to change it this way: https://github.com/informalsystems/apalache/pull/2648#discussion_r1264830100
I can only comply with one of these mutually exclusive requests, so you two need to reach consensus.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,177 @@
...
+   */
+  override def revert(n: Int): Unit = {
+    require(level >= n, s"Can't pop $n levels from a cache of level $level.")
+    _level -= n
+
+    def isOld[A, B](mapEntry: (A, (B, Int))): Boolean =
+      mapEntry._2._2 <= level
+
+    _cache = cache.filter(isOld)
+    _reverseCache = reverseCache.filter(isOld)
```
thpani, line 145: `isOld` is a confusing name here. how about:

```suggestion
    def isRetained[A, B](mapEntry: (A, (B, Int))): Boolean =
      mapEntry._2._2 <= level

    _cache = cache.filter(isRetained)
    _reverseCache = reverseCache.filter(isRetained)
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Add constraints for all entries added at any level.
+   */
+  def addConstraints(ctx: SolverContext): Unit = addConstraintsFromLevel(ctx)(0)
+
+  /**
+   * Add constraints for all entries added at the current level.
+   */
+  def addConstraintsForCurrentLevel(ctx: SolverContext): Unit = addConstraintsFromLevel(ctx)(level)
```
thpani, line 175: I don't think these should go into an abstract cache implementation.

Introduce a separate trait?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/IntValueCache.scala
Kukovec: You're referencing a comment from an old version, it has since been updated to "Compute a target value and context update, based on the source value, but do not save the result."

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/Snapshottable.scala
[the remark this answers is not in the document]
Kukovec: Ok, Shon, you got outvoted 2:1so I'm going to rename things back to Stackable.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,177 @@
...
+  private var _cache: Map[SourceT, (TargetT, LevelT)] = HashMap()
+  // inherited classes can read, but not modify without going thorough the public interface methods
+  protected def cache: Map[SourceT, (TargetT, LevelT)] = _cache
+
+  // reverse mapping
+  private var _reverseCache: Map[TargetT, (SourceT, LevelT)] = HashMap()
+  // inherited classes can read, but not modify without going thorough the public interface methods
+  protected def reverseCache: Map[TargetT, (SourceT, LevelT)] = _reverseCache
+
+  def values(): Iterable[TargetT] = _cache.values.map(_._1)
```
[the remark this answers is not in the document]
Kukovec: Hmm, you're probably right that `reverseCache.keys` is strictly better, it was this way originally and since it's not technically wrong it didn't stand out to me, but I'll change it to your suggestion.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
Kukovec: I agree in principle, but I was specifically asked to move these methods into `Cache` in https://github.com/informalsystems/apalache/pull/2648#discussion_r1264830100

> It seems very special purpose given the last three methods. Perhaps we can do without the trait for now and just stick with the abstract class?

I'll move them back out

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/Snapshottable.scala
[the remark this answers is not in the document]
shonfeder: I’m happy to defer if everyone else agrees. 

I would be interested to understand your reasoning tho. I don’t see how this behaves like a stack tho. Seems clear to me that it is based on a stack-like approach, but it is not “stackable”. E.g., you cannot push anything to it. You cannot peek at top elements. It is a key-value store that uses a stack abstraction to wipe out some parts of the state stored. id be genuinely interested to learn why you three look at that and say “this is a stack” :) 


Sorry for the churn on your PR due to disagreeing reviewers!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
Kukovec: I don't have a final name in mind, but to me there exists something like `SmtSyncable`, i.e. a trait, that is shared by all entities (not just caches, in principle), which "have to" update SMT constraints for every operation they perform (possibly delayed).

thpani: Hmmm, these are completely independent. The rest of the cache doesn't interacte with them, and vice versa.

So from a cohesion POV, I think they belong in a separate type.

Kukovec: I think it really is as simple as having a layer between `Stacklike` (`Snaphottable` here) and `Cache`, which governs the connection between stack-reversions and obligations to SMT

shonfeder: Ah I see if these methods
Are going to
Be used by several different classes and they are, as Thomas points out, totally orthogonal to this chaching utility that makes sense. Thanks for explaining! Sounds like they don’t belong in the stacklike trait either then? That was my particular concern before, but having them in a separate trait entirely for these reasons makes sense to me!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/Snapshottable.scala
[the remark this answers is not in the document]
thpani: Yeah, maybe this trait is more general, but the intended use is to add items at `level`, and then use `revert`/`snapshot` to manipulate a stack.

imo `Stackable` is not so bad.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
Kukovec: I mean, classes/traits are ultimately just a bookkeeping tool, if we'd rather have 1 trait 2ith 6 methods over 2 with 3 each that's not really that bad. We're not building a general-purpose library here.

thpani: > I mean, classes/traits are ultimately just a bookkeeping tool, if we'd rather have 1 trait with 6 methods over 2 with 3 each that's not really that bad. We're not building a general-purpose library here.

Yeah, but then it should probably be `SmtSomethingCache`, not just `Cache` 😄

Kukovec: >afaict it's only going to be used by this one class. 

Transition exploration is also stacklike.

thpani: > So should I just get rid of the entire `Stacklike` trait and just have push/pop on the Cache or what? And then an extension of that `Cache` with the SMT stuff?

I think you have a better sense of what the final design / class hierarchy will look like.
What do you propose?

Kukovec: Well, as it stands, there'd be at most 2 broad categories of classes, which have reversals as part of their behavior, the caches and the transition explorer (and the former only need this behavior because of the latter (in the current implementation, there are 2 traits, `StackableContext` and `Recoverable[T]`).

I actually think it makes sense to have the stack-like interface shared; at the very least the concept of push and pop. If we're not doing multicore, we don't need snapshot/recover (from `Recoverable`). So I'd keep it as a simple interface, but the two categories obviously have different implementations of push/pop, and none of the implementations actually get reused.

thpani: > but the two categories obviously have different implementations of push/pop

Yeah, this bothers me... Why should they refine the same type then? What's the shared interface, other than that they share the same signature?

Kukovec: I mean, isn't that enough? A stack and a queue share the same push/pop interface, but they obviously have different implementations.

thpani: Not if you ask Liskov and Wing. One is LIFO, the other FIFO.

thpani: Actually, stack and queue comes up as the first example, if you look for behavioral subtyping: https://en.wikipedia.org/wiki/Behavioral_subtyping

Kukovec: That post specifically says that it's dangerous to have a stack be a subclass of Queue, not that they shouldn't both inherit from an interface with push/pop

thpani: You answered my question above about their shared specification

> What's the shared interface, other than that they have the same type signature?

with another question, which is answered in the post: 🤷 

> It is important to stress that whether a type S is a behavioral subtype of a type T depends only on the specification (i.e. the documentation) of type T; the implementation of type T, if it has any, is completely irrelevant to this question.

Kukovec: Roll it back a bit, because I'm not sure I understand what concern you're pointing out. Am I correct in interpreting that you don't want `Cache` and also `TrEx` in the future both extending `Stacklike`? Because `Stacklike`'s behavior is too vague?

thpani: Yes. There's no code sharing, because all defs in `Stacklike` are abstract. And the interface is vague, since it lacks a signature for adding elements. So I'd rather have the concrete defs of `push` and `pop` first defined in two places, with very concrete behavioral descriptions.

My real pet peeve here is the following: Let's say I Cmd-click on `push()` in this line:

https://github.com/informalsystems/apalache/blob/bd215d798f14befd05c3ec72dc66770cc76f0cc7/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SymbStateRewriterImpl.scala#L526

IntelliJ takes me into the `StackableContext` trait. Not good, I wanted to see the actual implementation. If I try to reverse-lookup implementations of `push()`, there's 9 to choose from, from caches to rewriters, to the actual solver context. For `Recoverable`, it's even worse, there's 11 implementations of `recover()` (including the actual TRex, caches, various execution contexts, rewriters, etc). And all I wanted is to follow as simple method call.

thpani: Or, in shorter form: The only reason to have a `Stacklike` here, is if you actually had a `def doSomething(x: Stacklike)` that we wanted to call with either a `Cache` or a `TrEx`. IMO that's not the case here, so no reason to have a common subtype.

Kukovec: > If I try to reverse-lookup implementations of push(), there's 9 to choose from, from caches to rewriters, to the actual solver context. For Recoverable, it's even worse, there's 11 implementations of recover()

Yeah, part of the problem is that there a few de facto chaches, that just don't extend `AbstractCache`, but reimplement push&pop in exactly the same way. Ideally there would be 2 sub-classes, one for caches and one for transitions.

> The only reason to have a Stacklike here, is if you actually had a def doSomething(x: Stacklike) that we wanted to call with either a Cache or a TrEx.

You know, I can get behind that

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Creates a stack-like bidirectional cache, which behaves in the following way: Suppose there exists a deterministic
+ * injective function `f: (ContextT, SourceT) -> (ContextT, TargetT)`, that is, a function which, given a value of type
+ * `SourceT`, computes a value of type `TargetT`, while mutating a context of type `ContextT`.
+ *
+ * [[Cache]] defines an entry-point `getOrCreate`, such that
+ *   - `getOrCreate(c, x) = f(c,x)`, if this is the first time `x` appears as the source value for `getOrCreate`.
+ *   - `getOrCreate(c, x) = (c, v)`, if `getOrCreate(c', x)` was previously called, and returned `(c'', v)` (for some
+ *     `c', c''`).
```
shonfeder, line 16: Is the "mutation" of `ContextT` in place? If so, why do we need to return a reference to it? If it is instead a functional update of `ContextT`, then the doc should be tweaked I guess, e.g.:

```suggestion
 * injective function `f: (ContextT, SourceT) -> (ContextT, TargetT)`, that is, a function which, given a value of type
 * `SourceT`, computes a value of type `TargetT`, while generating an updated context of type `ContextT`.
 *
 * [[Cache]] defines an entry-point `getOrCreate`, such that
 *   - `getOrCreate(c, x) = f(c',x)`, if this is the first time `x` appears as the source value for `getOrCreate`.
 *   - `getOrCreate(c, x) = (c, v)`, if `getOrCreate(c', x)` was previously called, and returned `(c'', v)` (for some
 *     `c', c''`).
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,170 @@
...
+  def get(srcValue: SourceT): Option[TargetT] = cache.get(srcValue).map(_._1)
+
+  /**
+   * Find the key that was used to create a given value.
+   * @param value
+   *   a value, for which the key should be found
+   * @return
+   *   the key, if exists
+   */
+  def findKey(value: TargetT): Option[SourceT] = reverseCache.get(value).map(_._1)
```
thpani, line 111: This is the first time you mention "key". Should it be "source value"?

thpani (commented):
Code looks great, in particular the Scaladoc on `Cache`!

I still have a few questions about the docs on `DelayedConstraintGenerator`:

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/DelayedConstraintGenerator.scala
```diff
@@ -0,0 +1,26 @@
+package at.forsyte.apalache.tla.bmcmt.stratifiedRules.aux
```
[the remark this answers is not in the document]
Kukovec: I put stuff here that is required _for_ rewriting rules, but isn't a rewriting rule (aux is short for auxiliary) . If you dislike it or want to reshuffle things, I'm not particularly attached to it.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/DelayedConstraintGenerator.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import at.forsyte.apalache.tla.bmcmt.smt.SolverContext
+
+/**
+ * A [[DelayedConstraintGenerator]] may, on demand, discharge SMT constraints.
```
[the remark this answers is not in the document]
Kukovec: Externally, though it's currently up for debate where exactly this happens. I'll outline that part in the next comment.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/DelayedConstraintGenerator.scala
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * The primary use-case for this trait lies with [[caches.Cache]]s. Instead of instantly discharging SMT constants as
+ * soon as an entry is added to the cache, which effectively makes all caches stateful and inextricably liked to
+ * [[at.forsyte.apalache.tla.bmcmt.smt.SolverContext]], we can manipulate caches first, then discharge a batch of
+ * constraints separately.
```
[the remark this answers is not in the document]
Kukovec: Well, that remains to be seen. Personally, I think rules should call caches, and never worry about cache-related assertions. Instead, adding cache constraints to SMT should happen as late as possible, specifically in the transition manager just before you call `push` or `checkSat`. The reason for this is that, if you do it this way, you can have a wrapper class around the `push-` or `check-` calls, which ensures that cache constraints are always added before a push, which is an invariant you want to have anyhow.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
Kukovec: Makes sense

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/DelayedConstraintGenerator.scala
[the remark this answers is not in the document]
thpani: Makes sense!

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/DelayedConstraintGenerator.scala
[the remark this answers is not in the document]
thpani: Can we add a TODO comment, or file an issue, to remind us to update this comment with an explanation of how things actually work? I think that's important for future minds tyring to understanding the new caching framework :)

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/Cache.scala
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Creates a stack-like bidirectional cache, which behaves in the following way: Suppose there exists a deterministic
+ * injective function `f: (ContextT, SourceT) -> (ContextT, TargetT)`, that is, a function which, given a value of type
+ * `SourceT`, computes a value of type `TargetT`, while generating an updated context of type `ContextT`.
+ *
+ * [[Cache]] defines an entry-point `getOrCreate`, such that
+ *   - `getOrCreate(c, x) = f(c',x)`, if this is the first time `x` appears as the source value for `getOrCreate`.
+ *   - `getOrCreate(c, x) = (c, v)`, if `getOrCreate(c', x)` was previously called, and returned `(c'', v)` (for some
+ *     `c', c''`).
```
thpani, line 16: One last comment:

Would it make sense to demo and test this behavior in a unit test for `IntValueCache`?

Kukovec: Yes, actually. I was planning on adding cache tests once there's more cache-code out there, but no reason not to have it now, for just the int cache.

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/IntValueCacheTest.scala
```diff
@@ -0,0 +1,75 @@
...
+    // Some extra calls, which shouldn't affect constraint generation
+    cache.getOrCreate(a0, i1)
+    cache.getOrCreate(a0, i1)
+    val (_, ci2) = cache.getOrCreate(a1, i2)
+    // Some extra calls, which shouldn't affect constraint generation
+    cache.getOrCreate(a1, i2)
+    cache.getOrCreate(a1, i2)
+
+    // should be obvious
+    assert(fakeCtx.constraints.isEmpty)
```
thpani, line 61: Since it's obvious, we don't need the comment? 🙃

```suggestion
    assert(fakeCtx.constraints.isEmpty)
```

thpani (approved):
Looks great! Thanks for going through all the iterations on this PR

PR #2654, opened 2023-07-20
Deserialize `QuintInt.value` as `BigInt`

thpani:
Deserialize `QuintInt.value` as `BigInt`

Together with https://github.com/informalsystems/quint/issues/1059, closes https://github.com/informalsystems/quint/issues/1055

Closes #2653

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [ ] ~Documentation added for any new functionality~
- [x] [Entries added to `./unreleased/`][changelog format] for any new functionality

[changelog format]: https://github.com/informalsystems/apalache/blob/main/CONTRIBUTING.md#how-to-record-a-change

PR #2661, opened 2023-07-24
(De-)serialize BigInt from/to JSON number in Quint import

thpani:
#2654 fixed the deserialization of `QuintInt.value` into a Scala BigInt.
However, this is not the only TypeScript `bigint` needing conversion:
All Quint IDs, plus the lookup table references, are also TypeScript bigints and should be deserialized into Scala BigInts.

upickle's built-in `BigInt` (de)serializer reads/writes BigInts only from/to JSON strings.
Thus, there's two options for Apalache <-> Quint compatiblity:
1. Read/write all `bigint`s as strings from Quint.
2. Extend upickle to support (de)serializing BigInts from/to JSON numbers, [which can be arbitrarily long](https://www.json.org/json-en.html). This is how [`json-bigint`](https://www.npmjs.com/package/json-bigint) already behaves on the Quint side.

A brief investigation showed that approach (1) is non-trivial; for example, it would break the JSON re-import inside Quint.

This PR implements approach (2):
* Extend `QuintDeserializer` with custom `BigInt{Reader,Writer}` implicits to read/write JSON numbers.
* Port the remaining Quint IDs to BigInts.
* Revert commit 2a073abb6001aa1b23772cc8134fba0448ddc443, which wrapped all `QuintInt.value`s in integration tests into strings as part of #2654. Since we now read/write JSON numbers, this has become unnecessary.

Closes https://github.com/informalsystems/quint/issues/1055


Kukovec (approved):
I assume the two json files are generated by something, so I didn't read them. The rest looks good.

thpani:
> I assume the two json files are generated by something, so I didn't read them. The rest looks good.

Aaah yes, that's the revert commit. Thanks!

shonfeder (commented):
Awesome work! Glad to have the root cause sorted out :)

PR #2673, opened 2023-07-27
`IntRangeCache` rework

### LINKED ISSUE #2665: Rewrite cache: `IntRangeCache`
Partially resolves #2339.

Kukovec:

closes #2665

Moves `FakeCtx` out of the int cache tests since it's reusable, and adds forgotten test annotations there.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/caches/IntRangeCache.scala
```diff
@@ -0,0 +1,45 @@
+package at.forsyte.apalache.tla.bmcmt.stratifiedRules.aux.caches
+
+import at.forsyte.apalache.tla.bmcmt.smt.SolverContext
+import at.forsyte.apalache.tla.bmcmt.types.CellT
+import at.forsyte.apalache.tla.bmcmt.{ArenaCell, FixedElemPtr, PureArena}
+import at.forsyte.apalache.tla.lir.{IntT1, SetT1}
+import at.forsyte.apalache.tla.types.tla
+
+/**
+ * Cache tuple domains as well as ranges a..b.
```
thpani, line 10: Not sure, but isn't one just a special case of the other?

```suggestion
 * Cache ranges a..b and, as a special case, tuple domains.
```

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/FakeCtx.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import at.forsyte.apalache.tla.bmcmt.smt.{SolverConfig, Z3SolverContext}
+import at.forsyte.apalache.tla.lir.TlaEx
+
+/**
+ * A fake context, which just intercepts any expression added via assertGroundExpr and stores it in a Seq.
+ *
+ * We can use it in tests, to observe constraint generation without needing to instrument SMT.
+ */
+class FakeCtx extends Z3SolverContext(SolverConfig.default) {
```
thpani, line 11: I don't have a strong preference here, but you could just use easymock's `mock`?

thpani, line 11: In any case, let's call this a mock, I think that's the established lingo for this kind of things.

```suggestion
 * A mock context, which just intercepts any expression added via assertGroundExpr and stores it in a Seq.
 *
 * We can use it in tests, to observe constraint generation without needing to instrument SMT.
 */
class MockZ3SolverContext extends Z3SolverContext(SolverConfig.default) {
```

thpani (approved):
LGTM

tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/stratifiedRules/aux/FakeCtx.scala
Kukovec: We could, but that's another dependency, and I tend to avoid those if possible.

thpani: Not sure I understand. Dependency on what? We're already using easymock.

Kukovec: Hmm, I checked it out, we're actually only using it in 2 places, and only for a class with 2 unit methods, which are trivial to mock. We could just get rid of it entirely

thpani: Erm, there's 5 tests using it. Also, I suggested that you **use** it, not that you **remove** it?

```sh
$ ag easymock
tla-typechecker/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestConstraintSolver.scala
8:import org.scalatestplus.easymock.EasyMockSugar

tlair/src/test/scala/at/forsyte/apalache/tla/types/TestTypeUnifier.scala
8:import org.scalatestplus.easymock.EasyMockSugar

tla-typechecker/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeAliasSubstitution.scala
6:import org.scalatestplus.easymock.EasyMockSugar

tla-typechecker/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestEtcTypeChecker.scala
7:import org.easymock.EasyMock
11:import org.scalatestplus.easymock.EasyMockSugar

passes/src/test/scala/at/forsyte/apalache/tla/typecheck/etc/TestTypeCheckerTool.scala
13:import org.easymock.EasyMock
17:import org.scalatestplus.easymock.EasyMockSugar
```

Kukovec: The mock sugars don't really do anything, you can delete the extends and the tests don't fail. 

But the point is, you said that it doesn't add any value for the current tests to avoid easymock, because it's already a dependency, which is true at present. However, we can remove easymock from the codebase, and slim our dependencies, at which point having the above mock class avoids having to reintroduce easymock.

PR #2704, opened 2023-08-22
Arena fix for Gen

### LINKED ISSUE #2702: Constraints applied to a set created with `Gen(_)` are ignored
Given the spec:

```tla
---------------------------- MODULE test --------------------------

EXTENDS Integers, Apalache

VARIABLE
  \* @type: Set(Int);
  gen

Init ==
  /\ gen = Gen(1)
  /\ \A x \in gen : x \in 1..5

Next == UNCHANGED gen
=================================================================
```
Apalache computes the initial state:
```tla
...
State0 == gen = {6}
...
```
It _appears_ that for `\A x \in gen : x \in a..b` the value generated for `gen` is always `{b+1}`.

Kukovec:

closes #2702 

Arenas for `Gen`-created sets were incorrectly always using fixed pointers, meaning that in the reconstruction (and subsequent output) these sets were shown to contain elements, which the underlying SMT constraints did not imply. The fix is to use a pointer with an unconstrained SMT value instead, and let the solver determine memberships, if any.


konnov (approved):
Looks good to me. Shall we add one more test for a number different from 1?

Issue #2713, opened 2023-08-28
Spec rejected due to being outside ReTLA throws stack trace instead of exiting cleanly


Kukovec:
The spec:
```tla
---------------------------- MODULE test --------------------------

EXTENDS Integers, Apalache


\* @type: Set(Int);
GEN == Gen(1)

Init == 0 \in GEN
Next == TRUE
Inv == TRUE
=================================================================
```

The command: `transpile test.tla`

Expected: `EXITCODE: ERROR (75)`
Actual:
```
java.lang.NoSuchMethodError: 'upickle.core.Types$TupleNWriter upickle.default$.Tuple2Writer(upickle.core.Types$Writer, upickle.core.Types$Writer)'
	at at.forsyte.apalache.tla.passes.pp.PreproPassPartial.postLanguageCheck(PreproPassPartial.scala:75)
	at at.forsyte.apalache.tla.passes.pp.PreproPassPartial.executeWithParams(PreproPassPartial.scala:87)
	at at.forsyte.apalache.tla.passes.pp.ReTLAPreproPassImpl.execute(ReTLAPreproPassImpl.scala:48)
	at at.forsyte.apalache.infra.passes.PassChainExecutor$.exec(PassChainExecutor.scala:64)
	at at.forsyte.apalache.infra.passes.PassChainExecutor$.$anonfun$runPassOnModule$3(PassChainExecutor.scala:53)
	at scala.util.Either.flatMap(Either.scala:352)
	at at.forsyte.apalache.infra.passes.PassChainExecutor$.$anonfun$runPassOnModule$1(PassChainExecutor.scala:51)
	at scala.collection.LinearSeqOps.foldLeft(LinearSeq.scala:183)
	at scala.collection.LinearSeqOps.foldLeft$(LinearSeq.scala:179)
	at scala.collection.immutable.List.foldLeft(List.scala:79)
	at at.forsyte.apalache.infra.passes.PassChainExecutor$.runOnPasses(PassChainExecutor.scala:44)
	at at.forsyte.apalache.infra.passes.PassChainExecutor$.run(PassChainExecutor.scala:34)
	at at.forsyte.apalache.tla.tooling.opt.TranspileCmd.run(TranspileCmd.scala:25)
	at at.forsyte.apalache.tla.Tool$.runCommand(Tool.scala:138)
	at at.forsyte.apalache.tla.Tool$.run(Tool.scala:118)
[...]

PR #2714, opened 2023-08-28
Transpile regression

Kukovec:

Regression test for #2713


thpani (commented):
Why are we adding a passing regression test when the issue (#2713) is still open?

Would be useful to record this context in the PR description.

Kukovec:
> Why are we adding a passing regression test when the issue (#2713) is still open?
> 
> Would be useful to record this context in the PR description.

For context: Shon and I were discussing this on slack, since he couldn't reproduce it on his device. We made a PR specifically so that CI would run it to see if it was something wrong with my environment, or whether there was a code issue. Given that it's passing, it must be the former.

shonfeder:
Thanks for providing the context, Jure!

I agree with Thomas that it would help us all collaborate and save time to ensure every or had at least a sentence or two (but as much as needed) noting the context.

PR #2720, opened 2023-09-08
PI-week: Refactor Transpile

Kukovec:

This PR attempts to change the way the internal transpilation is performed, and, to a lesser extent, the SMT term representation.

The main issue is, that SMT separates two concepts: terms, and declarations/definitions. In an effort to make the transpilation "side-effect free", `FunDef` was introduced as a `Term`, which represented a function, to be defined in the SMT header, but used (by name) wherever `FunDef` appeared in the syntax tree. Later, a `Collector` would traverse the tree again, and extract the function definitions, along with any uninterpreted literals, which also needed to be declared.
This is clunky, and does not scale with extensions to `Term`s. Instead, the new approach introduces rules as `State` computations over a class which holds SMT declarations. This way, rules conceptually still translate TLA to SMT terms directly, but the internal declarations class keeps track of any declarations or definitions, which need to be introduced along the way.

The PR also includes minor QoL changes, such as replacing case classes without parameters with case objects, and standardizing the use of `Seq` (instead of a mix of `Seq` and `List`).


konnov (approved):
Looks good!

rodrigo7491:
Thanks for the changes @Kukovec!

PR #2733, opened 2023-09-14
Revise the ITF format

### LINKED ISSUE #2732: Revise the ITF format
The current version of [ITF](https://github.com/informalsystems/apalache/blob/403eb1430186fe962d22b12343babcfc6f24492d/docs/src/adr/015adr-trace.md) introduces two ways of serializing not-too-big integers:

> - A JSON integer, e.g., 123. According to [RFC7159](https://datatracker.ietf.org/doc/html/rfc7159.html), JSON integers must be in the range: $[-2^{53}+1, 2^{53}-1]$. Integers in this range may be written as JSON integers.
>    
> - A big integer of the following form: `{ "#bigint": "[-][0-9]+" }`. We are using this format, as many JSON parsers impose limits on integer values, see [RFC7159](https://datatracker.ietf.org/doc/html/rfc7159.html). Big and small integers may be written in this format.

Back then, I thought it was a good idea, but it introduces branching in every parser, which distinguishes between `#bigint` field and an integer field.

I was thinking of just keeping the `#bigint` field.

konnov:
Closes #2732. In this PR, we further simplify the ITF format by always representing integers as `{ "#bigint": "num" }`, instead of having two cases, namely, `num` for small integers and `{ "#bigint": "num" }` for big integers. This should simplify the usage of ITF in tests.

### DIFF
--- a/.unreleased/features/2732.md
+++ b/.unreleased/features/2732.md
@@ -1,0 +1,1 @@
+- Revise the ITF format: Use only `{ "#bigint": "num" }`, even for small integers`
--- a/docs/src/adr/015adr-trace.md
+++ b/docs/src/adr/015adr-trace.md
@@ -1,8 +1,8 @@
 # ADR-015: Informal Trace Format in JSON
 
-| authors                                | proposed by                   | last revised    |
-| -------------------------------------- | ----------------------------  | --------------: |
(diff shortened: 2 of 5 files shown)


Kukovec (approved):
lgtm

PR #2741, opened 2023-09-18
Account for possible output in CLI server test

shonfeder:
It's possible for output to be printed after the welcome message in
these tests. This change accounts for that.

An example of the incorrect failures on this case is https://github.com/informalsystems/apalache/actions/runs/6225708423/job/16896714985?pr=2739

It's just a race condition on the output.

-------


thpani (approved):
Looks great, thanks for taking care of this!

Issue #2316, opened 2022-12-08
Solver restart from timeout in simulate mode


rnbguy:
## Is your feature request related to a problem? Please describe

The `simulate` mode can get ~stuck~ slow if the solver is taking too long to solve the next transitions or to construct an example run.

I know I can set `--tuning-options=search.smt.timeout=15`. But this terminates the search globally - and does not try for the next example runs. It would be nice if z3 is timed out, apalache continues trying for the next runs until `--max-run` many runs are tried - timed out or not. 

## Describe the solution you'd like

For first versions, something like `--tuning-options=search.smt.timeout=15:search.smt.timeout.ignore=true`

## Describe the impact on your work

I can set a high `--max-run` and a timeout for the solver to retry for new runs automatically.

## Describe alternatives you've considered

On bash shell, `until apalache-mc simulate --tuning-options=search.smt.timeout=15 ...; do ...; done`

thpani:
Partially addressed in #2758 – please reopen if this continues to be an issue.

PR #2758, opened 2023-10-16
Continue simulation on SMT timeout in enabledness check

thpani:
Prevent Apalache from halting the (global) search if an enabledness query times out in `simulate`.

This is a simple fix that considers a transition disabled if the corresponding SMT query times out in `simulate` mode.

A more comprehensive treatment of timeouts in `simulate` would be more invasive: It would require us to differentiate timeouts from other `UNKNOWN` Z3 results, and to define (or parameterize) the expected search behavior if invariant or deadlock checks time out during simulation.

Partially addresses #2316 and unblocks https://github.com/informalsystems/quint/issues/1196.

### DIFF
--- a/.unreleased/bug-fixes/2758-sim-timeout.md
+++ b/.unreleased/bug-fixes/2758-sim-timeout.md
@@ -1,0 +1,1 @@
+Continue simulation on SMT timeout in enabledness check, see #2758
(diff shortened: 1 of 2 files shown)


konnov (approved):
That's a nice and simple workaround! I guess, we cannot distinguish between UNKNOWN and TIMEOUT in this case, right?

PR #2782, opened 2023-11-24
Add case to handle oneOf outside of nondet

### LINKED ISSUE #2774: Add warning for misplaced occurrence of `oneOf` when parsing from quint
As per https://github.com/informalsystems/quint/issues/381#issuecomment-1804043526

Can be achieved by adding a case at https://github.com/informalsystems/apalache/blob/d46cc5e8664637e6d4b67c9ca4ade3cf67ce263b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala#L546-L552 that returns an error when `oneOf` is encountered.

shonfeder:
Closes #2774

This is a workaround for https://github.com/informalsystems/quint/issues/381. 

The nub of the issue is that quint currently allows `oneOf` to occur outside of
`nondet` bindings, but we don't want to support this behavior in verification.
Our expectation is that we will make this illegal in quint, but in the mean time
this change will ensure less cryptic errors until the quint type discipline for
the operator is improved.

-------


bugarela (approved):
Thanks for taking care of this!

PR #2783, opened 2023-11-26
Support quint sum types

shonfeder:
Closes informalsystems/quint#1034
(thanks to https://github.com/informalsystems/quint/pull/1276).

This adds support for conversion of the following quint constructs:

- sum types
- the builtin `variant` operator
- the builtin `matchVariant` operator

A notable change made to support this work was relaxing the rules of the builder library to allow constructing empty records and record types. This is a meaningful value and a well formed type, and is the current choice for the unit type in quint, used for variants that don't take a value with any information (e.g., the `A` in `type T = A | B(int)`).

This takes us ~75% of the way to full support for quint sum types, and enables
running `verify` on
https://github.com/informalsystems/quint/blob/main/examples/language-features/option.qnt

What remains to do:

- Add support for the (not yet implemented) user-defined polymorphic type
  declarations. See https://github.com/informalsystems/quint/issues/1073

I expect reviewing by commit is helpful, as they are logically isolated.

---

Thanks due to @bugarela, who found this problem and was a huge help in debugging. 

---


tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
```diff
@@ -371,29 +372,75 @@ class Quint(quintOutput: QuintOutput) {
...
-              throw new QuintIRParseError(s"Invalid argument given to Rec ${invalidArgs}")
-          }
-        variadicApp { tlaVals =>
-          val fieldsAndArgs = fieldNames.zip(tlaVals)
-          tla.rowRec(rowVar, fieldsAndArgs: _*)
-        }(quintVals)
+    def record(rowVar: Option[String]): Converter = { quintArgs =>
+      // The quint Rec operator takes its field and value arguments
+      // via a variadic operator requiring field names passed as strings to
+      // be alternated with values. E.g.,
```
konnov, line 378: Since the only change here is supporting empty records, I would suggest adding a comment that explains why empty records are ok (they are the unit value).

tla-io/src/test/scala/at/forsyte/apalache/io/quint/TestQuintEx.scala
```diff
@@ -518,20 +520,13 @@ class TestQuintEx extends AnyFunSuite {
...
 
-  test("converting builtin Rec operator constructing empty record fails") {
-    val exn = intercept[QuintUnsupportedError] {
-      val typ = QuintRecordT.ofFieldTypes()
-      convert(Q.app("Rec")(typ))
-    }
-    assert(exn.getMessage.contains(
-            "Unsupported quint input: Given empty record, but Apalache doesn't support empty records."))
-  }
-
```
konnov, line 534: rather than deleting the test, I would leave it and assert that it does not fail. This is an important assumption and it would be great to have it reflected in the test

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
```diff
@@ -107,9 +107,6 @@ case object StrT1 extends TlaType1 {
  *   unique name of the constant type
  */
 case class ConstT1(name: String) extends TlaType1 {
-  require(ConstT1.isUninterpreted(name) || ConstT1.isAliasReference(name),
-      "ConstT1 accepts identifiers in upper case or $aliasReference, found: " + name)
-
```
konnov, line 112: This was an important invariant about uninterpreted types. Why is it gone now?

tlair/src/test/scala/at/forsyte/apalache/tla/typecomp/TestFunBuilder.scala
```diff
@@ -47,11 +47,6 @@ class TestFunBuilder extends BuilderTest {
         )
     )
 
-    // test fail on n = 0
-    assertThrows[IllegalArgumentException] {
-      build(builder.recMixed())
-    }
-
```
konnov, line 54: Similar to the comment above, it would be great to keep the test and assert that it does not fail anymore.

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
shonfeder: I noted the reason for this in the commit message, but my understanding is that this is a purely syntactic check, which we deal with in the parser. I don't see why the AST should care how the string looks, and I think this requirement on the AST node constructor is basically an artifact of mixing up levels of abstraction: we require a certain input syntax for parsing out valid type constants -- this is part of valid input to tell us that we need to construct a `ConstT1`. But if we are directly constructing a `ConstT1` the string representation should be irrelevant. The fact that no tests fail in the absence of this (imo, overly strict) requirement seems to support that view. But do you have some additional concern or context to share?

This change is motivated by the fact that it creates conflicts with quint's strings for type constants. We could instead do text munging on the quint strings, but that seems inelegant and to make for an overly tight coupling, when we can instead accommodate via removing the unneeded constraint.

tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintTypeConverter.scala
```diff
@@ -112,27 +112,28 @@ private class QuintTypeConverter extends LazyLogging {
...
-    RowT1(fieldTypes: _*)
-  }
+  // private def unionToRowT1(variants: Seq[UnionRecord]): RowT1 = {
+  //   val fieldTypes = variants.map {
+  //     case UnionRecord(tag, row) => {
+  //       (tag, RecRowT1(rowToRowT1(row)))
+  //     }
+  //   }
+  //   RowT1(fieldTypes: _*)
+  // }
```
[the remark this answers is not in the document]
shonfeder: Oops, forgot to clean that up. Thanks :)

tlair/src/main/scala/at/forsyte/apalache/tla/lir/TlaType1.scala
konnov: This is an internal invariant of the Intermediate Representation. It could be completely different from the syntax, as reflected in the Abstract Syntax Tree (e.g., of TLA+). I don't remember the reason for having this invariant. But I would expect other code depend on it. So instead of removing it, because it is in the way, I would rather try to understand, why it is there and whether anything breaks in Apalache, when this invariant does not hold true.

shonfeder: I reverted this change.

tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
```diff
@@ -371,29 +372,75 @@ class Quint(quintOutput: QuintOutput) {
...
+    //
+    // matchVariant(expr, "F1", elim_1, ..., "Fn", elim_n)
+    //
+    // Where each `elim_i` is an operator applying to value wrapped in field `Fi` of a variant.
+    //
+    // This is converted into the following Apalache expression, using Apalache's variant operators:
+    //
+    // CASE VariantTag(expr) = "F1" -> elim_1(VariantGetUnsafe(expr))
+    //   [] ...
+    //   [] VariantTag(expr) = "Fn" -> elim_n(VariantGetUnsafe(expr))
```
bugarela, line 427: Should this be

```suggestion
    // CASE VariantTag(expr) = "F1" -> elim_1(VariantGetUnsafe("F1", expr))
    //   [] ...
    //   [] VariantTag(expr) = "Fn" -> elim_n(VariantGetUnsafe("Fn", expr))
```

?

bugarela (approved):
LGTM, thanks for the many super helpful comments :sparkles:

tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
shonfeder: Yep! Thanks for the catch :D

PR #2792, opened 2023-12-01
Add support for default cases in quint match expressions

shonfeder:
This is a followup to #2783, which omitted support for the "wildcard",
default match case in match expressions.

I found this problem when checking the `Paxos` spec in the quint repo.  In
addition to the unit test added here, I've verified that this solves the problem
on a few quint specs locally. Integration tests confirming this will be
forthcoming in the quint repo.

-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
@@ -416,7 +416,7 @@ class Quint(quintOutput: QuintOutput) {
 
     // the quint builtin operator representing match expressions looks like
     //
-    // matchVariant(expr, "F1", elim_1, ..., "Fn", elim_n)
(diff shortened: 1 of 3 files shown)


tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala
```diff
@@ -416,7 +416,7 @@ class Quint(quintOutput: QuintOutput) {
 
     // the quint builtin operator representing match expressions looks like
     //
-    // matchVariant(expr, "F1", elim_1, ..., "Fn", elim_n)
+    // matchVariant(expr, "F1", elim_1, ..., "Fn", elim_n, "_", defaultElim)
```
bugarela, line 419: Perhaps a notation like this would help indicate that this is optional
```suggestion
    // matchVariant(expr, "F1", elim_1, ..., "Fn", elim_n[, "_", defaultElim])
```

bugarela (approved):
LGTM!

PR #2855, opened 2024-03-07
Update flake and improve nix setup for development

bugarela:
Hello :octocat: 

This PR:
1. Updates `flake.nix` to remove sourcing of `.envrc` file which resulted in a circular dependency if `direnv` was used to load nix. Other changes in `flake.nix` are just auto-formatting.
2. Adds instructions to `CONTRIBUTING.md` with how to use nix with `direnv`.
3. Updates `flake.lock` (with `nix flake update`) so the pinned version of `metals` is compatible with the scala version declared in `build.sbt`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,7 +184,8 @@ it here!
 We provide a nix shell in case you want to use nix to manage your development
 environment and dependencies. The primary benefits of using the nix shell is
 that it allows us to keep environments consistent, and distribute updates to
-environment dependencies.
+environment dependencies. Alternatively, you can also use `direnv` to load
+dependencies from nix.
 
 #### Getting Nix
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -202,6 +203,19 @@ Once you have `nix` installed, build and enter the clean development shell with:
...
 ```
 
+#### Loading nix dependencies with direnv
+
+If you want to use direnv to setup your environment with nix (instead of using a
+shell), you will need to add `use flake;` to your `.envrc`, and then running
+`direnv allow`:
+
+```sh
+echo "use flake;" >> .envrc && direnv allow
```
shonfeder, line 213: I think the `.local-envrc` could be used for this without requiring dirtying the worktree?

shonfeder (approved):
LGTM!

CONTRIBUTING.md
bugarela: Ah right, that should be better! Thanks

PR #2856, opened 2024-03-08
Use type given by quint to build lambda bodies

bugarela:
Hello :octocat: 

Fixes https://github.com/informalsystems/quint/issues/1393

The issue happened because we were only using the types given by Quint in the type map to annotate parameters, but not the result type of lambdas. This way, there was a mismatch in variable names on Apalache type inference (used for the body) and Quint type inference (used for the parameters).

This PR ensures that we always use Quint types.

### DIFF
--- a/.unreleased/bug-fixes/quint-lambda-types-conversion.md
+++ b/.unreleased/bug-fixes/quint-lambda-types-conversion.md
@@ -1,2 +1,3 @@
-Use the type inferred by Quint to build lambda bodies, avoiding mismatches with
+When converting Quint lambdas, derive the return type from the Quint type inferred for 
+the lambda, rather the type inferred for the body expression, avoiding mismatches with
 Apalache type variables. (#2856)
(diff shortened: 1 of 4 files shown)


shonfeder:
> This PR ensures that we always use Quint types.

Just to be clear: we have always been using quint types, it's just a matter of where we were sourcing those types from.

.unreleased/bug-fixes/quint-lambda-types-conversion.md
```diff
@@ -0,0 +1,2 @@
+Use the type inferred by Quint to build lambda bodies, avoiding mismatches with
```
shonfeder, line 1:
```suggestion
When converting Quint lambdas, derive the return type from the Quint type inferred for 
the lambda, rather the type inferred for the body expression, avoiding mismatches with
```

shonfeder (approved):
LGTM!

Great work debugging and fixing here! That test is much more readable than what we were looking at before :D

PR #2859, opened 2024-03-13
Extend Variants when pretty printing TLA+ in Shai

shonfeder:
Fixes an error in the existing TLA+ generated from quint, thanks to report from @bugarela

-------

### DIFF
--- a/shai/src/test/scala/at/forsyte/apalache/shai/v1/TestCmdExecutorService.scala
+++ b/shai/src/test/scala/at/forsyte/apalache/shai/v1/TestCmdExecutorService.scala
@@ -126,7 +126,7 @@ object TestCmdExecutorService extends DefaultRunnableSpec {
         val expectedPayload =
           """|----------------------------------- MODULE M -----------------------------------
              |
-             |EXTENDS Integers, Sequences, FiniteSets, TLC, Apalache
+             |EXTENDS Integers, Sequences, FiniteSets, TLC, Apalache, Variants
              |
              |Foo == TRUE
              |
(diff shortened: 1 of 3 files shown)


shai/src/main/scala/at/forsyte/apalache/shai/v1/CmdExecutorService.scala
```diff
@@ -99,7 +100,8 @@ class CmdExecutorService(logger: Logger) extends ZioCmdExecutor.ZCmdExecutor[ZEn
   }
 
   private def tlaModuleToJsonString(module: TlaModule): ujson.Value = {
-    ujson.Str(PrettyWriter.writeAsString(module))
+    val extendedModules = "Variants" :: TlaWriter.STANDARD_MODULES
```
bugarela, line 103: Is there a reason why this is only added on this specific method, and not directly to the `STANDARD_MODULES`? Is there a place where we don't want this to be extended?

shonfeder: I am not sure, which is why I didn't add it to the standard modules. I think the rationale previously was that not all writers of TLA+ for apalache will be expected to use variants, so that should be opt in rather than a required dependency. If you think it should be made a standard dependency, I can add it in tho. WDYT?

bugarela: I think it should either be a standard dependency or something we dinamically add if a variant is found in the spec.

Since Apalache is already a standar dependency, then it is safe to assume that the user has the proper library folder linked, and that same folder should have the Variants module. So I don't see a reason to not extend it and save the user the trouble of figuring this out alone.

bugarela (approved):

PR #2860, opened 2024-03-13
Sanitize identifiers when pretty printing

### LINKED ISSUE #2858: Pretty printing the apalache IR into TLA+ does not sanitize identifiers
Apalache supports 2 ways of constructing its IR that don't require providing TLA+ as input: it can deserialize the JSON representation of the IR and it can convert the JSON representation of Quint's IR. However, both of these input methods allow identifiers that are not valid TLA+ syntax. But we don't take account of the allowance in the `TlaWriters` that pretty print the IR into TLA+. As a result, we can easily produce TLA+ with invalid identifiers. Here are the known examples of invalid identifiers we've turned up after testing with https://github.com/informalsystems/apalache/pull/2852

- Identifiers starting with underscores that aren't followed by letters (like `__123`) are invalid.
- Apalache allows identifiers to include `::`, but this TLA+.

shonfeder:
Closes #2858

-------

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
@@ -89,8 +89,7 @@ class PrettyWriter(
     prettyWriteDoc(declToDoc(decl) <> line <> line)
   }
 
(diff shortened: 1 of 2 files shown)


bugarela (approved):
LGTM!

PR #2864, opened 2024-03-15
Translate quint variable initialization as equality, not assignment

### LINKED ISSUE #2863: Assignments in quint `init` operators  need to be unprimed
We are currently translating all cases of quint's `assign` operator, `_' = _` as `_' := _` in the apalache IR:

https://github.com/informalsystems/apalache/blob/efb30eaf2f1fc76249e1116ee72e659485483823/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala#L614

But in the init operator these should not be primed.

shonfeder:
Fixes #2863

Note that this presupposes #2860, and should be rebased on main after that is merged in.

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
@@ -89,8 +89,7 @@ class PrettyWriter(
     prettyWriteDoc(declToDoc(decl) <> line <> line)
   }
 
-  private def identity(x: String): String = x
-  def write(expr: TlaEx): Unit = prettyWriteDoc(exToDoc((0, 0), expr, identity))
+  def write(expr: TlaEx): Unit = prettyWriteDoc(exToDoc((0, 0), expr, sanatizeID))
 
   def writeComment(commentStr: String): Unit = {
     prettyWriteDoc(wrapWithComment(commentStr) <> line)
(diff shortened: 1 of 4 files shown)


shonfeder:
This isn't a full fix for #2863 yet: it only ensures primes are not added to assignments that occur within the body of the init predicate. For the general case, we need to ensure primes do not appear in any operator that is used by the init predicate.

We can solve that this way:

- A function `referencedDeclarationNames : QuintOpDef => List[String]` that gathers just the non-builtin operator names that are referenced within the body of the give operator definition.
- call `referencedDeclarationNames` on the init pred, then find all definitions with those names, and recurse on the results, until all results are exhausted (only empty lists returned)
- This should produce a list of all and only the operators that are used in the init pred, and it should be in topopological order.
- Set the `isInit` int the context (instead of doing it at https://github.com/informalsystems/apalache/blob/22b8270302da7acec0d3e4dbd5a7af619922d222/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala#L790-L792)
- Fold over those gathered operators with this context, using the same logic as https://github.com/informalsystems/apalache/blob/22b8270302da7acec0d3e4dbd5a7af619922d222/tla-io/src/main/scala/at/forsyte/apalache/io/quint/Quint.scala#L815-L817
- Finally, set `isInit` to `false`, and the continue to process the rest of the declarations

Some additional care will need to be take about where the init-relevant operators are placed back into the sequence of converted operators, in case there other operators that refer to them, which must occur later. Placing the init-relevant operators after all variable declarations but before any others should be safe, since they are guaranteed not to refer to other operators.

Alternatively, a different strategy could be employed on the quint side to remove the need for this complexity, as suggested on #2863.

Unfortunately, I won't have time to complete this work, as I am losing my company access to github etc. now.

<3

PR #2891, opened 2024-05-03
Fixes for compilation to TLA+

bugarela:
Hello :octocat: 

This includes 4 fixes for the Quint/JSON -> TLA+ compilation.
1. `slice` was incorrectly making an increment on the last argument. From the Quint manual:
```bluespec
replaceAt(l, start, end)
// Slice a list from `start` until `end`.
// Equivalent to SubSeq(lst, start + 1, end) in TLA+.
```
2. The produced TLA+ module was lacking type annotations required by Apalache. We now include them. 
3. Apalache transforms all operators given to folds in let in expressions. See: https://github.com/informalsystems/apalache/blob/9295b5b6cdc76b991e99a5546a6c62e62e7dfd39/tlair/src/main/scala/at/forsyte/apalache/tla/typecomp/unsafe/UnsafeApalacheBuilder.scala#L121-L129
However, this is not valid TLA+ and won't be parsed by SANY. To avoid touching Apalache internals, and as I can't really invest in forking the translation process right now, I'm simply fixing this with a regex replacement before printing. Not a great solution, but super low-risk as this only affects what is printed by the compilation command, which is only used by Quint.
4. We had some leftover `::` separators in our names, which is not parseable by SANY. They were coming from exists/forall expressions, so I'm making sure we also sanitize the names on those.

  - I couldn't add tests for (2) and (3). The setup we have for testing server commands doesn't have enough instrumentation, and I'd have to learn a lot of things in order to make it work with tests for this. 


tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -209,7 +209,7 @@ class PrettyWriter(
         val sign = PrettyWriter.bindingOps(op)
         val doc =
           group(
-              group(text(sign) <> space <> text(x.toString) <> space <>
+              group(text(sign) <> space <> text(sanatizeID(x.toString)) <> space <>
```
p-offtermatt, line 212: nit: should it be sanitize?

p-offtermatt (approved):
LGTM!

PR #2894, opened 2024-05-06
Remove cronjob for prepare-release workflow

bugarela:
Hello :octocat: 

We have this cronjob preparing a release every Monday, but it doesn't make sense anymore as there isn't a continuous flow of changes to this project currently. It is annoying because it creates a branch for the next release, and whenever I actually want to cut a release, the branch is already taken and it errors out - or something like that - and I have to remember to close the bot's PR and delete it's branch.

This simply removes that cronjob. We can easily add it back if it starts making sense again.


p-offtermatt (approved):
Good change for now! Maybe it needs to be added back in the future :)

PR #2927, opened 2024-08-08
Introduce the `Repeat` Apalache operator

Kukovec:

Extends the suite of Apalache operators, by adding `Repeat`, a repeated iteration operator. Semantically,
```tla
Repeat(Op, N, x) = ApaFoldSeqLeft(Op, x, MkSeq(N, LAMBDA i: i))
```
However, in the encoding, we can avoid constructing the `1` to `N` sequence explicitly, for an efficiency gain.


test/tla/RepeatBad.tla
```diff
@@ -0,0 +1,10 @@
+------------------------ MODULE RepeatBad ------------------------------------
```
konnov, line 1: Could you add a comment on why it is bad?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/RepeatRule.scala
```diff
@@ -0,0 +1,78 @@
...
+import at.forsyte.apalache.tla.lir._
+import at.forsyte.apalache.tla.lir.oper.ApalacheOper
+import at.forsyte.apalache.tla.lir.transformations.impl.IdleTracker
+import at.forsyte.apalache.tla.lir.transformations.standard.IncrementalRenaming
+import at.forsyte.apalache.tla.lir.values.TlaInt
+import at.forsyte.apalache.tla.pp.Inliner
+import at.forsyte.apalache.tla.types.tla
+
+/**
+ * Rewriting rule for Repeat. This rule is similar to [[FoldSeqRule]].
```
konnov, line 13: Please add a comment on why this operator is more efficient than `FoldSeqRule`

konnov (approved):
Looks good! Please add a few comments (see below). I believe that if we had a separate rewriting framework, we probably could just do that purely by rewriting.

PR #2936, opened 2024-08-14
add `--timeout-smt` to `check` and `simulate`

konnov:
In our experiments, it sometimes happens that `apalache-mc simulate` gets stuck in a randomly selected schedule, without progressing for many hours. Since we would like to get quicker feedback with `simulate`, this PR add the CLI option `--timeout-smt` that limits the time spent on a single SMT query. Also, we are removing the old tuning option `search.smt.timeout`, which is hard to discover.

### DIFF
--- a/docs/src/apalache/tuning.md
+++ b/docs/src/apalache/tuning.md
@@ -22,14 +22,6 @@ The following options are supported:
 `smt.randomSeed=<int>` passes the random seed to `z3` (via `z3`'s parameters
 `sat.random_seed` and `smt.random_seed`).
 
-##  Timeouts
-
-`search.smt.timeout=<seconds>` defines the timeout to the SMT solver in seconds.
-The default value is `0`, which stands for the unbounded timeout.  For instance,
-the timeout is used in the following cases: checking if a transition is enabled,
-checking an invariant, checking for deadlocks. If the solver times out, it
-reports 'UNKNOWN', and the model checker reports a runtime error.
-
 ## Invariant mode
 
 `search.invariant.mode=(before|after)` defines the moment when the invariant is
--- a/.unreleased/features/timeout-smt.md
+++ b/.unreleased/features/timeout-smt.md
@@ -1,0 +1,1 @@
+Add `--timeout-smt` to limit SMT queries (#2936)
--- a/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
+++ b/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
@@ -162,6 +162,8 @@ object Config {
    *   maximal number of Next steps
    * @param maxError
    *   whether to stop on the first error or to produce up to a given number of counterexamples
+   * @param timeoutSmtSec
+   *   the time limit on SMT queries in seconds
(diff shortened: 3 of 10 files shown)


docs/src/apalache/tuning.md
```diff
@@ -22,14 +22,6 @@ The following options are supported:
...
 `sat.random_seed` and `smt.random_seed`).
 
-##  Timeouts
-
-`search.smt.timeout=<seconds>` defines the timeout to the SMT solver in seconds.
-The default value is `0`, which stands for the unbounded timeout.  For instance,
-the timeout is used in the following cases: checking if a transition is enabled,
-checking an invariant, checking for deadlocks. If the solver times out, it
-reports 'UNKNOWN', and the model checker reports a runtime error.
-
```
Kukovec, line 32: Wouldn't it make more sense to move an explanation of the timeout flag somewhere, rather than deleting it outright?

Kukovec (approved):
LGTM

docs/src/apalache/tuning.md
konnov: Yes, this parameter is now visible via `check --help` and `simulate --help`

Kukovec:
As a question unrelated to the review, should we be supporting `timeout-smt` with `check`? Simulation is probabilistic anyway, so no real harm there, but setting a timeout may incorrectly give the impression that `check` succeeded, since the behavior is to assume invariants hold whenever timeouts happen.

konnov:
> As a question unrelated to the review, should we be supporting `timeout-smt` with `check`? Simulation is probabilistic anyway, so no real harm there, but setting a timeout may incorrectly give the impression that `check` succeeded, since the behavior is to assume invariants hold whenever timeouts happen.

`check --timeout-smt=n` finishes with a runtime error on timeout. Also, in case of `simulate`, the user gets a message that there were timeouts.

Kukovec:
> `check --timeout-smt=n` finishes with a runtime error on timeout. Also, in case of `simulate`, the user gets a message that there were timeouts.
I see a lot of `searchState.onResult(RuntimeError())` replaced with `searchState.onResult(SmtTimeout(...))`.
Is it this bit? https://github.com/apalache-mc/apalache/blob/f7aeb232442052916bca82b1708b8eb9bc594fa7/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala#L269-L274

konnov:
> > `check --timeout-smt=n` finishes with a runtime error on timeout. Also, in case of `simulate`, the user gets a message that there were timeouts.
> 
> I see a lot of `searchState.onResult(RuntimeError())` replaced with `searchState.onResult(SmtTimeout(...))`. Is it this bit?
> 
> https://github.com/apalache-mc/apalache/blob/f7aeb232442052916bca82b1708b8eb9bc594fa7/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala#L269-L274

Yeah. This is the clear distinction here. There are other cases where I am less certain. In any case, you will get a message that there were timeouts. I think we should play with this UX a bit and figure out what works the best.

PR #2943, opened 2024-08-15
Disable code coverage in CI

konnov:
Removing `test-coverage` from the CI workflows:

 - It slows down the build times
 - scoverage is not up-to-date, see [this build](https://github.com/apalache-mc/apalache/actions/runs/10401983974/job/28805552161?pr=2942)
 - I don't think anybody was looking at the coverage plots in years.

Without code coverage in tests:

image

With code coverage in tests:

image

The integration tests are still slower though, so the total build times are not changing.


konnov:
ah, actually, I cannot just yolo it :)

thpani (approved):

Issue #2932, opened 2024-08-13
Repair `docker-container` workflow


konnov:
This is broken due to moving the repository: https://github.com/apalache-mc/apalache/actions/workflows/container.yml

konnov:
It also does not work locally for me

konnov:
It was actually broken, probably, due to some changes in the `eclipse-temurin:17` docker image. See #2941

PR #2947, opened 2024-08-16
enable docker tests

### LINKED ISSUE #2931: Repair `docker-tests`
Due to the migration to `apalache-mc`, the workflow `docker-tests` broke. We should repair it.

konnov:
closes #2931. Since the docker container is now properly built #2932, the docker tests should be working again


thpani (approved):
🎉

Issue #2948, opened 2024-08-17
Bogus safety violation checking if a set is a subset of Nat


lemmy:
## Description

Bogus counterexample claiming that `{42} \in SUBSET Nat` is false.

## Input specification

```tla
------------------------- MODULE APASubRec ------------------------------           
EXTENDS Integers

VARIABLE
    \* @type: Set(Int);
    v

TypeOK ==
    v \in SUBSET Nat  \*  No bogus counterexample if {42} or Int is substituted for Nat.

Init ==
    v = {42}

Next ==
    UNCHANGED v
========================================================================== 
```

## The command line parameters used to run the tool

`apalache-mc check --init=Init --next=Next --inv=TypeOK APASubRec.tla`

<details>
 <summary>Apalache output</summary>

```
-> % apalache-0.44.11/bin/apalache-mc check --init=Init --next=Next --inv=TypeOK APASubRec.tla
[...]

PR #2960, opened 2024-08-22
Bogus safety violation checking if a set is a subset of Nat.

lemmy:
Fixes Github issue #2948
https://github.com/apalache-mc/apalache/issues/2948


lemmy:
5724d0d causes Apalache to report a proper error message: `SetInRule is not implemented for type PowSet[Set(Int)] (found in {1, 2, 3, 4} ∈ SUBSET Nat)`.

lemmy:
@thpani @konnov https://github.com/apalache-mc/apalache/commit/c69c355877ea306e38eaf53212ba779efa14318b and https://github.com/apalache-mc/apalache/commit/830caa71596caf488b7717b39ba2a4e535ddb779 cause Apalache to no longer hit this code path for the given inputs (the expressions are rewritten elsewhere). Would you like to add the safeguard below anyway?  If you do, I'll update the tests to expect the corresponding error message.

https://github.com/apalache-mc/apalache/pull/2960/files#diff-455ee0089c33ede12710ae233adedc17473d8ac7cc6a2ca7246f127a7482512aR58-R60

lemmy:
@konnov This PR does not aim to extend the encoding. Instead, it ensures that Apalache throws an error when an infinite set is encountered in a context where it isn't properly handled.

konnov:
Ah, sorry, I missed that. Yes, this makes sense  to add. Thanks a lot!

konnov (approved):
A safeguard would definitely help!

lemmy:
This is ready for review and merging. Although commits c69c355877ea306e38eaf53212ba779efa14318b and 830caa71596caf488b7717b39ba2a4e535ddb779 prevent any of the three encodings from reporting a bogus invariant violation for expressions like `{1, 2, 3} \in SUBSET Nat` (due to rewriting), this update ensures that Apalache generates a clear error message when handling infinite sets, should the rewrites ever be removed in the future.

konnov (approved):
This makes sense! Thanks for adding the safeguards

konnov:
I cannot add an entry in `.unchanged`, since it's your PR. Could you add one?

lemmy:
What should that entry be?

konnov:
> What should that entry be?

Oh, just one line describing the nature of the change, see https://github.com/apalache-mc/apalache/blob/main/CONTRIBUTING.md#how-to-record-a-change

I guess, it should technically go into [bug-fixes](https://github.com/apalache-mc/apalache/tree/main/.unreleased/bug-fixes).

lemmy:
done

thpani (approved):
LGTM

PR #2966, opened 2024-08-23
Fix broken links globally

konnov:
This is an automatic replacement of `apalache.informal.systems` with `apalache-mc.org`


thpani (approved):
lgtm

PR #2969, opened 2024-08-25
improve the error message about expanding powersets

konnov:
This PR simply improves error reporting for the case when `SUBSET _` has to be expanded to a very large set:
 
- [x] Reporting a normal error `RewriterKnownLimitationError` instead of `RewriterException`, which would ask the user to file a bug report
- [x] Printing a few hints about fixing the issue

Currently, the source location of the problematic code is not properly identified, due to multiple iterations of the rewriter. We could try to recover the source location when unwinding the rewriting stack. However, this would require a more careful refactoring. 

### DIFF
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/PowSetCtor.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/PowSetCtor.scala
@@ -40,15 +40,15 @@ class PowSetCtor(rewriter: SymbStateRewriter) extends LazyLogging {
     }
 
     rewriter.solverContext.log("; %s(%s) {".format(getClass.getSimpleName, state.ex))
-    val powSetSize = BigInt(1) << elems.size
-    if (powSetSize > Limits.POWSET_MAX_SIZE) {
-      logger.error("This error typically occurs when one enumerates all subsets:"
+    if (elems.size > Limits.POWSET_MAX_BASE_SIZE) {
+      logger.error("This error typically occurs when one enumerates all subsets of a set:"
         + " \\A S \\in SUBSET T: P or \\E S \\in SUBSET T: P")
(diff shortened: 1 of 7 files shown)


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/PowSetCtor.scala
```diff
@@ -40,8 +41,13 @@ class PowSetCtor(rewriter: SymbStateRewriter) {
...
     rewriter.solverContext.log("; %s(%s) {".format(getClass.getSimpleName, state.ex))
     val powSetSize = BigInt(1) << elems.size
-    if (powSetSize >= Limits.POWSET_MAX_SIZE) {
-      throw new RewriterException(s"Attempted to expand a powerset of size 2^${elems.size}", state.ex)
+    if (powSetSize > Limits.POWSET_MAX_SIZE) {
+      logger.error("This error typically occurs when one enumerates all subsets:"
+        + " \\A S \\in SUBSET T: P or \\E S \\in SUBSET T: P")
+      logger.error("Try to decrease the cardinality of the base set T, or avoid enumeration.")
+      val msg =
+        s"Attempted to expand SUBSET of size 2^${elems.size}, whereas the built-in limit is ${Limits.POWSET_MAX_SIZE}"
```
thpani, line 49: Printing 1048576 here may look confusing to the user.

Why not store `Limits.POWSET_MAX_SIZE` as the number of elements?
Then you can directly compare `elems.size` above and print it here as a power of 2.

thpani (approved):
LGTM, although I don't understand why we keep the bound as the size of the expanded set, see below

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/aux/PowSetCtor.scala
konnov: That's true. I will fix that :)

PR #2974, opened 2024-08-28
update the landing page

konnov:
The Apalache landing page is pretty much outdated. This PR aims at updating the most important information on the landing page. Since Informal Systems is no longer supporting development of Apalache, they receive our thanks on the landing page and in [FUNDING.md](./FUNDING.md), but they are now removed from the sponsors. Also, I have added my profile to `FUNDING.yml`. If the other main contributors (e.g., @Kukovec and @thpani) wish to add their Github handles, we should definitely do that!


bugarela (approved):
LGTM! Just to clarify: Informal Systems and Vienna Business Agency were removed from sponsorship links, which, in my understanding, means Apalache shouldn't receive sponsorships through them. If anyone want's to fund Apalache work, they should sponsor the maintainers directly, which I think makes perfect sense!

PR #2990, opened 2024-09-18
Propagate z3 parameters in the tuning options

### LINKED ISSUE #2315: Use Z3 in parallel mode
## Is your feature request related to a problem? Please describe

It would be nice to expose some z3 configuration at Apalache CLI.

For one, I want to run z3 in parallel mode. This can be done via setting [`parallel.enable` to `true`](https://theory.stanford.edu/~nikolaj/programmingz3.html#sec-parallel-z3).

## Describe the solution you'd like

This can be exposed via `--tuning-options`.

Eg. `apalache-mc simulate --tuning-options=search.solver.z3.parallel.enable=true ...`

## Describe the impact on your work

Faster z3 search on multi-core machines.

konnov:
Closes #2315.
This PR adds the ability to pass [z3 tuning parameters](https://microsoft.github.io/z3guide/programming/Parameters/#sat) right in the Apalache fine tuning parameters. This is a feature that we should have implemented long time ago.

The parameters are now parsed in `FineTuningParser`. Initially, I tried to use [pureconfig](https://pureconfig.github.io/). However, it is too opinionated and it needs more workarounds than it helps. Writing a trivial parameter parser is much easier than fiddling with pureconfig, e.g., propagating product hints via implicits.


docs/src/apalache/tuning.md
```diff
@@ -91,6 +86,47 @@ steps.
...
+Technically, Apalache propagates all the parameters that start with `z3.` to Z3.
+However, some of these parameters fail in the solver, even when they work in the
+command line. Hence, you have to experiment with the choice of parameters.
+
+## Randomization
+
+**DEPRECATED:** You can now pass the seeds directly to Z3 via [Z3 Tuning Parameters](#z3-tuning-parameters).
+
+`smt.randomSeed=<int>` passes the random seed to `z3` (via `z3`'s parameters
+`sat.random_seed` and `smt.random_seed`).
```
thpani, line 128: `smt.randomSeed` propagates to more than just `sat.random_seed` and `smt.random_seed`:

https://github.com/apalache-mc/apalache/blob/8630a60cb48b3c6f1d88c8e43f470182edae083d/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala#L1029-L1035

I don't think we should deprecate this.

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
```diff
@@ -24,21 +24,23 @@ class LogbackConfigurator(runDir: Option[Path], customRunDir: Option[Path]) exte
     setContext(loggerContext)
     runDir match {
       case Some(_) => configure(loggerContext)
-      case None    => configureSilent(loggerContext)
+      case None    => configureConsoleOnly(loggerContext)
     }
   }
 
-  def configureSilent(loggerContext: LoggerContext): Unit = {
+  def configureConsoleOnly(loggerContext: LoggerContext): Unit = {
```
thpani, line 31: AFAIU, this only logs at and above `WARN`?

I would prefer to reflect this in the name, something like

```suggestion
  def configureConsoleOnlyWarn(loggerContext: LoggerContext): Unit = {
```

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
```diff
@@ -24,21 +24,23 @@ class LogbackConfigurator(runDir: Option[Path], customRunDir: Option[Path]) exte
     setContext(loggerContext)
     runDir match {
       case Some(_) => configure(loggerContext)
-      case None    => configureSilent(loggerContext)
+      case None    => configureConsoleOnly(loggerContext)
```
thpani, line 27: Are you sure this doesn't mess with tests?

Where else do we end up here without a `runDir`?

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
```diff
@@ -82,6 +82,8 @@ object Tool extends LazyLogging {
    *   the exit code; as usual, 0 means success.
    */
   def run(args: Array[String]): Int = {
+    // Configure the silent logger first. Otherwise, Apache Commons spills a lot of text to the console.
+    new LogbackConfigurator(None, None).configureDefaultContext()
```
thpani, line 86: Why do we need this? Why wasn't it needed before?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
```diff
@@ -75,7 +76,12 @@ class BoundedCheckerPassImpl @Inject() (
 
     val smtProfile = options.common.smtprof
     val smtRandomSeed = tuning.getOrElse("smt.randomSeed", "0").toInt
-    val solverConfig = SolverConfig(debug, smtProfile, smtRandomSeed, smtEncoding)
+    // Parse the tuning parameters that are relevant to Z3. Some passes add more fields, which cannot be parsed.
```
thpani, line 79: What does this mean: `Some passes add more fields, which cannot be parsed.` ?

thpani (approved):
Generally LGTM, I left a few comments

docs/src/apalache/tuning.md
konnov: Yep, I was not sure about this either. The only issue is that some of the above settings conflict with more restricted logics

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
konnov: It does not mess with the tests. This setup is only needed before the actual configuration is read. If something happens when reading the configuration, e.g., parsing the tuning parameters fails, we should be able to see the error messages.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
konnov: I think we did not have any behavior in the tests that would require error messages when parsing the input configuration. In other words, potential configuration parsing issues were silently ignored.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
konnov: What I noticed is that `tuning` as a map contains some options that were not present in the passed options file. I am not sure how exactly configuration options are processed. Still on my todo list.

PR #3005, opened 2024-09-27
produce true on distinct for less than 2 elements

### LINKED ISSUE #2964: Do not produce distinct on singleton
This is a minor annoyance. Apalache produces `(distinct ..)` even for singletons. This is what `cvc5` complains about. We should simply fix it, so it would be possible to repay the SMT log with `cvc5`.

konnov:
Closes #2964. Simply translate `distinst` into `true`, when we have less than two subexpressions. It's only relevant for replaying SMT logs in CVC5, but it was sufficiently annoying for me to fix it.

### DIFF
--- a/.unreleased/bug-fixes/distinct-on-singleton.md
+++ b/.unreleased/bug-fixes/distinct-on-singleton.md
@@ -1,1 +1,1 @@
-Do not produce `(distinct ...)` for singletons (see #3005)
+Do not produce `(distinct ...)` for singletons, see #3005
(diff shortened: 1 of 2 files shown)


.unreleased/bug-fixes/distinct-on-singleton.md
```diff
@@ -0,0 +1 @@
+Do not produce `(distinct ...)` for singletons (see #3005)
```
thpani, line 1: I think we always used this style

```suggestion
Do not produce `(distinct ...)` for singletons, see #3005
```

thpani (approved):
Excellent! We should appease the SMT-LIB gods :D

PR #3006, opened 2024-09-27
refactor: remove repetitive folds in `Z3Context`

konnov:
I have noticed a few repetitive occurrences of summing up longs. Refactored that to a single call.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -767,26 +767,17 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext with LazyL
       case OperEx(ApalacheInternalOper.distinct, args @ _*) =>
         val (es, ns) = (args.map(toExpr)).unzip
         val distinct = z3context.mkDistinct(es: _*)
-        (distinct.asInstanceOf[ExprSort],
-            ns.foldLeft(1L) {
-              _ + _
-            })
+        (distinct.asInstanceOf[ExprSort], sumExprCount(ns))
```
thpani, line 770:
```suggestion
        (distinct.asInstanceOf[ExprSort], ns.sum + 1)
```

thpani (commented):
Would it be possible / easier to use [`Seq.sum`](https://www.scala-lang.org/api/2.12.2/scala/collection/immutable/Seq.html#sum:A)?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
konnov: Oh, I have missed that `Seq.sum` existed. That's definitely a better alternative!

konnov:
> Would it be possible / easier to use [`Seq.sum`](https://www.scala-lang.org/api/2.12.2/scala/collection/immutable/Seq.html#sum:A)?

Fixed that. Please review

thpani (approved):

PR #3028, opened 2024-10-31
update variants.md to new alias syntax

nano-o:


konnov (approved):
Thanks a lot for fixing this!

PR #3041, opened 2024-11-29
Quint -> TLA+ transpilation fixes

bugarela:
Hello :octocat: 

I was running into some issues when using Apalache to transpile Quint into TLA+. There were even parsing errors with invalid parenthesis. After much debugging, I got to fix all of the ones I know of, except for the assignments in init (which are not a big deal because there's only one init per spec so it's always easy to manually fix those).

Here are the issues:

1. Whenever we had a default eliminator on a match, we would produce something like `QUINT_LAMBDA1([])`, but `[]` is not a valid TLA+ expression. I've replaced it with `{}`.
2. When accessing a map (function) with lists as domains, we need to do something like `map[<<1, 2>>]` but the pretty printer was removing the tuple wrapper on this and producing `map[1, 2]` instead which is not right.
  - I don't know why this existed. On the test, seems like it could have something to do with recursive operators, which as I understand should have been removed. It was also introduced in a recursive-operator related PR: https://github.com/apalache-mc/apalache/pull/312/files#diff-9f17d2b8582a6882dd7eb194ab597fa265a8f5423e02b724b767af7a3c444a01
3. For Quint's unit type, we were producing "U_OF_UNIT" and this was causing a TLC-specific error: TLC tried to compare some value like `[ "tag" |-> "Some", value |-> 1 ]` with `"U_OF_UNIT"` for its fingerprint generation and it failed at comparing. Replacing `"U_OF_UNIT"` with `[ "tag" |-> "UNIT" ]` fixed this problem, so that's what I did.
4. Lastly, the most complicated one: applying operators to LET-IN arguments doesn't work on TLC, and the Apalache representation transforms a lot of things into this format. I didn't want to touch anything that could affect Apalache's internals, so I only changed how this things get printed. Something that has this structure on Apalache IR:
```tla
Foo(LET A(param1, param2) == param1 + param2 IN A(1, 2))
```

will be printed as:
```tla
LET A(param1, param2) == param1 + param2 IN Foo(A(1, 2))
```

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
@@ -465,14 +465,17 @@ class PrettyWriter(
         wrapWithParen(parentPrecedence, op.precedence, group(doc))
 
       case OperEx(op @ TlaOper.apply, NameEx(name), args @ _*) =>
+        val (decls, newArgs) = extractDecls(args)
+
         // apply an operator by its name, e.g., F(x)
-        val argDocs = args.map(exToDoc(op.precedence, _, nameResolver)).toList
+        val argDocs = newArgs.map(exToDoc(op.precedence, _, nameResolver)).toList
         val commaSeparated = ssep(argDocs, "," <> softline)
+
         val doc =
           if (args.isEmpty) {
             text(parseableName(name))
           } else {
-            group(parseableName(name) <> parens(commaSeparated))
+            group(ssep(decls, line) <> line <> parseableName(name) <> parens(commaSeparated))
           }
 
         wrapWithParen(parentPrecedence, op.precedence, doc)
@@ -486,16 +489,18 @@ class PrettyWriter(
         wrapWithParen(parentPrecedence, op.precedence, doc)
 
       case OperEx(op, args @ _*) =>
-        val argDocs = args.map(exToDoc(op.precedence, _, nameResolver)).toList
+        val (decls, newArgs) = extractDecls(args)
+        val argDocs = newArgs.map(exToDoc(op.precedence, _, nameResolver)).toList
         val commaSeparated = ssep(argDocs, "," <> softline)
         // The standard operators may contain a '!' that refers to the standard module. Remove it.
         val lastIndexOfBang = op.name.lastIndexOf("!")
         val unqualifiedName = if (lastIndexOfBang < 0) op.name else (op.name.substring(lastIndexOfBang + 1))
+
         val doc =
           if (args.isEmpty) {
             text(unqualifiedName)
           } else {
-            group(unqualifiedName <> parens(commaSeparated))
+            group(ssep(decls, line) <> line <> unqualifiedName <> parens(commaSeparated))
           }
 
         wrapWithParen(parentPrecedence, op.precedence, doc)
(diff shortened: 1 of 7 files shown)


tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -297,14 +297,6 @@ class PrettyWriter(
...
         ) ///
 
-      // a function of multiple arguments that are packed into a tuple: don't print the angular brackets <<...>>
-      case OperEx(op @ TlaFunOper.app, funEx, OperEx(TlaFunOper.tuple, args @ _*)) =>
-        val argDocs = args.map(exToDoc(op.precedence, _, nameResolver))
-        val commaSeparatedArgs = folddoc(argDocs.toList, _ <> text(",") <@> _)
-        group(
-            exToDoc(TlaFunOper.app.precedence, funEx, nameResolver) <> brackets(commaSeparatedArgs)
-        ) ///
-
```
konnov, line 307: Why do we have to remove this? In TLA+, `f[x, y]` is a short-form for `f[<<x, y>>]`. Look at this example: https://gist.github.com/konnov/ce0ca9071dbe75b936e96dbd99224f0d

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -473,14 +465,19 @@ class PrettyWriter(
         wrapWithParen(parentPrecedence, op.precedence, group(doc))
 
       case OperEx(op @ TlaOper.apply, NameEx(name), args @ _*) =>
+        val (decls, newArgs) = extractDecls(args)
```
konnov, line 468: Do you have an example of what does not work exactly? For instance, this one works in TLC: https://gist.github.com/konnov/e7df8586a9865027dc4abfc07188e060

konnov (commented):
This is a nice set of fixes. Thanks a lot! I have a question about two of the changes, trying to pinpoint the cases where it is actually required.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
bugarela: It does work like that for tuples, but not for sequences:
```tla
--------------------------------- MODULE test ---------------------------------
EXTENDS Integers, Sequences

VARIABLES
  f

Init ==
  f = [ l \in {<<1>>, <<>>, <<1, 2>>} |-> Len(l) ]

Next ==
  UNCHANGED f

Inv ==
  f[<<1>>] = 1

Inv2 ==
  f[1] = 1

```

TLC fails to check Inv2 here: 
```
Error: Attempted to check equality of the function << >> with the value:
1
While working on the initial state:
f = (<<>> :> 0 @@ <<1>> :> 1 @@ <<1, 2>> :> 2)

Error: TLC was unable to fingerprint.
```

I believe there is no way to disambiguate between tuples and sequences at this point, so the only option was to remove it completely.

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
```diff
@@ -527,6 +528,25 @@ class PrettyWriter(
     }
   }
 
+  /**
+   * On TLA+ files, we can't have LET..IN expressions as arguments. This is a helper function to extract the
+   * declarations so they can be printed before the operator application that would use them as arguments.
```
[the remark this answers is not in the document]
bugarela: Not sure if I understand the question. I'm saying this one works:
```tla
LET A(x, y) == x + y IN F(A)
```

while this one doesn't:
```tla
F(LET A(x, y) == x + y IN A)
```

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
konnov: Right. This is because operators are not values in TLA+

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
konnov: Interesting. I think the short-form syntax only applies to pairs, triples, etc.

konnov: Yeah, these invariants work for your example:

```tla
Inv ==
  f[<<1, 2>>] = 2

Inv2 ==
  f[1, 2] = 2
```

konnov (approved):
After the clarifications, this looks good to merge!

tla-io/src/main/scala/at/forsyte/apalache/io/lir/PrettyWriter.scala
bugarela: Ah interesting. Yeah, the problem I hit was on a list of one element (from the hashes you wrote in ICS23 btw).

PR #3044, opened 2024-12-10
Fix conversion of Quint nullary polymorphic operators

bugarela:
Hello :octocat: 

Here's another fix for the conversion from Quint. I was hitting problems whenever `None` was used to construct values of different types inside the same operator. Upon debugging, I realized this was specific to nullary operators (such as `None`), and that the builder itself enforces that the same name has the same type in a same scope. A concrete example:

```bluespec
pure def f(x) = 
  pure val a = if (x > 2) Some(x) else None
  pure val b = if (x < 5) Some(true) else None

  (a, b)
```

One solution could be to use a polymorphic type (a type variable) for constructing the `None` name in TLA+, but that is too general and causes issues later on. So, instead, I wrote a new builder method `polymorphicName` that bypasses this check, and used that to construct names for nullary operators.

This is only a problem currently for nullary operators, as regular operators go through the application translation (not the name expression translation) and are dealt with differently.

### DIFF
--- a/.unreleased/bug-fixes/quint-nullary-operators.md
+++ b/.unreleased/bug-fixes/quint-nullary-operators.md
@@ -1,0 +1,1 @@
+Fix a problem when translating Quint nullary operators used polymorphically, see #3044
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -54,7 +54,7 @@ object Dependencies {
     // Libraries
     val junit = "junit" % "junit" % "4.13.2" % Test
     val scalacheck = "org.scalacheck" %% "scalacheck" % "1.18.1" % Test
-    val easymock = "org.easymock" % "easymock" % "5.4.0" % Test
+    val easymock = "org.easymock" % "easymock" % "5.5.0" % Test
 
     val scalaTestVersion = "3.2.15"
     val scalatest = "org.scalatest" %% "scalatest" % scalaTestVersion % Test
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -5,7 +5,7 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "1.2.0")
 // https://github.com/marcuslonnberg/sbt-docker
 addSbtPlugin("se.marcuslonnberg" % "sbt-docker" % "1.11.0")
 // https://github.com/scoverage/sbt-scoverage
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.2.1")
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.2.2")
 // https://github.com/sbt/sbt-buildinfo
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.13.1")
 // https://github.com/sbt/sbt-native-packager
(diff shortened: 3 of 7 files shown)


bugarela:
Although my fix is sufficient to get over the conversion phase and produce valid TLA+, I see now that it is not sufficient to make snowcat happy:
```
error: internal error: while type checking in Apalache:
'unknown location,__QUINT_LAMBDA111's type annotation ((None({ tag: Str }) | Some(a36), a36) => None({ tag: Str }) | Some(a29)) is too general, inferred: ((None({ tag: Str }) | Some(a29), a29) => None({ tag: Str }) | Some(a29))'
Please report an issue: https://github.com/informalsystems/quint/issues/new
```

So more to investigate. This PR still unblocks things, so might be worth moving foward with it at this point. I'm not sure.

konnov:
Yeah, the behavior of `None` is annoying in Quint. What I don't see is why this should be resolved at the level of TLA+. I thought that the Quint typechecker would assign proper types to the different instances of `None`. If this is the case, I believe that instead of propagating polymorphism to Apalache, I would rather produce different names for the different instances of `None` that carry different types. Otherwise, Apalache may not have sufficient context to infer the proper types.

konnov:
I guess, this is the related issue: https://github.com/informalsystems/quint/issues/1451

bugarela:
> I thought that the Quint typechecker would assign proper types to the different instances of None. If this is the case, 

it is

> I believe that instead of propagating polymorphism to Apalache, I would rather produce different names for the different instances of None that carry different types. Otherwise, Apalache may not have sufficient context to infer the proper types.

Hmmm but Apalache doesn't need to infer the types, as we annotate everything. However, in the end, it complains that our annotation is too general. I'll need to further debug to understand if it is actually too general and Quint has to provide a more specific type, or if Apalache is expecting something too strict.

Regarding the issue, I just confirmed that this PR fixed it. That spec works both with `quint compile` and `quint verify`. So that's more confirmation for me that this PR is worth moving foward with and that the "too general" issue is another problem to solve (probably the same as in https://github.com/informalsystems/quint/issues/1398)

konnov:
The most important question for me is whether there was a reason for not allowing something like `(F: Int)(...)` and `(F: Bool)(...)` in the builder, unless `F` is defined as polymorphic. Perhaps, @Kukovec could help us to understand it.

Kukovec:
> The most important question for me is whether there was a reason for not allowing something like `(F: Int)(...)` and `(F: Bool)(...)` in the builder, unless `F` is defined as polymorphic. Perhaps, @Kukovec could help us to understand it.

I mean, the reason it's not allowed is simple, the basic notion of socpe-safety is that one name is only allowed to have one type in a given scope. For polymorphic operators, `F()` may have types `Bool` and `Int` in different instances, but `F` should have `() => t` in both. 
Please don't add methods to the specifically scope-safe builder that makes it explicitly scope-unsafe, just use the unsafe builder.

konnov:
Right! If we expect the types to be correct, we can just use `UnsafeBuilder`.

bugarela:
Ok, thanks for the input, it makes sense! Took me some time to figure out how to do it as I couldn't use `point` in `Quint.scala`. I ended up having to change the imports in unexpected ways, but it worked out in the end.

konnov (approved):
Looks great! Thanks for fixing it and going the extra mile in using UnsafeBuilder!

konnov:
I will merge it right away to a cut a fresh release

PR #3046, opened 2024-12-13
protect against SANY race conditions on the filesystem

konnov:
Although rare, it may still happen that multiple CLI instances of Apalache hit concurrency issues that is filed in https://github.com/tlaplus/tlaplus/issues/688 (`apalache-mc server` has a separate mutex for that):

```
$ cat t.tla
------------------- MODULE t --------------------                               
EXTENDS Integers
VARIABLE
    \* @type: Int;
    x
Init == x = 0
Next == x' = x + 1
=================================================
$ seq 1 100 | parallel ~/devl/apalache/bin/apalache-mc parse --out-dir=c/{} t.tla
...
Encountered an exception while attempt to validate /var/folders/93/s11lf4f16c77bpq91r90njbm0000gn/T/Naturals.tla - /var/folders/93/s11lf4f16c77bpq91r90njbm0000gn/T/Naturals.tla (No such file or directory)
```

This is a simple fix that was already discussed in https://github.com/tlaplus/tlaplus/issues/688. We simply set `java.io.tmpdir` to a specific directory.

Actually, it seems like we have fixed it in #1959, but the above test shows otherwise.

### DIFF
--- a/src/universal/bin/apalache-mc
+++ b/src/universal/bin/apalache-mc
@@ -39,7 +39,11 @@ then
 fi
 
 # Avoid SANY concurrency issues: https://github.com/tlaplus/tlaplus/issues/688
-TD=`mktemp -d "$TMPDIR"/SANYXXXXXXXXXX`
+if [ -z "$TMPDIR" ]; then
+    TMPDIR="$(pwd)/tmp"
+    mkdir -p "$TMPDIR"
+fi
+TD=`mktemp -d -p $TMPDIR -t SANYXXXXXXXXXX`
 JVM_ARGS="${JVM_ARGS} -Djava.io.tmpdir=$TD"
 
 # Check whether the CLI args contains the debug flag
--- a/src/universal/bin/apalache-mc
+++ b/src/universal/bin/apalache-mc
@@ -39,7 +39,7 @@ then
 fi
 
 # Avoid SANY concurrency issues: https://github.com/tlaplus/tlaplus/issues/688
-if [ -z "$TMPDIR" ]; then
+if [ -z "${TMPDIR:-}" ]; then
     TMPDIR="$(pwd)/tmp"
     mkdir -p "$TMPDIR"
 fi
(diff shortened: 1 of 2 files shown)


Kukovec (approved):

src/universal/bin/apalache-mc
```diff
@@ -38,16 +38,24 @@ then
     JVM_ARGS="${JVM_ARGS} -Xmx4096m"
 fi
 
+# Avoid SANY concurrency issues: https://github.com/tlaplus/tlaplus/issues/688
+if [ -z "$TMPDIR" ]; then
+    TMPDIR="$(pwd)/tmp"
+    mkdir -p "$TMPDIR"
+fi
+JAVA_IO_TMPDIR=`mktemp -d -p $TMPDIR -t SANYXXXXXXXXXX`
```
thpani, line 46: Why not skip `-p` (and the above) and rely on it's default behavior:

```
-p temp-dir, --tmpdir=temp-dir
              temp  directory  for  the  file.  This option is a member of the
              tmpdir class of options.

              If this option is not provided, mktemp will use the  environment
              variable  TMPDIR to find a suitable directory.  If these are not
              available, it will fall back to ~/tmp  or  /tmp.   A  <file-pat>
              command line argument containing a directory component will con-
              flict with this option.
```

thpani (approved):
I have one question (see above), so I've disabled auto-merge.

But in general, LGTM

src/universal/bin/apalache-mc
konnov: For some reason, it does not work with docker this way. Maybe it's using nix and it has different behavior. But it does not even work with my fix.

konnov: The above test was failing due to extra protection flags, now using `${TMPDIR:-}`. This works. Let me see how it works without `-p`. It definitely did not work without the additional test for `TMPDIR` above

PR #3094, opened 2025-05-08
Add an inductive invariant for two-phase commit from the earlier experiments

konnov:
This PR adds the inductive invariant for two-phase commit. This inductive invariant was available in the early [apalache tests](https://github.com/apalache-mc/apalache-tests/blob/master/performance/two-phase/APATwoPhase.tla). Adding it to the main tests, to keep it for the future reference.

Also, fixing the checkout version in the github workflow.


thpani (approved):
Looks great!

PR #3120, opened 2025-07-28
Update 015adr-trace.md

konnov:
Add references to the existing implementations of ADR15, see the update to the markdown document.

### DIFF
--- a/docs/src/adr/015adr-trace.md
+++ b/docs/src/adr/015adr-trace.md
@@ -31,7 +31,11 @@ see [Consequences](#consequences).
 ## Implementations
 
  - [itf-py](https://github.com/konnov/itf-py): a Python library for (de-)serializing ITF traces
- - [itf-rs](https://github.com/informalsystems/itf-rs): a Rust library for (de-)serializing ITF traces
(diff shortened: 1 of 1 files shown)


docs/src/adr/015adr-trace.md
```diff
@@ -28,6 +28,13 @@ TLA+ and can be easily integrated into other tools.
 The use of JSON numbers is not allowed anymore. This simplifies custom parsers for ITF,
 see [Consequences](#consequences).
 
+## Implementations
+
+ - [itf-py](https://github.com/konnov/itf-py): a Python library for (de-)serializing ITF traces
+ - [itf-rs](https://github.com/informalsystems/itf-rs): a Rust library for (de-)serializing ITF traces
```
thpani: This could be mentioned as a TypeScript version?

https://github.com/informalsystems/vscode-itf-trace-viewer/blob/main/src/Trace.ts

thpani (approved):

docs/src/adr/015adr-trace.md
konnov: Good point!

konnov: Added links to the ITF trace viewer and its parser

PR #3121, opened 2025-07-30
Migrate to Sonatype Central Publisher Portal

konnov:
Since the currently used OSS Sonatype servers are [setting down](https://central.sonatype.org/news/20250326_ossrh_sunset/), we have to migrate to [Central Publisher Portal](https://central.sonatype.org/publish/publish-portal-guide/).

This PR:

 - migrates to Central Publisher Portal, this is easy, just use `Resolver.sonatypeCentralSnapshots`
 - pulls `tla2tools` directly from GitHub:

```scala
    val tla2tools = "org.lamport" % "tla2tools" % "1.7.4" from
      "https://github.com/tlaplus/tlaplus/releases/download/v1.7.4/tla2tools.jar"
```


thpani (approved):
👍

PR #3122, opened 2025-07-31
wip: experimental json-rpc server

konnov:
This is in-progress work on implementing a simple JSON-RPC server, which would let us experiment with other exploration strategies outside of Apalache codebase.

### DIFF
--- a/json-rpc/README.md
+++ b/json-rpc/README.md
@@ -1,0 +1,4 @@
+# json-rpc
+
+A simple JSON-RPC server module for Apalache, providing a health-check endpoint.
+
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -48,6 +48,11 @@ object Dependencies {
     // see https://scalapb.github.io/docs/faq/#i-am-getting-import-was-not-found-or-had-errors
     val scalapbRuntime =
       "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf"
+    // json-rpc dependencies
+    val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.19.2"
+    //val jacksonModuleScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.19.0"
+    val jettyServer = "org.eclipse.jetty" % "jetty-server" % "12.0.23"
+    val jettyServlet = "org.eclipse.jetty" % "jetty-servlet" % "11.0.25"
   }
 
   // Test only depenendencies
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -49,9 +49,10 @@ object Dependencies {
     val scalapbRuntime =
       "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf"
     // json-rpc dependencies
+    val jakartaServlet = "jakarta.servlet" % "jakarta.servlet-api" % "6.0.0" % Provided
     val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.19.2"
-    //val jacksonModuleScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.19.0"
-    val jettyServer = "org.eclipse.jetty" % "jetty-server" % "12.0.23"
+    val jacksonModuleScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.19.2"
+    val jettyServer = "org.eclipse.jetty" % "jetty-server" % "11.0.25"
     val jettyServlet = "org.eclipse.jetty" % "jetty-servlet" % "11.0.25"
   }
 
--- a/build.sbt
+++ b/build.sbt
@@ -238,6 +238,12 @@ lazy val shai = (project in file("shai"))
       ),
   )
 
+lazy val json_rpc = (project in file("json-rpc"))
+  .dependsOn(tlair, infra, tla_io, tla_typechecker, tla_bmcmt, passes)
+  .settings(
+      // Add any specific settings or dependencies for json-rpc here
+  )
+
 lazy val tool = (project in file("mod-tool"))
   .dependsOn(tlair, tla_io, tla_assignments, tla_typechecker, tla_bmcmt, shai, passes)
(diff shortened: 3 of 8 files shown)


json-rpc/README.md
```diff
@@ -0,0 +1,137 @@
...
+See the [JSON-RPC specification][] for more details. It is real short.
+
+This is work in progress. More methods to be added in the future.
+
+### 2.1. Method loadSpec
+
+Load a TLA+ specification from a list of base64-encoded source files.
+The server will parse the specification and store all the necessary
+objects in memory for further exploration. This includes the SMT solver,
+which may consume a lot of memory.
```
thpani: I'm a bit confused by the last sentence.
Surely the SMT context doesn't use (much) memory yet, when you just load a spec?

json-rpc/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+**Example**:
+
+```sh
+curl -X POST http://localhost:8822/rpc \
+  -H "Content-Type: application/json" \
+  -d '{"jsonrpc":"2.0","method":"loadSpec","params":{"sources": [
+    "LS0tLSBNT0RVTEUgSW5jIC0tLS0KRVhURU5EUyBJbnRlZ2VycwpWQVJJQUJMRSB4CkluaXQgPT0geCA9IDAKTmV4dCA9PSB4JyA9IHggKyAxCj09PT09PT09PT09PT09PT09PT09PQogICAgICA="
+    ]},"id":1}'
```
thpani: Can we change the example to use `echo "<tlastring>" | base64` instead?

I would like something copy-pastable for the future.

json-rpc/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+### 2.2. Method disposeSpec
+
+Dispose the state that is associated with a session, including the SMT solver.
+You should call this method when you are done with the session, unless you
+shutdown the server. The session identifier must have been returned by the
```
thpani: I don't understand what the "unless you shutdown the server" means.

project/Dependencies.scala
```diff
@@ -48,6 +48,12 @@ object Dependencies {
     // see https://scalapb.github.io/docs/faq/#i-am-getting-import-was-not-found-or-had-errors
     val scalapbRuntime =
       "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf"
+    // json-rpc dependencies
+    val jakartaServlet = "jakarta.servlet" % "jakarta.servlet-api" % "6.0.0" % Provided
+    val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.19.2"
+    val jacksonModuleScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.19.2"
+    val jettyServer = "org.eclipse.jetty" % "jetty-server" % "11.0.25"
+    val jettyServlet = "org.eclipse.jetty" % "jetty-servlet" % "11.0.25"
```
thpani, line 56: why do we need the servlet packages?

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
```diff
@@ -46,9 +46,12 @@ class LogbackConfigurator(runDir: Option[Path], customRunDir: Option[Path]) exte
...
     rootLogger.addAppender(consoleAppender)
-    // debug messages at the apalache level
-    val apalacheLogger = loggerContext.getLogger("at.forsyte.apalache")
-    apalacheLogger.setLevel(Level.DEBUG)
+    // debug messages at the legacy Apalache level
+    val legacyApalacheLogger = loggerContext.getLogger("at.forsyte.apalache")
+    legacyApalacheLogger.setLevel(Level.DEBUG)
+    // debug messages at the newer Apalache level
+    val newApalacheLogger = loggerContext.getLogger("com.github.apalachemc.apalache")
+    newApalacheLogger.setLevel(Level.DEBUG)
```
thpani, line 54: I don't understand what's happening here. afaiu these are configured by name elsewhere, so your addition doesn't do anything?

konnov: `LogbackConfigurator` is the place where all logging is configuring. This is done programmatically, as we had issues with the standard approaches loading XML configurations. Basically, third-party packages were poisoning the configuration with their own files.

json-rpc/README.md
konnov: Good point. I've rephrased this sentence.

json-rpc/README.md
konnov: True. Just fixed the readme to compute base64.

json-rpc/README.md
konnov: Yeah, it is out of place. I've removed this sentence.

project/Dependencies.scala
konnov: I've tried to removed it. It's actually a servlet container provided by Jetty. I was also confused by the nomenclature.

konnov:
@thpani it looks like I've addressed all of your comments

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
thpani: I know. What I'm saying is that I don't understand what you're doing here *because* the logger is configured elsewhere.

json-rpc/README.md
thpani: Ah yes, this makes sense now!

mod-infra/src/main/scala/at/forsyte/apalache/infra/log/LogbackConfigurator.scala
konnov: Do you mean `newApalacheLogger`? I have introduced the new module under `com.github.apalachemc`, as `at.forsyte.apalache` is a bit outdated now. Apart from that, this is the place where the loggers are actually configured.

thpani: Ah, I missed that you introduced a fresh package. All good 👍🏻

PR #3129, opened 2025-08-15
Adding exploration methods to JSON-RPC

konnov:
This PR adds two more methods to execute transitions via JSON-RPC. I did not implement the context recovery logic, as it would make the PR significantly larger.

- [x] The command `check` now supports `--algo=remote` that does not do any exploration but prepares the model checker context for remote exploration
- [x] To this end, we introduce `ModelCheckerContext` that stores the essential parameters and the state for remote exploration.
- [x] The main extensions are done in `ExplorationService` and `JsonRpcServlet`:
  - [x] The method "assumeTransition"
  - [x] The method "nextStep"


json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -108,6 +248,167 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+   * @param params
+   *   the parameters object that contains the session ID and the transition ID
+   * @return
+   *   either an error or [[AssumeTransitionResult]]
+   */
+  def assumeTransition(params: AssumeTransitionParams): Either[ServiceError, AssumeTransitionResult] = {
+    val transitionId = params.transitionId
+    val sessionId = params.sessionId
+    // validate the input parameters under the global lock
+    rwLock.readLock().lock()
```
thpani: why?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -108,6 +248,167 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+        val checkerContext = checkerModule.modelCheckerContext.get
+        // prepare the transition
+        val stepNo = checkerContext.trex.stepNo
+        val transitions = if (stepNo == 0) {
+          checkerContext.checkerInput.initTransitions
+        } else {
+          checkerContext.checkerInput.nextTransitions
+        }
+        if (transitionId < 0 || transitionId >= transitions.size) {
+          Left(ServiceError(-32602, s"Invalid transition ID: $transitionId not in [0, ${transitions.size})"))
```
thpani: What is -32602?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -51,55 +160,86 @@ case class ServiceError(code: Int, message: String)
 class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging {
   // the rwLock to prevent concurrent
   private val rwLock: ReadWriteLock = new java.util.concurrent.locks.ReentrantReadWriteLock()
```
thpani: "to prevent concurrent"?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -51,55 +160,86 @@ case class ServiceError(code: Int, message: String)
...
-  // a pRNG to generate session IDs
+  // a pRNG to generate session IDs, we keep it constant to make the tests deterministic
```
thpani: why do you need a pRNG to generate session IDs?
cannot you just increment them?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -108,6 +248,167 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+      case None =>
+        Left(ServiceError(-32602, s"Session not found: $sessionId"))
+    }
+    rwLock.readLock().unlock()
+
+    // Prepare the transition. Make sure that we do not update the SMT context concurrently.
+    validationResult
+      .map {
+        case (checkerContext, actionExpr) => {
+          checkerContext.synchronized {
```
thpani, line 287: I don't understand the synchronization in this code.

There is a RW lock, but I don't know what it protects.
Then, here you're locking on checkerContext?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
```diff
@@ -19,16 +20,18 @@ import at.forsyte.apalache.tla.lir.transformations.standard.{
...
 /**
  * The implementation of a bounded model checker with SMT.
  *
  * @author
  *   Igor Konnov
  */
+@Singleton()
 class BoundedCheckerPassImpl @Inject() (
     options: OptionGroup.HasChecker,
     exprGradeStore: ExprGradeStore,
```
thpani, line 37: I don't understand how this can be a singleton, given that it's options depend on the source?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -108,6 +248,167 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+            val stepNo = checkerContext.trex.stepNo
+            val snapshotBefore = checkerContext.trex.snapshot()
+            val snapshotBeforeId = snapshotBefore.contextSnapshot.rewriterLevel
```
thpani, line 290: Is this really the right ID? (given that you query it after `snapshot()`)

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -108,6 +248,167 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+   */
+  def assumeTransition(params: AssumeTransitionParams): Either[ServiceError, AssumeTransitionResult] = {
+    val transitionId = params.transitionId
+    val sessionId = params.sessionId
+    // validate the input parameters under the global lock
+    rwLock.readLock().lock()
+    val validationResult = sessions.get(sessionId) match {
+      case Some(injector) =>
+        // get the checker context from the injector
+        val checkerModule = injector.getInstance(classOf[BoundedCheckerPassImpl])
```
thpani, line 263: This looks really weird. You're declaring it as `@Singleton` below?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -108,6 +248,167 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+    validationResult
+      .map {
+        case (checkerContext, actionExpr) => {
+          checkerContext.synchronized {
+            val stepNo = checkerContext.trex.stepNo
+            val snapshotBefore = checkerContext.trex.snapshot()
+            val snapshotBeforeId = snapshotBefore.contextSnapshot.rewriterLevel
+            // Upload the transition into the SMT context.
+            // It returns the result of a simple check that does not check satisfiability.
+            val isEnabled = checkerContext.trex.prepareTransition(transitionId, actionExpr)
```
thpani: If this doesn't check sat, don't call it `isEnabled`?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: We have to access the `sessions` map. It is actually immutable, so maybe we don't need a lock in this case?

konnov: Perhaps, we should just use [TrieMap](https://www.scala-lang.org/api/2.12.1/scala/collection/concurrent/TrieMap.html) and forget about locks when it concerns the `sessions` map.

thpani: TrieMap looks good.

Otherwise, I would just put a comment what we're synchronizing with the lock. But using a concurrent DS from the start sounds better.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: yeah, I though it would look cool, but there is no clear value to that. I will change it to a simple counter.

konnov: actually, I've just introduced an `AtomicInteger` to keep this counter. This and `TrieMap` allows us to get rid of the explicit locks altogether. We still need `synchronized` on the checker context though, but these locks are per session, not absolutely global.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: The lock on `checkerContext` is specific to the session. Basically, there should be no concurrent updates in the same session.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: Yeah, these are JSON-RPC error codes. I have introduced a separate object to store these codes.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: The order does not really matter in this case, but I will swap them, to improve readability

konnov: ah, no, this is right. Basically, the rewriter level in the snapshot ID.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
konnov: Right. I had to double-check the [Guice scopes page](https://github.com/google/guice/wiki/Scopes) on that:

> Guice comes with a built-in @Singleton scope that reuses the same instance during the lifetime of an application within a single injector.

Since we have a unique injector for each session id, this should work as expected.

konnov: Guice also has `@SessionScoped` but I cannot find how it actually works

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: This is related to https://github.com/apalache-mc/apalache/pull/3129#discussion_r2381587089

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
konnov: I am not sure we want to go along this route: https://github.com/google/guice/wiki/Servlets

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: Good point, renamed it to `isApplicable`

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
thpani: Right, I missed that the ID is retrieved from the snapshot

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/passes/BoundedCheckerPassImpl.scala
thpani: Ah, then singleton is fine

PR #3134, opened 2025-08-19
Constant simplifications for sets of literals

konnov:
Simplify set operators over literals:                                                                                   
 - Set union `{1, 2, 3} \cup {4}` becomes `{1, 2, 3, 4}`
 - Set difference `{3, 4} \ {3}` becomes `{4}`
 - Set union `{1, 2, 3} \cap {1, 3}` becomes `{1, 3}`
 - Set cardinality `Cardinality({3, 4, 5})` becomes 3

These optimizations happen to be quite useful, especially, in conjunction with the operator `..`, e.g., `0..Cardinality({ 1, 2, 3} \union {4})`.

The set of optimizations is incomplete. It would be great to have a more general rewriting framework in the future.

### DIFF
--- a/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
+++ b/tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
@@ -16,6 +16,7 @@ import at.forsyte.apalache.tla.types.{tlaU => tla, BuilderT, BuilderUT}
 abstract class ConstSimplifierBase {
   private val boolTag = Typed(BoolT1)
   private val intTag = Typed(IntT1)
+  private val strTag = Typed(StrT1)
 
   private def trueEx = ValEx(TlaBool(true))(boolTag)
   private def falseEx = ValEx(TlaBool(false))(boolTag)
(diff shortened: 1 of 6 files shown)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -235,6 +236,60 @@ abstract class ConstSimplifierBase {
...
+    case originalExpr @ OperEx(TlaSetOper.setminus, OperEx(TlaSetOper.enumSet, args1 @ _*),
+            OperEx(TlaSetOper.enumSet, args2 @ _*)) =>
+      val literals1 = extractLiterals(args1)
+      val literals2 = extractLiterals(args2)
+      if (!literals1.forall(_.isDefined) || !literals2.forall(_.isDefined)) {
+        // at least one of the sets contains non-literals, so we cannot simplify
+        originalExpr
+      } else {
+        // all elements are literals, so we can statically compute the set difference
+        val setDifference = (literals1.flatten.diff(literals2.flatten)).distinct
```
thpani: I don't think this is correct. `.diff` is multiset difference, so you have

```
{1,1,2,3} \ {1} = {1,2,3}
```

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -244,4 +299,29 @@ abstract class ConstSimplifierBase {
     simplifyShallow(ex) // when using TlaEx
   def applySimplifyShallowToBuilderEx(ex: BuilderT): BuilderT =
     ex.map(simplifyShallow) // when using TBuilderInstruction
+
+  // extract basic literals from the arguments of a set constructor
+  private def extractLiterals(args: Seq[TlaEx]): Seq[Option[Any]] = {
```
thpani: Can you use a union type instead of `Any`?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
```diff
@@ -244,4 +299,29 @@ abstract class ConstSimplifierBase {
...
+  // construct a set enumerator from a sequence of literals
+  private def literalsToSet(literals: Seq[Any], typeTag: TypeTag): OperEx = {
+    if (literals.isEmpty) {
+      emptySet(typeTag)
+    } else {
+      OperEx(TlaSetOper.enumSet,
+          literals.map {
+            case s: String  => ValEx(TlaStr(s))(strTag)
+            case i: Int     => ValEx(TlaInt(i))(intTag)
+            case i: BigInt  => ValEx(TlaInt(i))(intTag)
```
thpani: I think `TlaInt(_)` always takes a `BigInt`, so the `Int` case is redundant?

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
konnov: Good catch! Using `SortedSet` over literals now

tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/ConstSimplifierBase.scala
konnov: Done

PR #3141, opened 2025-08-21
Relax bound expressions in apalache::generate

konnov:
Currently, `apalache::generate` expects a constant integer and throws an exception otherwise. This is counterproductive, as there are plenty of opportunities to simplify the bound expression in the preprocessing passes.

This PR relaxes the requirement to the bound expression in `apalache::generate`. If it is not reduced to a constant expression while reaching the rewriting engine, then an error is produced. Basically, we increase the set of accepted expressions, while the UX is about the same for the end user.


thpani (approved):
👍

PR #3151, opened 2025-08-27
Demote the generic types checker

konnov:
Apalache's type checker has a very useful feature that checks whether a generic type is precise, that is, it is using exactly as many variables as needed. Unfortunately, this does not play well with the Quint's typechecker, which still does not have this feature.

Now we produce a warning instead of a type error in this case. Otherwise, it is quite hard to deal with Quint specifications.

/cc: @bugarela 

### DIFF
--- a/.unreleased/breaking-changes/demote-generics-error.md
+++ b/.unreleased/breaking-changes/demote-generics-error.md
@@ -1,0 +1,1 @@
+Report a warning instead of an error when generic types are not tight (#3151)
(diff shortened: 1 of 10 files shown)


thpani (approved):
LVGTM

thpani:
@bugarela 

It's specifically something like

```
def f(arg: a): a {...}
```

which Quint seems to treat as

```
def f(arg: a): b {...}
```

So, not enforcing arg and return value to be of the same type.
I suppose the same happens for multiple arguments that are annotated with the same type variable.

bugarela:
@thpani do you happen to have a more specific example? I'm failing to reproduce it in my first attempts:

```
def f1(arg: a): a = Set(arg)
def f2(arg: a): a = 1
def f3(arg: a): a = if (arg) 1 else 2
```
all of these give me errors in Quint

konnov:
> @thpani do you happen to have a more specific example? I'm failing to reproduce it in my first attempts:
> 
> ```
> def f1(arg: a): a = Set(arg)
> def f2(arg: a): a = 1
> def f3(arg: a): a = if (arg) 1 else 2
> ```
> 
> all of these give me errors in Quint

We had quite a complex example today. Not sure how to construct and MWE out of it.

Issue #101, opened 2020-02-21
Store original transitions in the ModelChecker execution stack


andrey-kuprianov:
Currently, during the ModelChecker execution, it's stack stores states and oracles, the latter representing the transition picked to reach the state. This transition is represented, though, as a relative index in the array of transitions enabled in the previous state.

For reporting purposes (e.g. printing a counterexample) we need the stack to contain references to the original transitions from the specification. Please refactor ModelChecker for that purpose.

konnov:
This will be resolved in the multicore version, as it completely changes `ModelChecker`

konnov:
This is fixed long time ago in `TransitionExecutor`.

PR #3167, opened 2025-09-13
Add support for labels in transitions and invariants

### LINKED ISSUE #3019: Labels should not stop VC generator from decomposing an invariant
Normally, the VC generator decomposes an invariant into two smaller invariants:

```tla
Inv ==
  /\ A
  /\ B
```

`Inv` is decomposed into:

```tla
Inv_0 == A
Inv_1 == B
```

However, if we annotate the invariant with a label, which is quite useful to see what has been violated, then the VC generator does not go inside the expression under the label:

```tla
Inv ==
  InInv ::
    /\ A
    /\ B
```

This should be quite easy to fix.

konnov:
Closes #3019. This also fixes the long-standing usability issue, e.g., mentioned in #101.

**The problem.** Apalache prints transition numbers and invariant numbers. However, these numbers are not telling us much. Since transitions and invariants are decomposed automatically, it is hard to assign meaningful names to these fragments *automatically*.

**The solution.** We finally have a pragmatic solution to this issue. The users can add labels to their actions and invariants. These labels are printed by the model checker. While this solution requires human intervention, it is the easiest one and the users can tune the precision of their labels as much as they like.


tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/SeqModelChecker.scala
```diff
@@ -256,18 +261,18 @@ class SeqModelChecker[ExecutorContextT](
                 // recover from the snapshot
                 trex.recover(snapshot.get)
                 // pick one transition
-                logger.info(s"Step ${trex.stepNo}: randomly picked transition $no")
+                logger.info(s"Step ${trex.stepNo}: randomly picked transition $no$labels_s")
```
thpani: For consistency, should it be #$no?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/util/LabelsCache.scala
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * A label that marks an expression as belonging to a certain state invariant.
+ * @param index
+ *   invariant index
+ */
+case class VCKind(index: Integer) extends ExprToLabelKind {}
+
+/**
+ * A cache for labels that are attached to expressions. The cache is used to avoid repeated calls of
```
thpani, line 32: Can you add why is this useful/necessary?

thpani (approved):
Amazing!

PR #3168, opened 2025-09-15
Demangle quint definitions into TLA+ labels

konnov:
Following #3167, this PR introduces a convention for translating Quint definitions into TLA+ labels.

For example:

```scala
val __label_Foo = true; e
```

becomes

```tla
Foo:: e
```

This is especially useful for keeping track of the actions that took place and invariants that were evaluated or violated.

@bugarela would you like to add syntax for labels in Quint?

To see an example of why this is nice and important, see `test/quint/labels.qnt`.

The following command prints the checked and violated invariants in the server output:

```sh
$ quint verify --max-steps=20  --invariant=inv test/quint/labels.qnt
# in the server output
PASS #13: BoundedChecker                                    
State 0: Checking 2 state invariants                          
State 0: state invariant 0 [above_min] holds.                  
State 0: state invariant 1 [below_max] holds.                    
State 0: Checking 2 state invariants                         
State 0: state invariant 0 [above_min] holds.                   
State 0: state invariant 1 [below_max] holds.               
Step 0: picking a transition out of 2 transition(s)      
State 1: Checking 2 state invariants                      
State 1: state invariant 0 [above_min] holds.               
State 1: state invariant 1 [below_max] holds.            
State 1: Checking 2 state invariants                       
State 1: state invariant 0 [above_min] holds.       
State 1: state invariant 1 [below_max] holds.                 
Step 1: picking a transition out of 2 transition(s)           
State 2: Checking 2 state invariants               
State 2: state invariant 0 [above_min] holds.        
State 2: state invariant 1 [below_max] holds.
State 2: Checking 2 state invariants           
State 2: state invariant 0 [above_min] holds.           
State 2: state invariant 1 [below_max] holds.             
Step 2: picking a transition out of 2 transition(s)      
...
State 10: Checking 2 state invariants                         
State 10: state invariant 0 [above_min] holds.       
...
State 10: state invariant 1 [below_max] violated.
```

Similar to that, with randomized symbolic execution:

```sh
$ quint verify --random-transitions=true --max-steps=20  --invariant=inv test/quint/labels.qnt
# the server output
...
Symbolic runs left: 92       
State 0: Checking 2 state invariants                          
State 0: state invariant 0 [above_min] holds.               
State 0: state invariant 1 [below_max] holds.        
Step 0: randomly picked transition #1 [x1]               
Step 0: picking a transition out of 1 transition(s)           
State 1: Checking 2 state invariants                            
State 1: state invariant 0 [above_min] holds.                  
State 1: state invariant 1 [below_max] holds.                  
Step 1: randomly picked transition #0 [inc] 
Step 1: picking a transition out of 1 transition(s)
State 2: Checking 2 state invariants
State 2: state invariant 0 [above_min] holds.  
State 2: state invariant 1 [below_max] holds. 
Step 2: randomly picked transition #0 [inc] 
Step 2: picking a transition out of 1 transition(s) 
State 3: Checking 2 state invariants 
State 3: state invariant 0 [above_min] holds.
State 3: state invariant 1 [below_max] holds.
Step 3: randomly picked transition #0 [inc] 
Step 3: picking a transition out of 1 transition(s) 
State 4: Checking 2 state invariants
State 4: state invariant 0 [above_min] holds.
State 4: state invariant 1 [below_max] holds.
...
State 12: Checking 2 state invariants
State 12: state invariant 0 [above_min] holds.
...
State 12: state invariant 1 [below_max] violated.
```

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -51,7 +51,7 @@ object Dependencies {
     // json-rpc dependencies
     val jakartaServlet = "jakarta.servlet" % "jakarta.servlet-api" % "6.1.0" % Provided
     val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.19.2"
-    val jacksonModuleScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.19.2"
+    val jacksonModuleScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.20.0"
     val jettyServer = "org.eclipse.jetty" % "jetty-server" % "11.0.26"
     val jettyServlet = "org.eclipse.jetty" % "jetty-servlet" % "11.0.26"
   }


bugarela:
Hi @konnov! I do have labels in mind, but I'm not settled on them as a solution. I think adding them in this non-invasive way for now is great and we can potentially start using it and assess the value.

We added an `--invariants` feature this year that helps with the problem you solved with the invariants label:

```
$ quint verify ../local/test.qnt --invariants "x >= -10" "x <= 10"
An example execution:

[State 0] { x: 1 }

[State 1] { x: 2 }

[State 2] { x: 3 }

[State 3] { x: 4 }

[State 4] { x: 5 }

[State 5] { x: 6 }

[State 6] { x: 7 }

[State 7] { x: 8 }

[State 8] { x: 9 }

[State 9] { x: 10 }

[State 10] { x: 11 }

[violation] Found an issue (1124ms).
  ❌ x <= 10
error: found a counterexample
```

As you can see, it prints which invariant was violated. It does so by loading the last state of the ITF trace and evaluating all invariants there, which is super cheap. I was trying to extend this feature so we automatically break down which "pieces" of invariants broke, but it was not so trivial with quantifiers - but still should be doable. So, for invariants, I really like this as a solution.

For the other labels in init and step, I was just thinking about this last week. For the simulator, we have `q::debug` which, in quality level, seems similar to what you get with labels in the symbolic simulator. What I'd like is to only see these values for the actual counterexample (and not for all the simulations until we find one), so I'm envisioning a better feature here. For these, I think Apalache will be very limited, so it would be a simulator-only thing - unless there is actually a way for Apalache to extract this info from Z3 as well.

konnov:
@bugarela, evaluating the invariants is definitely a nice feature! Though labels help us with other things, too. For instance, if the model checker is running for too long, it helps us to see, which transitions or invariants are the least efficient ones.

bugarela:
Could we re-use the `q::debug` operator for this? So it works in the simulator, with Apalache, and then also with TLC's `Print`.

konnov:
> Could we re-use the `q::debug` operator for this? So it works in the simulator, with Apalache, and then also with TLC's `Print`.

`q::debug` needs concrete values, which are only available if there is a counterexample. It actually looks like TLA+ labels are a more general facility than `TLC!Print`. In their current form, we are only using labels to indicate which parts of the spec are used. However, we could also use labels with arguments to extract concrete values from an SMT model.

PR #3171, opened 2025-09-24
Introduce labels in the TLA+ counterexamples

konnov:
Following #3167, in this PR, we output labels in the TLA+ counterexamples. (No support for labels in ITF yet).

To this end, I had to eliminate a bit of a technical debt and do a few refactorings:

 - [x] `Counterexample` has been moved to the `tla-ir` package and was renamed to `Trace`. Also, this class is parameterized by data now, which is `TlaEx` most of the time, but it can also be `Unit` in the case of examples.
 - [x] `Trace` is carrying `labels` in addition to the `states`. These labels are populated by the model checker. In principle, they can be any kind of strings, attached to the states.


tla-io/src/test/scala/at/forsyte/apalache/io/lir/TestItfCounterexampleWriter.scala
konnov, line 1: This file is renamed to `tla-io/src/test/scala/at/forsyte/apalache/io/lir/TestItfTraceWriter.scala`, but Github did not pick this up

tla-io/src/main/scala/at/forsyte/apalache/io/lir/Trace.scala
```diff
@@ -0,0 +1,45 @@
...
+ *   - for trace invariants: the applied, negated trace invariant
+ *
+ * @author
+ *   Shon Feder, Igor Konnov
+ */
+case class Trace[T](
+    module: TlaModule,
+    states: IndexedSeq[Trace.State],
+    labels: IndexedSeq[SortedSet[String]],
+    data: T)
```
thpani, line 30: Maybe this should be called `TraceAndWitness` or something.

I was confused by `Trace[T]`, because it looks like "trace of `T`s", but in fact it's a "trace and a `T`".

konnov: yeah, this `data` is the invariant violation. It is still the most weird part of this data structure.

konnov: I have tried to rename it into `TraceWithData`, and it became even more confusing. So I would just keep it as it is.

PR #3172, opened 2025-09-26
JSON-RPC: add checkInvariant

konnov:
This PR further extends the logic of the JSON-RPC server. In its new form, it is rich enough to implement `apalache simulate` via an external script.

What is implemented:

 - Modify `ExplorationService#assumeTransition` to support the parameter `snapshotId` that specifies a snapshot to revert to. This allows us to implement backtracking without implementing a new method.
 - Add `ExplorationService#checkInvariant` that checks a single invariant against the current context.
 - Add `CheckerSnapshotsPerSession.scala` to manage the session snapshots.
 - Add `script/explorer.py`. This is a simple implementation of `apalache simulate`. The initial version was generated by Claude. I had to change it though, to make the behavior more useful. I recommend not paying too much attention to this script. It is mostly added for simple testing.

Since this PR is relatively large, it is easier to review it by looking at the individual commits.


Makefile
```diff
@@ -38,6 +38,11 @@ test-coverage:
 
 # Run the integration tests
 integration: package
+	@echo "Installing Python dependencies for integration tests..."
+	@python -m pip install --user requests || \
```
thpani: Feels odd to have the Makefile force a particular way of installing dependencies.
I think this should be handled elsewhere.

json-rpc/README.md
```diff
@@ -79,15 +90,15 @@ when loading multiple specifications in different sessions.
...
-      "nTraceInvariants": <number of trace invariants>,
-      "hasView": <true if there is a view operator>
+    "sessionId": "unique-session-identifier",
+    "snapshotId": snapshot-identifier-after-loading-the-spec,
+    "specParameters": {
+      "nInitTransitions": number-of-Init-transitions,
+      "nNextTransitions": number-of-Next-transitions,
+      "nStateInvariants": number-of-state-invariants,
+      "nActionInvariants": number-of-action-invariants,
+      "nTraceInvariants": number-of-trace-invariants,
```
thpani, line 100: Why are these Hungarian style?
Isn't it obvious from the JSON what type they have?

json-rpc/README.md
```diff
@@ -178,19 +190,25 @@ Given a session identifier and a transition identifier, prepare this transition
 SMT context and assume that this transition holds true. Additionally, if `checkEnabled`
 is set to `true`, the server checks, whether there is a state that is reachable via
 the current transition prefix, including the supplied transition. The parameter `timeoutSec`
-sets the timeout for this check in seconds. If this parameter is set to `0`, then
-the timout is infinite.
+sets the timeout for this check in seconds. If `timeout` is not set, or it is set to `0`, then
+the timeout is infinite.
+
+To avoid an additional call, we also allow `assumeTransition` to roll back the context
+to an earlier snapshot *before* assuming the transition. In this case, the `snapshotId` parameter
```
thpani: the rollback is unconditional if `snapshotId` is set?

json-rpc/README.md
```diff
@@ -255,8 +273,64 @@ curl -X POST http://localhost:8822/rpc \
...
 ```
 
+### 2.5. Method checkInvariant
+
+Given a session identifier and an invariant identifier, check whether this invariant can be violated
+by a concrete execution that follows the collected symbolic path. The invariant identifier is a number
+that follows the following rules:
+
+- If `0 <= invariantId < nStateInvariants`, then the identifier refers to a state invariant.
+- If `nStateInvariants <= invariantId` and `nStateInvariants + nActionInvariants`, then the identifier refers to the
```
thpani: isn't this unnecessarily complicated?
you could just include a `kind` field

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
```diff
@@ -0,0 +1,83 @@
...
+   *   true, if the snapshot was found and recovered; false, otherwise
+   */
+  def recoverSnapshot(sessionId: String, checkerContext: CheckerContext, snapshotId: Int): Boolean = {
+    val sessionSnapshots = snapshots.getOrElseUpdate(sessionId, mutable.HashMap.empty)
+    val snapshot = sessionSnapshots.get(snapshotId)
+    if (snapshot.isEmpty) {
+      false
+    } else {
+      // remove all snapshots that beyond snapshotId
+      sessionSnapshots.filterInPlace((id, _) => id <= snapshotId)
```
thpani, line 67: doesn't this modify the underlying mutable hashmap?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
```diff
@@ -0,0 +1,83 @@
...
+class CheckerSnapshotsPerSession extends LazyLogging {
+  type Snapshot = ExecutionSnapshot[IncrementalExecutionContextSnapshot]
+  type CheckerContext = ModelCheckerContext[IncrementalExecutionContextSnapshot]
+
+  // A map from session IDs to saved snapshots. Each session may have multiple snapshots,
+  // as the user may want to roll back to a previous snapshot.
+  // The outer map is a TrieMap to allow concurrent reads and updates.
+  // The inner map is a mutable HashMap, as we do not expect concurrent accesses to
+  // snapshots within a single session.
+  private val snapshots: TrieMap[String, mutable.HashMap[Int, Snapshot]] = TrieMap.empty
```
thpani, line 30: why a mutable hashmap?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/ExplorationServiceParams.scala
```diff
@@ -0,0 +1,91 @@
...
+ */
+case class DisposeSpecParams(sessionId: String) extends ExplorationServiceParams
+
+/**
+ * Parameters for preparing a symbolic transition in the solver context.
+ * @param sessionId
+ *   the ID of the previously loaded specification
+ * @param snapshotId
+ *   the snapshot ID for recovering the context before the transition is assumed. If it is negative, no snapshot
+ *   recovery is performed.
```
thpani: I think this needs a better name. Maybe `rollbackToSnapshotId`

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/ExplorationServiceResult.scala
```diff
@@ -0,0 +1,105 @@
...
+ * @param nActionInvariants
+ *   the number of action invariants
+ * @param hasView
+ *   whether a view predicate is present in the specification
+ */
+case class SpecParameters(
+    nInitTransitions: Int,
+    nNextTransitions: Int,
+    nStateInvariants: Int,
+    nActionInvariants: Int,
```
thpani, line 74: again, why Hungarian style?

requirements.txt
```diff
@@ -0,0 +1,2 @@
+requests>=2.28.0
```
thpani: Is this a top-level dependency?

thpani (commented):
Cool! In general LGTM.
I would like a better solution for the python dependencies.

Makefile
konnov: Right. This was a quick fix. The integration tests in the CI are running via nix, so they don't need this command. Alternatively, we could just complain in the test that `requests` is not installed, e.g., checking the prerequisites.

json-rpc/README.md
konnov: Would you prefer `initTransitions` instead of `nInitTransitions`? I don't want to think whether these are transitions or the number of them.

json-rpc/README.md
konnov: yes

konnov: added a comment about that

json-rpc/README.md
konnov: true, let's introduce `kind`, which defaults to `STATE`.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
konnov: Yes, this is exactly my intent: To remove all snapshots above `snapshotId`.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
konnov: I am not sure exactly how `TrieMap` works, but I think that just retrieving the reference to a hash map and modifying it should be easy, as opposed to updating the value every time when we update the inner map.

requirements.txt
konnov: It is really just needed for `script/explorer.py`. So maybe we don't make it a dependency at all. After all, this script is mainly for testing and demonstration purposes.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/ExplorationServiceResult.scala
konnov: Because I want to say that something is the number of things, not the things themselves. We have to agree on notation that suits both of us here.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
thpani: no, I mean the HashMap inside the TrieMap

json-rpc/README.md
thpani: I just figured that `"initTransitions": 5` is unambiguous?
This doesn't really appear anywhere without a value attached?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
thpani: oh... is that actually what you want?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/ExplorationServiceResult.scala
thpani: okay, here it works for me

requirements.txt
thpani: yeah, maybe we should move it to a directory, or just have a comment in `explorer.py`

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/ExplorationServiceParams.scala
konnov: Renamed

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
konnov: Yes, this was my intent. I was also suspicious about this, but it makes sense?

json-rpc/README.md
konnov: Actually, I planned to turn these fields into arrays in a future PR. This would be needed to extract the labels. So maybe we just keep them as they are now and modify in the near future?

requirements.txt
konnov: The funny thing is that this is exactly what Claude recommended to me. So I've removed the forced dependency installation and added a test right in the script.

json-rpc/README.md
thpani: Yeah, sounds good

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/CheckerSnapshotsPerSession.scala
thpani: Yeah, I realized now that you're actually mutating it

thpani (approved):

PR #3173, opened 2025-09-29
JSON-RPC: phase 4

konnov:
Fixing the servlet to support long-running async tasks, since this is crucial for a model checking server.

### DIFF
--- a/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
+++ b/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
@@ -13,16 +13,18 @@ import at.forsyte.apalache.tla.bmcmt.ModelCheckerContext
 import at.forsyte.apalache.tla.bmcmt.config.CheckerModule
 import at.forsyte.apalache.tla.bmcmt.passes.BoundedCheckerPassImpl
 import at.forsyte.apalache.tla.bmcmt.trex.IncrementalExecutionContextSnapshot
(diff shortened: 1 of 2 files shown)


json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -412,12 +414,61 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+  )
   // data mapper for JSON serialization/deserialization
   // using Jackson with Scala module for better compatibility with case classes
   private val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
 
   override def doPost(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+    // Start async and disable its (30s default) timeout for truly long jobs:
+    val async = req.startAsync
+    async.setTimeout(0) // 0 = no timeout
+    val runnable: Runnable = () => {
```
thpani: I don't understand the comment here. afaiu, this wraps the request in a runnable and submits it to the thread pool? Maybe the comment should say just that?

thpani (approved):

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: Good point. The comments are by Claude. For some reason, I tend to ignore them :))

PR #3174, opened 2025-10-01
JSON-RPC: add method `query`

konnov:
Extending JSON-RPC with a new method `query` that allows an external script to query for an example trace and the value of a special view operator. This required extension of the `TransitionExecutorImpl`, which did not provide any logic to implement querying for the current view.

Here are the modifications in this PR:

 - [x] extend `README.md` with the description of `query` (implemented in this PR) and `nextView` (not implemented)
 - [x] add method `query` to `ExplorationService`, accompanied with private utility methods
 - [x] add method `evaluate` to `TransitionExecutorImpl`, `TransitionExecutor`, `ConstrainedTransitionExecutor`, `FilteredTransitionExecutor`
 - [x] fix synchronization issues in `ExplorationService` by using `Lock` instead of `synchronized`. For some reason, `synchronized` did not really lock the context, perhaps, due to the code returning different objects on map access.
 - [x] extend `script/explorer.py` to support querying.


json-rpc/README.md
```diff
@@ -140,7 +162,7 @@ curl -X POST http://localhost:8822/rpc \
       "nStateInvariants": 1,
       "nActionInvariants": 0,
       "nTraceInvariants": 0,
-      "hasView": false
+      "hasView": true
```
thpani, line 165: Why do we need `hasView`?

json-rpc/README.md
```diff
@@ -200,6 +233,17 @@ must be set to the identifier of the snapshot to roll back to. If `rollbackToSna
 performed unconditionally. You have to be careful about landing in a state with the right
 snapshot.
 
+**Precondition.** This method should be called once before calling `nextState`, unless
+the previous call to `assumeTransition` returned with the status `"DISABLED"`.
+
+**Effect.** If `rollbackToSnapshotId >= 0`, the model checker context is rolled back
+to the snapshot with `rollbackToSnapshotId`. Then, the transition with `transitionId`
```
thpani: I wonder if we should make rollback its own action, just to separate them logically

json-rpc/README.md
```diff
@@ -322,13 +396,152 @@ execution that violates the invariant. This field encodes a trace in the [ITF Fo
...
 
+### 2.6. Method query
+
+Given a session identifier, query the current context for values of several kinds:
+
+ - `"TRACE"`: A concrete trace that follows the symbolic path constructed so far.
+   There may be multiple such traces. This call returns the trace that
+   was found by the SMT solver.
+ - `"VIEW"`: The value of the view operator. The name of this operator must be
+   specified in the `view` field of the `loadSpec` method. There may be multiple
```
thpani, line 471: is there a good reason to not take the view operator name as an argument here?

thpani, line 471: I don't understand what "such values" refers to.

json-rpc/README.md
```diff
@@ -322,13 +396,152 @@ execution that violates the invariant. This field encodes a trace in the [ITF Fo
...
+   such values. This call returns the view of the model that was found by the
+   SMT solver.
+ - TBD: More kinds to be added in the future.
+
+**Precondition.** This method is successful only if the current context has a
+model. Otherwise, the result is undefined. Below are the conditions under which the
+model is guaranteed to exist:
```
thpani: Should we check for these and return an error?

Seems like something that's easy to mess up and result in unexpected behavior.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/VCGenerator.scala
thpani, line 1: The class docstring needs to be updates, given that this not only generates VCs now.

Maybe also the class name?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/VCGenerator.scala
```diff
@@ -86,27 +84,37 @@ class VCGenerator(tracker: TransformationTracker) extends LazyLogging {
...
+   * Given a module and the name of a view operator, normalize this operator as `VCView\$0` in the module.
+   *
+   * @param module
+   *   an input module
+   * @param viewName
+   *   name of the view operator
+   * @return
+   *   a transformed module
+   */
+  def genView(module: TlaModule, viewName: String): TlaModule = {
```
thpani, line 101: We do we need to normalize the view names?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -165,67 +170,68 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
-            val stepNo = checkerContext.trex.stepNo
-            val transitions = if (stepNo == 0) {
-              checkerContext.checkerInput.initTransitions
-            } else {
-              checkerContext.checkerInput.nextTransitions
-            }
-            if (transitionId < 0 || transitionId >= transitions.size) {
+        withLock(params.sessionId) {
+          if (!sessions.contains(params.sessionId)) {
+            return Left(ServiceError(JsonRpcCodes.INVALID_PARAMS, s"Session not found: ${params.sessionId}"))
```
thpani, line 175: wouldn't this fail earlier – in `withLock` while trying to retrieve the lock?

(pattern appears a couple of times below)

thpani (approved):
LGTM

Why do we need to mangle the view operator names though?

json-rpc/README.md
konnov: Yeah, this is the old approach we have in `apalache simulate`. I think a better approach would be to enumerate all operators that can be evaluated later. Otherwise, they are removed in the preprocessing phase. Then we could simply call the operators we like.

json-rpc/README.md
konnov: I agree. It's tricky when we blend two calls into one. JSON-RPC supports batching, so this call optimization is probably premature.

json-rpc/README.md
konnov: Nope. It was an attempt to make things work. What we could do instead is to receive an expression in the Apalache IR. It would be a bit harder to write, but for simple things like calling a nullary operator this should work.

konnov: Well, I thought it was not obvious that the view operator may produce different values in different models. This has to be improved.

json-rpc/README.md
konnov: This is actually outdated. Even `query` requires a call to `sat`, as we have to rewrite an expression, and this extends the SMT context.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/VCGenerator.scala
konnov: It actually decomposes the invariants into smaller pieces. The method `genView` is a bit outside of this, true.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/VCGenerator.scala
konnov: We normalize all important names such as `Init` and `Next`, so we can stop tracking the exact names in later stages. In what we discuss, we would not have to do that. If we remove `apalache simulate` at some point, this will be no longer needed.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: This is the pattern I am getting rid of: `dispose` acquires the lock, another method tries to acquire the lock and gets blocked, `dispose` frees the lock, the other method acquires the lock. If the other method acquired the lock right after `dispose` unlocking, then the other method may start working against a disposed session.

json-rpc/README.md
konnov: Let's postpone evaluation of arbitrary expressions to another batch of work: https://github.com/apalache-mc/apalache/issues/3178. It is more or less clear what to do, but I don't know how much code it would involve.

json-rpc/README.md
konnov: Postponed for https://github.com/apalache-mc/apalache/issues/3178

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/VCGenerator.scala
konnov: Updated the javadoc a bit

PR #3179, opened 2025-10-06
JSON-RPC: phase 6

### LINKED ISSUE #3178: Evaluate arbitrary expressions instead of views in JSON-RPC
Following these discussions:

https://github.com/apalache-mc/apalache/pull/3174#pullrequestreview-3293444851

https://github.com/apalache-mc/apalache/pull/3174#discussion_r2397902866

konnov:
- [x] Closes #3178. Evaluating arbitrary expressions is too hard, as it would require us to the expressions inliner on every call. Instead, we are using nullary operators. They slightly generalize the very specific behavior of views.
- [x] Adds `nextModel` to enumerate models. 


json-rpc/README.md
```diff
@@ -87,17 +87,27 @@ be used in subsequent calls to refer to this session.
...
+      "exported-operator-M"
     ]
   }
 }
 ```
 
+Note that if you are going to evaluate operators in the subsequent calls,
+you should list them in the `exports` field. Otherwise, the specification
+preprocessor prunes the definitions that are not used by the standard
+operators such as `Init`, `Next`, and the invariants.
```
thpani, line 109: Ah damn 😅

json-rpc/README.md
```diff
@@ -87,17 +87,27 @@ be used in subsequent calls to refer to this session.
...
+      "exported-operator-1",
+      ...,
+      "exported-operator-M"
     ]
   }
 }
 ```
 
+Note that if you are going to evaluate operators in the subsequent calls,
+you should list them in the `exports` field. Otherwise, the specification
```
thpani, line 107: Maybe `retain` is a better name? Just a suggestion, not a request.

json-rpc/README.md
```diff
@@ -467,11 +474,18 @@ Given a session identifier, query the current context for values of several kind
...
    was found by the SMT solver.
- - `"VIEW"`: The value of the view operator. The name of this operator must be
-   specified in the `view` field of the `loadSpec` method. There may be multiple
-   such values. This call returns the view of the model that was found by the
-   SMT solver.
- - TBD: More kinds to be added in the future.
+ - `"OPERATOR"`: A value of the operator whose name is given in `operator`. This name
+   must be exported in the `loadSpec` method, unless it is also used in the standard
+   operators such as `Init`, `Next`, and the invariants. The operator `operator`
+   MUST be side-effect free, that is, this operator may not refer to primed variables.
```
thpani, line 480: Just curious, what happens if it's not?

json-rpc/README.md
```diff
@@ -517,56 +536,76 @@ Execute the following command:
...
-
-**NOT IMPLEMENTED YET**
+### 2.8. Method nextModel
 
-Given a session identifier, find a view that is different from the one in the
+Given a session identifier, find a model that is different from the one in the
 current context. This method requires a call to the SMT solver, so it may take
-some time. The parameter `timeoutSec` sets the timeout for this check in seconds.
+some time. To distinguish the new model from the previous one, the method
+requires the parameter `operator`, which has exactly the same semantics as in the
```
thpani, line 553: So `operator` defines the view/projection? Wasn't quite clear from the text.

json-rpc/README.md
```diff
@@ -517,56 +536,76 @@ Execute the following command:
...
+   if such a model exists (i.e., if `hasOld == "YES"`). Otherwise, `oldValue == null`.
+
 ```json
 {
   "result": {
     "sessionId": "session-identifier",
-    "status": (SATISFIED|VIOLATED|UNKNOWN)
+    "oldValue": expr-in-itf-or-null,
+    "hasOld": (YES|NO|UNKNOWN),
+    "hasNext": (YES|NO|UNKNOWN)
```
thpani: Stylistic comment, but I'd prefer TRUE|FALSE|UNKNOWN

json-rpc/README.md
konnov: If not, the operator will modify the state. I think it will even work, but I am not sure how one would use that.

json-rpc/README.md
konnov: Clarified that in the comment

json-rpc/README.md
konnov: I thought of `keep`, but that's too close to the guts of the tool. The name `exports` is more common in PL, though it's a bit of a stretch here.

PR #3184, opened 2025-10-12
Set the core pool size to the number of logical cores

konnov:
Currently, the core pool size is statically set to 16. This underutilizes a machine with a larger number of logical cores, e.g., 32. In this PR, we set the core pool size to the number of logical cores.

### DIFF
--- a/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
+++ b/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
@@ -648,6 +648,8 @@ class JsonRpcServlet(service: ExplorationService) extends HttpServlet with LazyL
         t
       },
   )
+  logger.info("Created thread pool with core pool size %d, max pool size %d, keep alive %d sec"
+        .format(Runtime.getRuntime.availableProcessors(), MAX_POOL_SIZE, KEEP_ALIVE_SEC))
   // data mapper for JSON serialization/deserialization
   // using Jackson with Scala module for better compatibility with case classes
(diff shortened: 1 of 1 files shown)


json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -638,7 +635,9 @@ class JsonRpcServlet(service: ExplorationService) extends HttpServlet with LazyL
 
   // Our own bounded pool, to avoid blocking Jetty threads
   private val executor = new ThreadPoolExecutor(
-      CORE_POOL_SIZE,
+      // The core pool size is about the number of the logical cores.
+      // Since the most tasks are CPU-bound (z3), we do not need more threads.
```
thpani:
```suggestion
      // Since most tasks are CPU-bound (z3), we do not need more threads.
```

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -638,7 +635,9 @@ class JsonRpcServlet(service: ExplorationService) extends HttpServlet with LazyL
...
+      Runtime.getRuntime.availableProcessors(),
```
thpani, line 640: Javadoc says this returns # of cores _available to JVM_.

Shall we maybe print it, just to avoid unexpected behavior?

konnov: Good point. Added a log message.

PR #3198, opened 2025-10-22
Fix stability of the assignment finder

konnov:
We have found that the assignment finder slices actions in different orders, when applied multiple times. A single-run version seemed to be working, as we had a fix for that in the past.

This PR take more principled approach of using `SortedMap`, `SortedSet`, and sorting the final transitions again. This fixes the issue.

The code is very old and hard to understand, so a refactoring would be really great.

### DIFF
--- a/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
+++ b/tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
@@ -8,9 +8,8 @@ import at.forsyte.apalache.tla.lir.values.TlaBool
 import scala.collection.immutable.{SortedMap, SortedSet}
 
 /**
- * Constructs symbolic transitions from an assignment strategy.
- * In the intermediate computations, we use sets of assignment selections to avoid sorting overhead.
- * The final result is sorted to produce deterministic results.
+ * Constructs symbolic transitions from an assignment strategy. In the intermediate computations, we use sets of
+ * assignment selections to avoid sorting overhead. The final result is sorted to produce deterministic results.
  */
 class SymbTransGenerator(tracker: TransformationTracker) {
 


json-rpc/src/test/scala/com/github/apalachemc/apalache/jsonrpc/TestExplorationService.scala
```diff
@@ -53,6 +53,35 @@ class TestExplorationService extends AnyFunSuite with BeforeAndAfter {
     }
   }
 
+  test("load spec multiple times") {
```
thpani, line 56: Did you confirm that this test triggers the bug if the fix is not present?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -5,27 +5,31 @@ import at.forsyte.apalache.tla.lir.oper._
...
+        Set.empty[SortedSet[ValType]]
       else if (p_sets.length == 1)
         p_sets.head
       else {
         val one = p_sets.head
         val rest = allCombinations(p_sets.tail)
 
-        (for { s <- one } yield rest.map(st => st ++ s)).fold(Set.empty[Set[ValType]])(_ ++ _)
+        (for { s <- one } yield rest.map(st => st ++ s))
+          .fold(Set.empty[SortedSet[ValType]])(_ ++ _)
```
thpani: Don't `s <- one` and `rest.map` still iterate over the unsorted set?

In general, it's impossible to follow the types here.
Maybe we could type-annotate `one` and `rest`, and introduce an intermediary definition for the for-yield? Or even better, rewrite for-yield in a functional way?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -5,27 +5,31 @@ import at.forsyte.apalache.tla.lir.oper._
...
 class SymbTransGenerator(tracker: TransformationTracker) {
 
   private[assignments] object helperFunctions {
-    type LabelMapType = Map[UID, Set[UID]]
-    type AssignmentSelections = Set[Set[UID]]
-    type SelMapType = Map[UID, AssignmentSelections]
-    type letInMapType = Map[String, (UID, SelMapType)]
+    type AssignmentSelections = Set[SortedSet[UID]]
+    type SortedAssignmentSelections = SortedSet[SortedSet[UID]]
+    type SelMapType = SortedMap[UID, AssignmentSelections]
```
thpani, line 22: Why do we still use the unsorted AssignmentSelections?

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -272,10 +291,15 @@ class SymbTransGenerator(tracker: TransformationTracker) {
...
         ex: TlaEx,
         strategy: StrategyType,
-        selections: SelMapType): Seq[SymbTrans] =
-      selections(ex.ID).map { s =>
+        selections: SortedSelMapType): Seq[SymbTrans] = {
+      // It is crucial to have selections sorted here, to have a deterministic output.
+      // Otherwise, `sliceWith` produces UIDs in different orders on different runs.
+      // As a result, multiple runs of the same spec would produce different orderings of transitions,
+      // which is highly undesirable.
+      selections(ex.ID).toSeq.map { s =>
```
thpani, line 307: Since selections(ex.ID) is a `SortedSet`, cannot we

```suggestion
      selections(ex.ID).map { s =>
```

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -272,10 +291,15 @@ class SymbTransGenerator(tracker: TransformationTracker) {
     def getTransitions(
```
thpani: should `getTransitions` be private?

I'm surprised that this exposes both `apply()` and `getTransitions()`

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
```diff
@@ -5,27 +5,31 @@ import at.forsyte.apalache.tla.lir.oper._
 import at.forsyte.apalache.tla.lir.transformations.{TlaExTransformation, TransformationTracker}
 import at.forsyte.apalache.tla.lir.values.TlaBool
 
+import scala.collection.immutable.{SortedMap, SortedSet}
+
 /**
  * Constructs symbolic transitions from an assignment strategy.
```
thpani: afaiu, you're now computing selections unordered, but sort them before returning.

Maybe we should say this in the class-level comment, to avoid confusion.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
thpani: Okay, after reading the whole PR, my understanding is that you're computing selections unsorted, then sort them late. Is that correct?

I think we need to say this somewhere at the top to avoid confusion.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
thpani: I would also add comments here, explaining where/how the sorted/unsorted types are used

json-rpc/src/test/scala/com/github/apalachemc/apalache/jsonrpc/TestExplorationService.scala
konnov: Nope. I've only added it post-mortem

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
konnov: Yes, they are sorted in the very end

konnov: ok, I have refactored and commented this method a bit. Honestly, I have very little understanding of what is going on here.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
konnov: Yep, added a few comments. This requires a focused effort to understand what's going on in the code though.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
konnov: The resulting type is `Seq[...]`, so we have to convert it.

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
konnov: Done

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
konnov: Yep, I've made it `private[Symb...]`, so the companion object can still access it.

konnov:
> Okay, I was very confused until I got to the bottom of your PR.
> 
> I have a few questions just to confirm I understand the actual fix.

I have made another pass. Please have a look

tla-assignments/src/main/scala/at/forsyte/apalache/tla/assignments/SymbTransGenerator.scala
thpani: Ah, so both apply() and getTransitions() are entrypoints? 😕

konnov: Not anymore, `getTransitions` is called by `apply`, which is a companion object, so it needs `private[...]`.

PR #3207, opened 2025-11-05
Fix transitive inlining of polytypes when receiving Quint as input

### LINKED ISSUE #3204: Error when inlining a generic operator into a generic operator
Hello, I am trying to verify a specification written in Quint and am running into a spurious error about the rewriter:

```
error: <unknown>: rewriter error: Unexpected type u when generating a default value
```

Unfortunately, I cannot share the specification publicly (yet), but I am wondering if you could provide pointers on how to figure out what causes the issue. In particular, the error itself does not point to any source where the error originates. Where can I get some more context on where the rewriter is used, and what usually causes this sort of errors?

konnov:
Closes #3204. This fixes the inlining bug that was evading us for long time. For some reason, it was triggered only in Quint specifications that used chains of polymorphic types. A similar specification in TLA+ worked without issues, perhaps, due to the fact that the typechecker for TLA+ infers the types of all intermediate expressions in any case.

Here is a very minimal working example in Quint (see the original MWE in #3204):

```scala
module test {                                                                    
  def P(x: a): Set[a] = Set(x)
  def Q(y: b): Set[b] = P(y)

  var x: Set[Set[int]]

  action init = x' = Set()
  action step = x' = Q(Set())
}
```

Since we could not reproduce the same behavior in TLA+, only adding new tests in `TestInliner`.

@bugarela, perhaps, it's worth adding the above test to the Quint integration tests?

### DIFF
--- a/.unreleased/bug-fixes/transitive-inlining.md
+++ b/.unreleased/bug-fixes/transitive-inlining.md
@@ -1,0 +1,1 @@
+ - Fix transitive inlining of polymorphic Quint definitions (#3207)


tla-pp/src/main/scala/at/forsyte/apalache/tla/pp/Inliner.scala
```diff
@@ -96,31 +96,45 @@ class Inliner(
...
-  // Given a declaration (possibly holding a polymorphic type) and a monotyped target, computes
-  // a substitution of the two. A substitution is assumed to exist, otherwise TypingException is thrown.
-  private def getSubstitution(targetType: TlaType1, decl: TlaOperDecl): (Substitution, TlaType1) = {
-    val genericType = decl.typeTag.asTlaType1()
-    val maxUsedVar = Math.max(genericType.usedNames.foldLeft(0)(Math.max), targetType.usedNames.foldLeft(0)(Math.max))
-    new TypeUnifier(new TypeVarPool(maxUsedVar + 1)).unify(Substitution.empty, genericType, targetType) match {
+  // Given a declaration `callee` (possibly holding a polymorphic type) and the type at a call site,
+  // which can also be a polytype, computes a substitution of the two. A substitution is assumed to exist,
+  // otherwise TypingException is thrown.
+  private def getSubstitution(callSiteType: TlaType1, callee: TlaOperDecl): (Substitution, TlaType1) = {
```
thpani, line 102: Thanks for fixing the names 🙏

PR #3212, opened 2025-11-06
Implement JSON-RPC method `assumeState`

konnov:
Finally, this PR implements `assumeState` in the JSON-RPC server.

In addition to the implemented logic, I had to refactor the existing JSON nomenclature (factory vs. adapter) and add the adapters for Jackson. It's a lot of changes but most of them are renaming.

### DIFF
--- a/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
+++ b/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
@@ -350,7 +350,8 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
           checkerContext.checkerInput.rootModule.varDeclarations
             .map(d => d.name -> d.typeTag.asTlaType1())
             .toMap
-        new ItfJsonToTla(ScalaFromJacksonAdapter).parseState(varTypes, new JacksonRepresentation(params.equalities)) match {
+        new ItfJsonToTla(ScalaFromJacksonAdapter)
+          .parseState(varTypes, new JacksonRepresentation(params.equalities)) match {
           case Right(equalities) =>
             Right((checkerContext, varTypes, equalities))
           case Left(msg) =>
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/tracee/UJsonTraceReader.scala
@@ -25,7 +25,8 @@ import scala.util.{Failure, Success, Try}
  * @author
  *   Jure Kukovec
  */
-class UJsonTraceReader(sourceStoreOpt: Option[SourceStore], tagReader: TypeTagReader) extends TraceReader[UJsonRepresentation] {
+class UJsonTraceReader(sourceStoreOpt: Option[SourceStore], tagReader: TypeTagReader)
+    extends TraceReader[UJsonRepresentation] {
   private val builder = new UJsonToTlaViaBuilder(sourceStoreOpt)(tagReader)
   private val itfToTla = new ItfJsonToTla[UJsonRepresentation](ScalaFromUJsonAdapter)
 


tla-io/src/main/scala/at/forsyte/apalache/io/json/jackson/JacksonRep.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import at.forsyte.apalache.io.json.JsonRepresentation
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.ObjectMapper
+import scala.jdk.CollectionConverters._
+
+/**
+ * A JsonRepresentation using Jackson's JsonNode.
+ */
+sealed case class JacksonRep(value: JsonNode) extends JsonRepresentation {
```
thpani: I wouldn't abbreviate the child if the parent is not

```suggestion
sealed case class JacksonRepresentation(value: JsonNode) extends JsonRepresentation {
```

tla-io/src/main/scala/at/forsyte/apalache/io/json/jackson/ScalaFromJacksonAdapter.scala
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Adapter for extracting Scala primitives from JacksonRep.
+ */
+object ScalaFromJacksonAdapter extends ScalaFromJsonAdapter[JacksonRep] {
+  override def asInt(json: JacksonRep): Int = {
+    if (json.value.isInt) {
+      json.value.asInt()
+    } else {
+      throw new JsonDeserializationError("Not an integer.")
```
thpani, line 14: How does it work with BigInt? Or is that handled higher up?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -326,6 +329,82 @@ class ExplorationService(config: Try[Config.ApalacheConfig]) extends LazyLogging
...
+                  snapshots.recoverSnapshot(sessionId, checkerContext, snapshotBeforeId)
+                }
+                val status = if (isSat) AssumptionStatus.ENABLED else AssumptionStatus.DISABLED
+                logger.info(s"Session=$sessionId Step=$stepNo Snapshot=$snapshotAfterId: assumeState $status")
+                AssumeStateResult(sessionId, snapshotBeforeId, status)
+
+              case None =>
+                // in case of timeout or unknown, we do not roll back the context, but return unknown
+                logger.info(s"Session=$sessionId Step=$stepNo Snapshot=$snapshotAfterId: assumeState UNKNOWN")
+                AssumeStateResult(sessionId, snapshotAfterId, AssumptionStatus.UNKNOWN)
```
thpani, line 402: Looks a bit like overloading the `UNKNOWN` result, since we return it both when when we do not check, and when the SAT check is inconclusive.

I guess it's fine, because the client can still distinguish based on the parameters? But semantically, I would prefer `UNCHECKED` - you never asked me to check, and `UNKONWN` - you tould me to check, but I couldn't decide

tla-io/src/main/scala/at/forsyte/apalache/io/json/jackson/ScalaFromJacksonAdapter.scala
konnov: Good point. This adapter treats JSON integers as integers. BigInts are treated separately in ITF via the encoding `{ "#bigint": "nnnn" }`.

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: True. On the client side it usually not a problem, since it is clear from the code, which path is taken. I just wanted to avoid too many different values in the union types.

tla-io/src/main/scala/at/forsyte/apalache/io/json/jackson/JacksonRep.scala
konnov: Renamed

PR #3222, opened 2025-11-19
Add the `hostname` option to `server`

konnov:
This PR adds the option `--hostname` to the `server` command. This is needed to expose the JSON RPC on `0.0.0.0` when running inside Docker.

**WARNING:** I could not find how to do that with the GRPC server. So this only works for `--server-type=explorer`.

### DIFF
--- a/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
+++ b/json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
@@ -897,8 +897,15 @@ object JsonRpcServerApp {
   }
 
   def main(args: Array[String]): Unit = {
-    val hostname = if (args.length >= 1) args(0) else "localhost"
-    val port = if (args.length >= 2) args(1).toInt else 8822
+    // Backward compatibility: if only one argument and it's an integer, treat as port (old behavior)
+    val (hostname, port) =
+      if (args.length == 1 && Try(args(0).toInt).isSuccess) {
+        ("localhost", args(0).toInt)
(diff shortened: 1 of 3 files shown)


json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -881,8 +882,8 @@ class JsonRpcServlet(service: ExplorationService) extends HttpServlet with LazyL
 }
 
 object JsonRpcServerApp {
-  def run(config: Try[ApalacheConfig], port: Int): Unit = {
-    val server = new Server(port)
+  def run(config: Try[ApalacheConfig], hostname: String, port: Int): Unit = {
+    val server = new Server(new InetSocketAddress(hostname, port))
```
thpani, line 886: I don't get why you need this. The jetty docs say

> If null or 0.0.0.0, then bind to all interfaces.

Have you tried calling `.getHost()` on the connector to see what it's binding to?

konnov: I think you are right. It binds on 0.0.0.0 by default. Shall we discard this PR? The current implementation is incorrectly pointing that it listens on localhost.

thpani: If it works that way, I think so, yes 👍

konnov:
Discard this PR, as JSON RPC is already binding on 0.0.0.0 by default

PR #3312, opened 2026-04-23
Add support for TLC's CHECK_DEADLOCK config keyword

### LINKED ISSUE #3311: Support TLC’s CHECK_DEADLOCK config keyword
Please support TLC’s `CHECK_DEADLOCK` keyword in Apalache config files.

In particular:

`CHECK_DEADLOCK FALSE` should behave like `--no-deadlock`
`CHECK_DEADLOCK TRUE` should preserve the default behavior

This came up in https://github.com/tlaplus/Examples/pull/209, which adds Apalache type annotations and verification to many TLA+ examples. This would improve compatibility with TLC-style .cfg files and reduce wrapper/tooling special cases.

lemmy:
The CHECK_DEADLOCK option in TLC `.cfg` files was previously parsed but ignored. Honor it now: CHECK_DEADLOCK FALSE behaves like passing --no-deadlock on the CLI, while CHECK_DEADLOCK TRUE preserves the default deadlock-checking behavior. The --no-deadlock CLI flag and apalache.cfg continue to take precedence over the value in the TLC config file, mirroring how INIT, NEXT, INVARIANT, and other TLC options are merged.

Closes #3311.


mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
```diff
@@ -704,7 +704,9 @@ object OptionGroup extends LazyLogging {
         algo <- checker.algo.toTry("checker.algo")
         discardDisabled <- checker.discardDisabled.toTry("checker.discardDisabled")
         length <- checker.length.toTry("checker.length")
-        noDeadlocks <- checker.noDeadlocks.toTry("checker.noDeadlocks")
+        // `noDeadlocks` defaults to `false` when neither the CLI/apalache.cfg nor a TLC
+        // config file sets it (a None value here means "user did not specify").
+        noDeadlocks = checker.noDeadlocks.getOrElse(false)
```
konnov: This is too restrictive. The option `checker.noDeadlocks` can be transferred via configuration options

konnov (commented):
In principle, looks good. However, we have to fix that to keep the configurations compatible

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala
konnov: I have reverted this change, as it breaks backward compatibility

konnov (approved):
This features works. I've added a few more tests to make sure that we do not break the old features and this new feature. The file `options.scala` was already hard to understand and it's getting even more complex. At some point, we will need refactoring.

konnov:
I had to fix this once again, since the configuration logic is quite confusing

lemmy:
Thanks!

PR #3330, opened 2026-05-11
Add devcontainer build environment

thpani:
Add a devcontainer build environment.

Makes it easier to jail build environments for those not on nix.


konnov (approved):
looks good!

PR #3323, opened 2026-05-09
Fix import of named assumptions used as operators

shalashaska117:
Fixes #3318.

## Summary

SANY represents named assumptions used in operator applications as `ASTConstants.ThmOrAssumpDefKind`. The importer previously did not handle this operator kind in `OpApplTranslator.translateConstOper`, causing a `scala.MatchError: 23` during the SANY AST -> TLA-IR translation pass.

This PR treats theorem/assumption definitions like user-defined operators in this translation path.

## Changes

- Handle `ASTConstants.ThmOrAssumpDefKind` in `OpApplTranslator.translateConstOper`
- Add a regression test for a named `ASSUME` used inside `Init`
- Add an unreleased bug-fix changelog entry

## Testing

- `sbt "tla_io/compile"`
- `sbt "tla_io/testOnly at.forsyte.apalache.tla.imp.TestSanyImporter"`

PR #3331, opened 2026-05-11
Fix inlining of named assumptions

### LINKED ISSUE #3326: Propagate named assumptions into the inliner
A follow-up of #3323.

> ...Unfortunately, the preprocessor fails further down the pipeline, as the definition inliner does not push named assumptions into the scope. This needs a careful surgical fix in the inliner. I will check how to fix that.

_Originally posted by @konnov in https://github.com/apalache-mc/apalache/issues/3323#issuecomment-4415864703_

shalashaska117:
Fixes #3326.

## Summary

This PR fixes a preprocessing failure that happens when a named `ASSUME` declaration is referenced from an operator body.

After #3323, the SANY importer correctly handles named assumptions used as operators. However, the inliner did not expose named assumptions in its scope, so references such as `NamedAssumption` could survive the inline pass and later fail in `FlatLanguagePred` as undeclared operators.

This change keeps named assumptions as `TlaAssumeDecl`s in the module declarations, but exposes them to the inliner as synthetic nullary operators.

## Changes

- Expose named assumptions as synthetic nullary operators in the inliner scope
- Add a regression test for a named assumption referenced from an operator body
- Add an unreleased bug-fix changelog entry

## Testing

- `sbt scalafmtCheckAll`
- `sbt "tla_pp/testOnly at.forsyte.apalache.tla.pp.TestInliner"`
- Manual end-to-end check with a TLA+ module containing a named assumption referenced from `Init`, resulting in `EXITCODE: OK`

### DIFF
--- a/.unreleased/bug-fixes/3326-named-assumption-inliner.md
+++ b/.unreleased/bug-fixes/3326-named-assumption-inliner.md
@@ -1,0 +1,1 @@
+Fixed a preprocessing failure when a named ASSUME declaration is referenced from an operator body, see #3326.
--- a/test/tla/InlineAssumptions3318.tla
+++ b/test/tla/InlineAssumptions3318.tla
@@ -1,0 +1,21 @@
+----------------------- MODULE InlineAssumptions3318 -------------------------
+(*
+  A regression test for the issue identified in:
+  https://github.com/apalache-mc/apalache/issues/3318
+*)
+EXTENDS Integers
+
+VARIABLE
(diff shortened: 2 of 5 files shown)


konnov (approved):
Thank you for a clean fix! I've added a regression test in `test/tla/InlineAssumptions3318.tla` and its call in `cli-integration-tests.md`. This is to make sure that the issue does not surface somewhere else in the pipeline.

PR #3335, opened 2026-05-11
Implement CVC5 solver backend for OOPSLA encoding

thpani:
## Summary

Add experimental support for CVC5 as an SMT solver backend for the OOPSLA19 encoding.

Add ADR-025 for the CVC5 solver backend decision.

`--smt-solver=cvc5` is accepted for OOPSLA19, while `cvc5` with `arrays` or `funArrays` is rejected with a user-facing error. CVC5's support for SMT arrays with extensionality differs significantly from Z3, so we leave Arrays/FunArrays theory support for later.

### Solver selection and configuration

- Add `--smt-solver=(z3|cvc5)` and `SMT_SOLVER` support.
- Plumb solver selection through `SolverConfig`, `RecordingSolverContext`, `SolverContextFactory`, and checker setup.
- Solver-specific tuning parsing selects only the active solver namespace: `z3.*` for Z3, `cvc5.*` for CVC5
- Keep shared tuning such as `smt.randomSeed` and `smt.statsSec`.

### CVC5 backend

- Add `Cvc5SolverContext`.
- Implement the OOPSLA19 solver-context operations needed for current model-checker paths (cell declaration, boolean/integer constants and terms, (dis)equality, ...)
- CVC5 is sensitive to theory selection, so we add default CVC5 SMT logic selection `QF_UFLIA` for OOPSLA19. Tuning-option `cvc5.smt.logic` overrides the default CVC5 logic.
- CVC5 receives the Apalache SMT seed via the two options it exposes (`seed` and `sat-random-seed`)

### SMT logging

- CVC5 debug logs begin with replayable SMT-LIB prologue lines:
```
(set-logic ...)
(set-option :random-seed ...)
(set-option :sat-random-seed ...)
```
- Z3 seed logging now emits replayable SMT-LIB `set-option` lines instead of comments for the active seed parameters.

### Tests

- Add CVC5 unit and integration coverage for OOPSLA19 paths: solver context, recording context, arena tests, rewriter tests, trex tests, bmc smoke tests
- Add a cross-solver PBT suite comparing Z3 and CVC5 on OOPSLA19 `SolverContext` behavior.
- Add one full Ubuntu/OOPSLA19/CVC5 integration matrix row
- Nonlinear examples that run under the CVC5 matrix are adjusted to pass `cvc5.smt.logic=QF_UFNIA`.
- Fix `make integration` so an empty `TEST_FILTER` calls `mdx-test.py --debug` with no empty-string positional argument.

### Documentation

User docs mention:
  - CVC5 as an experimental solver backend, limited to OOPSLA
  - `--smt-solver` / `SMT_SOLVER` and `cvc5.smt.logic`

### Dependencies and packaging

- Add `io.github.cvc5:cvc5:1.3.4`.
- Add explicit native classifier jars to pull native solver libraries (supports Linux, macOS, and Windows on x86_64/aarch64)

Checklist:


rodrigo7491:
Nice to see cvc5 being added as an alternative backend! Quick comment from my side regarding the arrays encoding.

When they were implemented, both `arrays` and `funArrays` used some elements, i.e., constant arrays and maps, supported by Z3 but not in SMT-LIB at the time. SMT-LIB has since been updated, and [cvc5 now supports constant arrays](https://cvc5.github.io/docs/cvc5-1.3.4/api/python/pythonic/array.html#cvc5.pythonic.ConstArray), but it seems that maps are not in there yet (unless I missed something); [Z3 map for reference](https://z3prover.github.io/api/html/namespacez3py.html#ad9d31de7a1d284cf2c8a3074af6ec6e0).

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/FineTuningParser.scala
```diff
@@ -33,6 +33,8 @@ object FineTuningParser {
       "rewriter.shortCircuit" -> BoolFieldType(),
       // keeping for backwards compatibility
       "smt.randomSeed" -> UnsignedIntFieldType(),
+      // setting the SMT logic in CVC5
+      "cvc5.smt.logic" -> StringFieldType(),
```
konnov, line 37: Wondering whether we should just expose this as `smt.logic`. The logics are standardized by SMT-LIB.

thpani:
Hi @rodrigo7491, thanks for the input!

Yeah, I ran a few smoke tests and it looks like `funArrays` may work, but we may not get `arrays` to work with CVC5 yet. Even for `funArrays`, CVC5 supports equality between store chains to constant arrays only in its experimental solver, but that may be fine since we will also mark cvc5 on Apalache as experimental for now.

In any case, I decided to split off OOPSLA for now and wanted to ping you about arrays once I'm confident enough in the CVC5+OOPSLA combo.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
```diff
@@ -0,0 +1,587 @@
...
+    closeLogs()
+    cellCache.clear()
+    inCache.clear()
+    cellSorts.clear()
+  }
+
+  override def declareCell(cell: ArenaCell): Unit = {
+    smtListener.onIntroCell(cell.id)
+
+    if (maxCellIdPerContext.head >= cell.id) {
```
konnov, line 116: We also have `checkConsistency` below, but it is checking for `maxCellIdPerContext.head > topId`. I wonder whether we should call `checkConsistency` here as well, but then we have to generalize `checkConsistency`

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
```diff
@@ -0,0 +1,587 @@
...
+    val (term, _) = toExpr(ex)
+    val value = solver.getValue(term)
+    if (value.isBooleanValue) {
+      tla.bool(value.getBooleanValue)
+    } else if (value.isIntegerValue) {
+      tla.int(value.getIntegerValue)
+    } else if (value.getSort.hasSymbol && value.getSort.getSymbol.startsWith("Cell_")) {
+      tla.name(value.toString, TlaType1.fromTypeTag(ex.typeTag))
+    } else {
+      flushAndThrow(throw new SmtEncodingException(s"SMT $id: Expected an integer or Boolean, found: $value", ex))
```
konnov:
```suggestion
      flushAndThrow(throw new SmtEncodingException(s"SMT $id: Expected a Boolean, integer, or name; found: $value", ex))
```

konnov (approved):
This is amazing! I really like how this solution fits into the existing architecture. I was expecting CVC5 to require more changes. Thanks for a very clean integration.

shalashaska117:
I don’t see an obvious blocker, but one coverage gap may be worth closing before merge.

`TestCrossSolverContextWithOOPSLA19` explicitly says that `selectInFun`, `storeNotInFun`, and the ternary/function-update form of `storeInSet` are not exercised directly. Since `Cvc5SolverContext.toExpr` contains dedicated OOPSLA19 handling for some of these internal membership/function paths, it would be useful to add at least one focused regression test covering the function-related cases, or to document why the current OOPSLA19 checker paths cannot emit the untested ternary/function-update form.

My concern is that CVC5 may work for the covered set-membership paths while still failing on a real function-update path with an `InvalidTlaExException`.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/FineTuningParser.scala
thpani: I've been thinking about it. But we'd probably need different defaults, since Z3 works well with `(set-logic ALL)` and CVC5 doesn't. So it felt easier to keep them separate. wdyt?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
thpani: yeah, but `checkConsistency` is part of the public API of the `SolverContext` trait; it really only needs an arena, which is the right choice I think

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
thpani: it's actually a cell name, right?

konnov: Right. Maybe we could also write "unexpected term". I don't know how it helps the user to know what was expected here.

mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/FineTuningParser.scala
konnov: Hmmm. This makes sense. Let's keep it as is

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
thpani: Adressed in [7436943](https://github.com/apalache-mc/apalache/pull/3335/commits/7436943f89311bf89008a420f22ac8e35d151018)

thpani:
> My concern is that CVC5 may work for the covered set-membership paths while still failing on a real function-update path with an `InvalidTlaExException`.

That's covered by the existing rewriter and e2e tests

PR #3339, opened 2026-05-12
Quiet docker integration wrapper

thpani:
## Problem

The docker integration job runs the mdx CLI tests with `USE_DOCKER=true`, which shadows `apalache-mc` with `script/run-docker.sh`.

The CVC5 CLI snippets added on https://github.com/apalache-mc/apalache/pull/3335 are strict: they pipe stdout through `sed | grep` and expect only the Apalache result lines. In docker mode, two unrelated stderr prints were captured by mdx:

- `script/run-docker.sh` always printed `# running command: docker run ...`.
- The image entrypoint bind-mounted host `~/.tlaplus` directly to `/home/apalache/.tlaplus`. Docker creates bind mounts before the entrypoint runs, so `/home/apalache` already existed when `useradd --create-home` ran. That made `useradd` print warnings about the existing home directory.

Older docker mdx runs did not expose this because many existing sections use ellipses that tolerate incidental output.

## Fix

- Make the docker wrapper command echo opt-in via `APALACHE_DOCKER_DEBUG=true`.
- Mount host `~/.tlaplus` at `/mnt/host-tlaplus` instead of under `/home/apalache`.
- Keep `useradd --create-home` in the container entrypoint, then symlink `/home/apalache/.tlaplus` to the mounted directory after the home exists.

This preserves the runtime UID/GID-matching `apalache` user while avoiding docker/mdx noise in CLI integration output.

## Validation

- `bash -n script/run-docker.sh`
- `bash -n bin/run-in-docker-container`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -248,7 +248,12 @@ jobs:
         run: make docker
       - name: Run integration tests in docker container
         run: |
-          nix develop -c bash -c "source ./test/.envrc && test/mdx-test.py --debug"
+          nix develop -c bash -c '
+            export HOME="$PWD/test/tla"
+            export JVM_ARGS="-Duser.home=/var/apalache"
+            source ./test/.envrc
+            test/mdx-test.py --debug
+          '
         env:
           USE_DOCKER: true
           APALACHE_TAG: latest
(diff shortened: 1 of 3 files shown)


konnov (approved):
Good fix!

PR #3341, opened 2026-05-12
Support Unicode parsing through SANY 1.8

### LINKED ISSUE #2995: Pull in updated version of SANY to support Unicode specs
Now that SANY supports parsing Unicode TLA+ specs, it should be fairly straightforward to support Unicode here by updating to a newer version of SANY. Possible bugs include if your translation code looks at actual string representation (for example, `"\\E"` instead of `TLAplusParserConstants.EXISTS`) which was the case in SANY itself.

konnov:
Closes #2995. Update tla2tools.jar to 1.8.0, which required small API changes. Add support for Unicode via the new SANY version. 

### DIFF
--- a/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
+++ b/tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
@@ -5,8 +5,9 @@ import at.forsyte.apalache.tla.imp.src.SourceStore
 import at.forsyte.apalache.tla.lir.TlaModule
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.commons.io.output.WriterOutputStream
-import tla2sany.drivers.SANY
+import tla2sany.drivers.{SANY, SanySettings}
 import tla2sany.modanalyzer.SpecObj
+import tla2sany.output.{LogLevel, SimpleSanyOutput}
 
 import java.io._
 import java.nio.file.Files
@@ -33,11 +34,18 @@ object SANYSyncWrapper {
         .setWriter(errBuf)
         .setCharset(StandardCharsets.UTF_8)
         .get()
-      SANY.frontEndMain(
-          specObj,
-          file.getAbsolutePath,
+      val out = new SimpleSanyOutput(
           new PrintStream(outStream),
+          LogLevel.INFO,
       )
+      SANY
+        .parse(
+            specObj,
+            file.getAbsolutePath,
+            out,
+            SanySettings.defaultSettings(),
+        )
+        .code()
     }
   }
 }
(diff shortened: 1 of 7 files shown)


tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
```diff
@@ -195,14 +203,6 @@ class SanyImporter(sourceStore: SourceStore, annotationStore: AnnotationStore) e
     } yield (moduleName, tempFile)
 
   private def throwOnError(specObj: SpecObj): Unit = {
-    val initErrors = specObj.getInitErrors
-    if (initErrors.isFailure) {
-      throw new SanyAbortException(initErrors.toString)
-    }
-    val contextErrors = specObj.getGlobalContextErrors
```
thpani, line 202: these no longer exist?

thpani:
Would be interesting if this fixes #1963 and we can remove the lock

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
konnov: It looks like SANY finally does not have global objects. Good question though, needs double checking

konnov: @lemmy Could you comment on this. Are we using the new API correctly? Thanks!

lemmy: @ahelwer has been refactoring SANY's frontend.

konnov: @ahelwer could you comment on this?

ahelwer: Both of these were for capturing errors that can happen during static initialization functions that had to be called. That initialization is now done on process load, in java `static` blocks and should never throw.

ahelwer: The current recommended SANY entrypoint for external users is this function:

https://github.com/tlaplus/tlaplus/blob/209713870170c7da34d4a9a51439ab3ade6e5040/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java#L182-L186

konnov: Given that it still initializes the global environment, I guess, it is probably not thread safe? For the older versions, we had a mutex, so it looks we still have to keep it.

ahelwer: There are unfortunately still three remaining not-entirely-thread-safe static things in SANY that I've yet to remove:
- standard modules are copied to a system temp dir
- context static init; this can probably be removed but I need to figure out how to test it
- comment-attach for operator definitions is static for whatever reason

So for now SANY is not thread-safe, no.

konnov:
> Would be interesting if this fixes #1963 and we can remove the lock

Not yet. See: https://github.com/apalache-mc/apalache/pull/3341#discussion_r3228863269

thpani (approved):
LGTM

tla-io/src/main/scala/at/forsyte/apalache/tla/imp/SanyImporter.scala
thpani: @ahelwer excellent, thanks for the follow-up!

Issue #2972, opened 2024-08-28
<unknown>: checker error: Unexpected equality test over types CellTFrom(Set(Int)) and InfSet[CellTFrom(Int)]


lemmy:
```tla
---- MODULE Foobar ----
EXTENDS Apalache, Integers

VARIABLE
    \* @type: { p : Set({ q : Set(Int) }) };
    v

TypeOK ==
    v \in [ p: SUBSET [ q: SUBSET Int ] ]

Init ==
    /\ v = [ p |-> {[ q |-> { 1 } ]} ]
    /\ TypeOK    \* Conjoining  TypeOK  triggers the error.

GenInit ==
    /\ v = Gen(1)
    /\ TypeOK    \* Conjoining  TypeOK  triggers the error.

Next ==
    UNCHANGED v

====
```

### Init
```shell
-> % ~/apalache/apalache-0.45.3/bin/apalache-mc check --init=Init --next=Next Foobar.tla
17:37:44.762 [main] INFO at.forsyte.apalache.tla.tooling.opt.CheckCmd -- Loading configuration
17:37:44.787 [main] INFO at.forsyte.apalache.tla.tooling.opt.CheckCmd --   > TLC config file found in specification directory. To enable it, pass --config=Foobar.cfg.
# Usage statistics is ON. Thank you!
# If you have changed your mind, disable the statistics with config --enable-stats=false.

Output directory: /Users/markus/src/TLA/_specs/MSFT/pbft-tlaplus/_apalache-out/Foobar.tla/2024-08-27T17-37-44_41841269002290680
# APALACHE version: 0.45.3 | build: 0b3ae56                       I@17:37:44.978
Tuning: search.outputTraces=false                                 I@17:37:45.006
PASS #0: SanyParser                                               I@17:37:45.146
Parsing file /Users/markus/src/TLA/_specs/MSFT/pbft-tlaplus/Foobar.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/Apalache.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/Integers.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/Naturals.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/__rewire_sequences_in_apalache.tla
[...]

PR #3355, opened 2026-05-22
Reject powersets over infinite sets

lemmy:
Report SUBSET Int and similar powersets over infinite case sets as input errors before expansion, avoiding the confusing equality-type crash from #2972.

Adresses Github issue #2972
https://github.com/apalache-mc/apalache/issues/2972

- [ ] ~Documentation added for any new functionality~ Not new functionality


konnov:
I think we should address it properly, as described in #3356. The current PR is only fixing the symptoms, not the core issue.

PR #3358, opened 2026-05-22
Update html anchors in the docs

### LINKED ISSUE #3040: Links from CLI point to missing anchors
I got a CLI error pointing me to https://apalache-mc.org/docs/apalache/running.html#supplying-jvm-arguments but this anchor doesn't exist, as it was changed into https://apalache-mc.org/docs/apalache/running.html#13-supplying-jvm-arguments (there's a section number now `#13-*`)

This is probably the same for all other links to anchors.

thpani:
Fixes #3040

- Introduce some manual anchors that are used in the docs and source code.
- Inline anchors into the headings, to avoid rendering a separate `<p>` around them.
-  Uniformly use `<a id="...">` instead of `<a name="...">`.

### DIFF
--- a/docs/src/HOWTOs/howto-write-type-annotations.md
+++ b/docs/src/HOWTOs/howto-write-type-annotations.md
@@ -79,8 +79,7 @@ Run the type checker again. You should see the following message:
 ...
 ```
 
-<a id="recipe2"></a>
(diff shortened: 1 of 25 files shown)


konnov (approved):
Thanks for fixing!

PR #3384, opened 2026-06-28
Reject FoldSet over an infinite set instead of silently folding empty (#1691)

### LINKED ISSUE #1691: Unsound folds over `Nat`
An issue found by Markus:

```
I changed Sum(f, S) == FoldFunctionOnSet(+, 0, f, S) to Sum(f, S) == FoldFunctionOnSet(+, 0, f, Nat)
```

Apalache should have complained about `Nat`, but it interpreted `Nat` as an empty set.

shalashaska117:
## What

Fixes #1691.

`ApaFoldSet` (and the community-module folds built on it) over an **infinite set** such as `Nat` or `Int` was **unsound**: it silently treated the set as empty and returned the base value. For example:

```tla
ApaFoldSet(Plus, 0, Nat)   \* evaluated to 0, with no error
```

`Init == result = ApaFoldSet(Plus, 0, Nat)` followed by `Inv == result = 0` reported `NoError` before this change.

## Why

In `FoldSetRule`, the set argument is rewritten to an arena cell and its members are read via `arena.getHas(setCell)`. Infinite sets `Nat`/`Int` are represented by fixed cells of type `InfSetT` that point to **no** member cells, so the fold ran over zero elements and returned the base value.

## How

Detect an `InfSetT` set cell in `FoldSetRule` and raise a `RewriterKnownLimitationError`, which is surfaced as a clean, source-located error (`NormalErrorMessage`) rather than a silent wrong result or an internal crash:

```
foldnat.tla:10:38-10:40: known limitation: FoldSet is not supported over an infinite set such as Nat or Int, found: Nat. Fold over a finite set instead.
```

## Testing

- **New unit tests** in `TestSymbStateRewriterFoldSet` asserting `RewriterKnownLimitationError` for folding over `Nat` and over `Int` (run across the OOPSLA19 and arrays encodings).
- Existing FoldSet tests still pass (12 succeeded across encodings, 0 failed).
- Manual CLI verification:
  - `ApaFoldSet(Plus, 0, Nat)` -> clean `known limitation` error (previously `NoError`, result silently `0`).
  - `ApaFoldSet(Plus, 0, Int)` -> clean `known limitation` error.
  - `ApaFoldSet(Plus, 0, {1, 2, 3})` -> still folds correctly to `6`.

## Changelog

Added `.unreleased/bug-fixes/1691-fold-infinite-set.md`.


konnov (approved):
Thank you for handling this case!

Issue #1452, opened 2022-03-08
Expanding a set of functions


konnov:
In rare cases, we have to expand a set of functions. For instance in Queens:

https://github.com/informalsystems/apalache/blob/9b88fe7a6339135799a09d59f4daab62bd5308e5/test/tla/QueensTyped.tla#L46-L47

Currently, the model checker throws an exception:

```
QueensTyped.tla:46:32-46:45: rewriter error: Trying to expand a set of functions. This will blow up the solver. E@11:42:01.742
at.forsyte.apalache.tla.bmcmt.RewriterException: Trying to expand a set of functions. This will blow up the solver.
	at at.forsyte.apalache.tla.bmcmt.rules.SetExpandRule.apply(SetExpandRule.scala:27)
...
```

It looks like there is no workaround for Queens that would let us check the spec without expanding the set `[1..N -> 1..N]`.

konnov:
As a first step, we should add an explanation about why it is expensive to expand a set of functions

PR #3386, opened 2026-06-28
Expand the set argument of FoldSet (#3385)

### LINKED ISSUE #3385: FoldSet silently folds over zero elements for non-enumerated sets (SUBSET, function sets), returning the base value
## Description

`ApaFoldSet` (and the community-module folds built on it) **silently folds over zero elements and returns the base value** when the set argument is one that Apalache keeps *lazy / non-enumerated* rather than an explicit enumeration. This affects at least:

- a **powerset** `SUBSET S`, and
- a **function set** `[S -> T]`.

The fold produces a wrong result with no warning. This is a silent soundness problem in the same family as #1691 (folds over `Nat`/`Int`), but it is **not** caught by the fix for #1691: those sets are infinite (`InfSetT`), whereas a powerset/function set is *finite* and is represented by a different lazy cell (`PowSetT` / a function-set cell), whose `arena.getHas(...)` returns no member cells.

This is *more* dangerous than #1691, because the set really is finite, so a user has every reason to expect the fold to work, and the wrong value (the base value) is often a perfectly plausible number.

## Impact

Silent incorrect results from `FoldSet` over `SUBSET`/function sets. Any spec that aggregates over a powerset or a function set (counting, summing, building a record, etc.) can get a wrong answer and a bogus "no error" verdict. A workaround exists (`Expand`, see below), but the failure is silent, so users will not know to apply it.

## Input specification

**MWE 1 - fold over a powerset (expected 8, the number of subsets of a 3-element set):**

```tla
---------------------------- MODULE foldpow --------------------------
EXTENDS Integers, FiniteSets, Apalache
VARIABLE
  \* @type: Int;
  result

\* @type: (Int, Set(Int)) => Int;
CountOne(acc, s) == acc + 1

Init == result = ApaFoldSet(CountOne, 0, SUBSET {1, 2, 3})
Next == result' = result

\* Sound result is 8. The fold silently returns the base value 0 instead.
Inv == result = 8
=================================================================
```

**MWE 2 - fold over a function set (expected 4, the number of functions in `[{1,2} -> {1,2}]`):**

```tla
---------------------------- MODULE foldfunset --------------------------
EXTENDS Integers, FiniteSets, Apalache
VARIABLE
  \* @type: Int;
  result

\* @type: (Int, Int -> Int) => Int;
CountOne(acc, f) == acc + 1

Init == result = ApaFoldSet(CountOne, 0, [{1, 2} -> {1, 2}])
Next == result' = result

Inv == result = 4
=================================================================
```

```
apalache-mc check --inv=Inv --length=0 foldpow.tla
apalache-mc check --inv=Inv --length=0 foldfunset.tla
```

(default encoding `oopsla19`, solver `z3`)

## Expected behavior

Either:
1. fold over the actual elements of the (finite) set and return the correct result (`8` and `4` respectively) - this is what happens if the set is expanded first; or
2. if folding over a non-enumerated set is intentionally unsupported, fail loudly with a clear error, exactly like the fix for #1691 does for infinite sets.

What must **not** happen is silently returning the base value.

## Actual behavior

Both invariants are reported **violated** because `result = 0` (the base value), not `8` / `4`:

```
State 0: state invariant 0 violated.
The outcome is: Error
EXITCODE: ERROR (12)
```

Counterexample value (ITF) for MWE 1:

```json
"result": { "#bigint": "0" }
```

and likewise `0` for MWE 2.

## Workaround

Wrapping the set in `Apalache!Expand` forces enumeration and yields the correct result:

```tla
Init == result = ApaFoldSet(CountOne, 0, Expand(SUBSET {1, 2, 3}))   \* result = 8, Inv holds
```

## Likely root cause

`FoldSetRule` (`tla-bmcmt/.../rules/FoldSetRule.scala`) folds over `setState.arena.getHas(setCell)`. For a lazily-represented set cell (`PowSetT` for `SUBSET S`, the function-set cell for `[S -> T]`) `getHas` returns no member cells, so the fold iterates zero times and returns the base value - the same mechanism as the infinite-set case addressed in #1691, but for finite lazy sets. A guard analogous to the #1691 fix (reject, or force expansion) would resolve it.

## Log files

The relevant line from `detailed.log` / console for MWE 1: `state invariant 0 violated` with `result = 0`. Full output and ITF counterexample available on request.

## System information

- Apalache version: `main` @ `cf765e0` (VERSION `0.58.3-SNAPSHOT`, built on top of the `0.58.2` release)
- OS: Ubuntu (WSL2 on Windows 11)
- JDK version: OpenJDK Temurin 17.0.19

## Additional context

Related: #1691 (silent unsound folds over `Nat`, fixed in #3384) - this is the finite-lazy-set sibling of the same problem.

shalashaska117:
## What

Fixes #3385.

`ApaFoldSet` over a non-enumerated set silently returned the **base value** instead of the correct result:

- `ApaFoldSet(CountOne, 0, SUBSET {1,2,3})` returned `0` (should be `8`);
- `ApaFoldSet(CountOne, 0, [{1,2} -> {1,2}])` returned `0` (should be `4`).

This is a silent soundness problem in the same family as #1691, but for *finite* lazily-represented sets, so the #1691 fix does not catch it.

## Why

`FoldSetRule` folds over `state.arena.getHas(setCell)`. For a lazily-represented set cell - a powerset (`SUBSET S`) or a function set (`[S -> T]`) - `getHas` returns no member cells, so the fold iterated zero times and returned the base value. The root cause is one level up: `ExpansionMarker` decides which sets must be expanded into the arena (it already forces expansion for quantifiers, `filter`, `cup`/`UNION`, etc.), but `FoldSet` fell through to the generic case that propagates `shallExpand = false`, so the fold's set argument was never marked for expansion.

## How

Add an explicit `ApalacheOper.foldSet` case to `ExpansionMarker` that marks the **set argument** for expansion (`transform(true)`), leaving the operator and base value as-is. This mirrors how quantifiers and `filter` are handled.

Result:
- folds over a powerset now produce the **correct** result (`8`);
- folds over a set whose expansion is unsupported (function sets, see #1452) now **fail loudly** with `Trying to expand a set of functions...` instead of silently returning a wrong answer.

## Testing

- **New unit tests** in `TestExpansionMarker`: `ApaFoldSet(Op, v, SUBSET S)` and `ApaFoldSet(Op, v, [S -> T])` both mark the set with `Expand`.
- `TestExpansionMarker` (10) and the `FoldSet` rewriter tests pass; no regressions.
- Manual CLI verification (default `oopsla19`/`z3`):
  - `ApaFoldSet(CountOne, 0, SUBSET {1,2,3})` -> `result = 8` (was `0`).
  - `ApaFoldSet(CountOne, 0, [{1,2}->{1,2}])` -> loud `expand a set of functions` error (was silent `0`).
  - `ApaFoldSet(Plus, 0, {1,2,3})` (already-enumerated set) -> still `6`, unaffected.

## Changelog

Added `.unreleased/bug-fixes/3385-fold-non-enumerated-set.md`.

### DIFF
--- a/.unreleased/bug-fixes/3385-fold-non-enumerated-set.md
+++ b/.unreleased/bug-fixes/3385-fold-non-enumerated-set.md
@@ -1,0 +1,1 @@
+Fixed `FoldSet` silently returning the base value when folding over a non-enumerated set such as `SUBSET S` or `[S -> T]`. The set is now marked for expansion, so folds over powersets produce the correct result, and folds over sets whose expansion is unsupported fail loudly instead of returning a wrong answer, see #3385.
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExpansionMarker.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/analyses/ExpansionMarker.scala
@@ -88,6 +88,13 @@ class ExpansionMarker @Inject() (tracker: TransformationTracker) extends TlaExTr
       val tag = ex.typeTag
       OperEx(op, args.map(transform(true)): _*)(tag)
 
+    case ex @ OperEx(op @ ApalacheOper.foldSet, operEx, baseEx, setEx) =>
+      // FoldSet folds over the concrete elements of the set, so the set must be expanded.
+      // Otherwise a lazily-represented set (e.g. SUBSET S or [S -> T]) contributes no elements
+      // and the fold silently returns the base value. See #3385.
+      val tag = ex.typeTag
+      OperEx(op, transform(shallExpand)(operEx), transform(shallExpand)(baseEx), transform(true)(setEx))(tag)
+
     case ex @ OperEx(op @ TlaSetOper.filter, name, set, pred) =>
       // For the moment, we require the set to be expanded. However, we could think of collecting constraints on the way.
       val tag = ex.typeTag
--- a/tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/analyses/TestExpansionMarker.scala
+++ b/tla-bmcmt/src/test/scala/at/forsyte/apalache/tla/bmcmt/analyses/TestExpansionMarker.scala
@@ -117,4 +117,25 @@ class TestExpansionMarker extends AnyFunSuite with BeforeAndAfterEach {
 
     assert(input.build == output)
   }
+
+  // #3385: FoldSet folds over the concrete elements of the set, so the set must be expanded,
+  // otherwise a lazily-represented set contributes no elements and the fold silently returns the base value.
+  test("""marked: ApaFoldSet(Op, v, SUBSET S)""") {
+    val S = tla.name("S", intSetT)
+    def op = tla.lambda("A", tla.int(0), tla.param("acc", IntT1), tla.param("x", intSetT))
+    val input = tla.foldSet(op, tla.int(0), tla.powSet(S))
+    val output = marker.apply(input)
+    val expected = tla.foldSet(op, tla.int(0), tla.expand(tla.powSet(S)))
+    assert(expected.build == output)
+  }
+
+  test("""marked: ApaFoldSet(Op, v, [S -> T])""") {
+    val S = tla.name("S", intSetT)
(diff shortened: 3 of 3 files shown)


konnov (approved):
Thank you for discovering this issue and submitting a fix.

PR #3392, opened 2026-06-28
Simplify more set expressions in ConstSimplifier (#1238)

### LINKED ISSUE #1238: [FEATURE] Constant simplification over sets
`ConstSimplifier` already handles many arithmetical and logical simplifications, but does nothing over sets.

Examples:
- `{1, 2} \union {3}` could be simplified to `{1,2,3}`
- `{1, 2} \setminus {2}` could be simplified to `{1}`

And many more, see set operators reference: https://github.com/informalsystems/apalache/blob/unstable/tlair/src/main/scala/at/forsyte/apalache/tla/lir/oper/TlaSetOper.scala

shalashaska117:
## What

Fixes #1238.

`ConstSimplifier` already folds `\cup` / `\cap` / `\setminus` and `Cardinality` over set **enumerations of literals**. This adds simplifications that hold for an **arbitrary** set `S`, which the literal-only rules do not cover:

- `x \in {}` -> `FALSE`
- `{} \subseteq S` -> `TRUE`, `S \subseteq S` -> `TRUE`
- `{} \cup S` -> `S`, `S \cup {}` -> `S`, `S \cup S` -> `S`
- `{} \cap S` -> `{}`, `S \cap {}` -> `{}`, `S \cap S` -> `S`
- `{} \ S` -> `{}`, `S \ {}` -> `S`, `S \ S` -> `{}`

These are all sound semantic identities and apply even when `S` is a variable or a non-enumeration expression (the existing literal rules bail unless both operands are concrete enumerations).

## How

Added cases to `ConstSimplifierBase.simplifyShallow`, placed before the existing literal-only set rules so the empty-set / idempotence patterns fire first. The empty set is matched as the zero-argument enumerator `OperEx(TlaSetOper.enumSet)` (the same idiom already used elsewhere in the file), and `S \ S` returns an empty set carrying the original expression's type tag.

## Testing

- New unit tests in `TestConstSimplifier` covering all the cases above (`x \in {}`, the two `\subseteq` cases, and union/intersection/setminus with the empty set and with `S` itself).
- Full `TestConstSimplifier` suite passes (40 tests, 0 failures).
- Note: the change lives in `ConstSimplifierBase`, shared with `ConstSimplifierForSmt`, so the SMT-side simplifier benefits too; the additions are pure semantic identities.

## Changelog

Added `.unreleased/features/1238-const-simplify-sets.md`.

### DIFF
--- a/.unreleased/features/1238-const-simplify-sets.md
+++ b/.unreleased/features/1238-const-simplify-sets.md
@@ -1,0 +1,1 @@
+`ConstSimplifier` now simplifies more set expressions: membership in the empty set (`x \in {}` becomes `FALSE`), trivial subset checks (`{} \subseteq S` and `S \subseteq S` become `TRUE`), empty-set identities for an arbitrary set (`{} \cup S`, `S \cup {}`, `{} \cap S`, `S \cap {}`, `{} \ S`, `S \ {}`), and self-operations (`S \cup S`, `S \cap S`, `S \ S`), see #1238.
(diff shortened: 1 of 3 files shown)


konnov (approved):
Nice! Thanks for handling this issue!

PR #3393, opened 2026-06-28
Add unit tests for DefaultValueCache (#1544)

### LINKED ISSUE #1544: Introduce tests for DefaultValueCache
Although the code of `DefaultValueCache` is quite simple, we had a hard-to-debug bug in it #1543. We should introduce unit tests for this cache.

shalashaska117:
## What

Fixes #1544.

`DefaultValueCache` had a hard-to-debug bug (#1543) but no unit tests. This adds them.

## How

Added a `TestDefaultValueCache` trait (in `tla-bmcmt`, package `...bmcmt.caches`), mixed into the existing `TestRewriterWithOOPSLA19` and `TestRewriterWithCvc5OOPSLA19` suites alongside `TestDefaultValueFactory`. It asserts the core cache contract:

- `getOrCreate` for the **same** type returns the **same** cell (caching works);
- `getOrCreate` for **distinct** types returns **distinct** cells;
- `get` returns `None` before creation and the cached cell afterwards, and the cached cell has the requested type.

## Testing

- The 3 new tests pass under the OOPSLA19 (Z3) suite (and are wired into the cvc5 suite too).
- Tests-only change: no observable behavior changes, so no `.unreleased/` entry is added (per CONTRIBUTING).
- `make fmt-fix` applied; test sources compile cleanly with `APALACHE_FATAL_WARNINGS=true`.


konnov (approved):
Having basic tests for `DefaultValueCache` helps. Thank you for adding them

PR #3394, opened 2026-06-28
Allow string or integer keys for TLCGet/TLCSet (#3274)

### LINKED ISSUE #3274: Type checking incompatible with TLCSet/TLCGet(Str) and TLCSet/TLCGet(Int)
`TLCGet` and `TLCSet` accept both `Str` and `Int` as their (first) argument — `Str` for named properties (`TLCGet("level"), TLCGet("config")`) and `Int` for numbered registers (`TLCGet(2)`). Snowcat cannot assign a single type to the parameter, so any operator that mixes both call forms fails type checking.

shalashaska117:
## What

Fixes #3274.

`TLCGet`/`TLCSet` accept both a **string** (named register, e.g. `TLCGet("level")`) and an **integer** (numbered register, e.g. `TLCGet(2)`) as their key argument. Their type annotations in `src/tla/__rewire_tlc_in_apalache.tla` fixed the key to `Int`, so Snowcat (which has no union types) rejected any operator mixing both forms:

```
The operator TLCSet of type ((Int, a) => Bool) is applied to arguments of incompatible types in TLCSet("level", 7):
Argument "level" should have type Int but has type Str.
```

## How

Make the key argument polymorphic, as suggested by @konnov / @lemmy in the issue:

- `TLCGet`: `Int => a` -> `b => a`
- `TLCSet`: `(Int, a) => Bool` -> `(b, a) => Bool`

Each call site instantiates the key type `b` independently (`Str` or `Int`), while the value type `a` is inferred as before. This is strictly more permissive, so existing specs using numbered registers are unaffected.

## Testing

- Reproduced the failure first: a spec mixing `TLCSet(1, 42)`, `TLCSet("level", 7)`, and `TLCGet(1) = TLCGet("level")` failed type checking on `main`; it now type-checks.
- Edge cases verified via `apalache-mc typecheck`:
  - Int key + Int value used arithmetically -> OK
  - Str key + Str value -> OK
  - Str key + Int value -> OK
- Added a regression fixture `test/tla/Bug3274.tla` and a `typecheck` integration test in `test/tla/cli-integration-tests.md` ("typecheck accepts string and integer TLC registers").
- `TestSanyImporterStandardModules` (which imports the TLC module) still passes (10 tests).

## Changelog

Added `.unreleased/bug-fixes/3274-tlc-register-types.md`.

### DIFF
--- a/.unreleased/bug-fixes/3274-tlc-register-types.md
+++ b/.unreleased/bug-fixes/3274-tlc-register-types.md
@@ -1,0 +1,1 @@
+Fixed type checking of `TLCGet`/`TLCSet`: the register key may be a string (named register, e.g. `TLCGet("level")`) or an integer (numbered register, e.g. `TLCGet(2)`). The key argument is now polymorphic, so an operator mixing both forms type-checks, see #3274.
--- a/src/tla/__rewire_tlc_in_apalache.tla
+++ b/src/tla/__rewire_tlc_in_apalache.tla
@@ -68,18 +68,22 @@ JavaTime ==
     123
 
 \* Apalache does not support this operator, but the type checker does.
+\* The register key may be an integer (numbered register) or a string (named
+\* register), so its type is left polymorphic, see #3274.
 \*
-\* @type: Int => a;
+\* @type: b => a;
 TLCGet(__i) ==
     LET \* @type: () => Set(a);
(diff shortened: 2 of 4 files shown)


thpani (approved):
LGTM

PR #3395, opened 2026-06-28
Skip the initial-state invariant check when --init equals --inv (#1825)

### LINKED ISSUE #1825: Don't check Inv in state 0 with --init=Inv
Apalache checks that `Inv` holds in state 0 even with option `--init=Inv`. This is unnecessary but takes time, so I suggest skipping this check.

shalashaska117:
## What

Fixes #1825.

When checking an inductive invariant, one runs `--init=IndInit --inv=IndInv` with
`IndInit == TypeOK /\ IndInv`, or `--init=IndInv --inv=IndInv` as in the manual. Apalache then
checks `IndInv` in state 0, where it holds by construction. On large specs that query dominates
the run: the issue reports losing half a day to it, and the workaround so far was
`--tuning-options='search.invariantFilter=1->.*'`.

This PR makes it automatic: an invariant that is a conjunct of the init predicate is not checked
in state 0.

## How

- `VCGenerator.findInvariantsImpliedByInit` splits the init predicate into conjuncts with the same
  `splitInv` that produces the verification conditions, and returns the numbers of the conditions
  whose body is one of those conjuncts. It runs in `VCGenPass`, which is where
  `IndInv == A /\ B /\ C` is decomposed into `A`, `B`, and `C`.
- The bodies are compared right after inlining. Two copies of a conjunct then differ only in the
  counters that unique renaming gave to their bound variables (`\A i$1` vs `\A i$2`), so both are
  renamed from scratch before the comparison. Only syntax is compared, so the search finds fewer
  invariants than a semantic check would, never more.
- `VCGenPass` stores the numbers in `DerivedPredicates`, `BoundedCheckerPass` hands them to
  `ModelCheckerParams`, and `SeqModelChecker` leaves them out of the state invariants in the Init
  step. Later steps are untouched: `mayChangeAssertion` still forces a check at every step whose
  transition may change the invariant.

### Why it is sound

The initial states satisfy the init predicate, hence each of its conjuncts. `splitInv` only
descends into `/\`, `\A`, and labels, so a conjunct is never taken from under a `\/`.

The skip set holds numbers, not names, because VCGen numbers the conditions of each `--inv`
operator from 0: with `INVARIANT A B` in a TLC config, `VCInv$0` exists twice. The numbers come
from `ModuleAdapter.getDeclsFromSpec`, the same ordering that `BoundedCheckerPass` uses to build
the list of invariants.

## Testing

- `TestVCGenerator`: conjunct of init, partial match, no match, init under a disjunction, bound
  variables renamed apart, action invariant.
- `TestSeqModelCheckerTrait`: a state-0 violation is skipped, a violation after `Next` is still
  reported, only the listed invariants are skipped.
- `test/tla/IndInv1825.tla` with two CLI tests in `cli-integration-tests.md`.
- On `MC_n6t1f0_inductive.tla` from
  [konnov/apalache-examples](https://github.com/konnov/apalache-examples/tree/main/ben-or83), all
  15 verification conditions of `IndInv` are recognized as following from `IndInit`, and
  `check --init=IndInit --inv=IndInv --length=0` takes 1 min 40 s. On `main`, the same command was
  still on the 4th of the 15 state-0 checks after 17 minutes.
- By hand: `--init=IndInv --inv=IndInv`, the untouched `--init=Init --inv=IndInv --length=0`
  query, a partial match where the remaining conjunct is violated in state 0, two invariants from
  a TLC config where only one follows from init, both settings of `search.invariant.mode`, and
  `--discard-disabled=false`.
- `make test` passes, no warnings with `APALACHE_FATAL_WARNINGS=true`, `make fmt-fix` applied.

## Note for maintainers

The issue discussion also floated a broader `--profile=inductive-step`. This PR stays with the
smallest automatic behavior that matches the issue title. Happy to put it behind a flag or fold it
into a profile instead.

## Changelog

`.unreleased/features/1825-skip-initial-inv-check.md`.


konnov:
I do not see how it is supposed to work with practical inductive invariants. Typically, an inductive invariant is a conjunction of smaller predicates like `IndInv == A /\ B /\ C`. When this is the case, `VCGenerator` produces smaller invariants for `~A`, `~B`, `~C`. Hence, checking that `Init` and the invariant are the same thing in `BoundedCheckerPassImpl` is too late. A lot of preprocessing has been done already.

Hence, the proposed solution probably works on small specifications, but not on those where it matters. For example, you could try it on [Ben_or83_inductive.tla](https://github.com/konnov/apalache-examples/blob/main/ben-or83/Ben_or83_inductive.tla) and see whether the proposed fix works as expected.

shalashaska117:
> I do not see how it is supposed to work with practical inductive invariants. Typically, an inductive invariant is a conjunction of smaller predicates like `IndInv == A /\ B /\ C`. When this is the case, `VCGenerator` produces smaller invariants for `~A`, `~B`, `~C`. Hence, checking that `Init` and the invariant are the same thing in `BoundedCheckerPassImpl` is too late. A lot of preprocessing has been done already.
> 
> Hence, the proposed solution probably works on small specifications, but not on those where it matters. For example, you could try it on [Ben_or83_inductive.tla](https://github.com/konnov/apalache-examples/blob/main/ben-or83/Ben_or83_inductive.tla) and see whether the proposed fix works as expected.

I will try that and update this PR, but I have to check all the pipeline of the preprocessing done, just to understand where to solve the problem better

shalashaska117:
You were right, the check never fired for a real inductive invariant. I ran
`MC_n6t1f0_inductive.tla` from your examples repo: with `--init=IndInit --inv=IndInv` the two names
differ, so the flag stayed `false`. And even for equal names, `BoundedCheckerPassImpl` was the
wrong place, as you said, since by then `IndInv == Lemma2 /\ ... /\ Lemma1c` has already become 15
verification conditions.

The comparison now happens in `VCGen`, where the split happens, and per conjunct:

- `VCGenerator.findInvariantsImpliedByInit` splits the init predicate with the same `splitInv` that
  produces the verification conditions, and returns the numbers of the conditions whose body is one
  of those conjuncts.
- The bodies are compared right after inlining, where two copies of a conjunct differ only in the
  counters that unique renaming gave to their bound variables (`\A i$1` vs `\A i$2`), so both are
  renamed from scratch first.
- `VCGenPass` stores the numbers in `DerivedPredicates`, and `SeqModelChecker` leaves those
  invariants out in the Init step only. A conjunct that does not occur in the init predicate is
  still checked in state 0.

It keeps numbers, not names, because VCGen numbers the conditions of each `--inv` operator from 0:
with `INVARIANT A B` in a TLC config, `VCInv$0` exists twice.

On `MC_n6t1f0_inductive.tla`, `check --init=IndInit --inv=IndInv --length=0`:

```
> VCGen produced 15 verification condition(s)
> 15 verification condition(s) follow from IndInit
State 0: skipping 15 state invariant(s) that follow from Init
The outcome is: NoError
```

That run takes 1 min 40 s, nearly all of it in the Init transition itself. On `main`, the same
command was still on the 4th of the 15 state-0 checks (`Lemma11a`) after 17 minutes, so I stopped
it.

Also checked: `--init=IndInv --inv=IndInv` (the form in `running.md`) skips all conditions in state
0; the IndInit query `--init=Init --inv=IndInv --length=0` is untouched; when `IndInit` omits one
conjunct of `IndInv`, that conjunct is still checked in state 0 and its violation is reported, and
the same holds for two invariants from a TLC config where only one follows from init. Both settings
of `search.invariant.mode` and `--discard-disabled=false` behave the same.

New tests: `TestVCGenerator` for the matching, `TestSeqModelCheckerTrait` for the skip set, and two
CLI tests on `IndInv1825.tla`. `make test` passes. I could not run the mdx tests here (no
`ocaml-mdx` in my environment), so those two blocks only match output that I captured by hand.

On authorship: this PR was written with Claude Code, and that covers all of the text as opposed to
the code, namely the `.md` files (the changelog entry, the note in `running.md`, the test
descriptions in `cli-integration-tests.md`), the commit messages, and this comment.

konnov:
@shalashaska117 before you ask your AI agent for a fix, let's discuss the problem that you are trying to fix and the potential solutions to this problem. The issue #1825 is real, though this specific issue was hit by only 1-2 people in the past. The real reason for not fixing it is not that we could not use AI tools to produce a fix, but that we have not come up with a clean solution.

Certainly, the simplest solution would be to exclude invariant checking at step 0, when the invariant and the initialization predicate are the exactly same. This would solve the issue #1825 on the surface. However, it would not solve the deeper issue that we see in practice, namely, when the initialization predicate is stronger than the inductive invariant (when checking inductiveness).

I am happy to discuss the solution space in [GitHub discussions](https://github.com/apalache-mc/apalache/discussions). However, I would discuss this with a human, not an AI tool.

Closing this PR. As we have already discussed in the [AI policy](https://github.com/apalache-mc/apalache/blob/main/AI_POLICY.md), AI-generated comments do not help this project.

shalashaska117:
> @shalashaska117 before you ask your AI agent for a fix, let's discuss the problem that you are trying to fix and the potential solutions to this problem. The issue #1825 is real, though this specific issue was hit by only 1-2 people in the past. The real reason for not fixing it is not that we could not use AI tools to produce a fix, but that we have not come up with a clean solution.
> 
> Certainly, the simplest solution would be to exclude invariant checking at step 0, when the invariant and the initialization predicate are the exactly same. This would solve the issue #1825 on the surface. However, it would not solve the deeper issue that we see in practice, namely, when the initialization predicate is stronger than the inductive invariant (when checking inductiveness).
> 
> I am happy to discuss the solution space in [GitHub discussions](https://github.com/apalache-mc/apalache/discussions). However, I would discuss this with a human, not an AI tool.
> 
> Closing this PR. As we have already discussed in the [AI policy](https://github.com/apalache-mc/apalache/blob/main/AI_POLICY.md), AI-generated comments do not help this project.

ok, i will directly comment myself, I just wanted to get as formal as possible, being that english is not my first language. My AI agent is only used for draft and testing, the code is being written by me. I will look to improve my next PR pointing for good as first issues tasks.

Issue #3116, opened 2025-07-19
OptionCase without caseNone type annotation causes a confusing "Unexpected type" RewriterException


edwardcwang:
## Description

When using `OptionCase` without manually writing a type annotation on the `caseNone` lambda, we get a frustrating/confusing error message. If the error is intended (preferably it should work out of the box, as it works with TLC), the user is given no useful information that points towards a missing type annotation.

```
problemOperator(self) ==
  LET q1 == operator1(self) IN
  IF IsNone(q1) THEN None ELSE (
    LET v1 == OptionGetOrElse(q1, self) IN
    OptionCase(operator2(self, v1),
      \* @type: Int => Some(<<Int, Int>>) | None(UNIT);
      LAMBDA v2: Some(<<v1, v2>>),
      \* Broken without below type annotation
      \* type: UNIT => Some(<<Int, Int>>) | None(UNIT);
      LAMBDA u: None
    )
  )
```

Error message:

```
PASS #13: BoundedChecker                                          I@22:59:22.961
Step 0: picking a transition out of 1 transition(s)               I@22:59:23.667
<unknown>: rewriter error: Unexpected type a140 when generating a default value E@22:59:23.756
at.forsyte.apalache.infra.AdaptedException: <unknown>: rewriter error: Unexpected type a140 when generating a default value
	at at.forsyte.apalache.infra.passes.PassChainExecutor$.run(PassChainExecutor.scala:39)
	at at.forsyte.apalache.tla.tooling.opt.CheckCmd.run(CheckCmd.scala:136)
	at at.forsyte.apalache.tla.Tool$.runCommand(Tool.scala:139)
	at at.forsyte.apalache.tla.Tool$.run(Tool.scala:119)
	at at.forsyte.apalache.tla.Tool$.main(Tool.scala:40)
	at at.forsyte.apalache.tla.Tool.main(Tool.scala)
Please report an issue at [https://github.com/informalsystems/apalache/issues]: at.forsyte.apalache.infra.AdaptedException: <unknown>: rewriter error: Unexpected type a140 when generating a default value
```

## Impact

Can be worked around, but very frustrating/confusing error message for the user.

## Input specification

```
\* apalache-mc check --debug --write-intermediate OptionCaseBug.tla 

------------------------------- MODULE OptionCaseBug -------------------------------
EXTENDS Integers, Sequences, FiniteSets, TLC, Option

VARIABLES
  \* @type: Int -> Str;
  myPc

[...]

PR #3396, opened 2026-06-28
Give a clear error for a polymorphic default value (#3116)

shalashaska117:
## What

Addresses #3116.

Two parts:

1. **The confusing error is fixed.** `DefaultValueFactory.makeUpValue` threw `Unexpected type a140 when generating a default value` as a `RewriterException` (which prints "Please report an issue", implying an internal bug) when it met an uninstantiated type variable. This happens for a polymorphic type, e.g. an un-annotated `caseNone` of `OptionCase` that returns `None`. It is not an internal bug, but a sign the spec is not fully monomorphic. It now raises a `RewriterKnownLimitationError` with an actionable message:

   ```
   Cannot generate a default value for the polymorphic type a140. This usually means a type annotation is missing, e.g., on a `None`/`caseNone` branch of a variant. Please add an explicit type annotation.
   ```

   The generic `RewriterException` is kept for genuinely unexpected (non-variable) types.

2. **The originally-reported spec now works.** On the current `main` (`cf765e0`), the spec from the issue **no longer reproduces the crash** - it model-checks to `NoError`. I added it as a regression fixture (`test/tla/Github3116.tla`) + a `check` integration test so it stays fixed.

## Testing

- Reproduced first: confirmed the issue's spec runs to `NoError` on `main` (no `Unexpected type` crash), i.e. the specific scenario is already resolved upstream.
- **New unit test** in `TestDefaultValueFactory`: `makeUpValue` on a polymorphic type (`VarT1`) throws `RewriterKnownLimitationError` (the actionable error), not a bug-report `RewriterException`.
- **New integration test** in `cli-integration-tests.md`: `check Github3116.tla` succeeds (`EXITCODE: OK`).
- DefaultValueFactory / rewriter tests pass; compiles cleanly with `APALACHE_FATAL_WARNINGS=true`.

## Note

If maintainers would rather make the polymorphic case "just work" (instantiate free type variables to a default monotype before model checking), that is a larger type-system change; this PR takes the smaller step the issue explicitly accepts ("a more user-friendly error instructing the user on how to fix the problem") and locks in the already-fixed scenario with a regression test.

## Changelog

Added `.unreleased/bug-fixes/3116-default-value-polymorphic-error.md`.

### DIFF
--- a/.unreleased/bug-fixes/3116-default-value-polymorphic-error.md
+++ b/.unreleased/bug-fixes/3116-default-value-polymorphic-error.md
@@ -1,0 +1,1 @@
+Generating a default value for a polymorphic (uninstantiated) type now raises a clear "known limitation" error advising to add a type annotation, instead of a confusing internal `Unexpected type ... when generating a default value` error that asked the user to report a bug (e.g. with an un-annotated `caseNone` of `OptionCase`), see #3116.
--- a/test/tla/Github3116.tla
+++ b/test/tla/Github3116.tla
@@ -1,0 +1,50 @@
+------------------------------- MODULE Github3116 -------------------------------
+\* Regression test for https://github.com/apalache-mc/apalache/issues/3116:
+\* OptionCase with an un-annotated caseNone (returning None) used to crash the
+\* model checker with a confusing "Unexpected type ... when generating a default value".
+EXTENDS Integers, Sequences, FiniteSets, TLC, Option
+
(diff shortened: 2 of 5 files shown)


thpani:
This was a type-checker bug that has been fixed, probably in #3207.
I don't see how this PR contributes anything useful, as the supplied regression test doesn't even trigger the newly introduced exception.

PR #3401, opened 2026-07-08
Fix SANY parse error reporting

konnov:
Closes #3400. After upgrade to tla2tools 1.8, `SanyImporter` started to produce unreadable error messages. Fixing that. The fix is done with Codex GPT 5.5.

- [x] Tests added for any new code
- [x] Ran `make fmt-fix` (or had formatting run automatically on all files edited)
- [x] [Entries added to `./unreleased/`][changelog format] for any new functionality

[changelog format]: https://github.com/apalache-mc/apalache/blob/main/CONTRIBUTING.md#how-to-record-a-change

Issue #2816, opened 2024-01-22
Typechecking crashes


nano-o:
## Description

I am trying to set things up to model-check [Voting.tla](https://github.com/nano-o/TLA--Examples/blob/ApaVoting_with_cfg/specifications/Paxos/Voting.tla) with Apalache. I would like to reuse Voting.tla as much as possible, so I created [ApaVoting.tla](https://github.com/nano-o/TLA--Examples/blob/ApaVoting_with_cfg/specifications/Paxos/ApaVoting.tla), which instantiates Voting.tla, and [ApaVoting.cfg](https://github.com/nano-o/TLA--Examples/blob/ApaVoting_with_cfg/specifications/Paxos/ApaVoting.cfg), which substitutes 2 definitions of Voting.tla that are problematic for Apalache. However Apalache crashes as below.

See this PR for context: https://github.com/tlaplus/Examples/pull/112

## Impact

I have to duplicate most of Voting.tla in order to analyze with Apalache instead of being able to reuse it. See Voting2.tla and ApaVoting2.tla

## The command line parameters used to run the tool

```
script/run-docker.sh typecheck  ApaVoting.tla
```

## Expected behavior

No error

## Log files

<details>

```
2024-01-22T23:36:12,394 [main] INFO  a.f.a.t.Tool\$ - # APALACHE version: 0.44.2 | build: v0.44.2
2024-01-22T23:36:12,401 [main] INFO  a.f.a.t.t.o.TypeCheckCmd - Type checking ApaVoting.tla
2024-01-22T23:36:12,491 [main] INFO  a.f.a.i.p.PassChainExecutor\$ - PASS #0: SanyParser
2024-01-22T23:36:12,822 [main] DEBUG a.f.a.i.p.PassChainExecutor\$ - PASS #0: SanyParser [OK]
2024-01-22T23:36:12,823 [main] INFO  a.f.a.i.p.PassChainExecutor\$ - PASS #1: TypeCheckerSnowcat
2024-01-22T23:36:12,823 [main] INFO  a.f.a.t.p.t.EtcTypeCheckerPassImpl -  > Running Snowcat .::.
2024-01-22T23:36:13,037 [main] ERROR a.f.a.t.Tool\$ - Unhandled exception
java.lang.IllegalArgumentException: Unsupported expression: ∀g: (∀h: (((C!WFInductiveDefines(g, S, Def)) ∧ (C!WFInductiveDefines(h, S, Def))) ⇒ (g = h)))
[...]

PR #3402, opened 2026-07-09
[release] 0.58.3

coffeeinprogress:
# Reviewer instructions

- Check the changelog to ensure the version increment is consistent with https://semver.org/.
  - If a different version should be released, follow [the instructions](https://github.com/informalsystems/apalache/blob/main/CONTRIBUTING.md#via-github).
- Review the changeset as a sanity check.
- Approve the PR, if it pases muster.
- **Merge** this branch (do not squash or rebase), but **DO NOT DELETE THE BRANCH**.
- After the release has been published to github, delete this branch.

---

# Release notes

### Features

- `ConstSimplifier` now simplifies more set expressions: membership in the empty set (`x \in {}` becomes `FALSE`), trivial subset checks (`{} \subseteq S` and `S \subseteq S` become `TRUE`), empty-set identities for an arbitrary set (`{} \cup S`, `S \cup {}`, `{} \cap S`, `S \cap {}`, `{} \ S`, `S \ {}`), and self-operations (`S \cup S`, `S \cap S`, `S \ S`), see #1238.

### Bug fixes

- Fixed type checking of `TLCGet`/`TLCSet`: the register key may be a string (named register, e.g. `TLCGet("level")`) or an integer (numbered register, e.g. `TLCGet(2)`). The key argument is now polymorphic, so an operator mixing both forms type-checks, see #3274.
- Fixed `FoldSet` silently returning the base value when folding over a non-enumerated set such as `SUBSET S` or `[S -> T]`. The set is now marked for expansion, so folds over powersets produce the correct result, and folds over sets whose expansion is unsupported fail loudly instead of returning a wrong answer, see #3385.
- Fix the SANY parsing errors (#3401)
- Fixed a crash (`IllegalArgumentException: Unsupported expression`) in the type checker on unbounded quantifiers `\A x: P` and `\E x: P`. These expressions are now type-checked like their bounded counterparts, and are reported with a proper, source-located error only if they reach the model checker, see #2816.


konnov (approved):
ok

PR #3404, opened 2026-07-29
Introduce random seed in simulate

### LINKED ISSUE #2053: Remove unused fields from `ModelCheckerParams`

### LINKED ISSUE #2054: `search.simulation` wired via tuning options

### LINKED ISSUE #2083: Add an option to seed the random seed in `simulate`
Currently, the command `simulate` is using randomization to pick the next action. To make the experiments reproducible, we should introduce an option to set the seed.

konnov:
Closes #2083. Closes #2054. Closes #2053. This PR adds `--seed=N` to the commands `simulate` and `check`. This seed is used by `simulate` for random selection of the next transition. In addition, both in `simulate` and `check`, if SMT seeds are not set, then `--seed=N` is also used to set the SMT seeds. Seed generation is now done at the configuration phase, if the seed is not specified.

To address #2054, we introduce `searchKind` and `maxRun` directly in the Patch/Validated pair, so the options of `simulate` are resolved by the configuration loader, instead of the late resolution in `FineTuningOptions`.

This feature is implemented with Codex GPT-5.6-sol xhigh/max and then manually adapted.


thpani (commented):
I don't understand why we need 3 different tuning options to set SMT seeds.

I think it makes more sense to make a breaking change and introduce a new global option that replaces the existing ones.

konnov:
> I don't understand why we need 3 different tuning options to set SMT seeds.
> 
> I think it makes more sense to make a breaking change and introduce a new global option that replaces the existing ones.

I agree. We can leave one option for setting the SMT seed. The issue is that `simulate --seed=N` sets the seed for transition selection. I thought of deriving the SMT seed from this one, if no explicit SMT seed is provided. Do you think we should keep the default SMT seed, if `--seed` is provided?

thpani:
If we have a first-class `--seed` argument, it should be the only way to set the seed, I think

konnov:
> If we have a first-class `--seed` argument, it should be the only way to set the seed, I think

This sounds reasonable. I don't think we have users who need particular seed settings for SMT solvers.

konnov:
OK, fixed this PR to use `--seed=N` and its internal mapping to `search.seed` as the only way to set the seeds of the transition selection and SMT.

konnov:
I've done the further options refactoring, since we've merged #3409

.unreleased/breaking-changes/typed-search-controls.md
```diff
@@ -0,0 +1,4 @@
+Replace the tuning keys `search.seed`, `search.simulation`,
+`search.simulation.maxRun`, and `search.outputTraces` with the typed checker
```
thpani: afaict, we are also retiring `smt.randomSeed`?

docs/src/apalache/config.md
```diff
@@ -86,6 +89,10 @@ In the table below, a default of "none" means that the value is optional.
 |               | `output`             | Write the processed module to this file.                     | path string                  | none; `.tla` and `.json` are supported      |
```
thpani, line 92: now that I'm reading this again, I'm not sure how the options are grouped. e.g., `source` and `output` are top-level, but `server` won't be using them?

docs/src/apalache/running.md
```diff
@@ -79,6 +80,8 @@ The arguments are as follows:
       and integer arithmetic; `arrays` (experimental) and `funArrays` (experimental) use SMT arrays with extensionality.
       This parameter can also be set via the `SMT_ENCODING` environment variable. See the [alternative SMT encoding
       using arrays] for details.
+    - `--seed=NUM` sets a nonnegative search seed. The seed initializes SMT solving and, with `simulate`, is also used
```
thpani: I think it would be useful to specify the datatype.
I frequently find myself wondering (for other tools) from what range I should sample seeds

mod-tool/src/test/scala/at/forsyte/apalache/tla/tooling/opt/TestCommandConfig.scala
```diff
@@ -69,25 +69,52 @@ class TestCommandConfig extends AnyFunSuite {
...
-    assert(explicit.checker.tuning.flatMap(_.get("search.outputTraces")).contains("false"))
+    assert(explicit.checker.seed.contains(42))
+    assert(explicit.checker.outputTraces.contains(false))
+  }
+
+  test("seed is validated during resolution") {
+    val invalid = new CheckCmd()
+    invalid.read(List("CommandConfig.tla"))
+    invalid.seed = Some(-1)
+    val invalidResult = ApalacheConfigResolver.resolveCheck(invalid.toConfig.requireValue())
```
thpani, line 110: probably also worth testing overflow

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/search/ModelCheckerParams.scala
```diff
@@ -34,7 +29,11 @@ object ModelCheckerParams {
 class ModelCheckerParams(
     checkerInput: CheckerInput,
     val stepsBound: Int,
-    tuningOptions: Map[String, String] = Map()) {
+    tuningOptions: Map[String, String] = Map(),
+    val searchKind: SearchKind = Check,
+    val seed: Int = 0,
```
thpani: perhaps my ignorance of the new options system, but why does this take a default parameter? wouldn't it more appropriate to force the caller to come op with a value? or alternatively, to accept Option[Int]?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/search/ModelCheckerParams.scala
```diff
@@ -34,7 +29,11 @@ object ModelCheckerParams {
...
+    val maxRun: Int = 100,
```
thpani: similar here; afaict this only applies to `simulate`? should it be `Option[Int]`?

I find it odd that the default `100` is overridden in `SearchState` for `.searchKind != Simulate`. Certainly doesn't help with debugging.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/search/ModelCheckerParams.scala
thpani: otherwise, we should say in the docs that the default seed is always fixed at 0

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigResolver.scala
```diff
@@ -101,7 +102,16 @@ object ApalacheConfigResolver {
     val solver = requireDefault(checkerWithDefaults.smtSolver, s"$CHECKER.$SMT_SOLVER")
     val encoding = requireDefault(checkerWithDefaults.smtEncoding, s"$CHECKER.$SMT_ENCODING")
     val maxError = requireDefault(checkerWithDefaults.maxError, s"$CHECKER.$MAX_ERROR")
+    val maxRun = requireDefault(checkerWithDefaults.maxRun, s"$CHECKER.$MAX_RUN")
 
+    configuredChecker.seed.foreach { seed =>
+      if (seed < 0) {
+        errors += s"Option $CHECKER.$SEED must be nonnegative, but got $seed."
+      }
```
thpani, line 116: what's the upper bound? I'm not even sure it's uniform between cvc5 and z3

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigResolver.scala
```diff
@@ -101,7 +102,16 @@ object ApalacheConfigResolver {
...
+    }
+    if (maxRun <= 0) {
+      errors += s"Option $CHECKER.$MAX_RUN must be positive, but got $maxRun."
```
thpani, line 119: should we also check `!(searchKind = simulate) => maxRun == 1`?

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigResolver.scala
```diff
@@ -273,6 +289,9 @@ object ApalacheConfigResolver {
   private def requireDefault[A](value: Option[A], field: String): A =
     value.getOrElse(throw new IllegalStateException(s"Missing built-in default for $field"))
 
+  /** Generate one concrete nonnegative seed for an otherwise unseeded checker run. */
+  private def generateSeed(): Int = ThreadLocalRandom.current().nextInt() & Int.MaxValue
```
thpani, line 304:
```suggestion
  /** Generate one concrete nonnegative seed for an otherwise unseeded checker run. */
  // Java `int`s are in two's complement, so sampling remains uniform
  private def generateSeed(): Int = ThreadLocalRandom.current().nextInt() & Int.MaxValue
```

docs/src/adr/025adr-cvc5-solver-backend.md
```diff
@@ -120,9 +120,9 @@ The initial implementation should be intentionally conservative:
 2. Preserve the current cell/sort naming conventions so that existing logging, recording, and model-decoding assumptions
    remain easy to compare against Z3.
 3. Add a solver choice to configuration separately from `smtEncoding`; those are independent dimensions.
-4. Keep backend-neutral SMT controls separate from solver-specific options. For example, `smt.randomSeed`,
-   `smt.statsSec`, and SMT timeouts are concepts Apalache can define once and map to each backend. Low-level solver
-   parameters should remain namespaced, for example `z3.*` and `cvc5.*`.
+4. Keep backend-neutral SMT controls separate from solver-specific options. For example, `checker.seed`, `smt.statsSec`,
+   and SMT timeouts are concepts Apalache can define once and map to each backend. Low-level solver parameters should
+   remain namespaced, for example `z3.*` and `cvc5.*`.
```
thpani: I'm a bit torn about modifying ADRs in this way.

This particular edit is certainly useful, but in principle we have to assume that ADRs are historical.

thpani (commented):
Looks mostly good, but I think we need another round on this

docs/src/adr/025adr-cvc5-solver-backend.md
konnov: True. If we want to be really strict, we can introduce a revision history like in standards. Though it looks like an overkill in this case.

docs/src/apalache/config.md
konnov: I realized that `source` and `output` are a real security risk in the server mode. Not that Apalache itself and its server APIs are really secure, but letting an arbitrary server request to read and write arbitrary files is just asking for trouble. I am not even sure about why we need `output`. In the case of CLI, `source` is unavoidable.

Looking at the configuration options, too, yeah, we could regroup it even further.

konnov: Also, `smtprof` and `profiling` are top-level, because we have `check` and `tracee` (the last one I would rather evict and make an external tool). I guess, the logic is whatever is shared between different commands goes to top-level. Previously, it was `common`, which was also confusing.

.unreleased/breaking-changes/typed-search-controls.md
konnov: Added to the notes

docs/src/apalache/running.md
konnov: Fixed:

```md
    - `--seed=NUM` sets a search seed in the inclusive range `0..2147483647`. The seed initializes SMT solving and, with
      `simulate`, is also used for transition selection. When omitted, Apalache generates and logs a fresh seed for the
      run.
```

mod-tool/src/test/scala/at/forsyte/apalache/tla/tooling/opt/TestCommandConfig.scala
konnov: done:

```scala
  test("seed parsing accepts the upper bound and rejects overflow") {
    val boundary = new CheckCmd()
    boundary.read(List(s"--$SEED=${Int.MaxValue}", "CommandConfig.tla"))
    val config = ApalacheConfigResolver.resolveCheck(boundary.toConfig.requireValue()).requireValue()
    assert(config.checker.seed == Int.MaxValue)

    val overflow = intercept[ParsingException] {
      new CheckCmd().read(List(s"--$SEED=2147483648", "CommandConfig.tla"))
    }
    assert(overflow.getMessage.contains("expected an Int"))
  }
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/search/ModelCheckerParams.scala
konnov: Good point. Now the parameters are passed directly, no defaults anymore. Also removed the unused constants that were mentioned in #2053

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/search/ModelCheckerParams.scala
konnov: No defaults in `ModelCheckerParams` anymore

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigResolver.scala
konnov: Added it directly

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigResolver.scala
konnov: yep, added that

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigResolver.scala
konnov: Added the bound in the error message. We don't have to check it in the code though, as `seed` is of type `Int`.

docs/src/adr/025adr-cvc5-solver-backend.md
konnov: Added a revision note (actually, Codex did that)

konnov:
@thpani I think I've fixed all of your comments (except the configuration layout still needs some thinking). Can you have another look?

**update:** when this is done, I want to have another look at #3419

PR #3409, opened 2026-07-30
Refactoring the options and configuration framework

### LINKED ISSUE #2756: Document Apalache config format + options
While addressing https://github.com/informalsystems/quint/issues/1196, I tried to write an Apalache config file, but struggled.

The best documentation I found is [this manual section](https://apalache.informal.systems/docs/apalache/config.html#file-format-and-supported-parameters). It documents the principle format, but I had to guess the JSON keys since there is no 1:1 correspondence to CLI flags.

@shonfeder Maybe I missed something and there is documentation of the supported config file options somewhere?

The second thing I struggled with is setting `--tuning-options` / `checker.tuning` via the config file, which is passed as a `:`-separated key/value list on the CLI, but (at least it seems so) needs to be a nested JSON object in the config file. We should confirm this, and document how to set tuning options via config.

### LINKED ISSUE #2999: Refactor option configurations
The current version relies on pureconfig. As a result, when passing parameters from `quint verify`, we have a configuration file like this:

```
{
    "common": {
      "write-intermediate": true,
      "debug": true
    },
    "checker": {
        "discard-disabled": false,
        "no-deadlocks": true,
        "algo": "incremental",
        "tuning": {
            "search.invariant.mode": "after",
            "search.invariantFilter": "(1->.*)"
        }
    }
}
```

On the other hand, the option processing code is processing options like this:

https://github.com/apalache-mc/apalache/blob/9d89eb39ba9fffc29c7119ac548fc995ca82d048/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala#L683-L692

What is worse, if I pass `discardDisabled: true` instead of `discard-disabled: true` in the configuration file, the configuration parser does not complain.

We have to streamline configuration parsing and get rid of unnecessary complexity. Currently, it is very hard to understand what parameters have to be passed, even by reading the code.

konnov:
Closes #2999. Closes #2756. This PR presents a significant refactoring of the options and configuration framework. Before looking at the code, check ADR 026 in the companion PR DCO #3410.

The main achievement is that we no longer depend on `pureconfig` and its workarounds. Also, HOCON is replaced with plain JSON.

The change size looks scary. However, except a few things, the changes are mechanical, e.g., changing imports. Here are *the few things* to look at:

- The classes in the package `at.forsyte.apalache.io.config`:
   - The refactored **data model of the configuration file** in `ApalacheConfig` and `ConfigPatch`. These are plain data objects now. They represent the inputs to the configuration framework (CLI, JSON file, JSON configs via RPC).
   - The refactored **data model of the resolved configuration options** in `ResolvedOptions`. These are plain data objects that are consumed by the Apalache passes.
   - The new `ApalacheConfigJsonParser`. This is a replacement of the `pureconfig` parser. All generated, but tested.
   - `ApalacheConfigResolver` does command-specific validation and produces `ResolvedOptions` for specific Apalache commands.
   - Finally, `ApalacheConfigLoader` contains the entry point to the framework that use by the CLI tooling and the RPC code.
   - In addition, `README.md` in this package summarizes the above info. I believe that this is the way to go in the future, since looking for ADRs when writing code is not optimal.

- The CLI command classes in the package `at.forsyte.apalache.tla.tooling.opt`. They use the refactored framework. Also, the default values are now directly produced from the default values in the configuration framework.

For anyone who wants to review this PR, I recommend looking at the above classes in an IDE. Once you understand those changes, the rest should be easier to understand.

The refactoring is mainly done with Codex GPT 5.6-sol max. It took me multiple iterations to guide the AI tool, in order to achieve a meaningful design and code layout. Some pieces I had to change manually. I've spent about 1.5 day on this refactoring, so it's not purely vibe-coded.

The breakdown of the code changes by Codex:

  - Production code: ~830 mechanical, ~3,789 essential.
  - Tests: ~220 mechanical, ~562 essential.
  - Documentation adds another 496 changed lines, excluded above.
  - Entire tracked diff: 5,897 lines.

-------------------------------------


.unreleased/breaking-changes/json-only-configuration.md
```diff
@@ -0,0 +1,3 @@
+Configuration files now use strict JSON and the canonical `.apalache.json` /
+`$HOME/.tlaplus/apalache.json` names (migrate the old .cfg). HOCON is no longer
```
thpani: What does _migrate the old .cfg_ mean?

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
-
-The order of precedence of the sources follows their numbering: i.e., and any
-configuration set in an earlier numbered source overrides a configuration set in
-a later numbered source.
+Apalache configuration files use **strict JSON**. The supported filenames are
+`.apalache.json` for a project file and `$HOME/.tlaplus/apalache.json` for a user-wide file.
+
+## Loading and precedence
+
+Values are applied in this order, from highest to lowest precedence:
```
thpani: I think this is contradictory.

If they are applied in this order, the later ones would take presendence, but you list them from high to low precedence?

@copilot suggest better wording

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+
+1. command-line arguments;
+2. environment variables used by command-line options;
+3. the file passed with `--config-file`, or the nearest `.apalache.json` found by searching the current directory and
+   its parents;
+4. `$HOME/.tlaplus/apalache.json`;
```
thpani: this suggests `$HOME/.tlaplus/apalache.json` is applied even `.apalache.json` is found or `--config-file` is given?

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+5. built-in defaults.
+
+An explicit `--config-file` disables the search for a local file, but not the global fallback. Lists and ordinary values
```
thpani: Why? I cannot think of a case where I want to give an explicit `--config-file` and still use a global fallback.

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+replace lower-precedence values.
+`checker.tuning` objects are merged by key.
```
thpani: out of place?

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+
+The old `.apalache.cfg` and `$HOME/.tlaplus/apalache.cfg` application configuration filenames are rejected, even when
+they contain valid JSON or a
+`.json` file exists beside them. Rename them to `.apalache.json` and
+`$HOME/.tlaplus/apalache.json`, respectively. Explicit `--config-file` paths ending in `.cfg` are rejected as well.
+HOCON syntax is no longer supported.
```
thpani: This entire paragraph needs to be rewritten with the reader in mind.

```
Apalache previously supported `.cfg` configuration files in HOCON syntax.
These are no longer supported.
...
```

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+
+This does not affect TLC configuration files passed with `--config`, which files conventionally use the `.cfg`
+extension.
```
thpani:
```suggestion
This does not affect TLC configuration files passed with `--config`, that conventionally use the `.cfg` extension.
```

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+      "port": 8822,
+      "server-type": "checker"
+   }
+}
+```
 
-## Command line arguments and environment variables
+## Supported keys
+
+An absent default of “none” means that the value is genuinely optional.
```
thpani:
```suggestion
In the table below, a default of "none" means that the value is optional.
```

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+|               | `run-dir`            | Also write this run's output directly to this directory.     | path string                  | none                                        |
+|               | `debug`              | Enable detailed logging and a configuration snapshot.        | boolean                      | `false`                                     |
+|               | `smtprof`            | Write an SMT constraint profile.                             | boolean                      | `false`                                     |
+|               | `write-intermediate` | Save intermediate representations produced by passes.        | boolean                      | `false`                                     |
+|               | `profiling`          | Write profiling data for transformation rules.               | boolean                      | `false`                                     |
+|               | `features`           | Enable experimental language or checker features.            | array of strings             | `[]`; names shown by CLI help               |
```
thpani, line 84: Didn't know we have this 😂 

afaict, it's only used for `rows`? Should we remove it in a future PR?

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
-### File format and supported parameters
+## Deprecated JSON aliases
 
-Local configuration files support JSON and the JSON superset
-[HOCON](https://github.com/lightbend/config/blob/master/HOCON.md).
+The following old spellings remain accepted with warnings:
 
-Here's an example of a valid configuration for commonly used
-parameters, along with their default values:
+| Deprecated                | Canonical                |
```
thpani: since we anyhow break compatiblity, should we just retire them now?

docs/src/apalache/config.md
```diff
@@ -1,74 +1,172 @@
...
+| `checker.timeout-smt-sec` | `checker.timeout-smt`    |
+| `checker.no-deadlocks`    | `checker.no-deadlock`    |
+| `checker.temporal-props`  | `checker.temporal`       |
+| `typechecker.inferpoly`   | `typechecker.infer-poly` |
 
-```yaml
-common  {
-    # Directory in which to write all log files and records of each run
-    out-dir = "${PWD}/_apalache-out"
+Old object-shaped enum and source values using a `type` field are also read with a warning. Setting a canonical key and
```
thpani: I have no idea what this means.

Probably better to skip it in the docs.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/AbstractCheckerCmd.scala
```diff
@@ -4,47 +4,76 @@ import org.backuity.clist.{arg, opt}
...
 abstract class AbstractCheckerCmd(val name: String, description: String)
     extends ApalacheCommand(name, description) with LazyLogging {
 
   var file: File = arg[File](description = "a file containing a TLA+ specification (.tla or .json)")
   var config =
-    opt[Option[File]](name = "config", default = None, description = "configuration file in TLC format")
+    opt[Option[File]](name = "config", default = None,
+        description = descriptionWithDefault(
+            "configuration file in TLC format",
+            configDefaults.checker.tlcConfig,
```
thpani: It looks like we somehow pass the default twice in these?

Once as literal `None`, and once as `configDefaults.checker.tlcConfig`?

Or maybe I don't understand the code.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/ParseCmd.scala
```diff
@@ -19,22 +21,39 @@ class ParseCmd
...
-    logger.info("Parse " + file)
+  override def toConfig: ConfigParseResult[ApalacheConfig] = {
+    val base = super.toConfig
+    if (!base.isSuccess) {
+      ConfigParseResult.failureFrom(base)
+    } else {
+      val source = InputSource.FileSource(file)
+      if (!source.isSuccess) {
+        ConfigParseResult.failureFrom(source)
+      } else {
```
thpani, line 38: It looks like `toConfig` is somehow repeated between `Cmd`s?

Is there any way to avoid the duplication?

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TestCmd.scala
```diff
@@ -28,51 +28,54 @@ class TestCmd
...
-      ),
-      typechecker = cfg.typechecker.copy(
-          inferpoly = Some(true)
-      ),
-  )
-
-  def run() = {
-    val cfg = configuration.get
-    val options = OptionGroup.WithCheckerPreds(cfg).get
+    val seed = Math.abs(System.currentTimeMillis().toInt)
```
thpani, line 44: Really?

tla-io/src/main/scala/at/forsyte/apalache/io/config/ConfigEnums.scala
```diff
@@ -0,0 +1,104 @@
...
+  final def displayName: String =
+    if (details.isEmpty) name else s"$name ($details)"
+
+  final override def toString: String = name
+}
+
+object Algorithm {
+  case object Incremental extends Algorithm("incremental", "")
+  case object Offline extends Algorithm("offline", "")
+  case object Remote extends Algorithm("remote", "used by explorer")
```
thpani: I don't think `used by explorer` helps. I'd just remove the `details` arg

tla-io/src/main/scala/at/forsyte/apalache/io/config/ConfigPatch.scala
```diff
@@ -0,0 +1,61 @@
+package at.forsyte.apalache.io.config
+
+import at.forsyte.apalache.io.InputSource
+import at.forsyte.apalache.tla.lir.Feature
+
+import java.nio.file.Path
+
+/** A sparse patch to one section of [[ApalacheConfig]]. */
+sealed trait ConfigPatch
```
thpani: No idea what these are for. Please document.

tla-io/src/main/scala/at/forsyte/apalache/io/config/ValidatedOptions.scala
thpani, line 1: some kinda comment explaining what `ResolvedOptions` are would be appreciatec

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigLoader.scala
```diff
@@ -0,0 +1,217 @@
...
+ * `ApalacheConfigResolver`.
+ *
+ * @param workingDirectory
+ *   Starting directory for local configuration discovery.
+ * @param homeDirectory
+ *   Home directory containing the user-wide configuration.
+ */
+final class ApalacheConfigLoader(
+    workingDirectory: Path = Paths.get("").toAbsolutePath.normalize(),
+    homeDirectory: Path = Paths.get(System.getProperty("user.home"))) {
```
thpani: Java 21 docs:

_It is recommended to obtain a Path via the Path.of methods instead of via the get methods defined in this class as this class may be deprecated in a future release._

Also, for readability:

```suggestion
 *   Home directory containing the user-wide configuration.
 */
 
private def currentWorkingDirectory: Path =
     Path.of("").toAbsolutePath.normalize

   private def userHomeDirectory: Path =
     Path.of(System.getProperty("user.home")).toAbsolutePath.normalize
 
final class ApalacheConfigLoader(
    workingDirectory: Path = currentWorkingDirectory,
    homeDirectory: Path = userHomeDirectory) {
```

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigLoader.scala
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  /** Read and decode one file, appending either its configuration or its diagnostics to the supplied accumulators. */
+  private def addDecodedFile(
+      path: Path,
+      label: String,
+      errors: ListBuffer[String],
+      warnings: ListBuffer[String],
+      configurations: ListBuffer[ApalacheConfig]): Unit = {
```
thpani: Passing mutable data structures and returning `Unit` looks rather error-prone / un-Scala

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
```diff
@@ -0,0 +1,671 @@
...
+          val aliases = Set("timeout-smt-sec", "no-deadlocks", "temporal-props")
+          rejectUnknown(
+              obj,
+              path,
+              Set(
+                  "tuning",
+                  "algo",
+                  "config",
+                  "discard-disabled",
+                  "cinit",
```
thpani: It really doesn't look very maintainable to list these exhaustively as literals

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
```diff
@@ -0,0 +1,671 @@
...
+    /** Decode all recognized top-level keys and sections; malformed values contribute errors to this decoder. */
+    def config(root: ObjectNode): ApalacheConfig = {
+      rejectUnknown(
+          root,
+          "$",
+          Set(
+              "command",
+              "config-file",
+              "out-dir",
+              "run-dir",
```
thpani: It really doesn't look very maintainable to list these exhaustively as literals

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
```diff
@@ -0,0 +1,671 @@
...
+                  "smt-solver",
+                  "smt-encoding",
+                  "temporal",
+                  "view",
+              ) ++ aliases,
+          )
+
+          val timeoutNode = aliased(obj, "timeout-smt", Seq("timeout-smt-sec"), path)
+          val deadlockNode = aliased(obj, "no-deadlock", Seq("no-deadlocks"), path)
+          val temporalNode = aliased(obj, "temporal", Seq("temporal-props"), path)
```
thpani: I would really just remove these.

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
```diff
@@ -0,0 +1,671 @@
...
+      Some(values.result())
+    }
+
+    private def stringList(obj: ObjectNode, field: String, parent: String): Option[List[String]] =
+      stringListNode(Option(obj.get(field)), s"$parent.$field")
+
+    private def stringListNode(node: Option[JsonNode], path: String): Option[List[String]] =
+      node match {
+        case None =>
+          None
```
thpani: Bit subtle: if there are partial errors, this returns `Some(...)` AND records errors?

I would at least add a comment

tla-io/src/main/scala/at/forsyte/apalache/io/config/README.md
```diff
@@ -0,0 +1,88 @@
...
+`ApalacheConfig` to the command. Commands must not cache or reload configuration.
+
+Resolved command types also expose `source` and `output` directly.
+`ModuleIoOptions` is only the two-field adapter used by Guice when constructing frontend passes.
+
+## Files
+
+| File                            | Responsibility                                                                                        |
+|---------------------------------|-------------------------------------------------------------------------------------------------------|
+| [ApalacheConfig.scala]          | Top-level configuration data model, explicit merge rules, and defaults used for diagnostic snapshots. |
```
thpani: what are diagnostic snapshots?

thpani (commented):
Aaah, this looks very clean! 🎉

Bunch of comments below.

My main design gripe:

I don't think `~/.tlaplus/apalache.json` and local `.apalache.json` should be merged; at least I don't want "sparse" behavior between a local and a user-wide config.

I think such behavior is okay for frequently-used tools like git, where I expect sort-of uniform behavior with a few local tweaks.

For a verification tool, where configuration has serious effects, I expect the project-local config to be authoritative, and not have any "scoped-elsewhere" effects applied from a user-wide config.

My suggestion is: **only** consider `~/.tlaplus/apalache.json` if **no** `.apalache.json` is found.

konnov:
> I don't think `~/.tlaplus/apalache.json` and local `.apalache.json` should be merged; at least I don't want "sparse" behavior between a local and a user-wide config.
> 
> I think such behavior is okay for frequently-used tools like git, where I expect sort-of uniform behavior with a few local tweaks.
> 
> For a verification tool, where configuration has serious effects, I expect the project-local config to be authoritative, and not have any "scoped-elsewhere" effects applied from a user-wide config.
> 
> My suggestion is: **only** consider `~/.tlaplus/apalache.json` if **no** `.apalache.json` is found.

I agree. I would even go further and remove support for `~/.tlaplus/apalache.json`. However, we know that some of the earlier Apalache maintainers like having a global config. There is definitely a danger of overriding the defaults with a global config and introduce a hard-to-detect deviation.

.unreleased/breaking-changes/json-only-configuration.md
konnov: The HOCON configs had the suffix `.cfg`.

docs/src/apalache/config.md
konnov: Yeah. This whole policy is confusing. Let me check how it is actually done. I would also rather prefer not merging configuration files. It's easy to merge them, but this behavior is confusing. Shall we agree on merging CLI options with the config files (CLI options take priority over the config files), but using only a single config files, whichever wins?

docs/src/apalache/config.md
konnov: Totally. This is also on my refactoring list https://github.com/apalache-mc/apalache/issues/3405

docs/src/apalache/config.md
konnov: I wanted to keep them for compatibility with quint, but let me check, whether we actually need them

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/AbstractCheckerCmd.scala
konnov: This is the default value that is printed in the CLI help. I realized that our defaults were not synced with the CLI help. Now they are.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/ParseCmd.scala
konnov: It is repeated, but the fields are different. I decided to keep it this way, so we can immediately see what's going on without visiting the helper methods (and libraries). This is a compromise of course.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TestCmd.scala
konnov: I even do not understand where this command is coming from. Shall we remove it?

konnov: Oh, wow. It looks like I introduced it long time ago. We should just remove it. All kind of rare conveniences should be done with custom scripting and JSON-RPC. https://apalache-mc.org/docs/adr/006rfc-unit-testing.html#32-testing-actions

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigLoader.scala
konnov: I think this is my fault. I prompted Codex that it should be more Java-like Scala :))

tla-io/src/main/scala/at/forsyte/apalache/io/config/ConfigEnums.scala
konnov: True. It is propagated into CLI help, but it looks totally out of place.

.unreleased/breaking-changes/json-only-configuration.md
thpani: I meant that this particular phrasing makes no sense in the changelog. It's not clear if "migrate" is imperative or infinitive

thpani: It could also be read as "this change migrates the old .cfg"

docs/src/apalache/config.md
thpani: Yes, this sounds good to me 👍🏻

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/TestCmd.scala
konnov: I am planning to delete this command in follow up PRs at some point: #3412

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
konnov: True. There are a lot of duplicates there. We can introduce constants. But the lists will still be there? Do you want to change it further beyond the constants?

.unreleased/breaking-changes/json-only-configuration.md
konnov: I see. Fixed it this way:

> Configuration files now use strict JSON and the canonical `.apalache.json` /
`$HOME/.tlaplus/apalache.json` names. HOCON .cfg files are no longer
supported. The JSON structure is simplified. Check the manual for details. PR #3409.

docs/src/apalache/config.md
konnov: Ugh. This whole discussion surface a potential security issue. The current implementation allows the user to specify inputs and outputs in the configurations. Also, fixing that.

konnov: Fixed all of the things discussed here.

tla-io/src/main/scala/at/forsyte/apalache/io/config/ConfigEnums.scala
konnov: Fixed that

tla-io/src/main/scala/at/forsyte/apalache/io/config/ValidatedOptions.scala
konnov: Add comments, also renamed them to `ValidatedOptions`.

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigLoader.scala
konnov: Fixed that in all places

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigLoader.scala
konnov: This is gone, since we have significantly simplified the loading policy

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
konnov: Introduced `Constants`, now they are used across the codebase

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
konnov: I will remove "timeout-smt-sec" and "no-deadlocks", but "temporal-props" and "input.source" are used by Quint. Shall we keep those with a deprecation warning, or remove. In the latter case, the Quint team would have to update their code. Apalache is pinned there, so it's should not be a big issue

tla-io/src/main/scala/at/forsyte/apalache/io/config/README.md
konnov: This is a confusing name for the config file that is written with `--debug`. No idea why Codex started to call it this way.

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
konnov: On a second thought, let's remove this backward compatibility. Quint can be updated later.

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
konnov: Got rid of that. Now we are using `Either` in the parser. Since it's inside the parser, this should be fine

konnov:
@thpani I think I went through all of your comments. This is quite a refactoring

docs/src/apalache/config.md
thpani: Are these still present? 
We could at least make a follow-up issue.

mod-tool/src/main/scala/at/forsyte/apalache/tla/tooling/opt/AbstractCheckerCmd.scala
thpani: But why do you need both `configDefaults.checker.tlcConfig` (as a variable) and `default = None` (as a literal)

tla-io/src/main/scala/at/forsyte/apalache/io/config/ApalacheConfigJsonParser.scala
thpani: Ah, this was a misunderstanding.
I would actually prefer to read the literals rather than constant names :D

What I meant is: It seems like you could reconstruct this set from the `update()` calls below.

It's not clear to me what the consequence is of missing one literal in the `Set()`, or in `update()`s. In any case, what's the workflow for adding a parameter?

thpani (approved):
Approving; I have two outstanding questions that are marked unresolved in the discussion above.

konnov:
> Approving; I have two outstanding questions that are marked unresolved in the discussion above.

Thanks! I've made a separate issue #3419 to keep track of these two unresolved questions. I find it very hard to navigate through the current large PR. So let's merge this one and address #3419 in a separate smaller PR.

thpani:
> Thanks! I've made a separate issue #3419 to keep track of these two unresolved questions. I find it very hard to navigate through the current large PR. So let's merge this one and address #3419 in a separate smaller PR.

@konnov Actually, I wanted those questions addressed.
Now I see another PR (#3404), and I have the same questions again.

konnov:
> > Thanks! I've made a separate issue #3419 to keep track of these two unresolved questions. I find it very hard to navigate through the current large PR. So let's merge this one and address #3419 in a separate smaller PR.
> 
> @konnov Actually, I wanted those questions addressed. Now I see another PR (#3404), and I have the same questions again.

Sorry, my bad!

PR #3411, opened 2026-07-31
Add a Scala Steward config, including sign-offs

konnov:
Since we have the DCO sign-off policy now, we need a config for Scala Steward that enables sign-offs. Adding this one. Adapted the template config for our purposes.


thpani:
I'm a bit lost - was Scala Steward using default config so far?

thpani (approved):

konnov:
> I'm a bit lost - was Scala Steward using default config so far?

Yeah, there was no config in our repo, so, yes, some kind of default

PR #3413, opened 2026-07-31
Move the SANY importer to tla_parser

konnov:
This PR lays the grounds for publishing `tlair` and `tla_io` on Maven Central.

 - [x] Moved the SANY importer to its own module called `tla_parser`. This is needed as it depends on `tla2tools.jar` that is not published on Maven Central
 - [x] Moved a few support classes from `tla_io` to `tlair`, to break cyclic dependencies between `tla_io` and `tla_parser`. 
 - [x] Set the Maven `groupId` to `org.apalache-mc`, which the project owns.This is required for publishing a package. The Scala packages remain `at.forsyte.apalache.*` (though we should consider renaming the root package finally).

This PR is moving files around. It is done with Claude Opus 4.9.


konnov:
@lemmy Are there plans for publishing versioned `tla2tools.jar` on Maven? The current approach with GitHub works only partially: We cannot publish Apalache libraries that transitively use `tla2tools.jar`.

lemmy:
`tla2tools.jar` is already published to Sonatype’s Maven snapshots repository:

https://central.sonatype.com/repository/maven-snapshots/org/lamport/tla2tools

Thus, I assume that you are asking for immutable, versioned releases that can be used as *fixed* transitive dependencies by artifacts published to Maven Central?

The TLA+ tools currently follow a rolling-release model rather than a traditional versioned-release model. We could assign version identifiers, i.e., timestamps, but those identifiers would not imply semantic-versioning or compatibility guarantees. The substantive difference between any two such releases would be the corresponding Git diff.

konnov:
> `tla2tools.jar` is already published to Sonatype’s Maven snapshots repository:
> 
> https://central.sonatype.com/repository/maven-snapshots/org/lamport/tla2tools

Does it still work for you? My browser returns 404, and sonatype does not find tla2tools.

> Thus, I assume that you are asking for immutable, versioned releases that can be used as _fixed_ transitive dependencies by artifacts published to Maven Central?

Yes, we need versioned dependencies to publish a package.

> The TLA+ tools currently follow a rolling-release model rather than a traditional versioned-release model. We could assign version identifiers, i.e., timestamps, but those identifiers would not imply semantic-versioning or compatibility guarantees. The substantive difference between any two such releases would be the corresponding Git diff.

Yes, any kind of fixed versioning would help, if it works with Maven.

konnov:
> The TLA+ tools currently follow a rolling-release model rather than a traditional versioned-release model. We could assign version identifiers, i.e., timestamps, but those identifiers would not imply semantic-versioning or compatibility guarantees. The substantive difference between any two such releases would be the corresponding Git diff.

You could probably use the commit hash and the timestamp, e.g. something like `1.8.0-dev.20260801+g1a2b3c4`. Anything that allows us to trace a concrete package back to the source code and guarantees that the package contents stays the same and does not break builds, caches, etc.

lemmy:
Maven *releases* are heavyweight compared to snapshots: they are immutable and validated, so they need signatures, sources and javadoc jars, and complete POM metadata. A snapshot is simply overwritten by the next build. Releases also count against the monthly publishing quota that Central [starts enforcing on October 1, 2026](https://central.sonatype.org/publish/maven-central-publishing-limits/), which master's build volume would exceed.

More fundamentally, the TLA+ tools rolling-release model has no notion of a release at all. `1.8.0-SNAPSHOT` *is* the release, so promoting some builds to Maven releases means inventing a distinction that does not exist upstream.

tlaplus/tlaplus#1165 proposes bringing back point releases, but nobody has stepped up yet.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/StandardAnnotations.scala
```diff
@@ -5,6 +5,11 @@ package at.forsyte.apalache.io.annotations
  */
 object StandardAnnotations {
 
+  /**
+   * The name of the meta annotation that contains the text of the comments free of annotations.
+   */
+  val FREE_TEXT: String = "#freeText"
```
thpani, line 11: Can we remove this?

thpani (approved):

konnov:
> Maven _releases_ are heavyweight compared to snapshots: they are immutable and validated, so they need signatures, sources and javadoc jars, and complete POM metadata. A snapshot is simply overwritten by the next build. Releases also count against the monthly publishing quota that Central [starts enforcing on October 1, 2026](https://central.sonatype.org/publish/maven-central-publishing-limits/), which master's build volume would exceed.

I guess, I don't understand the complexity of publishing to Maven yet. I will publish two modules without the parser, just to get the feeling of the task.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/StandardAnnotations.scala
konnov: I will remove this in a follow-up PR.

PR #3418, opened 2026-08-02
 Refactor NameReplacementMap

### LINKED ISSUE #1872: Refactor: NameReplacementMap should be instantiated via dependency injection, not a static global mutable singleton
See the discussion in #1823

I think it is fine to have a "global" store, as long as the store is passed as a class parameter and instantiated by the dependency injection framework. If we do it this way, the class is still easily testable and mockable. Usually the problems arise with static singletons.

_Originally posted by @konnov in https://github.com/informalsystems/apalache/issues/1823#issuecomment-1164090443_

konnov:
Close #1872: This is an old refactoring request to make `NameReplacementMap` a non-singleton. Fixed with Codex GPT 5.6-sol high. This change is entirely mechanical: Introduce `VariableDescriptionsStore` and wire it properly.


konnov:
@p-offtermatt since this refactoring was initially requested by you, would you like to review it? It's quite mechanical

PR #3423, opened 2026-08-03
Remove `#freeText` annotations

konnov:
As noticed in #3413, removing the old feature that was never used, namely, comments stored as `#freeText` annotations.


thpani (approved):
LGTM, but please check the falling unit test

PR #3425, opened 2026-08-04
Migrate to Java 25

konnov:
Upgrading the build pipeline to Java 25. Scala already works with Java 25. This will allow us to have better compatibility with Java. Also, fixing warnings that are already shown on JRE 25.

In addition, addition a smoke build on Windows.

This is done with Codex GPT 5.6-sol max.


.unreleased/breaking-changes/java-25.md
```diff
@@ -0,0 +1,2 @@
+Raised the minimum supported Java runtime from Java 17 to Java 25. Released application and library artifacts now
+require Java 25 or newer.
```
thpani: I'm not sure the changelog script will pick up separate lines?

```suggestion
Raised the minimum supported Java runtime from Java 17 to Java 25. Released application and library artifacts now require Java 25 or newer.
```

docs/src/apalache/installation/jvm.md
```diff
@@ -15,10 +15,12 @@ better, add the `./bin` directory to your `PATH` and run `apalache-mc`.
 Alternatively, you can run Java directly with
 
 ```
-java.exe -jar ./lib/apalache.jar <args>
+java --sun-misc-unsafe-memory-access=allow -jar ./lib/apalache.jar <args>
```
thpani, line 18: do we have an issue for fixing this?

imo it requires updating / removing Guice and probably something with the native SMT solver libs

test/tla/cli-integration-tests.md
```diff
@@ -525,7 +527,7 @@ EXITCODE: ERROR (12)
 ### check Fix531.tla reports no error: regression for issue 531 (array-encoding)
 
 ```sh
-$ apalache-mc check --length=1 Fix531.tla | sed 's/I@.*//'
+$ output="$(apalache-mc check --length=1 Fix531.tla 2>&1)"; if printf '%s\n' "$output" | grep -Eiq 'terminally deprecated|restricted method'; then exit 1; fi; printf '%s\n' "$output" | sed 's/I@.*//'
```
thpani: I don't like filtering something away that may indicate a bug or failure in the future;
can we include it in the expected output? this way, it would just fail the test when we fix relying on `--sun-misc-unsafe-memory-access=allow`, which would be expected and easy to change.

flake.nix
```diff
@@ -12,7 +12,9 @@
   # dependencies.
   inputs = {
     # Nix Inputs
-    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+    # Nixpkgs 26.11 dropped x86_64-darwin. Stay on the current stable line so
+    # the development shell remains available on all four default systems.
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-26.05";
```
thpani: why do we need `x86_64-darwin`?

.github/workflows/main.yml
```diff
@@ -121,6 +121,56 @@ jobs:
...
+        run: |
+          $archive = Get-ChildItem package/apalache-*.zip | Select-Object -First 1
+          Expand-Archive -LiteralPath $archive.FullName -DestinationPath windows-smoke
+          $launcher = Get-ChildItem windows-smoke -Recurse -Filter apalache-mc.bat | Select-Object -First 1
+          $output = & $launcher.FullName version --debug 2>&1
+          $output | Write-Output
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          if ($output -match "terminally deprecated|restricted method") {
+            throw "Java 25 runtime warnings were emitted by the packaged launcher"
+          }
```
thpani: I'm not sure, is this a slopcode leftover? I'm not sure what it does

.github/workflows/main.yml
```diff
@@ -121,6 +121,56 @@ jobs:
...
+          path: target/universal/apalache-*.zip
+          if-no-files-found: error
+
+  windows-package-smoke:
+    needs: windows-package
+    runs-on: windows-latest
+    steps:
+      # The repository contains a directory named `aux`, which Windows cannot
+      # check out because AUX is a reserved device name. Build on Linux and
+      # exercise the platform-independent package on Windows instead.
```
thpani, line 150: this seems like something we should fix; is it really caused by Windows not supporting directories named `aux`? (I can kinda believe that, but at the same time I cannot 😄)

But in general, I think it would be better to just rename those packages to something supported on all platforms; shouldn't be a large code-change.

thpani (approved):
LGTM

flake.nix
thpani: It seems like we could force the macOS runners to `arm64` by pinning them to `macos-latest`: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/main.yml
konnov: I also forgot about MSDOS file restrictions. We should not call packages com1, com2, and lpt as well :)

I've created a separate issue on that: https://github.com/apalache-mc/apalache/issues/3429

.github/workflows/main.yml
konnov:
```suggestion
```

docs/src/apalache/installation/jvm.md
konnov: This is caused by Guice, the SMT solver libs are handled properly. Looking at the issues on Guice, it is one more sunset Google project. We should move away from it: https://github.com/apalache-mc/apalache/issues/78

flake.nix
konnov: Good point. Moved back to nixpkgs-unstable and excluded x86_64-darwin. We should drop Nix anyways #3427

test/tla/cli-integration-tests.md
konnov: Yep, Codex piggy backed another test for this purpose. Now we have:

```md
### check command emits no Java 25 compatibility warnings

The packaged launcher suppresses warnings from bundled dependencies that use
`sun.misc.Unsafe` and restricted native-access methods. Check the minimal no-variable specification to exercise both
application startup and native solver loading.

```sh
$ apalache-mc check --length=1 NoVars.tla 2>&1 | grep -Ei 'terminally deprecated|restricted method'
[1]
```

```

konnov:
I am going to publish a separate minor release with this one, as it is a big change, and it's good to have a clean version tag, in case things start to break.

Issue #3429, opened 2026-08-05
Rename the package aux


konnov:
As it does not work on Windows: https://github.com/apalache-mc/apalache/pull/3425

PR #3435, opened 2026-08-05
Downgrade bytecode compatibility to Java 17

konnov:
Since our migration to Java 25 [broke] the CI pipeline of [tlaplus/examples](https://github.com/tlaplus/examples), this PR downgrade the bytecode compatibility to Java 17. We are still going to target Java 25 at the source level.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -121,7 +121,7 @@ jobs:
           find $HOME/.cache/coursier/v1                -name "ivydata-*.properties" -delete || true
           find $HOME/.sbt                              -name "*.lock"               -delete || true
 
-  windows-package:
+  universal-package:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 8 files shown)


thpani (commented):
I don't like that the smoke test runs in a workflow called `windows-package`. Wouldn't the universal package workflow be the better target? 

Also the matrix with Java {17, 25} only runs on Windows. If we only pick one target, I'd prefer Linux.

thpani (commented):
On a second thought, why do we need these smoke tests? Won't the integration tests run the Java 17 bytecode?

konnov:
> On a second thought, why do we need these smoke tests? Won't the integration tests run the Java 17 bytecode?

In principle, yes, the integration tests will run Java 17 bytecode, but not using JRE 17.

konnov:
Re-did the smoke tests:
  1. Check Counter.tla with z3 and cvc5 on Java 17/25 and Ubuntu,
  2. Check Counter.tla with z3 and cvc5 on Windows.

The workflow for Windows is a bit complex due to #3429.

thpani (approved):
LGTM

Issue #3427, opened 2026-08-04
Speed-up integration tests and drop mdx and nix


konnov:
Our integration tests have three interconnected issues:
 
 1. The integration tests are slow and this makes working on a pull request painful (18-36 min per workflow).
 2. Our tests are written in Markdown (nice!) and are checked by `mdx`. It has to be pinned to a specific version and pulls OCaml.
 3. Since `mdx` is pinned to a specific version, we run tests in Nix, which has a maintenance cost of its own.

Here are a few ideas on how we could improve our tests and decrease the maintenance cost:

 1. Replace `mdx` with a JVM-native solution. Codex suggested JUnit/ScalaTest + [CommonMark](https://github.com/commonmark/commonmark-java). I would prefer JUnit in the spirit of decreasing dependency on Scala. CommonMark is just a Markdown parser, so we would need a vibe-coded (or hand-written) replacement of `mdx`, which does not seem to be hard to do.
 2. Run most of the integration tests inside the same JVM. We already have `Tool` which is mostly well-behaving. We still have one singleton #1048, but this can be fixed. If we run most of the tests within the same JVM, we do not have to spend time on starting the JVM for each test. Of course, we have to be careful about some tests having serious side effects.
 3. Drop Nix. Our Nix setup is not taken care of, as the maintainers are not using Nix. If someone wants to use Apalache with Nix, they can do it in a separate project.

Related: #1689, #2566, #1552, #527

**Drop Markdown?** On a second thought, maybe we don't even need a markdown parser. Most of our integration tests are testing the output. Moreover, 311 out of 314 uses of `sed` could be fixed by using a proper logger configuration that does not introduce timestamps. A few file operations could be done directly in Scala/Java.

If we worry about missing side effects when running all tests in the same JVM, we can introduce a configuration where every test actually forks the process and executes the tool in a separate JVM. This configuration would run once per day or week, not all the time.

This is how a test could look like (this is Java, but it can be Scala):

```java
  @Nested
  @DisplayName("running the check command")
  class CheckCommandTests {
      @Test
      @Tag("array-encoding")
      @DisplayName("factorization finds a counterexample")
      void factorizationFindsCounterexample() {
          apalache.run(
                  "check",
                  "--tuning-options=cvc5.smt.logic=QF_UFNIA",
                  "--length=2",
                  "--inv=Inv",
                  "factorization.tla")
              .expectExit(12)
              .expectOutput("""
                  ...
                  The outcome is: Error
[...]

PR #3436, opened 2026-08-05
Refactor OutputManager

### LINKED ISSUE #1048: [FEATURE] Make `OutputManager` a non-singleton
Currently, #1047 practically necessitates that `OutputManager` exists as a singleton, but we want to move to an DI-based approach for parallelization in the future.

konnov:
Closes #1048. This is a long overdue refactoring of `OutputManager`:

 - [x] It is no longer a singleton but a properly injected instance. Importantly, the RPC calls spawn their own instances.
 - [x] The class is fully rewritten, as it had plenty of coding issues.
 - [x] Several methods that do not belong to its responsibilities were evicted.
 - [x] `OutputManager` is renamed to `OutputWorkspace` to better reflect the nature of what is going on.

Most of the changes are pretty mechanical: change in the imports, propagate an instance of `OutputWorkspace`.

This opens up the way to do #3427.

The refactoring was done with Codex GPT-5.6-sol high/xhigh, but I had to conduct it a lot.

-----------------------------


mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
```diff
@@ -48,33 +48,35 @@ object Tool extends LazyLogging {
...
 
-  private def outputAndLogConfig(cmd: ApalacheCommand, cfg: ApalacheConfig): Try[Unit] = for {
+  private def outputAndLogConfig(cmd: ApalacheCommand, cfg: ApalacheConfig): Try[OutputWorkspace] = for {
     initialization <- {
       val result = ApalacheConfigResolver.resolveCommandInitialization(cfg)
       if (result.isSuccess) Try(result.requireValue())
       else Failure(new ConfigurationError(result.errors.mkString("; ")))
     }
-    _ <- Try(OutputManager.configure(initialization))
+    outputWorkspace <- Try(new OutputWorkspace(initialization))
```
thpani: Can we avoid `Try`?

passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/ParserModule.scala
```diff
@@ -15,8 +16,9 @@ import com.google.inject.TypeLiteral
  * @author
  *   Igor Konnov
  */
-class ParserModule(options: ValidatedParseOptions) extends ToolModule {
+class ParserModule(options: ValidatedParseOptions, outputWorkspace: OutputWorkspace) extends ToolModule {
   override def configure(): Unit = {
+    bind(classOf[OutputWorkspace]).toInstance(outputWorkspace)
```
thpani, line 21: Hmmm. You set up dependency injection here, yet you pass it explicitly in a lot of places, including the `*Cmd`s, `*Module`s, `CmdExecutorService`?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
```diff
@@ -221,8 +222,9 @@ class Cvc5SolverContext(val config: SolverConfig) extends SolverContext with Laz
 
   private def initLogs(): Iterable[PrintWriter] = {
     val filePart = s"log$id.smt"
-    val writers =
-      (OutputManager.runDirPathOpt ++ OutputManager.customRunDirPathOpt).map(OutputManager.printWriter(_, filePart))
+    val writers = outputWorkspace.toSeq.flatMap { workspace =>
+      (Some(workspace.runDir) ++ workspace.additionalRunDir).map(dir => workspace.openWriter(dir.resolve(filePart)))
+    }
```
thpani: idk, I cannot read this `.toSeq.flatMap`
perhaps this is more readable:

```suggestion
val writers = outputWorkspace match {
  case Some(workspace) => ...
  case None            => Seq.empty
}
```

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/SolverContextFactory.scala
```diff
@@ -1,19 +1,20 @@
...
+import at.forsyte.apalache.io.OutputWorkspace
 import at.forsyte.apalache.io.config.SMTSolver
 
 /**
  * Creates concrete SMT solver contexts from solver configuration.
  */
 object SolverContextFactory {
 
-  def create(config: SolverConfig): SolverContext = {
+  def create(config: SolverConfig, outputWorkspace: Option[OutputWorkspace] = None): SolverContext = {
```
thpani: why is it optional in some places?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -424,12 +425,13 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext with LazyL
...
 
   /**
-   * Initialize up to two log writers, one in the run directory and one in the custom run directory, if those are set.
+   * Initialize up to two log writers, one in the run directory and one in the additional run directory, if set.
    */
   private def initLogs(): Iterable[PrintWriter] = {
     val filePart = s"log$id.smt"
-    val writers =
-      (OutputManager.runDirPathOpt ++ OutputManager.customRunDirPathOpt).map(OutputManager.printWriter(_, filePart))
+    val writers = outputWorkspace.toSeq.flatMap { workspace =>
```
thpani: same, just match `Some(_)` / `None`?

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
```diff
@@ -91,8 +91,15 @@ class ExplorationService(config: ConfigParseResult[ApalacheConfig]) extends Lazy
           ))
     }
     val options = optionsResult.requireValue()
+    val initialization = CommandInitializationOptions(SERVER, options.common, Some(options.source))
+    val outputWorkspace =
+      try new OutputWorkspace(initialization)
```
thpani: @copilot when would this ever fail?

tla-io/src/main/scala/at/forsyte/apalache/io/ReportGenerator.scala
```diff
@@ -1,31 +1,36 @@
...
-  private val reportFile = "BugReport.md"
-
-  private def getFileOrEmptyStr(filename: String) = OutputManager.readContentsOfFileInRunDir(filename).getOrElse("")
-
-  def getLog(): String =
-    Matcher.quoteReplacement(getFileOrEmptyStr("detailed.log")) // handle $s in log
+  def getLog(outputWorkspace: OutputWorkspace): String = {
+    val path = OutputWorkspace.detailedLogPath(outputWorkspace.runDir)
+    Matcher.quoteReplacement(Files.readString(path, StandardCharsets.UTF_8).trim)
+  } // handle $s in log
```
thpani: what is this supposed to tell me?

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
```diff
@@ -160,17 +162,25 @@ object Tool extends LazyLogging {
   }
 
   // Execute the program specified by the subcommand cmd, handling errors as needed
-  private def runCommand(cmd: ApalacheCommand, config: ApalacheConfig): ExitCodes.TExitCode =
+  private def runCommand(
+      cmd: ApalacheCommand,
+      config: ApalacheConfig,
```
thpani, line 167: Why is this using sparse `ApalacheConfig`. I thought they should all be resolved

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
```diff
@@ -160,17 +162,25 @@ object Tool extends LazyLogging {
...
+      outputWorkspace: OutputWorkspace): ExitCodes.TExitCode = {
+    // a helper for submitting bug reports
+    def readSourceCode(): Option[String] = config.source.flatMap(_.readUtf8.value)
```
thpani: I wouldn't eagerly read this string here and pass it around

shai/src/main/scala/at/forsyte/apalache/shai/v1/TransExplorerService.scala
```diff
@@ -176,7 +178,9 @@ class TransExplorerService(connections: Ref[Map[UUID, Conn]], logger: Logger)
           throw new IllegalArgumentException(resolved.errors.mkString("; "))
         }
         val options = resolved.requireValue()
-        PassChainExecutor(new ParserModule(options))
+        val initialization = CommandInitializationOptions(SERVER, options.common, Some(options.source))
```
thpani, line 181: What is `SERVER`? Is it `SERVER_VERSION`? `SERVER_TYPE`?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/profiler/RuleStatLocator.scala
```diff
@@ -11,7 +11,7 @@ import scala.collection.immutable.SortedMap
  * @author
  *   Igor Konnov
  */
-class RuleStatLocator {
+class RuleStatLocator(outputWorkspace: Option[OutputWorkspace] = None) {
```
thpani: why is it optional?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/profiler/RuleStatListener.scala
```diff
@@ -8,8 +9,8 @@ import at.forsyte.apalache.tla.lir.{OperEx, TlaEx}
  * @author
  *   Igor Konnov
  */
-class RuleStatListener extends SmtListener {
-  val locator = new RuleStatLocator()
+class RuleStatListener(outputWorkspace: Option[OutputWorkspace] = None) extends SmtListener {
```
thpani: why is it optional?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/RecordingSolverContext.scala
```diff
@@ -13,8 +14,9 @@ object RecordingSolverContext {
    */
   def create(
       parentLog: Option[SmtLog],
-      config: SolverConfig): RecordingSolverContext = {
-    val solverImpl = SolverContextFactory.create(config)
+      config: SolverConfig,
+      outputWorkspace: Option[OutputWorkspace] = None): RecordingSolverContext = {
```
thpani: optional

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/PrettyWriterWithAnnotationsFactory.scala
```diff
@@ -13,7 +14,8 @@ import java.io.PrintWriter
 class PrettyWriterWithAnnotationsFactory @Inject() (
     sourceStore: SourceStore,
     changeListener: ChangeListener,
-    store: AnnotationStore)
+    store: AnnotationStore,
+    override protected val outputWorkspace: OutputWorkspace)
```
thpani: what is `override protected val`?

tla-io/src/main/scala/at/forsyte/apalache/io/lir/CounterexampleWriter.scala
thpani, line 181: can we just inline this?

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+    val joinedPath = parts.foldLeft(dir)(_.resolve(_))
+    withWriter(joinedPath)(writeFun)
+  }
+
+  private def ensureDirExists(path: Path): Path = {
+    try {
+      Files.createDirectories(path)
+    } catch {
+      case e: IOException =>
+        throw new ConfigurationError(s"Could not find or create directory ${path.toAbsolutePath}: ${e.getMessage}")
```
thpani: why "find"?

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Join `dir` and `parts`, and call `withWriter` with this path and `writeFun`.
+   */
+  private def withWriterInJointPath(dir: Path, parts: Seq[String], writeFun: PrintWriter => Unit): Unit = {
+    // Join the parts, starting with `dir`. The standard `Path.of` works only over strings.
+    val joinedPath = parts.foldLeft(dir)(_.resolve(_))
+    withWriter(joinedPath)(writeFun)
+  }
+
+  private def ensureDirExists(path: Path): Path = {
```
thpani:
```suggestion
  private def createDir(path: Path): Path = {
```

thpani: can we move the helper above all other code?

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+ * @author
+ *   Jure Kukovec, Shon Feder, Igor Konnov
+ */
+final class OutputWorkspace(initialization: CommandInitializationOptions) {
+
+  /** Namespace shared by all runs of the same input file or command. */
+  private val outDir: Path = {
+    val fileName = initialization.source match {
+      case Some(InputSource.FileSource(path, _)) => path.getFileName.toString
+      case _                                     => initialization.command
```
thpani: I assume the `command` case is for server?

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+    }
+    ensureDirExists(initialization.common.outDir.resolve(fileName).toAbsolutePath)
```
thpani: I was staring at this for minutes. could use a helper

`initialization.common.outDir.resolve(fileName)`

basically creates `{outDir}/{fileName}`

thpani: Note that this is only safe because we do `.getFileName` above.

If `fileName` was an absolute path, JavaDoc says:
_If the other parameter is an [absolute](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#isAbsolute--) path then this method trivially returns other._

I think this deserves at least a comment.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+ * @param initialization
+ *   resolved values that determine this execution's output locations and enabled output features
+ *
+ * @author
+ *   Jure Kukovec, Shon Feder, Igor Konnov
+ */
+final class OutputWorkspace(initialization: CommandInitializationOptions) {
+
+  /** Namespace shared by all runs of the same input file or command. */
+  private val outDir: Path = {
```
thpani: afaict, this is not used anywhere except inside `runDir`. perhaps we can inline it there?

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+  /** Intermediate-output directory inside `runDir`, present only when intermediate output is enabled. */
+  private val intermediateDirOpt: Option[Path] =
+    if (initialization.common.writeIntermediate) {
+      Some(ensureDirExists(runDir.resolve(OutputWorkspace.IntermediateDirName)))
+    } else {
+      None
+    }
+
+  /** Intermediate-output directory inside the additional run directory, when both options are enabled. */
+  private val additionalIntermediateRunDirOpt: Option[Path] =
```
thpani: consistent naming would be
```suggestion
  /** Intermediate-output directory inside the additional run directory, when both options are enabled. */
  private val additionalIntermediateDirOpt: Option[Path] =
```

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+    intermediateDirOpt.flatMap(_ =>
+      additionalRunDir.map { path =>
+        ensureDirExists(path.resolve(OutputWorkspace.IntermediateDirName))
+      })
+
+  /** Open a UTF-8 writer that the caller must close. Prefer [[withWriter]] for scoped writes. */
+  def openWriter(path: Path): PrintWriter = {
+    new PrintWriter(Files.newBufferedWriter(path, StandardCharsets.UTF_8))
```
thpani: UTF-8 is already the standard charset

```suggestion
    new PrintWriter(Files.newBufferedWriter(path))
```

thpani: I don't think `PrintWriter` is the right option here:

_Methods in this class never throw I/O exceptions, although some of its constructors may. The client may inquire as to whether any errors have occurred by invoking [checkError()](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/io/PrintWriter.html#checkError())._

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  /** Intermediate-output directory inside the additional run directory, when both options are enabled. */
+  private val additionalIntermediateRunDirOpt: Option[Path] =
+    intermediateDirOpt.flatMap(_ =>
+      additionalRunDir.map { path =>
+        ensureDirExists(path.resolve(OutputWorkspace.IntermediateDirName))
+      })
+
+  /** Open a UTF-8 writer that the caller must close. Prefer [[withWriter]] for scoped writes. */
+  def openWriter(path: Path): PrintWriter = {
```
thpani: I'm also confused that this just lets us write anywhere?
Is it really needed as a public API?

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
```diff
@@ -0,0 +1,155 @@
...
+    new PrintWriter(Files.newBufferedWriter(path, StandardCharsets.UTF_8))
+  }
+
+  /** Apply `f` to a UTF-8 writer for `path`, then close the writer. */
+  def withWriter(path: Path)(f: PrintWriter => Unit): Unit = {
```
thpani: similar here, should this be public?

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
```diff
@@ -221,8 +222,9 @@ class Cvc5SolverContext(val config: SolverConfig) extends SolverContext with Laz
 
   private def initLogs(): Iterable[PrintWriter] = {
     val filePart = s"log$id.smt"
-    val writers =
-      (OutputManager.runDirPathOpt ++ OutputManager.customRunDirPathOpt).map(OutputManager.printWriter(_, filePart))
+    val writers = outputWorkspace.toSeq.flatMap { workspace =>
+      (Some(workspace.runDir) ++ workspace.additionalRunDir).map(dir => workspace.openWriter(dir.resolve(filePart)))
```
thpani: you say in `.openWriter` that _Open a UTF-8 writer that the caller must close_

I don't see where these are being closed.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
```diff
@@ -424,12 +425,13 @@ class Z3SolverContext(val config: SolverConfig) extends SolverContext with LazyL
...
+      (Some(workspace.runDir) ++ workspace.additionalRunDir).map(dir => workspace.openWriter(dir.resolve(filePart)))
```
thpani: nowhere closed?

thpani (commented):
Honestly, I think there are too many issues here. There are comments below, but I'm far from finished here.
We should decide if it makes sense to follow up on this or just close it.

I think this is the problem when an AI agent just writes a fresh implementation of something that hasn't been designed upfront.

tla-io/src/main/scala/at/forsyte/apalache/io/ReportGenerator.scala
konnov: Well, we have to ask the original author. This comment is copied verbatim from the old code.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
konnov: Cvc5SolverContext and Z3SolverContext are opening long-term writers that are closed in `dispose`. These two contexts are actually the only call sites of `openWriter`.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Z3SolverContext.scala
konnov: Cvc5SolverContext and Z3SolverContext are opening long-term writers that are closed in `dispose`. These two contexts are actually the only call sites of `openWriter`.

konnov:
> Honestly, I think there are too many issues here. There are comments below, but I'm far from finished here. We should decide if it makes sense to follow up on this or just close it.
> 
> I think this is the problem when an AI agent just writes a fresh implementation of something that hasn't been designed upfront.

I don't think this is only the issue with an AI agent. A lot of code that you flagged was written long time ago, before AI agents were used. (There are also strange artifacts by Codex, thanks for flagging them!) I suggest we first settle on `OutputWorkspace`, which is a direct refactoring of `OutputManager`.

A lot of code surrounding the call sites of `OutputManager` collected plenty of technical debt. I worry that refactoring all of  these call sites in one PR would drive us down the rabbit whole.

shai/src/main/scala/at/forsyte/apalache/shai/v1/TransExplorerService.scala
konnov: This is constant that is defined in `Constants`. I think the issue with PRs like this one is that GH diffs are suboptimal here:

```scala
  val SERVER = "server"
```

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
konnov: It's actually a function. Its only call site is the report generator, which is only called on exception.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
konnov: This is actually the function in the CLI tool that does config resolution in a later call to a specific command.

passes/src/main/scala/at/forsyte/apalache/tla/passes/imp/ParserModule.scala
konnov: Yep. This is what we see after `OutputManager` stopped being a singleton. Wherever it can be injected in passes, stores, etc., it is injected. `*Cmd`, `*Module` are not configured by Guice. The same thing applies to the options.

mod-tool/src/main/scala/at/forsyte/apalache/tla/Tool.scala
konnov: We probably can. Again, look at the original return type. Someone used `Try[Unit]` before.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: I missed that, we should restrict `path` to the directories managed by this workspace. Good point.

konnov: We can specialize this to the intermediate directory, as it is only called from solver contexts

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: This one is used in one call site. It needs a bit of investigation, but this should be fixable

json-rpc/src/main/scala/com/github/apalachemc/apalache/jsonrpc/JsonRpcServer.scala
konnov: `OutputWorkspace` creates new directories, as Copilot wrote

tla-io/src/main/scala/at/forsyte/apalache/io/lir/CounterexampleWriter.scala
konnov: We can, and I already refactored this somewhere else. However, this is not even in the change set. We could make a separate PR on `*Writer` refactoring.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: This is a very good finding! I completely forgot about this behavior of `PrintWriter`, and I don't know why we used it everywhere in the first place.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/PrettyWriterWithAnnotationsFactory.scala
konnov: Good point. We should move `outputWorkspace` to the methods, instead of having it as part of the state.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: So I made an experimental PR #3447 with this change. It's totally doable. However, I think we should have it in a separate PR to avoid increasing the change set. Also, we better refactor our own `*Writer` classes, as they currently have inconsistent APIs, which could be simplified.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/SolverContextFactory.scala
konnov: Yeah, I also wondered. Now, when `OutputWorkspace` is passed as a parameter, we can see that something strange is going on. The old code basically used `None` values returned by `OutputManager` in the unit tests to avoid IO. So the proper refactoring is to make `OutputWorkspace` a trait and introduce the default implementation as well as a mock. Good that we see it now. I will do it in this PR.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: The thing is that `ensureDirExists` does not always create a directory. If it already exists, it simply returns it. E.g., it is used to locate `_apalache-out`. So the proper name would be `findOrCreateDir`.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: See the comment above. It actually finds or creates a directory.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: This one is harder to get rid of, as it used to implement `parse --output=<file>`. We can, of course, move the code to where it is used. However, this will be against the idea of centralizing all output-related code in one place. I suggest to rename the method, so it suggests not to use it unless really needed.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: I will do that in the IDE directly

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: Yes. This is quite confusing though. I will add a comment.

tla-io/src/main/scala/at/forsyte/apalache/io/annotations/PrettyWriterWithAnnotationsFactory.scala
konnov: I've made `OutputWorkspace` an explicit parameter in the methods of `TlaWriterFactory`. The thing is that `TlaWriterFactory` is more like an abstract class rather than a trait.

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: Renamed it to `withWriterOutsideWorkspace`. Also, added a clear comment.

tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/smt/Cvc5SolverContext.scala
konnov: This is gone

tla-io/src/main/scala/at/forsyte/apalache/io/OutputWorkspace.scala
konnov: Renamed to `findOrCreateDir`

konnov:
> Honestly, I think there are too many issues here. There are comments below, but I'm far from finished here. We should decide if it makes sense to follow up on this or just close it.
> 
> I think this is the problem when an AI agent just writes a fresh implementation of something that hasn't been designed upfront.

@thpani Thank you for your comments! They highlighted a number of deficiencies. I've made another pass that focuses on fixing `OutputWorkspace`, the `Option[...]` issue, the issues in `TlaWriterFactory`. In particular, `OutputWorkspace` is a trait now, and `OutputWorkspaceFileSystem` and `OutputWorkspaceNoopMock` are the two implementations. Since we don't have an `object` now, this is possible.

I suggest we agree on these changes first. Refactoring `PrintWriter` to `BufferedWriter` is a very good next step, but I think we should make it in a separate PR, see the issue #3448.

Refactoring the code that involves `Try[...]` would open a can of worms in this PR, so I suggest to postpone this one. I would also like to remove `Try[...]` and scalaz as much as possible.

konnov:
Closing in favor of #3454

PR #3443, opened 2026-08-07
Introduce in-process integration tests

### LINKED ISSUE #3427: Speed-up integration tests and drop mdx and nix

konnov:
Closes #3427. This PR adds an experimental integration testing framework to replace the current mdx-based e2e tests. The goal is to speed-up test execution as well as to get rid of dependency on MDX and Nix. For more details, see #3427. To focus on the design, the tests include a small subset of 20 actual integration tests.

Using the standard ScalaTest allows us to use the standard features. For instance, see the run times plots: https://github.com/apalache-mc/apalache/actions/runs/31311764237

Currently, we run `integration-tests-*` in parallel with `scala-integration-*`. In the future, we will remove `integration-tests-*`.

Once we agree on the testing framework, an AI tool can easily transfer the remaining tests to this framework. Also, we are not fully exploiting the potential for parallelization in this PR. We will do this later. The current implementation is already offering a 6x speedup over the mdx tests.

The code is written with multiple iterations of Codex GPT 5.6-sol.


konnov:
@copilot can you help me to resolve conflicts against this stacked PR

konnov:
@copilot+claude-sonnet-5 resolve the conflicts, the changes on main are in priority

PR #3462, opened 2026-08-18
Recognize Quint labels prefixed with namespaces

### LINKED ISSUE #3458: Recognize namespace-prefixed labels in the Quint IR
The Quint IR reader recognizes expression labels only when a local binder name
starts directly with `__label_`. Current Quint output can qualify that binder,
for example as `main::replica::__label_proposer`. Apalache then treats it as an
ordinary `LET` binding and loses the label.

This prevents JSON-RPC clients from using the transition labels returned in
`loadSpec.specParameters` to select the relevant decomposed transition.

## Current behavior

Apalache uses the convention:

```quint
pure val __label_proposer = true
<expression>
```

to encode a labeled expression. The Quint reader currently detects this with
logic equivalent to:

```scala
binding.name.startsWith("__label_")
```

and obtains the label with:

```scala
binding.name.stripPrefix("__label_")
```

After flattening an imported or instantiated module, Quint may emit:

```text
main::replica::__label_proposer
```

The prefix check fails. The specification still loads and the transition
semantics are unchanged, but JSON-RPC reports empty label sets for all affected
transitions:

```json
{"index": 0, "labels": []}
```

## Expected behavior

Label recognition should operate on the unqualified final name segment. Both
of these binders should encode the label `Foo`:

```text
__label_Foo
main::__label_Foo
```

konnov:
Closes #3458. Recognize Quint labels prefixed with namespaces.

Done with GPT-5.6-sol high/max.


thpani (approved):

konnov:
@gabrielamafra fyi

PR #3480, opened 2026-09-03
Fix SetInRule

### LINKED ISSUE #3479: [BUG] Unhandled NoSuchElementException in SetInRule when a fold's lambda tests membership in a singleton set literal
## Impact

Not blocking: there is a one-line workaround (below). But the failure is an
unhandled Scala exception rather than a diagnostic, and the trigger is very hard
to see from the spec — the same specification runs fine under Quint's simulator,
and the crash names an arena cell (`$C$8`) that does not appear anywhere in the
input. In a Quint specification the offending set literal is usually at an
operator's *call site*, while the `fold` that makes it fatal is inside the
operator, so the two halves of the trigger can be in different files.

## Input specification

Thirteen lines, no Quint involved:

```tla
-------------------------- MODULE FoldSingletonIn --------------------------
EXTENDS Integers, Apalache

VARIABLE
  \* @type: Int;
  n

\* @type: (Int, Int) => Int;
Count(acc, i) == IF i \in {1} THEN acc + 1 ELSE acc

Init == n = 0
Next == n' = ApaFoldSet(Count, 0, {1, 2})
Inv  == n >= 0
```

Changing `{1}` to `{1, 3}` — a single character — makes it check cleanly. That
is the whole difference between a crash and `The outcome is: NoError`.

```
apalache-mc check --init=Init --next=Next --inv=Inv --length=2 FoldSingletonIn.tla
```

## Expected behavior

The invariant holds, so I expected `The outcome is: NoError`, as I get from the
`{1, 3}` variant. Failing that, a diagnostic naming the unsupported construct.

## Actual behavior

```
java.util.NoSuchElementException: key not found: $C$8
  at at.forsyte.apalache.tla.bmcmt.Binding.apply(Binding.scala:11)
  at at.forsyte.apalache.tla.bmcmt.rules.SetInRule.apply(SetInRule.scala:40)
  at at.forsyte.apalache.tla.bmcmt.rules.IfThenElseRule.apply(IfThenElseRule.scala:31)
  at at.forsyte.apalache.tla.bmcmt.rules.FoldSetRule.$anonfun$apply$1(FoldSetRule.scala:99)
  at at.forsyte.apalache.tla.bmcmt.rules.FoldSetRule.apply(FoldSetRule.scala:84)
  ...
```

`EXITCODE: ERROR (255)`. The full log is under "Log files" below.

## What does and does not trigger it

All of these were run against 0.56.1. The trigger needs three things at once: a
lambda parameter of a **fold** on the left, `\in` (or `contains`), and a
**singleton set literal** on the right.

| variation | result |
|---|---|
| `i \in {1}` inside `ApaFoldSet` | **crash** |
| `i \in {1}` inside `ApaFoldSeqLeft` (Quint `foldl`) | **crash** |
| `acc \in {0}` — the accumulator rather than the element | **crash** |
| `i \in {1, 3}` — two elements | ok |
| `i \in {1, 3, 5}` — three elements | ok |
| `i \in S` where `S` is a state variable | ok |
| `i \in {1}` under `map` / `filter` / `exists` instead of a fold | ok |
| `Cardinality({i, 9})`, `{i} \union {9}`, `{i} \subseteq {1, 2}` inside a fold | ok |
| `i \in {1}` where the singleton is bound by a `LET` outside the fold | ok |

The last row is the workaround: binding the set before the fold keeps the
right-hand side from being a syntactic `enumSet` at the point the rule matches.

## Workaround

Bind the singleton to a `LET` (in Quint, a `val` inside the action — a
module-level `pure val` does **not** work, because Quint inlines it and the
literal comes back):

```quint
// crashes
action step = n' = Set(1, 2).fold(0, (acc, i) => if (Set(1).contains(i)) acc + 1 else acc)

// checks
action step = {
  val s = Set(1)
  n' = Set(1, 2).fold(0, (acc, i) => if (s.contains(i)) acc + 1 else acc)
}
```

## How it turned up

In a Quint specification of a scheduling state machine, where a `fold` over a
map of records asks `scope.contains(i)` and callers passed `Set(theOneId)` at
the call site. Four actions crashed and three did not, which looked arbitrary
until it came down to whether the caller happened to bind the set to a `val`
first. `quint run` and `quint test` were unaffected throughout, so it only
showed up when moving from simulation to `quint verify`.

## System information

- Apalache version: `0.56.1`, build `70cdaf4` (as fetched by Quint 0.32.0)
- OS: Linux (x86-64, container)
- JDK: OpenJDK 21.0.10
- Reached both through `quint verify` and by running `apalache-mc` directly on
  the TLA+ above.

## Minor, while you are in here

The `BugReport.md` that Apalache generates is not quite valid Markdown:

- Dollar signs in the log are backslash-escaped, so the trace reads `\$C\$8` and
  class names read `a.f.a.t.Tool\$`. Thirty occurrences in a 187-line report.
- The "Input specification" block opens with ``` and closes with ````, so
  everything after it renders inside the code block.

Both are visible in the attached `BugReport.md`, which is committed verbatim.

[FoldSingletonIn.tla.txt](https://github.com/user-attachments/files/31580619/FoldSingletonIn.tla.txt)

[BugReport.md](https://github.com/user-attachments/files/31580622/BugReport.md)

konnov:
Closes #3479. When `SetInRule` was called inside a fold, it was passing an arena cell as a name. This failed in the optimization case. The fix is to guard this optimization case with a check that the name is not a cell. Also, adding a similar test to `PrimeRule`.

Fixed with GPT-5.6-sol medium.

### DIFF
--- a/.unreleased/bug-fixes/3479-fold-singleton-membership.md
+++ b/.unreleased/bug-fixes/3479-fold-singleton-membership.md
@@ -1,0 +1,1 @@
+Fixed a crash when a fold's operator tests membership in a singleton set literal, see #3479.
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/PrimeRule.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/PrimeRule.scala
@@ -21,6 +21,9 @@ class PrimeRule extends RewritingRule {
 
   override def apply(state: SymbState): SymbState = {
     state.ex match {
+      case OperEx(TlaActionOper.prime, NameEx(name)) if ArenaCell.isValidName(name) =>
+        throw new RewriterException(s"Prime operator cannot be applied to arena cell $name", state.ex)
+
       case OperEx(TlaActionOper.prime, nEx @ NameEx(name)) =>
         state.setRex(tla.name(name + "'", TlaType1.fromTypeTag(nEx.typeTag)))
 
--- a/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRule.scala
+++ b/tla-bmcmt/src/main/scala/at/forsyte/apalache/tla/bmcmt/rules/SetInRule.scala
@@ -34,7 +34,8 @@ class SetInRule(rewriter: SymbStateRewriter) extends RewritingRule {
     state.ex match {
       // a common pattern x \in {y} that is equivalent to x = y, e.g., the assignment solver creates it
       case OperEx(op, NameEx(name), OperEx(TlaSetOper.enumSet, rhs))
-          if op == TlaSetOper.in || op == ApalacheInternalOper.selectInSet =>
+          if (op == TlaSetOper.in || op == ApalacheInternalOper.selectInSet) &&
+            !ArenaCell.isValidName(name) =>
         val nextState = rewriter.rewriteUntilDone(state.setRex(rhs))
         val rhsCell = nextState.arena.findCellByNameEx(nextState.ex)
         val lhsCell = state.binding(name)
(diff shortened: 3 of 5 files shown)


thpani (approved):
LGTM, thanks for the fix 👍

